PR #9, opened 2020-01-15
Add Dockerfile-dev

tsiegleauq:
Adds the Dockerfile-dev which executes the developing starting commands
in a docker container.
Required to attach other services using docker environment.

Adds the docker building instructions to the makefile

### DIFF
--- a/Dockerfile-dev
+++ b/Dockerfile-dev
@@ -1,0 +1,16 @@
+FROM python:3.8-alpine
+
+RUN apk add build-base
+
+WORKDIR /srv/code
+
+EXPOSE 8000
+
+COPY requirements.txt .
+COPY requirements_production.txt .
+RUN pip install -U -r requirements.txt
+
+COPY gunicorn_conf.py .
+COPY openslides_backend/ ./openslides_backend/
+
+CMD [ "gunicorn", "--config=python:gunicorn_conf", "--reload", "openslides_backend.wsgi:application" ]
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,9 @@ __pycache__/
 .pytest_cache/
 .coverage
 htmlcov/
+
+# Include when developing application packages.
+.vscode/launch.json
+.vscode/settings.json
+.vscode/tasks.json
+.vimrc
--- a/Makefile
+++ b/Makefile
@@ -17,3 +17,9 @@ test:
 
 test_all:
 	OPENSLIDES_BACKEND_RUN_ALL_TESTS=1 pytest
+
+docker-build-dev:
+	docker build -f Dockerfile-dev . -t os4-backend
+
+docker-run-dev:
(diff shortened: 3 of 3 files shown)


Dockerfile-dev
```diff
@@ -0,0 +1,19 @@
+FROM python:3.8
+
+RUN apt-get -y update && \
```
FinnStutzenstein: if `wait-for-it` not used in an entrypoint script, this RUN-command can be removed.

normanjaeckel:
How can I develop inside a docker container (and commit changes and so on)? Maybe we have to map the code into a volume that is the dev repo outside.

Dockerfile-dev
tsiegleauq: you still need the update and upgrade, no?

FinnStutzenstein: Bo, the image provides all needed (up-to-date) software. The update is necessary to fetch the apt-repositories, so apt can find wait-for-it. I think upgrade wasn't necessary at all. But maybe you ask Gernot for a confirmation on this.

tsiegleauq:
> How can I develop inside a docker container (and commit changes and so on)? Maybe we have to map the code into a volume that is the dev repo outside.

Edited the `docker-run-dev` command to:
`docker run -it -v ${PWD}/openslides_backend:/srv/code/openslides_backend -p 8000:8000 --rm os4-backend`

Running `docker-run-dev` now starts a (not silent) server that refreshes once you make changes inside the "openslides_backend"

Dockerfile-dev
tsiegleauq: @gsiv ?

tsiegleauq: Gernot and I removed the commando and changed to image back to alpine

Dockerfile-dev
```diff
@@ -0,0 +1,17 @@
+FROM python:3.8-alpine
+
+RUN apk add build-base
```
normanjaeckel: I think even this run command is not necessary.

tsiegleauq: pip will fail without GCC, which alpine does not bring. It seems that Build-base the recommended way to get GCC.

https://wiki.alpinelinux.org/wiki/GCC

tsiegleauq:
- Rebased on current master
- Adjusted the Docker-dev from using `start.py` to gunicorn, which seems to reload way faster

normanjaeckel:
Sorry for the new merge conflict. I want to play a little bit with docker and developing inside it and then take your commit. You don't have to care about at the moment. Thanks.

normanjaeckel:
Rebased and changed in #15

tsiegleauq:
could you explain what are you doing?

normanjaeckel:
I rebase your commit and added another. Because I did not want to force push into your branch I just opened a new pull request.

PR #15, opened 2020-01-21
Dockerfile dev

normanjaeckel:


Dockerfile
```diff
@@ -1,5 +1,7 @@
 FROM python:3.8-alpine
 
+RUN apk add build-base
```
tsiegleauq: When I build the productive dockerfile, using only `requirements_production.txt` without `requirements.txt`, the build-base package was not required

normanjaeckel: Sure? I thought the ``dependency_injector`` requires building with gcc or somethink like this because there is no wheel online.

PR #19, opened 2020-01-24
Added restrictions view.

normanjaeckel:


openslides_backend/restrictions/restrictions.py
```diff
@@ -0,0 +1,148 @@
...
+                    "description": "Id of the user the given fields should be restricted for.",
+                    "type": "number",
+                    "mininmum": 0,
+                },
+                "fqfields": {
+                    "description": "An array of strings in the format of full qualified fields.",
+                    "type": "array",
+                    "items": {
+                        "type": "string",
+                        "pattern": f"^[a-z_]+{KEYSEPARATOR}[1-9]\\d*{KEYSEPARATOR}[a-z_]+$",
```
ostcar: Are you sure you don't want any numbers in the first and the last part?

openslides_backend/restrictions/restrictions.py
```diff
@@ -0,0 +1,148 @@
...
+        "type": "array",
+        "items": {
+            "type": "object",
+            "properties": {
+                "user_id": {
+                    "description": "Id of the user the given fields should be restricted for.",
+                    "type": "number",
+                    "mininmum": 0,
+                },
+                "fqfields": {
```
ostcar: I would suggest, that you call this "keys" since it is not only about the fields but about all of the key. But I can live with fqfields in this place

openslides_backend/restrictions/restrictions.py
```diff
@@ -0,0 +1,148 @@
...
+
+prepare_restrictions_map()
+
+
+payload_schema = fastjsonschema.compile(
+    {
+        "$schema": schema_version,
+        "title": "Schema for restrictions API",
+        "description": "An array of restriction blobs, i. e. bundles of user_id and full qualified fields.",
+        "type": "array",
```
ostcar: The autoupdate-service creates one request for one user. So the current code does not need the restricter to accept an array, it would only the the first element in the array.

It would be a challenge to bundle all requests from all connected users from one event. I understand, that there are performance improvements on the backend, when the requests are bundled. But I don't think this is possible with the current design of the autoupdate service.

Do you think, you can get the same performance improvements by using some sort of caching?

Do you think it is a big performance problem, when the autoupdate-service sends one request per user? Should I redesign the autoupdate service one request for all users are possible?

openslides_backend/restrictions/restrictions.py
```diff
@@ -0,0 +1,148 @@
...
+                #     fqfield, self.user_id
+                # )
+                blob_response[fqfield] = self.dummy_restriction()
+            response.append(blob_response)
+        self.logger.debug("Restricted data ready.")
+        return response
+
+    def dummy_restriction(self) -> Any:
+        # Just a dummy restrictor, that restricts nothing but always send the same message.
+        return "This is a restricted field content made by dummy restrictor."
```
ostcar: The autoupdate service only send values to the client when the value as changed. Thins this dummy restricter always returns the same data, the autoupdate sends no new data to a (test-) client.

Could you add some variants in the output? For example the current time, a random number or a ascending counter.

openslides_backend/restrictions/restrictions.py
normanjaeckel: I do not know if it is a big performance loss if you have to send me an array even if it always contains only one element.

Some kind of caching might be difficult, at least for the first draft we should deliver until March 1st or April 1st. I have no idea at the moment. For the first draft we just might scale up with more workers (it would be again the network stack that might be the bottle neck).

So I propose that the backend accepts arrays and the autoupdate service later decides whether to bundle or not. (And we even can later decide to remove the array in my code. This should be easier than the other way.)

openslides_backend/restrictions/restrictions.py
normanjaeckel: Of course. The dummy client will send whatever you want (and code in your next pull request ;-) ).

openslides_backend/restrictions/restrictions.py
ostcar: I don't think that it is any problem if I always send an array with one element.

I am also thinking about an optional caching-throtteling in the autoupdate-service. So an request to the restricter is not send immediately but the cache waits for some time (for example 10ms). All other requests in this timeframe are send together. You are right, that this is not necessary for the frist draft.

What do you think about not using an array but an object where the key is the user_id and the value the list of keys?
```
{
  5: ["motion/1/title", "user/5/name"],
  2: ["motion/1/title", "agenda/2/title"],
}
```
I think it is easier this way than keeping track of an index of an array.

The problem with the array is, that when (for any reason) one element of the response get lost, than the wrong connection gets the values for a different user.

normanjaeckel: New format would be ok for me.

ostcar: If it is not urgent for you, then let me think a bit more about it. The disadvantage of the new format is, that when a user has many open connections, the autoupdate service has to distribute the keys of one user to the correct connections. With the array, I could skip this step.

PR #22, opened 2020-01-26
Fix dockerfile

ostcar:
The current Dockerfile has two problems.

1. It can not be build because some of your dependencies need gcc but there is not gcc installed in alpine-linux. I fixed it be using `python:3.8` as distribution, of curse, there are other possibilities. I think the best solution would be to use [multi-stage builds](https://docs.docker.com/develop/develop-images/multistage-build/)

2. You changed the name of the "main" file from `wsgi.py` to `main.py`. I changed it in the README and the Dockerfile accordingly.

3. There where one requirement missing in the `requirements_production.txt`

Is it possible to test with travis or another tool, that the dockerfile can be build and started?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,6 @@ __pycache__/
 .pytest_cache/
 .coverage
 htmlcov/
+
+# Editors
+.vscode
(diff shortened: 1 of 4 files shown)


normanjaeckel:
Duplicate, see #15.

Issue #28, opened 2020-02-04
Change Create and Update event structure


normanjaeckel:
The current spec says that a create event should contain fqid and fields parameters. The current implementation uses only a bunch of fqfields.

Who wants to refactor this (and use this opportunity to get to know my nice codebase)?

@FinnStutzenstein or @tsiegleauq or @emanuelschuetze: Maybe you find a newcomer in your office who wants to have a short look into this repo.

normanjaeckel:
The same for update events of course.

normanjaeckel:
See #35

PR #33, opened 2020-02-13
Refactoring of actions to use plain fieldnames and separate fqid in the action

ThomasJunk:
I looked into #28 and talked to @FinnStutzenstein about the current and future structure. 
And I came up with this changeset.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,4 +46,3 @@ To start it run
 * OPENSLIDES_BACKEND_EVENT_STORE_URL
 
   URL of event store service. Default: http://localhost:9003/
-n
--- a/tests/actions/test_committee.py
+++ b/tests/actions/test_committee.py
@@ -90,40 +90,33 @@ class CommitteeCreateActionPerformTester(BaseCommitteeCreateActionTester):
             self.action.perform(payload, user_id=self.user_id)
 
     def test_perform_correct_1(self) -> None:
+        expected = [
+            {
+                "events": [
+                    {
+                        "type": "create",
+                        "fqid": get_fqid("committee/42"),
+                        "fields": {"organisation_id": 1, "title": "title_ieth5Ha1th",},
+                    },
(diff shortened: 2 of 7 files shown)


README.md
```diff
@@ -46,4 +46,3 @@ To start it run
 * OPENSLIDES_BACKEND_EVENT_STORE_URL
 
   URL of event store service. Default: http://localhost:9003/
-n
```
normanjaeckel: ??? Maybe just remove it.

tests/actions/test_committee.py
```diff
@@ -28,7 +28,8 @@ class CommitteeCreateActionUnitTester(BaseCommitteeCreateActionTester):
     def setUp(self) -> None:
         super().setUp()
         self.action = CommitteeCreate(PermissionTestAdapter(), DatabaseTestAdapter())
-        self.action.user_id = 7668157706  # This user has perm COMMITTEE_CAN_MANAGE.
+        # This user has perm COMMITTEE_CAN_MANAGE.
```
normanjaeckel: Please revert this change. It is much easier for me to copy past a user_id with its short description in test code, if it remains in one line. Please revert also in all other cases.

tests/utils.py
```diff
@@ -69,7 +69,7 @@ def get_fqid(value: str) -> FullQualifiedId:
     """
     Returns a FullQualifiedId parsed from the given value.
     """
-    collection, id = value.split(KEYSEPARATOR)
+    collection, id, *catch_trailing_slash = value.split(KEYSEPARATOR)
```
normanjaeckel: Please do not catch more slashes here. Instead we should check an raise an error if someone tries to use an invalid value here.

ThomasJunk:
I undid the changes.

normanjaeckel:
See #35

PR #36, opened 2020-04-03
Imply guest if auth cannot be reached

tsiegleauq:
Try to connect to the auth server.
If the auth server cannot be reached, treat the user as a guest

Refactors some code

### DIFF
--- a/openslides_backend/services/authentication.py
+++ b/openslides_backend/services/authentication.py
@@ -5,6 +5,9 @@ from simplejson.errors import JSONDecodeError  # type: ignore
 from ..shared.exceptions import AuthenticationException
 from ..shared.interfaces import Headers, LoggingModule
 
+# TODO: something should tell this service what the correct user id of the guest is
+gusest_user_id = 0
+
 
 class AuthenticationHTTPAdapter:
     """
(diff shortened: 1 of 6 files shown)


openslides_backend/services/authentication.py
```diff
@@ -23,24 +26,41 @@ def get_user(self, headers: Headers) -> int:
...
+    def is_auth_accessible(self, url: str) -> bool:
+        """
+        Checks if authentication service is accessible.
+        """
         try:
-            user_id = body["user_id"]
-        except (TypeError, KeyError):
-            raise AuthenticationException(
-                "Empty or bad response from authentication service."
+            response = requests.head(url)
```
ostcar, line 60: Are you sure you want to send two request to the auth service in the healthy situation?

You can also send the real request immediately and check if it works. There is no caching etc, so there is no need for an head request.

Also, your code does not work in all cases. If the auth service gets down after the head request is send, but before the real request is send, then you will get a AuthenticationException.

tsiegleauq: > Are you sure you want to send two request to the auth service in the healthy situation?
> You can also send the real request immediately and check if it works. There is no caching etc, so there is no need for an head request.

I tried that, but did not find any meaningful exceptions in that case. 

> Also, your code does not work in all cases. If the auth service gets down after the head request is send, but before the real request is send, then you will get a AuthenticationException.

Sounds not too realistic.

Issue #32, opened 2020-02-07
Action Response


normanjaeckel:
At the moment our actions do not respond anything. But the spec expects some kind of response. So this has to be implemented.

normanjaeckel:
@ThomasJunk This is also a small issue to get to know my first draft for actions service. It is not a must have but if you like you may have a look at this. Feel free.

FinnStutzenstein:
THe most common response would be the empty one since the read-path of the data flow is going through the autoupdate service. There are some exceptions (I definitely missed some):
- When creating an object, the id is returned (so that the client can open the detail page)
- Import views have typically return values indicating which entries have been imported and which have errored.
- sending emails to users

FinnStutzenstein:
Additionally, the interface must be improved: The response is not just a message, but can be arbitrary json (https://github.com/OpenSlides/OpenSlides/blob/openslides4-dev/docs/interfaces/actions-service.txt)

ThomasJunk:
This issue has #27 as a prerequisite.
So I am deferring this to the point when the discussion on #27 settled

ThomasJunk:
Independently, I made a suggestion regarding the [interface description](https://github.com/OpenSlides/OpenSlides/pull/5343).

jsangmeister:
I think this is obsolete now

PR #40, opened 2020-04-27
Implementing new presenter interface.

ThomasJunk:
Implementation according to
https://github.com/OpenSlides/OpenSlides/blob/openslides4-dev/docs/interfaces/presenter-service.txt

I added the "initial-data"-presenter and the "whomai"-presenter (which may be dropped as @FinnStutzenstein mentioned).

Regarding the registry pattern, I had a look into 
https://stackoverflow.com/questions/5189232/how-to-auto-register-a-class-when-its-defined
which looks like a nice alternatice to the decorator. But I decided against it because one is in need of a class variable which looks more magical than a simple argument in a decorator.

I improved the construct found in https://github.com/OpenSlides/openslides-backend/blob/master/openslides_backend/actions/actions.py to the extent, that I placed the registry alongside the base class https://github.com/ThomasJunk/openslides-backend/blob/new_presenter/openslides_backend/presenter/base.py and did the import in the module entry https://github.com/ThomasJunk/openslides-backend/blob/new_presenter/openslides_backend/presenter/__init__.py

so no extra call to a "register" function is needed. Only the python module loading.

Tests were added too

Please review.

### DIFF
--- a/openslides_backend/http/views.py
+++ b/openslides_backend/http/views.py
@@ -33,10 +33,15 @@ class BaseView:
         self.logger = logging.getLogger(__name__)
 
     def get_user_id_from_headers(self, headers: Headers) -> int:
+        """
+        Returns user id from authentication service using HTTP headers.
+        """
         try:
-            return self.services.authentication().get_user(headers)
+            user_id = self.services.authentication().get_user(headers)
         except AuthenticationException as exception:
             raise ViewException(exception.message, status_code=400)
+        self.logger.debug(f"User id is {user_id}.")
+        return user_id
 
 
 class ActionsView(BaseView):
@@ -52,8 +57,11 @@ class ActionsView(BaseView):
         Dispatches request to the viewpoint.
         """
         self.logger.debug("Start dispatching actions request.")
+
+        # Get user id.
         user_id = self.get_user_id_from_headers(headers)
-        # Setup payload
+
+        # Setup payload.
         payload: ActionsPayload = body
 
         # Handle request.
@@ -83,14 +91,16 @@ class PresenterView(BaseView):
         """
         self.logger.debug("Start dispatching presenter request.")
 
-        # Setup payload
+        # Get user_id.
+        user_id = self.get_user_id_from_headers(headers)
+
+        # Setup payload.
         payload: PresenterPayload = body
 
         # Handle request.
         handler: Presenter = PresenterHandler(
             logging=self.logging, services=self.services,
         )
-        user_id = self.get_user_id_from_headers(headers)
         try:
             presenter_response = handler.handle_request(payload, user_id)
         except PresenterException as exception:
--- a/openslides_backend/http/views.py
+++ b/openslides_backend/http/views.py
@@ -32,6 +32,12 @@ class BaseView:
         self.logging = logging
         self.logger = logging.getLogger(__name__)
 
+    def get_user_id_from_headers(self, headers: Headers) -> int:
+        try:
+            return self.services.authentication().get_user(headers)
+        except AuthenticationException as exception:
+            raise ViewException(exception.message, status_code=400)
+
 
 class ActionsView(BaseView):
     """
@@ -46,22 +52,14 @@ class ActionsView(BaseView):
         Dispatches request to the viewpoint.
         """
         self.logger.debug("Start dispatching actions request.")
-
-        # Get request user id.
-        try:
-            self.user_id = self.services.authentication().get_user(headers)
-        except AuthenticationException as exception:
-            raise ViewException(exception.message, status_code=400)
-
+        user_id = self.get_user_id_from_headers(headers)
         # Setup payload
         payload: ActionsPayload = body
 
         # Handle request.
-        handler: Actions = ActionsHandler()
+        handler: Actions = ActionsHandler(logging=self.logging, services=self.services)
         try:
-            result = handler.handle_request(
-                payload, self.user_id, self.logging, self.services
-            )
+            result = handler.handle_request(payload, user_id)
         except ActionException as exception:
             raise ViewException(exception.message, status_code=400)
         except PermissionDenied as exception:
@@ -89,11 +87,12 @@ class PresenterView(BaseView):
         payload: PresenterPayload = body
 
         # Handle request.
-        handler: Presenter = PresenterHandler()
+        handler: Presenter = PresenterHandler(
+            logging=self.logging, services=self.services,
+        )
+        user_id = self.get_user_id_from_headers(headers)
         try:
-            presenter_response = handler.handle_request(
-                payload, self.logging, self.services
-            )
+            presenter_response = handler.handle_request(payload, user_id)
         except PresenterException as exception:
             raise ViewException(exception.message, status_code=400)
         self.logger.debug("Presenter request finished successfully. Send response now.")
--- a/openslides_backend/presenter/__init__.py
+++ b/openslides_backend/presenter/__init__.py
@@ -1,1 +1,3 @@
+from . import initial_data  # noqa
+from . import whoami  # noqa
 from .presenter_interface import *  # noqa
--- a/openslides_backend/presenter/base.py
+++ b/openslides_backend/presenter/base.py
@@ -1,33 +1,19 @@
-from typing import Any
-
-presenters = {}
+from typing import Any, Dict
 
 
-def register_presenter(name: str) -> Any:
-    """Decoratorfunction
-       @register_presenter is used for
-       automagic adding presenters to the registry (presenters)
-    Arguments:
-        name {string} -- registry the decorated class
+class PresenterBase:  # pragma: no cover
+    """
+    Abstract base class for presenters.
     """
 
-    def wrapper(clazz: object) -> object:
-        """Wrapper / inner function of decorator
-
-        Arguments:
-            clazz {object} -- The actual decorated class
-
-        Returns:
-            [object] -- The decorated class
-        """
-        presenters[name] = clazz
-        return clazz
-
-    return wrapper
+    @property
+    def data(self) -> Dict[Any, Any]:
+        ...
 
 
-class PresenterBase:
-    """Baseclass for Presenters
+class Presenter(PresenterBase):
+    """
+    Base clase for presenters.
     """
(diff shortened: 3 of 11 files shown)


openslides_backend/presenter/presenter.py
```diff
@@ -120,16 +117,13 @@ def parse_presentations(self, payload: Payload) -> PresenterResponse:
...
         for presentation_blob in payload:
-            # user_id = presentation_blob["user_id"]
-            if presentation_blob["presentation"] == "dummy":
-                response.append(self.dummy_presentation())
+            Presenter = presenters.get(presentation_blob["presenter"])
+            if Presenter is not None:
+                p = Presenter()
+                response.append(p.data)
             else:
                 raise PresenterException(
```
FinnStutzenstein, line 125: Hi, one note about the `PresenterException` (see https://github.com/OpenSlides/OpenSlides/blob/openslides4-dev/docs/interfaces/presenter-service.txt#L24):

You must differentiate between a server-error vs. client-error (analog to HTTP 400 vs 500). A presenter may raise an exception (you have to decide if it is the `PresenterException` or something else) on user errors. E.g. a number was expected as the payload but something else was given. This must result in a HTTP 200 response with `[{error: "a number was expected"}]` as the payload. A HTTP 500 can be raised, if e.g. a dependend service failed.

For non-existent presenters (e.g. the clients request the presenter `not-existent`) there are two possibles: Either use the error response as stated above. This allowes for the execution of other presenters, if multiple were requested. The other option is to check for validity of all presenters before executing them. Then you may raise a 400-error with a hint that some presenters are unknown.

openslides_backend/http/views.py
```diff
@@ -89,11 +89,12 @@ def dispatch(self, body: RequestBody, headers: Headers) -> ResponseBody:
         payload: PresenterPayload = body
 
         # Handle request.
-        handler: Presenter = PresenterHandler()
+        handler: Presenter = PresenterHandler(
+            logging=self.logging, services=self.services,
+        )
+        user_id = "Test"
```
normanjaeckel, line 95: See line 51 how to get the user id. this should already work and give an integer.

openslides_backend/http/views.py
```diff
@@ -89,11 +89,12 @@ def dispatch(self, body: RequestBody, headers: Headers) -> ResponseBody:
         payload: PresenterPayload = body
 
         # Handle request.
-        handler: Presenter = PresenterHandler()
+        handler: Presenter = PresenterHandler(
```
normanjaeckel, line 92: Why not like for ActionHandler? If your way is better then we should change the way of action handler. Then please open an extra issue for this.

openslides_backend/presenter/__init__.py
```diff
@@ -1 +1,3 @@
 from .presenter_interface import *  # noqa
+from . import initial_data  # noqa
```
normanjaeckel, line 2: I strongly prefer that the init does not import more than the inferface file. As I did it for action module.

openslides_backend/presenter/base.py
```diff
@@ -0,0 +1,13 @@
+presenters = {}
```
normanjaeckel, line 1: Please add typing and docstrings like in actions/actions.py and actions/base.py. Of course we can discuss whether to have actions/actions.py and base.py or only one of these files. We should do this for presenter equally. We can have a phone call if this is easier.

openslides_backend/presenter/presenter.py
```diff
@@ -120,16 +117,13 @@ def parse_presentations(self, payload: Payload) -> PresenterResponse:
         )
         response = []
         for presentation_blob in payload:
-            # user_id = presentation_blob["user_id"]
-            if presentation_blob["presentation"] == "dummy":
-                response.append(self.dummy_presentation())
+            Presenter = presenters.get(presentation_blob["presenter"])
+            if Presenter is not None:
+                p = Presenter()
```
normanjaeckel, line 122: Maybe you have to inject the dependencies (database and permission) here too. Of course this is not required at the moment we can already prepare for it because I am sure we sill need then.

openslides_backend/presenter/presenter_interface.py
```diff
@@ -18,6 +18,6 @@ class Presenter(Protocol):  # pragma: no cover
     """
 
     def handle_request(
-        self, payload: Payload, logging: LoggingModule, services: Services,
+        self, payload: Payload, logging: LoggingModule, services: Services, user_id: str
```
normanjaeckel, line 21: See above: Dependency injection here or at init level?

openslides_backend/presenter/base.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+def register_presenter(name):
+    def wrapper(c):
```
normanjaeckel, line 5: Just a stylish remark: I prefer self explaning variable names not just single characters like `c`. See the same function in action. Please also add typings.

openslides_backend/presenter/presenter.py
normanjaeckel, line 122: Here the same. I would prefere `presenter` or `presenter_instance` or something like that instead of `p`. But this is personal style of course I won't insist on that.

openslides_backend/presenter/presenter.py
normanjaeckel: Ah I am not sure about this. See actions module for the same case. The module may always use Python exceptions maybe with different code names so that the http component may decide whether to send HTTP 4xx or 5xx. If a user sends a bad payload he should get HTTP 4xx and not HTTP 200 with error message. in my opinion. But that is still discussable. For the first implementation I would recommend the way we use in action module so we have a first consistent implementation here.

openslides_backend/presenter/__init__.py
normanjaeckel: Of course we can discuss this. My issue was that one may import the interface without "loading" all the stuff.

openslides_backend/presenter/presenter.py
ThomasJunk: Regarding HTTP-Status-Codes I share the view of @normanjaeckel on this.
There is a relatively clear semantics on this: 

`HTTP 200` means `OK` 
See https://tools.ietf.org/html/rfc7231#section-6.3

`HTTP 400`: is a formally malformed request; that excludes questions of validity.
See https://tools.ietf.org/html/rfc7231#section-6.5.1 
> malformed request syntax, invalid request message framing, or deceptive request routing

`HTTP 500`: is a formally wellformed request but the server can not deal properly with the content. This would include questions on validation.

See https://tools.ietf.org/html/rfc7231#section-6.6 

I have a look at how this was done in the actions module :]

openslides_backend/http/views.py
ThomasJunk: Thanks. I moved the extraction part to the base class. See https://github.com/OpenSlides/openslides-backend/pull/40/commits/9e5bf7ff16b0b7375eaee4d92c6c4fcd6d21c35f

openslides_backend/http/views.py
ThomasJunk: > Then please open an extra issue for this.
See https://github.com/OpenSlides/openslides-backend/issues/43

openslides_backend/presenter/__init__.py
ThomasJunk: > My issue was that one may import the interface without "loading" all the stuff.
May you elaborate on this one? What are the reasons behind this? Are there performance reasons?

openslides_backend/presenter/base.py
ThomasJunk: See https://github.com/OpenSlides/openslides-backend/pull/40/commits/37d88fca9da06527257d44eea8bebf66562f94a0#diff-45ecc7a634e3cdfde96d2afe26f4af0d

openslides_backend/presenter/presenter.py
ThomasJunk: ATM I do not understand the role of the `Presenter` (or if you will _pars pro toto_ the architecture) as such. For the moment I took the Presenter as a `ViewModel` presenting not necessarily retrieving the data required.

openslides_backend/presenter/presenter_interface.py
ThomasJunk: At `init`-level (at least for now).

openslides_backend/presenter/base.py
ThomasJunk: See https://github.com/OpenSlides/openslides-backend/pull/40/commits/213db5f4841c9b12a29f0ed2c22d44e2bb5b3091

openslides_backend/presenter/presenter.py
ThomasJunk: Usually I decide on scope: the wider the scope of a variable the more descriptive its name. But I adapt in order to provide a consistent style.

openslides_backend/presenter/presenter.py
FinnStutzenstein: We are using the status codes for different reasons here, since one can send multiple presenters. presenter-scoped arrors should not stop the execution of other presenters and must so reported as error dicts.

HTTP status codes, on the other hand, are request-scoped, meaning the payload of all presenters. If e.g. a non-existing presenter is specified it is a request-scoped error resulting in a 400er.

We should divide the both layers where errors can happen.

ThomasJunk: I opt for leaving this out of this pull request. Otherwise this PR gets _feature creep_ and I start fixing too many things at once. 

Dealing with results of Actions or Presenters is part of https://github.com/OpenSlides/openslides-backend/issues/32#event-3279225558
Perhaps this is a better place to talk about this - or even a separate issue.

normanjaeckel:
Ok, nice. Can you please rebase all commits to one commit. Then I will add a second one only with stylish stuff (docstrings and thelike) I do not want to bother you and merge it. The point in init file we can discuss separately.

openslides_backend/presenter/__init__.py
normanjaeckel: Lets make an extra issue for that.

ThomasJunk:
I stashed the commits :smile:

PR #48, opened 2020-05-07
First draft of datastore implementation ready for merging

ThomasJunk:
I suggest integrating the current version of the datastore implementation in order to avoid possible drift between repositories. This is a _irst draft_ of the implementation which needs to be reviewed. 

Basically there are unit test which have the main focus of securing stable APIs. Functionally nothing exciting is happening. 

I made some additions to the previos codebase:

* Adding more types
* Separating `Adapter` and `Engine` in their respective subpackages each having its own interface
* Substitution of the `Union` based type declaration with an abstract base class for `filters`
  due to mypy complaining _The type alias to Union is invalid in runtime context _
* Enhancement of the `Filter` classes with a `to_dict()` method in order to leverage _chaining_.
* Adding tests to the `Filter`

Known ToDo:
Writing inegration tests.


ThomasJunk:
I added integration tests.
In order to trigger the test manually you could execute `TESTCONTEXT=INTEGRATION pytest` locally. Since I have no idea atm how to set up a complete testing environment the tests assume things I have locally on my machine like collections `a` and `b` and having several members `1`, `2`, `3`, `4`,`5` each having `f`` as field set to various numbers. For a first draft I think it does the trick.

normanjaeckel:
Ok, very nice. I will have a look at it the next days.

One first remark: We have the adapter interface in the file shared/interfaces.py. This is used in http/views.py via the Services interface. The idea behind this was that the interface is separated from the implementation in a separate module (@FinnStutzenstein right?). So I think you may move the new adapter and engine interfaces to this file.

ThomasJunk:
>One first remark: We have the adapter interface in the file shared/interfaces.py. This is used in http/views.py via the Services interface. 

So many people - so many customs :]

>  The idea behind this was that the interface is separated from the implementation in a separate module

I fully support the idea of having distinct modules for implementation and contract (interface). 

But I would argue having one module where every interface is to be found is a violation of the [single responsibility principle](https://en.wikipedia.org/wiki/Single-responsibility_principle).

But @FinnStutzenstein should decide on how he wants this to be structured.

normanjaeckel:
Very nice. I think we can use this for the moment without major changes. Some cosmetics I can do myself another day (like docstrings starting not at the same line as the `"""`).

Two big points: Can you please clean up this pull request: There are only three commits that contain your code. All other commits are merge fragments. I propose that you use a clean checkout from commit 14202b9403bc483bc49c387b43d9548b35b9f2dc for this. Then you may rebase them all to one commit. We have been using this style in OpenSlides for some time to keep the commit history as clean as possible.

Second: Please open an issue where we can discuss the open question about where the interface should live. At the moment we have two interfaces (yours and mine) and both are "used" in some kind. This should be harmonized.

Thank you very much.

ThomasJunk:
So, I hope my git fu helped reintegrating with upstream. I cleaned my master and put eveything into one commit.

PR #49, opened 2020-05-09
Agenda model

normanjaeckel:


normanjaeckel:
@ostcar: [This](https://github.com/OpenSlides/openslides-backend/pull/49/files#diff-3c3d8eb43c9f8dbb633b2ae3730de347R19) is broken. I need an OrderedDict here but mypy complains and I do not know how to solve.

normanjaeckel:
Fixed.

PR #51, opened 2020-05-14
Change relations

normanjaeckel:


normanjaeckel:
I will merge this even the code can be better. I will have a look at it the next days (e. g. to enhance typings).

PR #53, opened 2020-06-08
Connecting components to datastore. Combined reader and writer and re…

normanjaeckel:
…factored the service adatper.

There is still some hard coded stuff in it.


normanjaeckel:
Attention: This pull request brings some hard coded stuff within but we can fix this one by one ...

PR #58, opened 2020-06-10
(WIP) Changes required for integrating all services.

FinnStutzenstein:
- Changed loading of environment variables
- Renamed makefile targets

@normanjaeckel When the first integration works, someone should move the tests into a docker container.


normanjaeckel:
Can we please reduce complexity in setting up all environment variables? Just lets write a list with all vars and their defaults without magic combining functions, automatic suffixes for host and port and thelike.

FinnStutzenstein:
@normanjaeckel is it better now?

Dockerfile-dev
```diff
@@ -7,20 +7,15 @@ ENV OPENSLIDES_BACKEND_DEBUG 1
 
 WORKDIR /srv/code
 
-RUN apk add --no-cache gcc musl-dev linux-headers git \
```
FinnStutzenstein, line 10: Note: For the dev setup, we do not need an extra user, since changes in this container (with correct mounted volumes) should reflect on the host system, e.g. when running black

normanjaeckel:
I rebased and fixed some stuff but also changed the Makefile a little bit. Because the dockerfile does not work with current directory but only wiht Github master I will merge and then test everything.

FinnStutzenstein:
Hi, Please undo the changed environment variables. At least the `<service>_HOST` and `<service>_PORT` must be evaluated. This is to unify the setup between all services, so I do not have to declare special parameters for each service. The goal ist to have the productive setup as easy as possible, because to manage and integrate so many services is a lot of work.

normanjaeckel:
I do not see the advantage to split up the endpoints (urls) into its components (procotol, host, port, path) but I can so so if you like.

FinnStutzenstein:
Other services also rely on the exact same env variables. The urls and schemata are constant (if you do them as env variables, it is ok), but the host and port are controlled by the surrounding setup, so these things must be ijectable from outside. And this should be identical for every service, otherwise it is a configuration-mess.

PR #73, opened 2020-07-22
Implement dev fixes

jsangmeister:


openslides_backend/action/relations.py
```diff
@@ -96,7 +96,7 @@ def perform(self) -> Relations:
                 ],
                 lock_result=True,
             )
-            rels = response[target]
+            rels = response[target] if target in response else {}
```
jsangmeister, line 99: If I call a `topic.create` from the client, the backend tries to fetch mediafiles from the datastore but provides an empty array of ids, which leads to a `KeyError` here. This was the easiest fix for me, feel free to improve it.

openslides_backend/services/authentication.py
```diff
@@ -50,9 +50,8 @@ def get_user(self, headers: Headers) -> int:
             try:
                 user_id = body["user_id"]
             except (TypeError, KeyError):
-                raise AuthenticationException(
-                    "Empty or bad response from authentication service."
-                )
+                # TODO: remove this and throw an AuthentificationException when permissions are implemented.
+                user_id = 1
```
jsangmeister, line 54: To not be dependent on the right communication between auth and backend and since at first, we want to build the app admin-only for the first milestone, I added this.

FinnStutzenstein: Can you move this up: Just do not do an request. We have to wait a bit for the auth service.

jsangmeister: Done

FinnStutzenstein:
Please fix travis - then we can merge this since there was no further response from anyone else.

jsangmeister:
Commented the tests out for now. Can put back in once auth is fixed

FinnStutzenstein:
Travis still does not succeed.

normanjaeckel:
The remaining travis error just looks like a problem of isort. Try "make isort" to fix it.

PR #75, opened 2020-07-28
Add Settings

jsangmeister:
Test in conjunction with https://github.com/OpenSlides/openslides-client/pull/19
Settings are now fully functional.

### DIFF
--- a/openslides_backend/action/meeting/create_update_delete.py
+++ b/openslides_backend/action/meeting/create_update_delete.py
@@ -15,10 +15,10 @@ class MeetingActionSet(ActionSet):
 
     model = Meeting()
     create_schema = DefaultSchema(Meeting()).get_create_schema(
-        properties=["committee_id", "name"],
+        properties=["committee_id", *Meeting().get_settings_keys()],
         required_properties=["committee_id", "name"],
     )
     update_schema = DefaultSchema(Meeting()).get_update_schema(
-        properties=["name", "motion_poll_default_group_ids"]
+        properties=Meeting().get_settings_keys()
     )
     delete_schema = DefaultSchema(Meeting()).get_delete_schema()
--- a/openslides_backend/models/meeting.py
+++ b/openslides_backend/models/meeting.py
@@ -1,3 +1,5 @@
+from typing import Iterable
+
 from ..shared.patterns import Collection
 from . import fields
 from .base import Model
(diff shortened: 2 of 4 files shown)


openslides_backend/models/fields.py
```diff
@@ -11,51 +11,62 @@ class Field:
...
-        return dict(description=self.description, type="integer",)
+        return self.extend_schema(super().get_schema(), minimum=1)
 
 
-class PositiveIntegerField(Field):
+class IdField(PositiveIntegerField):
+    pass
+
+
+class TimestampField(Field):
```
normanjaeckel, line 43: This should inherit from PositiveIntegerField

openslides_backend/models/fields.py
```diff
@@ -120,46 +131,43 @@ def __init__(
...
             items={"type": "integer", "minimum": 1},
             uniqueItems=True,
         )
+
+
+class ManyToManyArrayField(ManyIdField):
+    type = "m:n"
+
+
+class ManyToOneArrayField(ManyIdField):
```
normanjaeckel, line 172: Please remove or at least comment this type of field. I strongly feel that can not use it but build such relations always from the reverse side as ForeignKey.

openslides_backend/models/meeting.py
```diff
@@ -16,10 +18,254 @@ class Meeting(Model):
     collection = Collection("meeting")
     verbose_name = "meeting"
 
+    def get_settings_keys(self) -> List[str]:
```
normanjaeckel, line 21: Please change return type to Iterable

openslides_backend/action/meeting/create_update_delete.py
```diff
@@ -12,8 +12,10 @@ class MeetingActionSet(ActionSet):
 
     model = Meeting()
     create_schema = DefaultSchema(Meeting()).get_create_schema(
-        properties=["committee_id", "name"],
+        properties=["committee_id", "name", *Meeting().get_settings_keys()],
```
normanjaeckel, line 15: Remove "name" here because it is included in settings keys.

openslides_backend/models/fields.py
normanjaeckel: Please also add a short comment or docstring what this integer is about (the UNIX timestamp I think).

openslides_backend/models/meeting.py
```diff
@@ -16,10 +18,254 @@ class Meeting(Model):
...
+
     id = fields.IdField(description="The id of this meeting.")
     committee_id = fields.RequiredForeignKeyField(
         description="The id of the committee of this meeting.",
         to=Collection("committee"),
         related_name="meeting_ids",
     )
-    name = fields.RequiredCharField(description="The name of this meeting.")
+
+    welcome_title = fields.CharField()
```
normanjaeckel, line 129: I dont check this big list but trust you that you created it carefully. ;-)

openslides_backend/models/meeting.py
```diff
@@ -16,10 +18,254 @@ class Meeting(Model):
     collection = Collection("meeting")
     verbose_name = "meeting"
 
+    def get_settings_keys(self) -> List[str]:
+        return [
+            "name",
```
normanjaeckel, line 23: Missing welcome title and welcome text.

openslides_backend/services/authentication.py
```diff
@@ -50,9 +50,8 @@ def get_user(self, headers: Headers) -> int:
             try:
                 user_id = body["user_id"]
             except (TypeError, KeyError):
-                raise AuthenticationException(
-                    "Empty or bad response from authentication service."
-                )
+                # TODO: remove this and throw an AuthentificationException when permissions are implemented.
```
normanjaeckel, line 53: Please revert this change because this is done in #73.

openslides_backend/models/meeting.py
```diff
@@ -16,10 +18,254 @@ class Meeting(Model):
...
+
+    assignment_poll_default_type = fields.TextField()  # TODO: PollType enum
+    assignment_poll_default_method = fields.TextField()  # TODO: PollMethod enum
+    assignment_poll_default_100_percent_base = (
+        fields.TextField()
+    )  # TODO: PercentBase enum
+    assignment_poll_default_majority_method = (
+        fields.TextField()
+    )  # TODO: MajorityMethod enum
+    assignment_poll_default_group_ids = fields.ManyToOneArrayField(
```
normanjaeckel, line 243: Please remove this setting here and in the list above. We should define this as boolean field in group model.

openslides_backend/models/meeting.py
```diff
@@ -16,10 +18,254 @@ class Meeting(Model):
...
+    motions_amendments_multiple_paragraphs = fields.BooleanField()
+    motions_supporters_min_amount = fields.IntegerField(minimum=0)
+    motions_supporters_enable_autoremove = fields.BooleanField()
+
+    motion_poll_default_type = fields.TextField()  # TODO: PollType enum
+    motion_poll_default_100_percent_base = fields.TextField()  # TODO: PercentBase enum
+    motion_poll_default_majority_method = (
+        fields.TextField()
+    )  # TODO: MajorityMethod enum
+    motion_poll_default_group_ids = fields.ManyToOneArrayField(
```
normanjaeckel, line 222: Please remove this setting here and in the list above. We should define this as boolean field in group model.

openslides_backend/models/meeting.py
```diff
@@ -16,10 +18,254 @@ class Meeting(Model):
...
+    list_of_speakers_amount_next_on_projector = fields.IntegerField(minimum=-1)
+    list_of_speakers_couple_countdown = fields.BooleanField()
+    list_of_speakers_show_amount_of_speakers_on_slide = fields.BooleanField()
+    list_of_speakers_present_users_only = fields.BooleanField()
+    list_of_speakers_show_first_contribution = fields.BooleanField()
+
+    motions_default_workflow_id = fields.RequiredForeignKeyField(
+        to=Collection("workflow"), related_name="default_workflow_meeting_id",
+    )
+    motions_default_statute_amendments_workflow_id = fields.RequiredForeignKeyField(
```
normanjaeckel, line 179: Please remove this setting here and in the list above. We should define this as boolean field in workflow model.

openslides_backend/models/meeting.py
```diff
@@ -16,10 +18,254 @@ class Meeting(Model):
...
+    agenda_show_internal_items_on_projector = fields.BooleanField()
+
+    list_of_speakers_amount_last_on_projector = fields.IntegerField(minimum=0)
+    list_of_speakers_amount_next_on_projector = fields.IntegerField(minimum=-1)
+    list_of_speakers_couple_countdown = fields.BooleanField()
+    list_of_speakers_show_amount_of_speakers_on_slide = fields.BooleanField()
+    list_of_speakers_present_users_only = fields.BooleanField()
+    list_of_speakers_show_first_contribution = fields.BooleanField()
+
+    motions_default_workflow_id = fields.RequiredForeignKeyField(
```
normanjaeckel, line 176: Please remove this setting here and in the list above. We should define this as boolean field in workflow model.

tests/action/test_meeting.py
```diff
@@ -428,6 +428,10 @@ def test_prepare_dataset_1(self) -> None:
                         "committee_id": None,
                         "motion_ids": None,
                         "topic_ids": None,
+                        'assignment_poll_default_group_ids': None,
```
normanjaeckel, line 431: After you fixed my comments above you may also revert this change.

FinnStutzenstein:
> Please remove this setting here and in the list above. We should define this as boolean field in group model.

This was done intentionally: With using a relation, the access for those special objects are instant. You do not have to search for them. E.g. for groups there are still boolean flags, which I'll convert to relations, since I noticed that we need this in the client. Also relations are autoupdate-friendly: I can just request the default for X or a special thing for Y wihtout the need of having all objects and searching for them.

openslides_backend/models/fields.py
jsangmeister: I needed this to be able to update the relations like `assignment_poll_default_group_ids` from the meeting side. As @FinnStutzenstein already said, it's better and wanted to have these as relations. I tried to define a `ForeignKey` on groups but with this I was not able to update it from the meeting side since the field does not exist there. I agree with you that this would be the cleaner solution, but I could not figure out what I needed to change to make it work. Please feel free to make the reverse relation fully functional, then we can remove this.

openslides_backend/models/meeting.py
jsangmeister: Left it out intentionally, since they don't belong to the traditional settings since they are not on the settings page. They can be added in a future PR. However, I do concur (if nobody says something against it) that they can be handled in the same action.

openslides_backend/services/authentication.py
jsangmeister: Yes, rebased on master

jsangmeister:
Fixed remarks and added custom translations.

openslides_backend/models/fields.py
FinnStutzenstein: Wait, reverse relations do currently not work in the action service??

FinnStutzenstein:
I do not get a json response on errors: You can set motions supporters min amount to -1 to test it. When you adjusted the response, please care about error handling in the client.

openslides_backend/models/meeting.py
normanjaeckel: Maye lets avoid creating an extra action just for these two fields.

openslides_backend/models/fields.py
normanjaeckel: A ManyToOne-Field is not fully implemented in the relations.py logic. But I think is is possible to use the reverse relation by just adding a list of values to e. g. `assignment_poll_default_group_ids`. This case should be handles e. g. [here](https://github.com/OpenSlides/openslides-backend/blob/master/openslides_backend/action/relations.py#L254).

openslides_backend/models/meeting.py
normanjaeckel: Maybe avoid adding an extra action just for these two fields.

normanjaeckel:
Using foreign keys instead of boolean fields does not look very clean but if you like we can try this way.

normanjaeckel:
But maybe change to OneToOne field instead.

FinnStutzenstein:
> OneToOne

It is an O2O field (just to clarify, I do not know your abstraction/implementation).

openslides_backend/models/fields.py
jsangmeister: Yes, that is what I wanted to do initially. If I add a group model with

    used_as_motion_poll_default_id = fields.ForeignKeyField(
        to=Collection("meeting"), related_name="motion_poll_default_group_ids"
    )

and give `motion_poll_default_group_ids` in the create/update schema of a meeting, the backend fails to boot with `ValueError: motion_poll_default_group_ids is not a field of meeting`. That's why I wrote the `ManyToOneField`. @normanjaeckel If you fix the relations, we can remove it again.

normanjaeckel: Please merge #87 first and then set this pull request on top. Then the problem should be solved.

jsangmeister: Does not work right now. I added a test `test_settings` which fails because of the bugged relation handling. The code seems to think that the `1` side of the relation is an array and the `m` side is a single id...
Other way to reproduce another error: startup the main repos setup and try to update the default group ids in the settings, it fails on https://github.com/OpenSlides/openslides-backend/blob/19de8c82603df7e35fdd43d7823cf38dbc9113d3/openslides_backend/action/relations.py#L283 because the value is not a list.

normanjaeckel:
This pull request has to be rebased on top of master to solve conflicts.

PR #76, opened 2020-07-28
Restructure presenters and add GetMediafileId presenter

jsangmeister:


openslides_backend/presenter/base.py
```diff
@@ -1,19 +1,43 @@
...
+    
+    def validate(self) -> None:
+        """ Validates the given data. If self.schema is not set, assumes that no data should be given. """
+        if self.schema:
+            if self.data is None:
+                raise PresenterException("No data given.")
+            try:
+                # unfortunately, python injects self if we use the bounded method, so
+                # we have to use the class method
+                self.__class__.schema(self.data)
```
normanjaeckel, line 34: In BaseAction I use `type(self).schema(...)`. Which solution is better?

openslides_backend/presenter/get_mediafile_id.py
```diff
@@ -0,0 +1,60 @@
...
+from ..shared.filters import FilterOperator
+import fastjsonschema
+from ..shared.schema import schema_version
+from ..shared.exceptions import PresenterException
+
+
+get_mediafile_id_schema = fastjsonschema.compile(
+    {
+        "$schema": schema_version,
+        "type": "object",
```
normanjaeckel, line 15: Please add short title and description here (maybe we want to use this schema later also for documentation purposes).

openslides_backend/presenter/get_mediafile_id.py
```diff
@@ -0,0 +1,60 @@
...
+class GetMediafileId(BasePresenter):
+    """
+    Retrieve an Id for the given mediafiel or None if it not exists or access is forbidden.
+    """
+    
+    schema = get_mediafile_id_schema
+
+    def get_result(self) -> Any:
+        # TODO: filter by meeting id
+        filter = FilterOperator("path", self.data["path"], "=")
```
normanjaeckel, line 39: Operator is `==` not `=` I think.

openslides_backend/presenter/get_mediafile_id.py
```diff
@@ -0,0 +1,60 @@
...
+        result = self.datastore.filter(
+            collection=Mediafile().collection,
```
FinnStutzenstein, line 41: Do you have to instantiate `Mediafile` to get the collection?

openslides_backend/presenter/base.py
jsangmeister: Ahh, didn't see that, that would have saved me quite some trouble... I think yours is cleaner, will change it, thanks

openslides_backend/presenter/get_mediafile_id.py
jsangmeister: It's `=`: https://github.com/OpenSlides/OpenSlides/blob/8c57d911a30f22b41bbfe3b2b494bd805ffc6af4/docs/interfaces/datastore-service.txt#L266

jsangmeister: (Btw, it's this way because we inject it directly into the query and SQL also uses `=`)

jsangmeister:
@normanjaeckel I wrote tests for the presenter and refactored a little. I think your approach to system/integration tests is way too heavy: In your `fake_services/databse.py`, you are basically rewriting the whole datastore. This is not the intention of a system test. The better way would be to just patch the right method(s), in this case the `get`/`filter`/... methods of the adapter. You have to then specifiy in each test what they should return so  you can test the actions/presenters with different outputs. I could not figure out how to access the DI from the tests, else I would have implemented an example. Please ask if you have any questions.

Also, some of the datastore signatures were wrong, I fixed all places I could find, please double check if I missed something. Normally, this wuld belong in another PR but since multiple test were failing because of this I did it in this one.

tests/fake_services/http_engine.py
```diff
@@ -82,10 +82,10 @@ def filter(self, data: str) -> Tuple[bytes, int]:
         mapped_fields = payload.get("mapped_fields")
         if payload.get("meeting_id"):
             raise ValueError("The argument meeting_id is not supported here.")
-        result: List[Dict[str, Any]] = []
-        if not filter.get("operator") == "==":
+        result: Dict[int, Dict[str, Any]] = {}
+        if not filter.get("operator") == "=":
```
FinnStutzenstein, line 86: Can you define constants for these? The fact, that there was a mistake and several places needed adjustments ist a good opportunity for it ;)

FinnStutzenstein:
I created an issue to track the need of refactoring, so just necessary changes can be done here and other refactoring in another branch.

FinnStutzenstein:
@jsangmeister I added some quick fixes. Please check, if this is ok. I still like GET more, but browsers are unable to send body data in GET requests. The best would be to allow both get and post; but this is not important I guess.

openslides_backend/presenter/get_mediafile_id.py
normanjaeckel: Ok then I have some bugs in other places of the code. See #79.

tests/fake_services/http_engine.py
jsangmeister: Wait, do you mean the operators? I think they are sufficiently verbose that no constants are needed. It's especcially nice that you can do `FilterOperator(field, "=", value)` which is about as verbose as it can get, while it would be less verbose to write `FilterOperator(field, Operator.EQUALS, value)` IMO

jsangmeister:
Since the refactoring should not be done here, I just fixed the tests to pass travis. Will need to fix/improve them later

tests/fake_services/http_engine.py
FinnStutzenstein: Yes I mean exactly this. Otherwise there is room for mistakes. It is more verbose, but maybe you should do it this way and do not use `FilterOperator` directly: Make child classes (Is this a class?) like `EqualOperator`, and so on. Encapsulate the usage. Imagine we change `=` to `~` -> You hav to change may places in the code. With the constant it is easy. With the more abstration it is even better. E.g. special docs can be provided for each operator, how they behave. Like: What happens, if you compare a number and a string. It is specified in the DS interface, but it might be done here in the docs, too, so foreign developers can directly see how the `EqualOperator` behaves.

jsangmeister: I see where you are coming from, but I don't agree. We will most likely never touch these again, except to add more operators. If you still think it is that important, I can do it, but I don't see the necessity :D

PR #77, opened 2020-07-29
Move cleanup & tests inside docker container

jsangmeister:


jsangmeister:
@FinnStutzenstein I can't seem to figure out where the error is, can you take a look at it? pytest fails inside the docker container

Makefile
```diff
@@ -46,13 +28,25 @@ build-dev:
...
+run-dev:
+	docker run $(dev_args)
+
+run-dev-interactive run-bash:
+	docker run $(dev_args) sh
+
+run-cleanup:
+	docker run $(dev_args) sh cleanup.sh
+
+run-tests:
```
FinnStutzenstein, line 45: A dependency like `build-dev` is missing. same for `run-cleanup`.

Makefile
```diff
@@ -46,13 +28,25 @@ build-dev:
...
+	docker run $(dev_args) sh OPENSLIDES_BACKEND_RUN_ALL_TESTS=1 pytest
```
FinnStutzenstein, line 46: sh: can't open 'OPENSLIDES_BACKEND_RUN_ALL_TESTS=1': No such file or directory

FinnStutzenstein:
```suggestion
	docker run $(dev_args) sh "OPENSLIDES_BACKEND_RUN_ALL_TESTS=1 pytest"
```

Correct quoting does the trick.

FinnStutzenstein: Another thing: I got an INTERNALERROR `AssertionError: /home/finn/OpenSlides/OpenSlides/openslides-backend/tests/integration/test_database.py` inside the container -> you have to remove all `__pycache__` folders and `*.pyc` files. Those compiled artifacts were mounted inside the container and confused pytest.

Makefile
FinnStutzenstein: As far as I can tell the dependency is missing everywhere.

jsangmeister: Yes, these weren't there beforehand. I assumed @normanjaeckel didn't want that dependency to be executed on every run but to have manual control over it. But if he agrees, I can sure add it

Makefile
jsangmeister: > you have to remove all __pycache__ folders and *.pyc files.

Ahh, that's what I was missing, thanks.

> docker run $(dev_args) sh "OPENSLIDES_BACKEND_RUN_ALL_TESTS=1 pytest"

Didn't work for me, the `-c` switch did the trick. But after pytest didn't work, I didn't bother to look at that.

jsangmeister:
Done. The authentification tests all fail, is that a problem with the docker container?

Makefile
FinnStutzenstein: At least for `run-tests` (a required target from the main repo), it just should work without any extra steps. And docker caches image layers. If there is nothing to rebuild, this should be done in seconds.

Dockerfile-dev
```diff
@@ -12,11 +9,13 @@ RUN apk add --no-cache gcc musl-dev linux-headers git
...
 COPY requirements.txt .
 COPY requirements_production.txt .
+COPY cleanup.sh .
+COPY setup.cfg .
 
 RUN pip install --no-cache-dir --requirement requirements.txt
 
-COPY openslides_backend/ ./openslides_backend/
-COPY setup.cfg .
+ARG OPENSLIDES_BACKEND_ACTION_PORT=9002
```
normanjaeckel, line 17: Why dont you have to copy the code into the container?

Makefile
```diff
@@ -1,26 +1,8 @@
...
-
-test:
-	pytest
-
-run-tests:
-	echo "This has to be fixed."
-	OPENSLIDES_BACKEND_RUN_ALL_TESTS=1 pytest
-	# TODO: Run tests inside a container.
+all:
+	$(MAKE) run-cleanup
```
normanjaeckel, line 4: Please revert these changes and get my make targets back. I use them every time.

Makefile
normanjaeckel: You can change "run-tests" as you like and also with build dependency.

cleanup.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
```
normanjaeckel, line 1: Please add this single use commands to your run-cleanup make target so you can use them inside the container and I can use my former make targets outside.

normanjaeckel:
Of course I to not want to stress you but I think the make targets can be made clean for both uses (inside and outside container).

FinnStutzenstein:
> The authentification tests all fail

I cannot reproduce it. They do all work for me. Two possibilities: You have old containers or there is something depending on external things (which won't be good...).

Dockerfile-dev
jsangmeister: Because you mount the directories in the makefile with the `-v` switch. `docker run -v foo:/srv/code/bar` automatically creates the folder `/srv/code/bar` in your container and mounts the folder `foo` there.

Makefile
jsangmeister: Mmh, I think we should be strict about only working inside the docker container. Would it be fine for you if the commands get executed inside the container?

cleanup.sh
jsangmeister: See other comment for the makefile. Regardless of that I would like to have this bash file since I use that all the time when I'm in the docker container ;)

jsangmeister:
> I cannot reproduce it. They do all work for me. Two possibilities: You have old containers or there is something depending on external things (which won't be good...).

The test binds to ports. Since we use `--network=host`, if the port is already bound on the host machine, the tests will fail. I was still running an old instance of the auth container in the background, so they failed. I do not think this should be this way.

jsangmeister:
Fixed you remarks. @normanjaeckel As stated in a comment above, I added your black etc. commands in the same fashion as the `run-cleanup`, so you can use them alle as you did before, but they now execute inside the container. I hope this suits your needs.

jsangmeister:
Merging it for now to have it available. If you have any other remarks, we can fix them in a seperate PR.

normanjaeckel:
Sorry but this I can not accept. Please revert this pull request. I so not want to develop inside the docker container. If you do this is ok for me but we have to find a cooperative way for all of us.

Makefile
normanjaeckel: No.

normanjaeckel:
#88 should be acceptable for all of us.

PR #85, opened 2020-08-07
Better testing

jsangmeister:
Setting up on #76, I tried to improve the integration tests. I'm mocking the `retrieve` method of the datastore since that's the connecting point between the backend layer and the raw http response. If this looks good, I can move it up a layer to use it in all tests.

@FinnStutzenstein I noticed while doing this that the integration and unit tests of the presenters are extremely similar since they only consist of one function. Should we cut one of them and focus on the other?


normanjaeckel:
Please rebase this pull request on top of master so we can see the real changes better.

normanjaeckel:
I think this kind of patching won't work for actions because we have multiple datastore calls during one action request.

jsangmeister:
Closing in favor of #95.

PR #87, opened 2020-08-07
Added reverse schemas.

normanjaeckel:


normanjaeckel:
Please merge this and set your meeting settings pull request on top of it.

(I know that there is some double code here but this was for me an easy way to fix the problem and show how it could work.)

PR #88, opened 2020-08-07
Get back dev tools for using without docker.

normanjaeckel:


jsangmeister:
And please adjust the docs, forgot to do that in my PR. The default way (and the way described in the docs) should be via docker.

jsangmeister:
Done.

PR #94, opened 2020-08-12
Add motion statute paragraph action set and tests.

reiterl:
Add the action set and some tests.

### DIFF
--- a/openslides_backend/action/motion_statute_paragraph/create_update_delete.py
+++ b/openslides_backend/action/motion_statute_paragraph/create_update_delete.py
@@ -1,0 +1,41 @@
+from typing import Any, Dict
+
+from ...models.motion_statute_paragraph import MotionStatuteParagraph
+from ..action import register_action_set
+from ..action_set import ActionSet
+from ..default_schema import DefaultSchema
+from ..generics import CreateAction, DeleteAction, UpdateAction
+
+
+class MotionStatuteParagraphCreateAction(CreateAction):
+    """
+    Create action to set weight default in update_instance.
+    """
+
+    def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
+        """
+        set default for weight.
+        """
+        instance["weight"] = instance.get("weight", 0)
+        return instance
+
+
+@register_action_set("motion_statute_paragraph")
+class MotionStatuteParagraphActionSet(ActionSet):
+    """
+    Actions to create, update and delete motion statute paragraph.
+    """
+
+    model = MotionStatuteParagraph()
+    create_schema = DefaultSchema(MotionStatuteParagraph()).get_create_schema(
+        properties=["title", "text"], required_properties=["title"],
+    )
+    update_schema = DefaultSchema(MotionStatuteParagraph()).get_update_schema(
+        properties=["title"]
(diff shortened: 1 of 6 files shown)


openslides_backend/models/motion_statute_paragraph.py
```diff
@@ -0,0 +1,29 @@
...
+    meeting_id = fields.RequiredForeignKeyField(
+        description="The id of the meeting of this motion statute paragraph.",
+        to=Collection("meeting"),
+        related_name="motion_statute_paragraph_ids",
+    )
+    title = fields.RequiredCharField(
+        description="The title of this motion statute paragraph."
+    )
+    text = fields.TextField(description="The text of this statute paragraph.")
+    weight = fields.PositiveIntegerField(
```
normanjaeckel, line 27: Weight can be negative. Please use `IntegerField`.

openslides_backend/models/motion_statute_paragraph.py
```diff
@@ -0,0 +1,29 @@
...
+from . import fields
+from .base import Model
+
+
+class MotionStatuteParagraph(Model):
+    """
+    Model for motion statute paragraph.
+
+    Reverse fields:
+    motion_ids
```
normanjaeckel, line 11: Please write:
```
Reverse fields
    motion_ids: (motion/statute_paragraph_id)[];
```

openslides_backend/action/motion_statute_paragraph/create_update_delete.py
```diff
@@ -0,0 +1,20 @@
...
+    """
+
+    model = MotionStatuteParagraph()
+    create_schema = DefaultSchema(MotionStatuteParagraph()).get_create_schema(
+        properties=["title", "text"], required_properties=["title"],
+    )
+    update_schema = DefaultSchema(MotionStatuteParagraph()).get_update_schema(
+        properties=["title"]
+    )
+    delete_schema = DefaultSchema(MotionStatuteParagraph()).get_delete_schema()
```
normanjaeckel, line 20: Please see https://github.com/OpenSlides/openslides-backend/blob/master/openslides_backend/action/agenda_item/create.py#L58 how to set the default for weight field in create action. If this does not fit with the ActionSet because you can not patch the correct `update_instance` method (only for create and not for update or delete), we should add some small logic to the model fields to set the default overthere.

normanjaeckel (changes requested):
Some remarks that may help you. Thank you very much.

openslides_backend/models/motion_statute_paragraph.py
reiterl: Done. Changes into IntegerField.

openslides_backend/models/motion_statute_paragraph.py
reiterl: Done, changed the notation here.

openslides_backend/action/motion_statute_paragraph/create_update_delete.py
reiterl: Created a Create-Action, which sets the weight default and add it to the action set.

normanjaeckel (approved):

PR #95, opened 2020-08-12
Restructure tests and add first system test for agenda item

jsangmeister:
First version of the system tests. Runs with the full setup, so you can just write data to the datastore and then verify that the backend does the right things. Will restructure it to move everything to a base class so creating and verifying models is as easy as can be.

For some reason, I cannot reach reader and writer via localhost, but instead via their container names.

Also, @normanjaeckel: I would like to partition the tests into subfolders `unit` and `system` (and maybe `integration`). Currently, the test files are really clunky and overloaded, this would be improved with better seperation.


jsangmeister:
It would be nice to implement this quickly so the new actions can directly adjust to the new test schema.

FinnStutzenstein:
> TODO: Travis has to be adjusted to use docker for this to work.

We should generally disable travice and use GH actions.

Does `make run-tests` checks the formatting? You should be able to copy the GH action config from the client. There it is all configured to use `make run-tests`.

jsangmeister:
> > TODO: Travis has to be adjusted to use docker for this to work.
> 
> We should generally disable travice and use GH actions.
> 
> Does `make run-tests` checks the formatting? You should be able to copy the GH action config from the client. There it is all configured to use `make run-tests`.

Done in #100

jsangmeister:
Merging this tomorrow.

tests/system/test_general.py
```diff
@@ -0,0 +1,52 @@
...
+        )
+        self.assertEqual(response.status_code, 400)
+        self.assertIn(
+            "data[0] must contain [\\'action\\', \\'data\\'] properties",
+            str(response.data),
+        )
+
+    def test_request_no_existing_action(self) -> None:
+        response = self.client.post(
+            "/", json=[{"action": "fuzzy_action_hamzaeNg4a", "data": [{}]}],
```
FinnStutzenstein, line 42: Why is `/` the action service? How do you differentiate form presenters? This can be done in another PR.

Makefile
```diff
@@ -56,13 +56,28 @@ run-mypy: | build-dev
...
-	docker run $(dev_args) sh -c "OPENSLIDES_BACKEND_RUN_ALL_TESTS=1 pytest"
 
-run-tests-fast: | build-dev
-	docker run $(dev_args) pytest
+# compose commands
+
+run-dev-compose: | build-dev
+	docker-compose -f docker-compose-dev.yml up -d
+
+run-tests: | run-dev-compose
```
FinnStutzenstein, line 67: This does not stop the compose setup, does it?

This target and `run-tests-cov` should stop it afterwards. For running pytest over and over again you can use `run-compose-bash`.

docker-compose-dev.yml
```diff
@@ -1,9 +1,70 @@
 version: "3"
```
FinnStutzenstein, line 1: Please rename this file to `docker-compose.dev.yml`.

FinnStutzenstein (approved):
Just little things to change.

tests/system/test_general.py
jsangmeister: Differentiation between actions and presenters is done in the Initialization of the application (depending on the view given [here](https://github.com/OpenSlides/openslides-backend/blob/8842bf1fdd0f1852aeeaae1e14a65d6071802a92/tests/system/util.py#L48). The routes are just `/` for both. I noticed that as well, but didn't have the time yet to look into it, since we said every route should have the prefix `/internal/<service>`, but on the other side, routing is handled by the haproxy anyway. Create an issue if this needs to be changed.

Makefile
jsangmeister: done

docker-compose-dev.yml
jsangmeister: done

Makefile
jsangmeister: Correction: If I add it to `make run-tests-cov-no-tty` (the command that gets executed in the GH actions), the first few tests fail for some reason. Leaving it out for now.

tests/system/test_general.py
FinnStutzenstein: This is no problem, if everything is public. I was just curios how this can be actions, if nowhere in this file the word `action` is used (except for some teststrings). EIther there is some magic (but even the file location does not hint that there are actions, not presenters), or some base-class name is not well choosen. Please have a look and make it clear, why here are actions and not presenters. Also: If this is action-specific, this should not be named `test_general` without any hint for actions.

jsangmeister: Good catch, I somehow even threw away all presenter tests... Well, they are back now and I moved the base classes/tests to the right folders.

jsangmeister:
Merging this now. @normanjaeckel if you want, you can double check that I did not delete some important tests.

normanjaeckel:
The "old" tests are now disabled because by using a MagicMock the do not really test anything.

jsangmeister:
Yes, I know. They are not even executed since the file names do not begin with `test`. This is intentional, I will adopt them to the new schema this week, I just wanted this to be merged so the new actions could be implemented with the new tests.

PR #98, opened 2020-08-12
Add motion block action set.

reiterl:
Another action set. Not complete model. Add tests.

### DIFF
--- a/openslides_backend/models/motion_block.py
+++ b/openslides_backend/models/motion_block.py
@@ -1,0 +1,37 @@
+from ..shared.patterns import Collection
+from . import fields
+from .base import Model
+
+
+class MotionBlock(Model):
+    """
+    Model for motion block.
+
+    Reverse fields:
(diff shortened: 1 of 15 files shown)


openslides_backend/models/motion_block.py
```diff
@@ -0,0 +1,25 @@
...
+    verbose_name = "motion_block"
+
+    id = fields.IdField(description="The id of this motion block.")
+    meeting_id = fields.RequiredForeignKeyField(
+        description="The id of the meeting of this motion block.",
+        to=Collection("meeting"),
+        related_name="motion_block_ids",
+    )
+    title = fields.RequiredCharField(description="The title of this motion block.")
+    # TODO internal: boolean;
```
jsangmeister, line 24: Boolean fields added in #101.

reiterl: Merged #101 and add the internal field.

PR #100, opened 2020-08-13
Move travis to github actions

jsangmeister:
Travis has to be deactivated in order to merge this.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,15 +5,30 @@ on:
     branches: [ master ]
 
 jobs:
-  build:
+  build-and-test-dev-image:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
-    # repair the tests
-    #- uses: actions/setup-python@v2
-    
-    #- name: pytest
-    #  run: make test
-        
-    - name: docker build
+
+    - name: create docker dev image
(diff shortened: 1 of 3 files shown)


FinnStutzenstein:
Does `make run-tests` executes everything: `pytest` and all format checking tools?

jsangmeister:
> Does `make run-tests` executes everything: `pytest` and all format checking tools?

No, currently only `pytest` is executed, similar to the datastore.

FinnStutzenstein:
> No, currently only `pytest` is executed, similar to the datastore.

I think this also should be done (after `pytest`), so `make run-tests` tests everything.

jsangmeister:
> > No, currently only `pytest` is executed, similar to the datastore.
> 
> I think this also should be done (after `pytest`), so `make run-tests` tests everything.

Sure, can do. Should I also adjust the datastore then?

jsangmeister:
On second thought: If you have `make run-tests` execute everything, it will terminate on the first error, which is typically not what you want in a CI, so we have to have explicit commands anyway and these are the ones getting exectuted by github actions. So when would you even use the all-in-one command?

jsangmeister:
Merging this now; create an issue if you want `run-tests` to execute everything so we can discuss it there

PR #101, opened 2020-08-13
Restructure the fields for less code duplication and add some fields

jsangmeister:
Copied the changes to fields from #75 to make them available now.


jsangmeister:
Merging it directly, should be fine since already reviewed. If you have any other remarks, I will implement them in a seperate PR.

Issue #90, opened 2020-08-11
Change motion.(create|update|delete) to ActionSet


FinnStutzenstein:
Currently, motion has seperate actions for create, update and delete. They should be changed to the new ActionSet.

reiterl:
I looked at it. I don't know a way to add code to an action from an action_set.

jsangmeister:
Would probably need to be injected in the `get_actions` method. Question is, is this worth the effort or should we stick with seperate actions?

FinnStutzenstein:
Maybe @normanjaeckel explains the reason he implemented action sets. I currently do not see the advantage of creating logically dependencies between actions, that can live by their own.

reiterl:
Ah, I have found the hook to include code into an action of an action set.
The action set saves some lines code, but adds some complexity.

reiterl:
Add pull request #102 which include the action set to motion actions.

normanjaeckel:
I implemented ActionSets to provide a way of simple Create-Update-Delete actions without any extra stuff. I also provide a hook I someone wants to hack in but it was not my intention that every model has to have an action set. E. g. for motion if would be ok for me not to have an ActionSet and handel every (complex) action in their own class.

jsangmeister:
Okay, then this was my misunderstanding, but I came to the same conclusion with @FinnStutzenstein.

FinnStutzenstein:
Yes, we were afraid of highly coupling, that in some scenarios just doesn't make sense. For "stupid" models it is fine though.

PR #102, opened 2020-08-14
Change the motion actions into an action set.

reiterl:
Change the motion create, update, delete into an action_set, uses the action code in the action_set.


jsangmeister:
Closing this, see #90

PR #107, opened 2020-08-20
Fix handling of reverse schemata

jsangmeister:
fixes the errors encountered in #75. May also fix https://github.com/OpenSlides-action-crew/openslides-backend-coordination/issues/75, please test @GabrielInTheWorld

### DIFF
--- a/openslides_backend/action/relations.py
+++ b/openslides_backend/action/relations.py
@@ -251,7 +251,7 @@ class RelationsHandler:
         for rel_id, rel in sorted(rels.items(), key=lambda item: str(item[0])):
             new_value: Optional[Union[int, List[int]]]
             if rel_id in add:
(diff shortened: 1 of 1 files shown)


jsangmeister:
@normanjaeckel did you silenty accept or miss this? :D anyway, I would like your review on this. Is this right and you missed it when implementing reverse relations?

normanjaeckel:
In lines 63 to 65 i switch the type so your change should not be required. But maybe there is a misstake and we have to use `self.type` and not `self.field.type`.

jsangmeister:
Well, It didn't work before and now it works :D Maybe you want to look into it where the mistake lays? The relation handling is critical for the backend and I don't know enough about your code to debug this.

normanjaeckel:
Do you have a test or test szenario for me for the error case?

PR #110, opened 2020-08-20
Enhance system tests & fix agenda_item tests

jsangmeister:


normanjaeckel:
@jsangmeister Please check if this pull request is still valid according to current master.

PR #116, opened 2020-08-28
Changed dev tools.

normanjaeckel:


.github/workflows/continuous_integration.yml
```diff
@@ -0,0 +1,62 @@
...
+    - uses: actions/checkout@v2
+
+    - name: Create docker production image
+      run: docker build . --tag=openslides-backend
+
+    - name: Run docker production image
+      run: docker run --detach --publish 9002:9002 --publish 9003:9003 openslides-backend
+
+    - name: Sleep for some seconds
+      run: sleep 2
```
FinnStutzenstein, line 21: Can this be removed? it is very arbitrary. If you want to wait for your services to be online, use another tool.

Makefile
```diff
@@ -27,85 +30,26 @@ pip-check:
...
 
-run-prod:
+run-prod: | build-prod
 	docker run --interactive --tty \
 	--publish 9002:9002 --publish 9003:9003 --rm openslides-backend
 
 
-# Build and run development and testing docker container
-
-build-dev:
```
FinnStutzenstein, line 39: We need `build-dev`. It is required in the big setup.

Makefile
```diff
@@ -27,85 +30,26 @@ pip-check:
...
+	docker-compose -f dev/docker-compose.dev.yml down --volumes
 
-ci-isort-check: | build-dev
-	docker run $(action_args) isort --check-only --diff openslides_backend/ tests/
+start-dev-interactive:
+	docker-compose -f dev/docker-compose.dev.yml up  --build
 
-ci-flake8-check: | build-dev
-	docker run $(action_args) flake8 openslides_backend/ tests/
+run-dev run-bash:
```
FinnStutzenstein, line 51: There should be a dependency to start the containers. I cannot do `run-dev` stand alone anymore.

Makefile
```diff
@@ -27,85 +30,26 @@ pip-check:
...
-	--network=host --rm openslides-backend-dev
-dev_args=--tty $(action_args)
-
-run-dev: | build-dev
-	docker run $(dev_args)
-
-run-dev-interactive run-bash: | build-dev
-	docker run $(dev_args) sh
-
-run-cleanup: | build-dev
```
FinnStutzenstein, line 56: Can you add it back? We must have a make target to run the cleanup scripts inside the containers without starting everything (e.g. for ludwig - he does not have python installed).

dev/Dockerfile-dev
```diff
@@ -0,0 +1,18 @@
+FROM python:3.8.1-alpine3.11
+
+RUN apk add --no-cache gcc musl-dev linux-headers make curl git
+
+WORKDIR /srv/code
+COPY requirements.txt .
+COPY setup.cfg .
+COPY Makefile Makefile
```
jsangmeister, line 8: Mmh, I'm not a huge fan of this since many of the commands don't work any more. As said on the phone, we don't really need it in the container, either (since we have the `cleanup.sh`). Is there a specific reason you want the Makefile available in the container? If not, I would suggest to remove it. If yes, Maybe you can split the Makefile into the parts that are functional inside the container and those who are not? Maybe have a look at [include](https://www.gnu.org/software/make/manual/html_node/Include.html), I think with this it should be possible.

jsangmeister (changes requested):

.github/workflows/continuous_integration.yml
normanjaeckel: I will open an issue so that we can insert the usage of another tool later. I do not know a smart solution at the moment.

Makefile
normanjaeckel: Please explain what the "big setup" is. I thought we use everything together with docker compose using the production Dockerfile and not the development version (as we do in the new systems tests here concerning the datastore).

Makefile
normanjaeckel: I understood that you do not run stand alone dev container because you always use docker compose dev setup.

Makefile
normanjaeckel: Inside the container (started with docker compose) you can just type `make` to get the cleanup. `make run-cleanup` is not required any more.

Makefile
```diff
@@ -17,6 +17,9 @@ mypy:
 test:
 	pytest
 
+test-incomplete:
```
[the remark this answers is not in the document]
normanjaeckel: Ok, I will changed this.

.github/workflows/continuous_integration.yml
```diff
@@ -0,0 +1,62 @@
...
+
+jobs:
+  build-production-image:
+    name: Build and test production docker image
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Create docker production image
+      run: docker build . --tag=openslides-backend
```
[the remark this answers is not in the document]
normanjaeckel: It would be the first and last make command used in the GitHub Workflow and is less explicit but I can changed if you really really insist.

README.md
```diff
@@ -2,19 +2,49 @@
...
+    $ make run-debug
+
+To start it run
+
+    $ OPENSLIDES_BACKEND_DEBUG=1 python -m openslides_backend
+
+or just
 
-    $ make run-prod
+    $ make run-debug
```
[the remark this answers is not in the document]
normanjaeckel: Ok, I will fix this.

.github/workflows/continuous_integration.yml
```diff
@@ -0,0 +1,62 @@
...
+      run: black --check --diff openslides_backend/ tests/
+
+    - name: Check isort
+      run: isort --check-only --diff openslides_backend/ tests/
+
+    - name: Check flake8
+      run: flake8 openslides_backend/ tests/
+
+    - name: Check mypy
+      run: mypy openslides_backend/ tests/
```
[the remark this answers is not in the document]
normanjaeckel: I understand and I will fix this although this looks a little bit strange ...

dev/run-tests.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+
+DC="docker-compose -f dev/docker-compose.dev.yml"
+
+$DC up --detach --build
+$DC exec -T backend dev/wait.sh writer 9011
+$DC exec -T backend dev/wait.sh reader 9010
+$DC exec -T backend pytest --cov --cov-fail-under=86
```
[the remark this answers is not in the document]
normanjaeckel: Ok, I will fix this.

dev/Dockerfile-dev
normanjaeckel: I understand you idea but I strongly recommand to provide the Makefile inside the dev environment so everybody can use `make black` or `make run-debug`. Of course some commands won`t work but I can add a hint in a comment in the Makefile. These commands do not harm there. Splitting up is less explicit so I prefer not to do so.

normanjaeckel:
@FinnStutzenstein @jsangmeister Thank you very much for your constructive review. I will change the mentioned points and the merge it and also change the "Required" flag in GitHub settings. Possibly missing points can be added later of course.

.github/workflows/continuous_integration.yml
FinnStutzenstein: The question is: Why did you add it? What goes wrong when you leave it out?

Makefile
FinnStutzenstein: The dev-mode in the main repo required `build-dev`.

Makefile
FinnStutzenstein: > because you always use docker compose dev setup.

Yes, but if I do `make run-bash`, the setup isn't started and I get an error.

Makefile
FinnStutzenstein: Ok. This is just a comfy-thing for me personally. Sometimes I do not want to start everything to run cleanup ;)

Makefile
normanjaeckel: Ok I got it. But then it is more difficult to start the docker compose setup in an extra terminal and run everything with an extra command. But I can change this.

Makefile
normanjaeckel: ;-)

Ok I will provide you a private script in the dev directory ...

.github/workflows/continuous_integration.yml
normanjaeckel: The previous command finishs before the server is really ready. The next command comes to quickly.

FinnStutzenstein: You can use `wait.sh <host> <port>` for this case.

.github/workflows/continuous_integration.yml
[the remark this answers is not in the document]
jsangmeister: I don't see a problem with using make commands in the workflow and I think `make build-prod` is about as explicit as it gets ;) but it's not that important, just cleaner IMO

PR #120, opened 2020-09-08
Install auth-lib

GabrielInTheWorld:


normanjaeckel:
Please check Continuous Integration. Maybe rebase and push force to restart the run.

normanjaeckel:
@GabrielInTheWorld ping

normanjaeckel:
@GabrielInTheWorld @FinnStutzenstein @jsangmeister What is the status of this pull request?

jsangmeister:
@GabrielInTheWorld ...

dev/Dockerfile-dev
```diff
@@ -1,6 +1,6 @@
 FROM python:3.8.5-slim-buster
 
-RUN apt-get update && apt-get install --yes make git curl ncat
+RUN apt-get update && apt-get install --yes make git curl ncat vim
```
FinnStutzenstein, line 3: Debugging inside the lib in the container? ;)

openslides_backend/services/auth/adapter.py
```diff
@@ -1,87 +1,49 @@
...
-                )
-            access_token = response.headers.get(AUTHENTICATION_HEADER, None)
-        return user_id, access_token
+    def hash(self, toHash: str) -> str:
+        return self.auth_handler.hash(toHash)
+
+    def is_equals(self, hash: str, toCompare: str) -> bool:
+        return self.auth_handler.is_equals(hash, toCompare)
 
     def auth_is_down(self) -> int:
```
FinnStutzenstein, line 44: Can this be removed?

dev/Dockerfile-dev
jsangmeister: No, the static files like `setup.cfg` aren't mounted, but copied instead, and it bothered me that I had to restart the container for every change :D

FinnStutzenstein (approved):

dev/Dockerfile-dev
FinnStutzenstein: I would propose to mount them; but since they do not change often, it is ok to just install vim ;)

jsangmeister: We'd need to put them into a seperate directory then which we can mount and point all programs to this dir to look for the files instead of the base dir etc. It would mean a lot of restructuring

PR #137, opened 2020-09-18
Update motion_workflow delete action: check for default workflow.

reiterl:
Add check to default workflow into delete action of motion_workflow.
Add some tests.

### DIFF
--- a/openslides_backend/action/motion_workflow/create_update_delete.py
+++ b/openslides_backend/action/motion_workflow/create_update_delete.py
@@ -1,20 +1,69 @@
+from typing import Any, Dict
+
 from ...models.motion_workflow import MotionWorkflow
-from ..action import register_action_set
-from ..action_set import ActionSet
+from ...shared.exceptions import ActionException
+from ...shared.patterns import Collection, FullQualifiedId
+from ..action import register_action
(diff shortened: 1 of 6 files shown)


openslides_backend/action/motion_workflow/create_update_delete.py
```diff
@@ -1,20 +1,64 @@
...
+            ["meeting_id"],
+        )
+        meeting = self.fetch_model(
+            FullQualifiedId(Collection("meeting"), int(workflow["meeting_id"])),
+            [
+                "motions_default_workflow_id",
+                "motions_default_statute_amendment_workflow_id",
+            ],
+        )
+        if instance["id"] == meeting["motions_default_workflow_id"]:
```
normanjaeckel, line 60: there is a third one: "motions_default_amendment_workflow_id"

normanjaeckel (approved):
Just add the third case. Thanks.

openslides_backend/action/motion_workflow/create_update_delete.py
reiterl: Done, add the third default check, too. And add a test.

jsangmeister (approved):

PR #140, opened 2020-09-21
Move TreeSortMixin and vars into sort_generic.py

reiterl:
Move TreeSortMixin to sort_generic.

### DIFF
--- a/openslides_backend/action/sort_generic.py
+++ b/openslides_backend/action/sort_generic.py
@@ -1,0 +1,143 @@
+from typing import Any, Dict, Iterable, List, Set
+
+import fastjsonschema
+
+from ..models.base import Model
+from ..shared.exceptions import ActionException
+from ..shared.filters import FilterOperator
+from ..shared.interfaces import Event, WriteRequestElement
+from ..shared.patterns import FullQualifiedId
+from ..shared.schema import schema_version
+from .base import BaseAction, DataSet
+
+sort_node_schema = {
+    "$schema": schema_version,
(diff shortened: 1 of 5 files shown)


reiterl:
I tried to add simple test logic, but ran into problems.

reiterl:
Add a fix to the datastore "id as str" problem in tree_sort.
Add some system tests for motion sort.

openslides_backend/action/sort_generic.py
```diff
@@ -0,0 +1,143 @@
...
+
+        # Get all item ids to verify, that the user send all ids.
+        filter = FilterOperator("meeting_id", "=", meeting_id)
+        db_instances = self.database.filter(
+            collection=self.model.collection,
+            filter=filter,
+            mapped_fields=["id"],
+            lock_result=True,
+        )
+        all_model_ids = set(map(int, db_instances.keys()))
```
jsangmeister, line 70: Please move this fix to the datastore adapter since it will effect all actions which use the filter method.

jsangmeister (changes requested):
Looks good! Please add another test which tries to create a circle to have that covered as well. It kinda doubles with the `test_sort_not_a_tree` test, but thats okay.

openslides_backend/action/sort_generic.py
reiterl: Add a circle test and move the int fix to the datastore adapter.
Updated the tests.

jsangmeister (approved):

Issue #115, opened 2020-08-28
Tests for relation handling


jsangmeister:
While restructuring the tests, I noticed that there are no tests for the relation handling. This vital part of the backend should be tested thoroughly to prevent errors like #107 (whatever the error there is). @normanjaeckel will you do that? As said, I'm not deep enough into the code currently to do this. Depending on the detail level, unit and/or integration tests are both suited.

PR #144, opened 2020-09-24
Add relations tests

jsangmeister:


jsangmeister:
fixes #115

jsangmeister:
@normanjaeckel the PR contains two `xtests` which are commented out because they don't work the way I understood it. Take for example `xtest_O2O_replace`. What do you do if you have a case like this and how does the relation handler handle this?

normanjaeckel:
I have to make a deep look into this test. Please merge this pull request and open an issue for me.

PR #147, opened 2020-09-24
Add Decimal & Html field

jsangmeister:
Also add automatic html validation.

### DIFF
--- a/openslides_backend/action/base.py
+++ b/openslides_backend/action/base.py
@@ -111,6 +111,19 @@ class Action(BaseAction, metaclass=SchemaProvider):
         """
         return instance
 
+    def validate_fields(self, instance: Dict[str, Any]) -> Dict[str, Any]:
+        """
+        Validate all model fields according to the model definition.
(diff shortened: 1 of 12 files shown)


jsangmeister:
Tag list and validation style copied from OS3.

normanjaeckel:
Please wait with this PR until we have the new calculated models.py.

normanjaeckel:
The easiest solution I think is to use two fields: HTMLStrictField and HTMLPermissiveField. Therefor we have to change the types in the new models.yml and adjust the `dev/generate_models.py` (all after #153 is merged).

If we need more flexibility we can do this later.

jsangmeister:
Why do we generate the models now? I thought we agreed in our phone call that it's not necessary since we have most of the models finished already?

jsangmeister:
> Please wait with this PR until we have the new calculated models.py.

Merging this now so it is available for client work. The changes you mentioned can also be done in your PR if needed.

PR #150, opened 2020-09-24
Reorder Dockerfile-dev

jsangmeister:
The problem with running `make` inside the docker container is that then docker takes ownership of all edited files, so I want to have the cleanup script back on root level for easier access. Also, I recently changed some things in the `setup.cfg`, which led to the pip packages being installed every time, so I moved it behind the requirements section.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,18 +40,18 @@ run-prod: | build-prod
 # Build and run development docker container setup with docker compose (not usable inside docker container)
 
 start-dev:
-	docker-compose -f dev/docker-compose.dev.yml up --build --detach
+	USER_ID=$$(id -u $${USER}) GROUP_ID=$$(id -g $${USER}) docker-compose -f dev/docker-compose.dev.yml up --build --detach
 
 stop-dev:
 	docker-compose -f dev/docker-compose.dev.yml down --volumes
 
 start-dev-interactive:
-	docker-compose -f dev/docker-compose.dev.yml up  --build
+	USER_ID=$$(id -u $${USER}) GROUP_ID=$$(id -g $${USER}) docker-compose -f dev/docker-compose.dev.yml up --build
 
-run-dev-standalone:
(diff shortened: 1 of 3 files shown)


normanjaeckel:
Looks ok but why dont you just copy the complete dev dir?

Merge it if you like.

normanjaeckel:
Hm. Just for my interest: If you run `make` inside the docker container docker gets control of the changed files, right? Does this also happen if you run your script inside the container or is this just a make-problem? Outside the container you do not need your script because you can just run `make` ;-)

FinnStutzenstein:
https://medium.com/@mccode/understanding-how-uid-and-gid-work-in-docker-containers-c37a01d01cf

I had problems with cleanup scripts in OS3 and the university project. Since then I use this technique:

https://github.com/FinnStutzenstein/OpenSlides/blob/externalAutoupdateService/dc.dev.yml#L13

and 

https://github.com/FinnStutzenstein/OpenSlides/blob/externalAutoupdateService/Makefile#L7

does the trick. You run the same uid:gid in the container as in the host, so the permissions are correct in both environments.

jsangmeister:
> Looks ok but why dont you just copy the complete dev dir?

Because I want to copy different files at different points in time

jsangmeister:
@FinnStutzenstein Ah, I looked shortly into that but thought it would be more hacky than your solution actually is. Will try and implement it here later.

FinnStutzenstein:
> but thought it would be more hacky than your solution actually is

Yes, I thought that too (Also since I never worked with unix permissions that much it was like magic). But since it is not a VM (containers run on the host!) it is just a number to set right ;)

jsangmeister:
username now shows as `I have no name!` but otherwise it's working, thanks.

FinnStutzenstein:
> username now shows as `I have no name!` but otherwise it's working, thanks.

This is correct! (And a nice "easter egg" IMO)

PR #153, opened 2020-09-24
Generated models

normanjaeckel:
This PR is not ready yet. I have to fix relations and tests but you can already see what this is about ...


normanjaeckel:
I hope everything works fine with this. We now have all models in a generated file.

Sorry if some of the open pull request have to be adopted but I think we can save some work in the future with this.

jsangmeister:
Why do we generate the models now? I thought we agreed in our phone call that it's not necessary since we have most of the models finished already? As said before, I'm not a fan of generated code.

openslides_backend/models/models.py
```diff
@@ -0,0 +1,1493 @@
...
+class MotionSubmitter(Model):
+    collection = Collection("motion_submitter")
+    verbose_name = "motion submitter"
+
+    id = fields.IntegerField()
+    weight = fields.IntegerField(constraints={"default": 0})
+    user_id = fields.RelationField(
+        to=Collection("user"),
+        related_name="submitted_motion_$_ids",
+        structured_relation=["meeting_id"],
```
jsangmeister, line 831: This doesn't work since the `meeting_id` doesn't exist as a field

jsangmeister (changes requested):
I didn't look into the generation code. I have no idea what you mean by `replacement` or `index` in the structured relations. When fixing things on the models, one cannot simply make the change, but has to adjust the whole generation code instead (and, more importantly, understand it!). We generate much more work with this than we safe IMO...

normanjaeckel:
I remember that you said that you are not a fan of generated code but that for our case we sould do this so we do not have to define model definitions twice (in backend and also for autoupdate/restricter).

openslides_backend/models/models.py
normanjaeckel: We have to add e `through=[..]` keyword which is not implemented yet.

dev/generate_models.py
```diff
@@ -0,0 +1,310 @@
...
+
+COMMON_FIELD_CLASSES = {
+    "string": "CharField",
+    "number": "IntegerField",
+    "boolean": "BooleanField",
+    "JSON": "JSONField",
+    "HTML": "HTMLField",
+    "float": "FloatField",
+    "decimal(6)": "DecimalField",
+    "datetime": "DatetimeField",
```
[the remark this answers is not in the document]
normanjaeckel: This has to be done in the main repo in the models.yml. Then we can adjust/rename it here.

dev/generate_models.py
```diff
@@ -0,0 +1,310 @@
...
+)
+
+COMMON_FIELD_CLASSES = {
+    "string": "CharField",
+    "number": "IntegerField",
+    "boolean": "BooleanField",
+    "JSON": "JSONField",
+    "HTML": "HTMLField",
+    "float": "FloatField",
+    "decimal(6)": "DecimalField",
```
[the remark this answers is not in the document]
normanjaeckel: This is a question of `models.yml` in main repo. There we have both fields.

normanjaeckel:
I merge this now so new work can rebase on top of it. I will change the URI for the models.yml after my last fixes are merged into the main repo.

openslides_backend/models/models.py
normanjaeckel: Should be done now.

PR #160, opened 2020-09-28
Add check_dependant_action_execution_agenda_item

reiterl:
Add some tests for the function.
It checks agenda_item_creation and agenda_create and calculates, if an agenda item needs to be created,


jsangmeister (approved):

normanjaeckel:
Looks good. Please rebase on top of master. Maybe the conflict can be solved just by using the correct import of the model from new models.py.

PR #163, opened 2020-09-29
Exclude old tests from workflow

jsangmeister:
Since you just deleted the sections from #141, I'll do it in a new PR. I don't want to fix syntax errors in old tests when working on new code, so we have to exclude the old tests from the normal workflow with cleanup & github actions. You can still explicitly fix them by calling the commands with the `old` folder if you want to.


jsangmeister:
Merging this directly since #162 depends on it

PR #164, opened 2020-09-29
Add defaults

jsangmeister:


FinnStutzenstein:
Do we want to include the defaults in the model description in the main repo?

jsangmeister:
> Do we want to include the defaults in the model description in the main repo?

Generally speaking yes, although these are the "default defaults" if you want to, so I think it's only necessary to note the defaults which differ from these in the main repo.

FinnStutzenstein:
Ah yes, I haven't looked into the code. Do you really want all number fields to have 0 as the default default? -> null should be fine, but if we want a default value, it should be expliitly noted in the model definition.

jsangmeister:
> Do you really want all number fields to have 0 as the default default?

No, I just thought about that too, also it breaks everything since all ids are default 1 then... Will move that to the model definition :D

PR #167, opened 2020-09-30
Update the sort actions to use arrays in data.

reiterl:
Try to fix https://github.com/OpenSlides-action-crew/openslides-backend-coordination/issues/57

### DIFF
--- a/openslides_backend/action/action.py
+++ b/openslides_backend/action/action.py
@@ -25,19 +25,11 @@ payload_schema = fastjsonschema.compile(
                     "minLength": 1,
                 },
                 "data": {
-                    "oneOf": [
-                        {
-                            "description": "Data for the action (array)",
-                            "type": "array",
-                            "items": {"type": "object"},
-                            "minItems": 1,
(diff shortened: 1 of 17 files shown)


normanjaeckel:
I think this would only be a hotfix. What if the client sends more then one item inside the array? To do this properly we have to loop over the items of the array and create multiple results.

I really recommend that we do NOT change the behavior here and let the client send an object instead of an array. This would be cleaner at the moment and means no more work here.

@FinnStutzenstein ?

jsangmeister:
Let's keep the discussion in https://github.com/OpenSlides-action-crew/openslides-backend-coordination/issues/57#issuecomment-701542581

normanjaeckel:
I will care of this PR and merge it.

normanjaeckel:
I adjusted the ActionPayload to be always an array of data. @reiterl Can you please fix the last tests here for two other actions that formerly required a dict? Then this PR can be merged.

jsangmeister:
Doing that now so this is finally finished

PR #174, opened 2020-10-01
Repair default values

jsangmeister:


jsangmeister:
I thought jsonschema would automatically insert the defaults, but that's not the case. So I implemented the behaviour manually.

Also mounted the model generation into the container to be able to execute it there and included it in the cleanup scripts.

normanjaeckel:
I disabled the automatism of jsonschema because we do not want default values in update requests.

cli/generate_models.py
```diff
@@ -254,6 +263,8 @@ def get_code_for_normal(self, field_name: str) -> str:
             properties += "required=True, "
         if self.read_only:
             properties += "read_only=True, "
+        if self.default is not None:
+            properties += f"default={self.default.__repr__()}, "
```
normanjaeckel, line 267: I uses json.dumps two lines later so you maybe want to use it here too.

normanjaeckel (approved):

PR #182, opened 2020-10-06
Add auth service interface

jsangmeister:


jsangmeister:
@FinnStutzenstein ping

FinnStutzenstein:
Looks good, but in the meantime Gabriel did finish the real one: https://github.com/OpenSlides/OpenSlides/blob/openslides4-dev/docs/interfaces/auth-service.txt#L144

There are two functions: `toHash` generates one. `isEquals` has to be used to check, if a PW is correct, giving the user input and the hashed password

FinnStutzenstein:
This would now add the auth service as another component to the dev/test setup. So I guess there is a bit more todo.

jsangmeister:
Adapter implementation has to be adjusted.

FinnStutzenstein:
Ok, the interface is bigger than needed, since login, ... are directly routed to the auth service. I think you just need authenticate, hash and isEqual. You can leave them there and delete them later, if not used.

I would prefer to merge it and see, that #120 is done. Gabriel has done more work in the library lately which is also already merged. Especially the three functions are included, so you might just need to rewrite the adapter class to convert your data representation to the library ones and convert exceptions. @jsangmeister Please ping me at thursday if you are working, so we can concentrate on getting auth-related sutff working.

PR #185, opened 2020-10-07
Add motion.reset_recommendation action.

reiterl:
Add tests.

### DIFF
--- a/openslides_backend/action/motion/reset_recommendation.py
+++ b/openslides_backend/action/motion/reset_recommendation.py
@@ -1,0 +1,23 @@
+from typing import Any, Dict
+
+from ...models.models import Motion
+from ..default_schema import DefaultSchema
+from ..generics import UpdateAction
+from ..register import register_action
+
+
+@register_action("motion.reset_recommendation")
+class MotionResetRecommendationAction(UpdateAction):
+    """
+    Reset motion recommendation action.
+    """
+
+    schema = DefaultSchema(Motion()).get_update_schema(properties=[])
+    model = Motion()
+
+    def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
+        """
+        Set recommendation to None.
+        """
+        instance["recommendation_id"] = None
+        return instance
--- a/tests/system/action/motion/test_reset_recommendation.py
+++ b/tests/system/action/motion/test_reset_recommendation.py
@@ -1,0 +1,51 @@
+from tests.system.action.base import BaseActionTestCase
+
+
+class MotionResetRecommendationActionTest(BaseActionTestCase):
+    def test_reset_recommendation_correct(self) -> None:
+        self.create_model("meeting/222", {"name": "name_SNLGsvIV"})
+        self.create_model(
+            "motion_state/77",
+            {"name": "test1", "motion_id": 22, "motion_recommendation_ids": [22]},
+        )
+        self.create_model(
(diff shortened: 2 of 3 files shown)


openslides_backend/action/motion/reset_recommendation.py
```diff
@@ -0,0 +1,38 @@
+from typing import Any, Dict
+
+from ...models.models import Motion
+from ...shared.schema import schema_version
+from ..generics import UpdateAction
+from ..register import register_action
+
+reset_schema = {
```
jsangmeister, line 8: Since all schemas now are arrays, we can generalize this further. But this doesn't have to be done here, if you don't want to, I can also do it in its own PR

openslides_backend/action/motion/reset_recommendation.py
```diff
@@ -0,0 +1,38 @@
...
+    "$schema": schema_version,
+    "title": "Reset motion recommendation schema",
+    "description": "An array of objects with motion id",
+    "type": "array",
+    "items": {
+        "type": "object",
+        "properties": {"id": {"type": "integer"}},
```
jsangmeister, line 15: But you should at least use `Motion().get_properties("id")` here so that we don't double any field definitions

tests/system/action/motion/test_reset_recommendation.py
```diff
@@ -0,0 +1,36 @@
...
+            "motion/22", {"meeting_id": 222, "title": "test1", "recommendation_id": 77}
+        )
+        response = self.client.post(
+            "/", json=[{"action": "motion.reset_recommendation", "data": [{"id": 22}]}]
+        )
+        self.assert_status_code(response, 200)
+        model = self.get_model("motion/22")
+        assert model.get("recommendation_id") is None
+
+    def test_reset_recommendation_correct_2(self) -> None:
```
jsangmeister, line 21: That is exactly the same test as the first one, or am I missing something here? I'm guessing c&p error. Anyway, a negative test is missing.

jsangmeister (changes requested):

openslides_backend/action/motion/reset_recommendation.py
jsangmeister: Update: see other comment.

openslides_backend/action/motion/reset_recommendation.py
jsangmeister: Nevermind, just had a look at it again and you can just use the default update schema without any properties since this is the exact same schema you're using here.

tests/system/action/motion/test_reset_recommendation.py
reiterl: No, it is not: In the first test the recommendation_id was set to a state, in the second, it was set to None.

openslides_backend/action/motion/reset_recommendation.py
reiterl: Okay

openslides_backend/action/motion/reset_recommendation.py
reiterl: Okay, use the DefaultSchema instead.

tests/system/action/motion/test_reset_recommendation.py
jsangmeister: Ah, okay, missed that.

jsangmeister (approved):

PR #189, opened 2020-10-08
Added some tests. Increased coverage. Simplified default schema.

jsangmeister:


jsangmeister:
Also fixes schema for https://github.com/OpenSlides-action-crew/openslides-backend-coordination/issues/74

openslides_backend/action/default_schema.py
```diff
@@ -13,20 +13,22 @@ class DefaultSchema:
...
             "type": "array",
             "items": {
                 "type": "object",
-                "properties": self.model.get_properties(*properties),
+                "properties": self.model.get_properties(
+                    *required_properties, *optional_properties
+                ),
                 "required": list(required_properties),
                 "additionalProperties": False,
             },
```
normanjaeckel, line 34: Please add `"minItems": 1` and `"uniqueItems": True` here.

openslides_backend/action/agenda_item/delete.py
```diff
@@ -11,4 +11,4 @@ class AgendaItemDelete(DeleteAction):
     """
 
     model = AgendaItem()
-    schema = DefaultSchema(AgendaItem()).get_delete_schema()
+    schema = DefaultSchema(AgendaItem()).get_update_schema()
```
normanjaeckel, line 14: I understand that a delete schema is the same as an update schema without properties. Nevertheless it is much more explicit if we call it `get_delete_schema` here. So we also can readd title and description of the schema.

openslides_backend/action/assignment/create_update_delete.py
```diff
@@ -8,10 +8,9 @@
 from ..generics import DeleteAction, UpdateAction
 from ..register import register_action_set
 
-create_schema = DefaultSchema(Assignment()).get_create_schema(
-    properties=[
-        "title",
-        "meeting_id",
+create_schema = DefaultSchema(Assignment()).get_default_schema(
```
normanjaeckel, line 11: Here we can use `get_create_schema` which may use `get_default_schema` internally but readd title and description.

openslides_backend/action/default_schema.py
```diff
@@ -13,20 +13,22 @@ class DefaultSchema:
...
+    def get_default_schema(
+        self,
+        required_properties: Iterable[str] = [],
+        optional_properties: Iterable[str] = [],
     ) -> Dict:
         """
         Returns a default create schema with properties and required properties as given.
         """
         return {
             "$schema": schema_version,
```
normanjaeckel, line 25: Please readd title and description for create, update and delete schemas. For default schema it can be setup by optional parameters as you do for the single item schema.

normanjaeckel (changes requested):

openslides_backend/action/default_schema.py
jsangmeister: It's already there, seems like github clipped it ;)

openslides_backend/action/agenda_item/delete.py
jsangmeister: Seems reasonable. Will add `create/update/delete` schema methods back

PR #190, opened 2020-10-08
Check agenda_item.parent_id

jsangmeister:
As mentioned in https://github.com/OpenSlides-action-crew/openslides-backend-coordination/issues/80, check that the parent_id belongs to the same meeting.

### DIFF
--- a/openslides_backend/action/agenda_item/create.py
+++ b/openslides_backend/action/agenda_item/create.py
@@ -44,4 +44,12 @@ class AgendaItemCreate(CreateAction):
         if not content_object.get("meeting_id"):
             raise ActionException("Given content object has no meeting id.")
         instance["meeting_id"] = content_object["meeting_id"]
+        # assert parent_id is valid
+        if "parent_id" in instance:
+            parent_item = self.fetch_model(
+                FullQualifiedId(self.model.collection, instance["parent_id"]),
+                ["meeting_id"],
+            )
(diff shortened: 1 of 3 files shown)


openslides_backend/action/agenda_item/create.py
```diff
@@ -44,4 +44,12 @@ def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
...
             raise ActionException("Given content object has no meeting id.")
         instance["meeting_id"] = content_object["meeting_id"]
+        # assert parent_id is valid
+        if "parent_id" in instance:
+            parent_item = self.fetch_model(
+                FullQualifiedId(self.model.collection, instance["parent_id"]),
+                ["meeting_id"],
+            )
+            if not parent_item.get("meeting_id") == instance["meeting_id"]:
+                raise ActionException("Given parent_id is invalid.")
```
normanjaeckel, line 54: Optional: Extend error message to tell the user why it is invalid (because the parent does not belong to this meeting ..)

normanjaeckel (approved):

openslides_backend/action/agenda_item/create.py
jsangmeister: I see where you're coming from, but these error messages are not meant for the user: a client should never send an invalid parent_id or the client has an error somewhere. So I'll leave it as-is for now.

PR #192, opened 2020-10-08
Add motion.set_state and tests.

reiterl:


normanjaeckel (dismissed):

jsangmeister (approved):

jsangmeister:
> State changing behavior for motion is missing (e. g. motion gets a number if it is switched to some states.)

I don't really know what you mean by this? Can you provide a more detailed example? Is it something that is configurable for the user, since the states are created by them, or do default behaviours exist? Anyway, I think this can be handled in a seperate PR.

PR #197, opened 2020-10-09
Fixed decimal implementation

jsangmeister:
@FinnStutzenstein it's correct that decimals are always positive, right?

### DIFF
--- a/openslides_backend/action/user/update_temporary.py
+++ b/openslides_backend/action/user/update_temporary.py
@@ -25,7 +25,7 @@ class UserUpdateTemporary(UpdateAction):
             "number",
             "structure_level",
             "email",
-            # "vote_weight",
+            "vote_weight",
             "is_present_in_meeting_ids",
         ]
     )
--- a/openslides_backend/models/fields.py
+++ b/openslides_backend/models/fields.py
@@ -176,8 +176,16 @@ class FloatField(Field):
 
 
 class DecimalField(Field):
+    """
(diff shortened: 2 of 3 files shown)


FinnStutzenstein:
No, there are special values `-1` and `-2` for `votescast` and so on.

jsangmeister:
Mmh, right. So the final regex would be `^-?(\d|[1-9]\d+)(\.\d{6})$` to support all needed numbers. Anything missing, before I need to create yet another PR for this? :D

jsangmeister:
Or should I restrain negative numbers to just `-1` and `-2`? Then, it would be ^-1|-2|(\d|[1-9]\d+)(\.\d{6})$

FinnStutzenstein:
> Or should I restrain negative numbers to just -1 and -2?

Would be enough for our use case, but then name the decimal field class something like `VotingDecimalFieldWithSpecialValues` so so ;)

Also: Currently OS3 accepts these numbers as floats since JS does not have a native decimal implementation.

jsangmeister:
> Also: Currently OS3 accepts these numbers as floats since JS does not have a native decimal implementation.

What does this mean for OS4? Should we accept strings and floats?

FinnStutzenstein:
Ok, only strings. We'll adapt the client

jsangmeister:
Ok, final regex: ^-?(\d|[1-9]\d+)\.\d{6}$ (so all numbers with exactly 6 decimal places are accepted)

PR #202, opened 2020-10-12
Fix nested cascade deletion

jsangmeister:
If a meeting gets deleted, it cascades the deletion to all dependent models, some of which have set `OnDelete.PROTECT` on some relations. This has to be ignored.

### DIFF
--- a/openslides_backend/action/generics.py
+++ b/openslides_backend/action/generics.py
@@ -1,6 +1,7 @@
 import re
 from collections import defaultdict
-from typing import Any, Dict, Iterable, List, Set, Tuple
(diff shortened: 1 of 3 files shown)


normanjaeckel (approved):
The "meeting" problem still exists, right? Maybe open a new issue for this.

jsangmeister:
> The "meeting" problem still exists, right? Maybe open a new issue for this.

Do you mean the problem I described in the description of this PR? If so, then no, that was the whole point of this PR to fix this :D If not, then explain more detailed what you mean.

normanjaeckel:
Ok.

PR #211, opened 2020-10-13
Update speaker.create, add weight 10000 default.

reiterl:


jsangmeister:
This is not needed, we can add defaults in the `models.yml` (see for example https://github.com/OpenSlides/OpenSlides/blob/eded14f0665b4850d0df22e0beddc530f25e466e/docs/models.yml#L872)

jsangmeister:
see https://github.com/OpenSlides/OpenSlides/pull/5632

PR #212, opened 2020-10-13
Update motion.update action + tests.

reiterl:
Update update action
Don't test amendment_paragraph.

### DIFF
--- a/openslides_backend/action/motion/update.py
+++ b/openslides_backend/action/motion/update.py
@@ -1,7 +1,8 @@
 import time
+from typing import Any, Dict
 
 from ...models.models import Motion
-from ..base import ActionPayload, DataSet, DummyAction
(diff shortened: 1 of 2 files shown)


openslides_backend/action/motion/update.py
```diff
@@ -15,13 +16,19 @@ class MotionUpdate(UpdateAction):
...
-    def prepare_dataset(self, payload: ActionPayload) -> DataSet:
-        for instance in payload:
-            instance["last_modified"] = round(time.time())
-        return super().prepare_dataset(payload)
+        optional_properties=[
+            "title",
+            "number",
+            "text",
+            "reason",
+            "amendment_paragraph_",
```
normanjaeckel, line 24: This is wrong. It is a template field so you can not write directly into "amendment_paragraph_"

PR #214, opened 2020-10-14
Update payload of user.create_temporay and user.update_temporay

reiterl:
Update user.create_temporay.

Add vote_weight and default_password.
Add test.

Update user.update_temporay.

Add default_password and group__ids to payload.
Add a test.

### DIFF
--- a/openslides_backend/action/generics.py
+++ b/openslides_backend/action/generics.py
@@ -1,4 +1,3 @@
-import re
 from collections import defaultdict
 from copy import deepcopy
 from typing import Any, Dict, Iterable, List, Set, Tuple, Type
@@ -11,7 +10,7 @@ from ..models.fields import (
 )
 from ..shared.exceptions import ActionException
 from ..shared.interfaces import Event, WriteRequestElement
-from ..shared.patterns import FullQualifiedId
+from ..shared.patterns import ID_PATTERN, FullQualifiedId
 from ..shared.typing import DeletedModel, ModelMap
 from .actions_map import actions_map
 from .base import Action, ActionPayload, DataSet, merge_write_request_elements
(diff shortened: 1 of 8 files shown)


openslides_backend/action/user/update_temporary.py
```diff
@@ -27,5 +27,7 @@ class UserUpdateTemporary(UpdateAction):
             "email",
             "vote_weight",
             "is_present_in_meeting_ids",
+            "default_password",
+            "group__ids",
```
jsangmeister, line 31: This is wrong I think. `group_$_ids` is a structured field and what you are doing here is writing directly into the field `group__ids` instead of using it as a template field. We should wait for the discussion in https://github.com/OpenSlides/openslides-backend/issues/219 until we change anything here.

jsangmeister: Update: discussion in https://github.com/OpenSlides-action-crew/openslides-backend-coordination/issues/135

jsangmeister: Okay, seems like we came to a conclusion there. They payload will have a field `group_ids` which you need to translate internally to `group_$<meeting_id>_ids` which will then be written to the DS.

openslides_backend/action/user/create_temporary.py
```diff
@@ -25,7 +25,8 @@ class UserCreateTemporary(CreateAction):
             "number",
             "structure_level",
             "email",
-            # "vote_weight",
+            "vote_weight",
             "is_present_in_meeting_ids",
+            "default_password",
```
jsangmeister, line 30: `group_ids` is still missing here, alos

openslides_backend/action/user/update_temporary.py
normanjaeckel: Ok.

reiterl: Ok, do it then, so I have an example how to handle this case.

jsangmeister:
Added group_ids with general changes to write structured_fields elsewhere.

jsangmeister:
@normanjaeckel please check the new changes and merge if its fine with you

openslides_backend/action/generics.py
```diff
@@ -44,34 +44,26 @@ def create_action_prepare_dataset(self, payload: ActionPayload) -> DataSet:
...
+                if field_name in instance:
+                    if field.structured_relation:
+                        if instance.get(field.structured_relation[0]) is None:
+                            raise ActionException(
+                                "You must give both a relation field "
+                                "with structured_relation and its corresponding "
+                                "foreign key field."
+                            )
+                    relation_fields.append((field_name, field))
+                elif isinstance(field, BaseTemplateField):
```
normanjaeckel, line 56: THis should be BaseTemplateRelationField. Other template fields like amendmend_paragraph are not handled here.

openslides_backend/action/generics.py
```diff
@@ -44,34 +44,26 @@ def create_action_prepare_dataset(self, payload: ActionPayload) -> DataSet:
...
+                        field_name, field, instance
+                    )
+                    for instance_field, substitute in structured_fields:
                         relation_fields.append((instance_field, field))
                         template_field_name = (
                             field_name[: field.index] + "$" + field_name[field.index :]
                         )
                         additional_instance_fields[template_field_name].append(
-                            match.group(1)
+                            substitute
```
normanjaeckel, line 66: Maybe we should assert here that substitute does only contain digits.

openslides_backend/action/generics.py
```diff
@@ -164,33 +156,24 @@ def update_action_prepare_dataset(self, payload: ActionPayload) -> DataSet:
...
+                                "You must not try to update both a relation field "
+                                "with structured_relation and its corresponding "
+                                "foreign key field."
+                            )
+                    relation_fields.append((field_name, field))
+                elif isinstance(field, BaseTemplateField):
+                    structured_fields = self.get_structured_fields_in_instance(
+                        field_name, field, instance
+                    )
+                    for instance_field, replacement in structured_fields:
```
normanjaeckel, line 172: Here the same: replacement or substitute (maybe use always the same term) must not contain other chars than digits.

normanjaeckel (changes requested):

normanjaeckel:
Only some small remarks. Please also resolve conflicts.

openslides_backend/action/generics.py
jsangmeister: Why aren't they handled here? Shouldn't they be working the same way?

normanjaeckel: In line 46 we call get_relation_fields. This should not return non relation fields.

jsangmeister: Ah, I see. Okay, I'll look into it how to handle this then when looking at #226

PR #217, opened 2020-10-14
Update motion_workflow.create. Set the first_state of workflow.

reiterl:
Add the created new state to the first state.
Use first_state_of_workflow_id property of the motion_state.

### DIFF
--- a/tests/system/action/motion_workflow/test_create.py
+++ b/tests/system/action/motion_workflow/test_create.py
@@ -16,8 +16,10 @@ class MotionWorkflowSystemTest(BaseActionTestCase):
         self.assert_status_code(response, 200)
         workflow = self.get_model("motion_workflow/1")
         assert workflow.get("name") == "test_Xcdfgee"
+        assert workflow.get("first_state_id") == 1
         state = self.get_model("motion_state/1")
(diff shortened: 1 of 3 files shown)


tests/system/action/motion_workflow/test_create.py
```diff
@@ -16,6 +16,7 @@ def test_create(self) -> None:
         self.assert_status_code(response, 200)
         workflow = self.get_model("motion_workflow/1")
         assert workflow.get("name") == "test_Xcdfgee"
+        assert workflow.get("first_state_id") == 1
         state = self.get_model("motion_state/1")
         assert state.get("workflow_id") == 1
```
jsangmeister, line 21: Please check here that `first_state_of_workflow_id` is set correctly

jsangmeister (changes requested):
Nice! Thought this would be more complicated. Minor test missing

tests/system/action/motion_workflow/test_create.py
reiterl: Add this assert, too. Done.

normanjaeckel (approved):

jsangmeister (approved):

PR #216, opened 2020-10-14
Add mediafile.update action + tests.

reiterl:
Add action + tests.

PR #232, opened 2020-10-19
Add mediafile create_directory action + tests.

reiterl:


openslides_backend/action/mediafile/create_directory.py
```diff
@@ -0,0 +1,59 @@
...
+                len(instance["inherited_access_group_ids"]) > 0
+                or len(parent["inherited_access_group_ids"]) > 0
+            )
+        else:
+            instance["inherited_access_group_ids"] = instance.get(
+                "access_group_ids", []
+            )
+            instance["has_inherited_access_groups"] = (
+                len(instance["inherited_access_group_ids"]) > 0
+            )
```
jsangmeister, line 57: Since this should be the same as for `mediafile.update` (except the part about the children I commented there), you should move this to a seperate method/base class. You can wait until https://github.com/OpenSlides/openslides-backend/pull/216 is merged with that

jsangmeister (changes requested):

openslides_backend/action/mediafile/create_directory.py
reiterl: Done, with the code from mediafile.update.

tests/system/action/mediafile/test_create_directory.py
```diff
@@ -0,0 +1,303 @@
...
+        assert model.get("is_directory") is True
+
+    def test_create_directory_parent(self) -> None:
+        self.create_model(
+            "group/7", {"name": "group_LxAHErRs", "user_ids": [], "meeting_id": 110}
+        )
+        self.create_model("meeting/110", {"name": "meeting110"})
+        self.create_model(
+            "mediafile/110",
+            {"title": "title_srtgb199", "child_ids": [111], "meeting_id": 110},
```
jsangmeister, line 38: Why do you set `child_ids` here? This should be set automatically by the relation handling

tests/system/action/mediafile/test_create_directory.py
```diff
@@ -0,0 +1,303 @@
...
+                        }
+                    ],
+                }
+            ],
+        )
+        self.assert_status_code(response, 200)
+        model = self.get_model("mediafile/1")
+        assert model.get("title") == "title_Xcdfgee"
+        assert model.get("access_group_ids") == [7]
+        assert model.get("is_directory") is True
```
jsangmeister, line 29:
```
assert model.get("inherited_access_group_ids") == [7]
assert model.get("has_inherited_access_group_ids")  is True
```

tests/system/action/mediafile/test_create_directory.py
```diff
@@ -0,0 +1,303 @@
...
+        )
+        self.create_model(
+            "group/8", {"name": "group_sdfafd", "user_ids": [], "meeting_id": 110}
+        )
+        self.create_model("meeting/110", {"name": "meeting110"})
+        self.create_model(
+            "mediafile/110",
+            {
+                "title": "title_srtgb199",
+                "child_ids": [111],
```
jsangmeister, line 75: see above

jsangmeister (approved):
besides the minor remarks it looks good!

tests/system/action/mediafile/test_create_directory.py
normanjaeckel: I think this is correct here. The creation in favor of a test is not run through any relation handling code.

tests/system/action/mediafile/test_create_directory.py
reiterl: see above for Norman's comment.

tests/system/action/mediafile/test_create_directory.py
reiterl: Add the two missing asserts.

tests/system/action/mediafile/test_create_directory.py
jsangmeister: Yes, I know, but if this is needed here, there is a bug in the action since this situation can never happen. If this is not needed, it can be removed.

jsangmeister:
Cleaned up and squashed

PR #235, opened 2020-10-20
Add Curl

FinnStutzenstein:
This is needed as long as no management scripts are there. Curl is externally used in the container to gain access to the datastore to execute management commands


normanjaeckel:
I do not understand. Why does the productive container has to send curl commands to some other service? This should be done outside the backend container.

PR #236, opened 2020-10-20
Prevent mutation of mapped_fields and make it a Set

jsangmeister:


openslides_backend/action/generics.py
```diff
@@ -277,7 +277,11 @@ def update_action_create_instance_write_request_element(
         """
         fqid = FullQualifiedId(self.model.collection, element["instance"]["id"])
         information = {fqid: ["Object updated"]}
-        fields = {k: v for k, v in element["instance"].items() if k != "id"}
+        fields = {
+            k: v
+            for k, v in element["instance"].items()
+            if k != "id" and not k.startswith("meta_")
```
jsangmeister, line 283: Somehow the backend suddenly tried to write meta fields, so I had to remove them here

PR #245, opened 2020-10-22
Add meeting.delete_all_speakers_of_all_list action + tests.

reiterl:


openslides_backend/action/meeting/delete_all_speakers_of_all_lists.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+@register_action("meeting.delete_all_speakers_of_all_lists")
+class DeleteAllSpeakersOfAllListsAction(DeleteAction):
+    """
+    Action to delete all speakers of all lists of a meeting.
+    """
+
+    model = Speaker()  # we want to delete Speakers
+    schema = DefaultSchema(Speaker()).get_delete_schema()
```
normanjaeckel, line 16: This schema is not correct. It produces a missleading title and description.

normanjaeckel (changes requested):

openslides_backend/action/generics.py
```diff
@@ -307,7 +307,7 @@ def delete_action_prepare_dataset(self, payload: ActionPayload) -> DataSet:
         """
 
         data = []
-        for instance in payload:
+        for instance in self.update_instances(payload):
```
jsangmeister, line 310: Please rebase to get the newest changes here

openslides_backend/action/meeting/delete_all_speakers_of_all_lists.py
```diff
@@ -0,0 +1,36 @@
...
+                continue
+            for los in meeting["list_of_speakers_ids"]:
+                list_of_speakers = self.database.get(
+                    FullQualifiedId(Collection("list_of_speakers"), los),
+                    ["speaker_ids"],
+                )
+                for speaker in list_of_speakers.get("speaker_ids", []):
+                    new_payload.append({"id": speaker})
+        if not new_payload:
+            raise ActionException("No speakers to delete.")
```
jsangmeister, line 35: See #247

jsangmeister (changes requested):

openslides_backend/action/generics.py
reiterl: Rebased.

openslides_backend/action/meeting/delete_all_speakers_of_all_lists.py
reiterl: The underlying code cannot handle a empty payload case.

def merge_write_request_elements(
       [...]
        if user_id is None:
>           raise ValueError("At least one of the given user ids must not be None.")
E           ValueError: At least one of the given user ids must not be None.

openslides_backend/action/base.py:384: ValueError

normanjaeckel: Here we either have to insist in sending HTTP 400 or we have to add a new exception that is parsed to HTTP 200 or even better HTTP 204

openslides_backend/action/meeting/delete_all_speakers_of_all_lists.py
reiterl: Updated the schema.

openslides_backend/action/meeting/delete_all_speakers_of_all_lists.py
reiterl: Add an exception which returns a  HTTP 200.

openslides_backend/shared/exceptions.py
```diff
@@ -23,6 +23,10 @@ class ActionException(View400Exception):
     pass
 
 
+class NoContentException(BackendBaseException):
```
jsangmeister, line 26: I would suggest to subclass from `ViewException` and set the status code to 204. The rest should be handled automatically then and you can remove the try...catch block

openslides_backend/action/meeting/delete_all_speakers_of_all_lists.py
```diff
@@ -15,7 +15,11 @@ class DeleteAllSpeakersOfAllListsAction(DeleteAction):
     """
 
     model = Speaker()  # we want to delete Speakers
-    schema = DefaultSchema(Speaker()).get_delete_schema()
+    schema = DefaultSchema(Speaker()).get_default_schema(
```
jsangmeister, line 18: Although it makes no difference in this case, it is still not quite correct, since the base model for the schema is `meeting`. Strictly speaking, you are accesing the property `speaker.id` here, but want to have `meeting.id`. Of course, this is the same schema, but if for some reason other fields are added, it would produce wrong results. So please change it to `Meeting()`.

jsangmeister (changes requested):

openslides_backend/action/meeting/delete_all_speakers_of_all_lists.py
jsangmeister: ...or we could update the merge method to be able to handle empty write events? Throwing an exception in a success case seems really hacky. The only other thing to check is then at the end, if eyn events should be written to the datastore at all, which should be easy.

openslides_backend/shared/exceptions.py
jsangmeister: Or just throw away this exception altogether, see https://github.com/OpenSlides/openslides-backend/pull/245#discussion_r517444090

openslides_backend/action/meeting/delete_all_speakers_of_all_lists.py
normanjaeckel: Sounds reasonable.

openslides_backend/shared/exceptions.py
normanjaeckel: I would also prefer not to use an exception here (although I might have wrote something else some days ago).

openslides_backend/action/meeting/delete_all_speakers_of_all_lists.py
reiterl: Updated. Now I use Meeting instead Speaker in DefaultSchema.

jsangmeister:
I will implement this.

jsangmeister (approved):

openslides_backend/action/generics.py
```diff
@@ -432,6 +432,12 @@ def delete_action_prepare_dataset(self, payload: ActionPayload) -> DataSet:
 
         return {"data": data}
 
+    def update_instances(self, payload: ActionPayload) -> ActionPayload:
```
normanjaeckel, line 435: Please remove this method.

normanjaeckel (changes requested):

openslides_backend/action/generics.py
jsangmeister: Done

PR #249, opened 2020-10-23
Added end speach action.

normanjaeckel:


openslides_backend/action/speaker/end_speach.py
```diff
@@ -0,0 +1,51 @@
...
+                raise ActionException(
+                    f"Speaker {instance['id']} is not speaking at the moment."
+                )
+            instance["end_time"] = round(time.time())
+            data.append({"instance": instance})
+        return {"data": data}
+
+    def create_write_request_elements(
+        self, dataset: DataSet
+    ) -> Iterable[WriteRequestElement]:
```
jsangmeister, line 43: If you (rebase and) subclass from `UpdateAction`, you can use the `get_updated_models` function instead of `prepare_dataset` and don't have to overwrite this function. This should simplify your code by a good amount.

normanjaeckel: Good idea. I also change the return value from List to Iterable so I can use yield statement.

normanjaeckel:
@jsangmeister Please merge if everything is ok.

jsangmeister (approved):

PR #251, opened 2020-10-23
Add mediafile.delete action + tests

reiterl:

PR #252, opened 2020-10-23
Changed exception handling. Removed addendum server. Changed action p…

normanjaeckel:
…ayload type.

@jsangmeister Maybe this also fits your needs regarding exception debugging during tests.

Just set envirionment variable `OPENSLIDES_BACKEND_RAISE_4XX` to something truthy and the backend will raise the ActionExceptions and DatabaseExceptions and thelike instead of transforming them to HTTP 400. So you can see the traceback. Does this fit your needs?

### DIFF
--- a/openslides_backend/action/base.py
+++ b/openslides_backend/action/base.py
@@ -109,6 +109,22 @@ class Action(BaseAction, metaclass=SchemaProvider):
         """
         raise NotImplementedError
 
+    def get_updated_instances(self, payload: ActionPayload) -> ActionPayload:
+        """
+        By default this does nothing. Override in subclasses to adjust the updates
+        to all instances of the payload.
+        """
+        yield from payload
+
+    def validate_fields(self, instance: Dict[str, Any]) -> Dict[str, Any]:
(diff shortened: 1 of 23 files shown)


jsangmeister:
> Does this fit your needs?

Not really. But I guess we'll leave it this way for now, since there does not seem to be a simple solution

openslides_backend/action/base.py
```diff
@@ -109,17 +109,12 @@ def prepare_dataset(self, payload: ActionPayload) -> DataSet:
         """
         raise NotImplementedError
 
-    def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
+    def get_updated_instances(self, payload: ActionPayload) -> ActionPayload:
```
jsangmeister, line 112: Ah, I didn't see this until now, @reiterl did a similar change in https://github.com/OpenSlides/openslides-backend/pull/251, but maybe the additional base class was overkill. You can fix the conflicts however you want

jsangmeister (approved):

PR #263, opened 2020-10-29
Add models version check

jsangmeister:
In https://github.com/OpenSlides/openslides-backend/pull/262 I updated the models.py which did not happen for quite some time and led to a few unwanted errors. I implemented a simple check to assert that the models.py is always up-to-date.

### DIFF
--- a/cli/generate_models.py
+++ b/cli/generate_models.py
@@ -1,5 +1,7 @@
+import hashlib
 import os
 import string
+import sys
 from textwrap import dedent, indent
 from typing import Any, Dict, List, Optional, Union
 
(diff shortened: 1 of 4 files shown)


cli/generate_models.py
```diff
@@ -104,6 +115,8 @@ def main() -> None:
     MODELS = yaml.safe_load(models_yml)
     with open(DESTINATION, "w") as dest:
         dest.write(FILE_TEMPLATE)
+        breakpoint()
```
FinnStutzenstein, line 118: Please remove

jsangmeister: Done

PR #180, opened 2020-10-02
Create motion.create action and add to calculate state_id

reiterl:
Add draft motion.create action and tests.
Update motion.create: add state_id calculation.
Add two tests.

PR #264, opened 2020-10-29
Fix template field handling

jsangmeister:
Fix the template field handling in Create and UpdateAction to also handle structured tag fields correctly

### DIFF
--- a/openslides_backend/action/generics.py
+++ b/openslides_backend/action/generics.py
@@ -5,6 +5,7 @@ from typing import Any, Dict, Iterable, List, Set, Tuple, Type
 from ..models.fields import (
     BaseGenericRelationField,
     BaseRelationField,
+    BaseTemplateField,
     BaseTemplateRelationField,
     OnDelete,
 )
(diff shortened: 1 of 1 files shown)


openslides_backend/action/generics.py
```diff
@@ -50,8 +51,8 @@ def create_action_prepare_dataset(self, payload: ActionPayload) -> DataSet:
             # Collect relation fields and also check structured relations and template fields.
```
normanjaeckel, line 51: This comment is not up to date because the follwing code now handles not only relation fields but also other template fields.

normanjaeckel (changes requested):
Do you like to describe a test case for the purpose of this change?

jsangmeister:
Yes, see my commit in https://github.com/OpenSlides/openslides-backend/pull/180, it is meant to make structured tags like `amendment_paragraph_$` work

openslides_backend/action/generics.py
jsangmeister: I think the "and template fields" includes the new changes, doesn't it?

PR #267, opened 2020-10-30
Update black

jsangmeister:
Primarily because of the trailing comma issue (https://github.com/psf/black/issues/1288)


jsangmeister:
@normanjaeckel I saw that you commented out this version in the requirements, does this have a special reason? Because I would like to use the newer version

normanjaeckel:
I commented it out because I do not want to mix a "update black" pull request with a pull request containing real changes and I do not want to make to much noise during development (forcing others to rebase their commits e.g.). If this pull request does not contain any real changes it can be merged.

PR #270, opened 2020-10-30
Add set_password_self function with tests

jsangmeister:


jsangmeister:
waiting for https://github.com/OpenSlides/openslides-auth-service/issues/41

jsangmeister:
Finished integrating the auth service. Also added the `execute_other_action` method to encapsulate the delegation of actions.

openslides_backend/action/generics.py
```diff
@@ -336,9 +335,7 @@ def delete_action_prepare_dataset(self, payload: ActionPayload) -> DataSet:
             relation_fields: List[Tuple[str, BaseRelationField]] = []
             # Gather all delete actions with payload and also all models to be deleted
             delete_actions: List[Tuple[Type[Action], ActionPayload]] = []
-            additional_relation_models: ModelMap = deepcopy(
```
normanjaeckel, line 339: Please check this. Do we have to copy the given additional_relation_models here or not? I am not sure. If you are sure it is ok.

openslides_backend/action/user/set_password_self.py
```diff
@@ -0,0 +1,41 @@
...
+    schema = DefaultSchema(User()).get_default_schema(
+        additional_required_fields={
+            "old_password": {"type": "string", "minLength": 1},
+            "new_password": {"type": "string", "minLength": 1},
+        }
+    )
+
+    def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
+        if self.auth.is_anonymous(self.user_id):
+            raise PermissionDenied("Can't set password for anonymous")
```
normanjaeckel, line 27: This is not a permission denied here but a ActionException (400 not 403).

openslides_backend/shared/interfaces.py
```diff
@@ -79,6 +79,8 @@ class WSGIApplication(Protocol):  # pragma: no cover
     Interface for main WSGI application class.
     """
 
+    services: Any
```
normanjaeckel, line 82: Is this line necessary? I do no see which change in some other file would cause mypy fail.

openslides_backend/action/generics.py
jsangmeister: Yes, we have to copy them, but this was moved to the `execute_other_action` method (where the copying is done via he `**` operator).

openslides_backend/action/user/set_password_self.py
jsangmeister: Ok, changed.

openslides_backend/shared/interfaces.py
jsangmeister: Mypy complained when you had an object of type `WSGIApplication` and tried to access the services, I think I had to cast to `Any` every time.

PR #272, opened 2020-10-30
Remove obsolet dummy actions and add new delete ones

jsangmeister:


jsangmeister:
Should fix the cascade deletion error in https://github.com/OpenSlides/openslides-backend-coordination/issues/62. @FinnStutzenstein please check as I have not implemented a test since it's only dummy actions.

FinnStutzenstein:
Works for deleting assignments. Please set the coordination-issue to done, when this is merged.

PR #283, opened 2020-11-04
Update model of motion_state + test.

reiterl:
Add default of lightblue to css_class of motion state.

Need the update of the model before this is clean.


jsangmeister:
Whoops, deleted your branch by accident, but I think you delete them anyway, don't you?

Issue #269, opened 2020-10-30
Check readd_last


FinnStutzenstein:
-> If the last speaker is also in the waiting-list, there should be an error.

PR #285, opened 2020-11-04
Fix speaker create and list of speakers re add last.

normanjaeckel:
See #269

### DIFF
--- a/openslides_backend/action/speaker/create_update_delete.py
+++ b/openslides_backend/action/speaker/create_update_delete.py
@@ -1,4 +1,9 @@
+from typing import Any, Dict
+
 from ...models.models import Speaker
+from ...shared.exceptions import ActionException
+from ...shared.filters import And, FilterOperator
+from ...shared.patterns import Collection
 from ..create_action_with_inferred_meeting import CreateActionWithInferredMeeting
 from ..default_schema import DefaultSchema
 from ..generics import DeleteAction, UpdateAction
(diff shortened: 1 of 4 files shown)


openslides_backend/action/speaker/create_update_delete.py
```diff
@@ -14,6 +19,28 @@ class SpeakerCreateAction(CreateActionWithInferredMeeting):
...
     )
 
+    def validate_fields(self, instance: Dict[str, Any]) -> Dict[str, Any]:
+        """
+        Checks that a new speaker does not already exist on the list of speaker as
+        comming speaker (with begin_time == None)
+        """
+        filter_obj = And(
+            FilterOperator("list_of_speakers_id", "=", instance["list_of_speakers_id"]),
+            FilterOperator("end_time", "=", None),
```
jsangmeister, line 29: Can you add a speaker another time if he is currently speaking? If yes, the check for `begin_time == None` is missing here. If not, the comment above is false and should say `end_time`.

jsangmeister (changes requested):

openslides_backend/action/speaker/create_update_delete.py
normanjaeckel: Fixed.

normanjaeckel:
Fixed and rebased. Ready to merge.

jsangmeister (approved):

Issue #259, opened 2020-10-28
Return newly created ids for every create action


FinnStutzenstein:
The client needs the id to navigate to the newly create object. Reference in OS3: https://github.com/OpenSlides/OpenSlides/blob/master/server/openslides/utils/rest_api.py#L329

normanjaeckel:
New approach is given in #286.

jsangmeister:
This should be solved then

PR #286, opened 2020-11-04
Added information about new id to action result of create actions.

normanjaeckel:
See #259

### DIFF
--- a/openslides_backend/action/action.py
+++ b/openslides_backend/action/action.py
@@ -1,4 +1,4 @@
-from typing import Dict, Iterable, List, Tuple, Union
+from typing import Dict, Iterable, List, Optional, Tuple, Union
 
 import fastjsonschema
 
@@ -6,7 +6,12 @@ from ..shared.exceptions import ActionException, EventStoreException
 from ..shared.handlers.base_handler import BaseHandler
 from ..shared.interfaces.write_request_element import WriteRequestElement
 from ..shared.schema import schema_version
-from .action_interface import ActionResult, Payload
+from .action_interface import (
+    ActionResponse,
+    ActionResponseResults,
+    ActionResponseResultsElement,
+    Payload,
+)
 from .actions_map import actions_map
 from .base import merge_write_request_elements
 
@@ -57,7 +62,7 @@ class ActionHandler(BaseHandler):
             else:
                 yield name, action.schema
 
-    def handle_request(self, payload: Payload, user_id: int) -> List[ActionResult]:
+    def handle_request(self, payload: Payload, user_id: int) -> ActionResponse:
         """
         Takes payload and user id and handles this request by validating and
         parsing all actions. In the end it sends everything to the event store.
@@ -74,7 +79,7 @@ class ActionHandler(BaseHandler):
         # store for some time if event store sends ModelLocked Exception
 
         # Parse actions and creates events
-        write_request_element = self.parse_actions(payload)
+        write_request_element, results = self.parse_actions(payload)
 
         # Send events to datastore
         try:
@@ -86,9 +91,9 @@ class ActionHandler(BaseHandler):
         # TODO: This is a fake result because in this place all actions were
         # always successful.
         self.logger.debug("Request was successful. Send response now.")
-        return [
-            ActionResult(success=True, message="Action handled successfully")
-        ] * len(payload)
+        return ActionResponse(
+            success=True, message="Actions handled successfully", results=results
+        )
 
     def validate(self, payload: Payload) -> None:
         """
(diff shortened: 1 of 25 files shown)


normanjaeckel:
@GabrielInTheWorld Please check if this fits your needs. The action service now provides information about all actions and events and you can look for the string "Object create" at the new fqid.

FinnStutzenstein:
Hi, sadly, this does not fit into the needs. These are too much and too verbose, human-like information. Whats about just returning the created id? The thing is that in the client code we need the id without the knowledge about the created object; and when creating an assignment, multiple fqids matches "Object Created". Also: This string can change etc. Please just provide the created id

normanjaeckel:
It is not that easy just to return the id. I have a list of action. Each action creates a unknown amount of write request elements that are merged together at different points of code. Different objects are created during an action (e. g. topics, agenda items, list of speakers). The information I now provide is the same that goes to the datastore and can be used for the history. It would be an unnecessary extra work to provide another "information" which is more machine-like.

Proposal: We change the interface of the history information. Instead of a string "Object created" we provide an object like `{"code": "CREATED", "verbose": "Object created. Have a good day."}`.

FinnStutzenstein:
Then you should return an array of created ids. Somewhere in the create action you have to reserve the id - just return it (or however your code is structured). - If a single action can take multiple payloads I expect the action to return an array with results (or be omitted, if there is none). And the result of a single create action is the id, which is needed. The actions are designed to return information, if this is still not implemented, pelase do it now (it seems, that this part is not finished...).

normanjaeckel:
> The actions are designed to return information

This is done by this pull request.The information are the same that are send to the datastore. If you like to get a more machine like information object we should extend this information interface instead of creating a new one.

The current code uses generators to yield write requests that should be send to the datastore. During an action you can yield such WriteRequestElements as much as you like. Each element contains the relevant information (including FQID, including the new id of created objects). We use mixins and other things to yield such elements in different parts of the procedure (e. g. for creating agenda items, list of speakers, or cascade deleting of items and the like).

Of course I can yield with every WriteRequestElement an extra object with some kind of array index and some new kind of information. I just coded this and then got in mind that we already have all information we need in the information object of the WriteRequestelement.

FinnStutzenstein:
> > The actions are designed to return information
> 
> This is done by this pull request.

You misunderstood me: `information` is more general - I did not mean the datastore ones. I think the client should not deal with your "information" at this point but get a clear answer to each action (which may ne `null`, and can be omitted)

normanjaeckel:
Please lets have a short phone call on this which might be the easiest way to understand what to do and what is needed.

jsangmeister:
All that is needed is to somehow capture all created fqids and return them in the response. @normanjaeckel if you want, I can go ahead and implement it - the communication with @FinnStutzenstein would be a lot faster this way :)

FinnStutzenstein:
@jsangmeister We had a call today; Each action can give another result -> The open point is that I just want the id of the one thing created and normand wants to provide all ids; But he will implement and I do an update of the interface. We'll see how it goes.

jsangmeister:
Ok, sure

normanjaeckel:
@FinnStutzenstein Please check again this new approach.

openslides_backend/action/action.py
```diff
@@ -86,9 +91,9 @@ def handle_request(self, payload: Payload, user_id: int) -> List[ActionResult]:
         # TODO: This is a fake result because in this place all actions were
         # always successful.
         self.logger.debug("Request was successful. Send response now.")
-        return [
-            ActionResult(success=True, message="Action handled successfully")
-        ] * len(payload)
+        return ActionResponse(
+            success=True, message="Actions handled successfully", results=results
```
FinnStutzenstein, line 95: is `success=True` redundant? Since you have a (valid?) response, success is implied. I would expect to raise errors in error-cases and not usind `ActionResponse`, but I do not know the code well.

FinnStutzenstein:
@normanjaeckel This looks good. It fits into the proposed schema. Also the error (only tested without error_index) works. @jsangmeister Please have a look at the code. From a usability-point it works.

jsangmeister (commented):
Mmh, I don't like the handling of WriteRequestElements (WRE) vs. ActionResponseResultElements. I would prefer if we'd have one struct which is return everywhere (previously only WRE). This struct has a field for WRE and one which holds what is currently ActionResponseResultElements. Proposal:
```
ActionResponseElement = TypedDict("ActionResult", {
    "write_request_elements": Optional[Iterable[WriteRequestElement]], 
    "additional_information": Optional[Iterable[TODO]]
})
```
(however the type should be called, I'm not sure about that)

The advantages are:
1. We don't have to care about distinct cases when handling the return value of actions (WRE vs. ResultElement)
2. We can easily extend this schema if needed and don't have to change all lines where it is used in the code
3. it's generally cleaner than using Unions

What do you think @normanjaeckel?

jsangmeister:
And, if we don't go through with this suggestion, at least we should alias the `Union` that is used everywhere and rename the `ActionResponseResultElement`. We'd have to brainstorm about the last one, but the name is an abomination ;)

normanjaeckel:
Both proposals wre ok for me. But please merge this pull request first and provide the changes (if you like) in a new one.

FinnStutzenstein:
> We'd have to brainstorm about the last one

`ActionResponse` ;)

Ok, I'll merge the PR now so the Client can work with the information. For a restructure or issues, other PRs can be opened.

PR #289, opened 2020-11-06
The authlib expects to get the endpoint without the url

FinnStutzenstein:


jsangmeister:
The `AUTH_PATH` is overridden in the `docker-compose.dev.yml`. If I remove that entry, the auth cannot be reached anymore.

FinnStutzenstein:
black formatted a lot of stuff; maybe there was an update. I rebuilt all containers so it might be possible.

jsangmeister:
I guess you didn't rebase, black should automatically exluce the old integration tests to not have to deal with changes there in PRs.

normanjaeckel:
Does the authlib library magically set HOST, PORT and PATH of the auth service? I think it should be configurable ...

FinnStutzenstein:
> I think it should be configurable ...

It is, but the service does it (gets the variables from the env)

FinnStutzenstein:
@jsangmeister Ok, should these changes be redone?

jsangmeister:
> @jsangmeister Ok, should these changes be redone?

Na, it's ok, since these files shouldn't be used anymore. It just makes the PR more convoluted ^^ but it looks good, if you squash we can merge this.

jsangmeister:
Done.

normanjaeckel:
Please add the silently used environment variables to the README so that users of the openslides_backend know how to configure this service if they want to connect it via the authlib package to another instance of the auth service.

FinnStutzenstein:
https://github.com/OpenSlides/openslides-backend/pull/301

PR #295, opened 2020-11-09
Add user.reset_password_to_default_temporary + tests

reiterl:
Based on dev_reset_password_to_default.

### DIFF
--- a/openslides_backend/action/user/reset_password_to_default_temporary.py
+++ b/openslides_backend/action/user/reset_password_to_default_temporary.py
@@ -1,0 +1,21 @@
+from typing import Any, Dict
+
+from ..register import register_action
+from .check_temporary_mixin import CheckTemporaryMixin
+from .reset_password_to_default import UserResetPasswordToDefaultAction
+
+
+@register_action("user.reset_password_to_default_temporary")
+class UserResetPasswordToDefaultTemporaryAction(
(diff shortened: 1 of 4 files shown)


openslides_backend/action/user/reset_password_to_default_temporary.py
```diff
@@ -0,0 +1,24 @@
...
+
+    def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
+        """
+        Check for temporary user and call super().update_instance().
+        """
+        user = self.database.get(
+            FullQualifiedId(self.model.collection, instance["id"]), ["meeting_id"]
+        )
+        if not user.get("meeting_id"):
+            raise ActionException(f"User {instance['id']} is not temporary.")
```
jsangmeister, line 23: You should move this check into a base class/mixin since it is needed in every temporary action. Please also adjust existing temporary actions then

jsangmeister (changes requested):
Rest looks good

openslides_backend/action/user/reset_password_to_default_temporary.py
reiterl: Created a mixin for the check.

jsangmeister (approved):

jsangmeister:
Rebased and squashed.

PR #296, opened 2020-11-10
Add permission service

FinnStutzenstein:
@jsangmeister @normanjaeckel Can someone fix the tests? I think most of them relied on the actual HTTP-Adapter and not the test service. I do not have the overview to change this.

Also: Black formatted old tests again, but now I branched from master. Can someone run `./cleanup.sh` again and see, if it changes back?

This PR does only work in combination with https://github.com/OpenSlides/OpenSlides/pull/5646 in the main repo, if the complete setup is started there. With the complete setup, it can be seen in action. In [this file](https://github.com/OpenSlides/openslides-permission-service/blob/master/internal/core/queries.go) the current implemented queries are listed.

### DIFF
--- a/tests/integration/old/fake_services/permission.py
+++ b/tests/integration/old/fake_services/permission.py
@@ -1,4 +1,6 @@
-from typing import Any
+from typing import Any, Dict, List
+
+from openslides_backend.services.permission.interface import NotAllowed
 
 
 class PermissionTestAdapter:
@@ -8,8 +10,13 @@ class PermissionTestAdapter:
     implementation.
     """
 
-    def __init__(self, superuser: int, *args: Any, **kwargs: Any) -> None:
-        self.superuser = superuser
+    def __init__(self, superuser_id: int, *args: Any, **kwargs: Any) -> None:
+        self.superuser_id = superuser_id
 
-    def check_action(self, user_id: int, action: str, data: Any) -> bool:
-        return user_id == self.superuser
+    def is_allowed(
+        self, name: str, user_id: int, data: List[Dict[str, Any]]
+    ) -> List[Dict[str, Any]]:
+        if user_id == self.superuser_id:
+            return [{} for _ in data]
+        else:
+            raise NotAllowed("Not the superuser")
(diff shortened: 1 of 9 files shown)


jsangmeister:
Ahh, you use the `cleanup.sh`, I forgot to adjust that since I only use the makefile. That explains the diff

FinnStutzenstein:
I rebased -> the black files were changed in master, so no problems here. Still open:

One of @normanjaeckel or @jsangmeister should fix the tests

tests/integration/old/fake_services/permission.py
```diff
@@ -8,8 +10,13 @@ class PermissionTestAdapter:
...
 
-    def check_action(self, user_id: int, action: str, data: Any) -> bool:
-        return user_id == self.superuser
+    def is_allowed(
+        self, name: str, user_id: int, data: List[Dict[str, Any]]
+    ) -> List[Dict[str, Any]]:
+        if user_id == self.superuser_id:
+            return [{} for _ in data]
+        else:
+            raise NotAllowed("Not the superuser")
```
jsangmeister, line 22: This is the fake adapter from the old integration tests. The changes here will have no effect on the actual tests

FinnStutzenstein: Oh ok, but I'll leave them, if they do not hurt (It is more work to redo them instead of leaving as-is;) )

jsangmeister: Yes, just wanted to let you know :D

jsangmeister:
I can fix the tests.

jsangmeister:
> (I'm still a bit confused, that the actual http adapter is used in the tests even there is no perms service...)

After looking into it, I think that came from the fact that the actual implementation was the same as the now mocked adapter, as it always returned `True`. So there was just no need to mock a fake adapter.

Anyway, the tests should be fixed now.

normanjaeckel (commented):
Ok.

jsangmeister:
@FinnStutzenstein you can merge if you have no additional remarks.

PR #299, opened 2020-11-10
Add group.delete action + tests.

reiterl:


openslides_backend/action/group/delete.py
```diff
@@ -0,0 +1,36 @@
...
+        )
+        if group.get("default_group_for_meeting_id"):
+            raise ActionException(
+                f"Group {instance['id']} is a default group for a meeting."
+            )
+        if group.get("superadmin_group_for_meeting_id"):
+            raise ActionException(
+                f"Group {instance['id']} is a superadmin group for a meeting."
+            )
+        return instance
```
jsangmeister, line 36: This can also be done via the `models.yml`, just set these fields to `on_delete: PROTECT`. Remove this and then test again that it works as intended

jsangmeister: Only issue with this is that the error messages become too broad, this was already addressed in https://github.com/OpenSlides/openslides-backend-coordination/issues/96#issuecomment-720973153 and will be fixed with a general solution

jsangmeister (changes requested):

openslides_backend/action/group/delete.py
reiterl: Okay, we want it more implicitly by the model definition. Okay, done this, Check was okay.
Add PR https://github.com/OpenSlides/OpenSlides/pull/5697 for the model changes.

jsangmeister (approved):
Please sqash

PR #300, opened 2020-11-10
Added test.

normanjaeckel:
@jsangmeister Can you please fix this test?

Maybe it is related to https://github.com/OpenSlides/openslides-backend-coordination/issues/48 but I am not sure. The problem is a cascade deletion with a reverse reference: Workflow -> State -> Workflow.

Maybe the solution is to put the deleted model also to `additional_related_models` here https://github.com/OpenSlides/openslides-backend/blob/master/openslides_backend/action/generics.py#L412 and not only models of related fields.

### DIFF
--- a/openslides_backend/action/generics.py
+++ b/openslides_backend/action/generics.py
@@ -311,8 +311,9 @@ class DeleteAction(Action):
                 for field_name, field in self.model.get_relation_fields()
                 if field.on_delete != OnDelete.SET_NULL
             ]
+            this_fqid = FullQualifiedId(self.model.collection, instance["id"])
             db_instance = self.datastore.get(
-                fqid=FullQualifiedId(self.model.collection, instance["id"]),
+                fqid=this_fqid,
                 mapped_fields=relevant_fields,
                 lock_result=True,
             )
@@ -322,7 +323,7 @@ class DeleteAction(Action):
             relation_fields: List[Tuple[str, BaseRelationField]] = []
             # Gather all delete actions with payload and also all models to be deleted
             delete_actions: List[Tuple[Type[Action], ActionPayload]] = []
-            additional_relation_models: ModelMap = {}
+            additional_relation_models: ModelMap = {this_fqid: DeletedModel()}
             for field_name, field in self.model.get_relation_fields():
                 if field.structured_relation or field.structured_tag:
                     # TODO: We do not fully support these fields. So silently skip them.
@@ -350,7 +351,7 @@ class DeleteAction(Action):
                             ):
                                 raise ActionException(
                                     f"You can not delete {self.model} with id {instance['id']}, "
-                                    f"because you have to delete the related {str(field.to)} first."
+                                    f"because you have to delete the related model {fqid} first."
                                 )
                     else:
                         # field.on_delete == OnDelete.CASCADE
--- a/openslides_backend/action/relations.py
+++ b/openslides_backend/action/relations.py
@@ -7,7 +7,6 @@ from ..models.fields import (
     BaseRelationField,
     GenericRelationField,
     GenericRelationListField,
-    OnDelete,
     RelationField,
     RelationListField,
     TemplateRelationField,
@@ -390,16 +389,6 @@ class RelationsHandler:
                 )
             else:
                 assert rel_id in remove
-                if (
-                    self.type in ("1:1", "m:1")
-                    and self.reverse_field.on_delete == OnDelete.PROTECT
-                ):
-                    # Hint: There is no on_delete behavior in m:n cases. The reverse
-                    # field is always nullable. We just modifiy the related field list.
-                    raise ActionException(
-                        f"You are not allowed to delete {self.model} {self.id} as "
-                        "long as there are some required related objects."
-                    )
                 if self.type in ("1:1", "m:1"):
                     new_value = None
                 else:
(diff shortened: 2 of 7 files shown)


normanjaeckel:
See also https://github.com/OpenSlides/openslides-backend-coordination/issues/82

I added a commit for this problem but I am nut sure whether I understood your DeletedModel thing correctly.

tests/system/base.py
```diff
@@ -41,6 +44,14 @@ def setUp(self) -> None:
         )
         self.client = self.create_client(ADMIN_USERNAME, ADMIN_PASSWORD)
 
+    def load_example_data(self) -> None:
+        self.datastore.truncate_db()
+        example_data = json.loads(requests.get(self.EXAMPLE_DATA).content)
+        for collection, models in example_data.items():
+            for model in models:
+                fields = {k: v for k, v in model.items() if k != "id"}
```
jsangmeister, line 52: Why do you not write the id into the datastore? for some tests it's needed

openslides_backend/action/generics.py
```diff
@@ -336,7 +337,7 @@ def delete_action_prepare_dataset(self, payload: ActionPayload) -> DataSet:
             relation_fields: List[Tuple[str, BaseRelationField]] = []
             # Gather all delete actions with payload and also all models to be deleted
             delete_actions: List[Tuple[Type[Action], ActionPayload]] = []
-            additional_relation_models: ModelMap = {}
+            additional_relation_models: ModelMap = {this_fqid: DeletedModel()}
```
jsangmeister, line 340: This was missing and the reason why the test failed

openslides_backend/action/relations.py
```diff
@@ -390,16 +389,6 @@ def prepare_result_to_id(
...
-                if (
-                    self.type in ("1:1", "m:1")
-                    and self.reverse_field.on_delete == OnDelete.PROTECT
-                ):
-                    # Hint: There is no on_delete behavior in m:n cases. The reverse
-                    # field is always nullable. We just modifiy the related field list.
-                    raise ActionException(
-                        f"You are not allowed to delete {self.model} {self.id} as "
-                        "long as there are some required related objects."
-                    )
```
jsangmeister, line 402: I removed this since it threw an error when actually the related model was being deleted by the cascade deletion. Structurally, it also makes more sense to have this kind of check in the delete action, since it knows best about which models are deleted and which not. It seems to not have had any unforseen side effects, but please double check if it's ok to remove this here @normanjaeckel.

tests/system/action/motion_workflow/test_delete.py
```diff
@@ -40,6 +40,28 @@ def test_delete_with_states(self) -> None:
...
+        # needed because you can't delete the last workflow of a meeting
+        self.create_model("motion_workflow/100", {"meeting_id": 1})
+        response = self.client.post(
+            "/",
+            json=[{"action": "motion_workflow.delete", "data": [{"id": 2}]}],
+        )
+        self.assert_status_code(response, 200)
+        self.assert_model_deleted("motion_workflow/2")
+        self.assert_model_deleted("motion_state/3")
+
```
jsangmeister, line 64: This was the key point why the test failed, I extracted it into it's own test to not have to load all example data.

tests/system/base.py
normanjaeckel: It just because we do not do this in most (or all) other tests. I have no problem if you do so here.

openslides_backend/action/generics.py
normanjaeckel: Ok, nice.

tests/system/base.py
jsangmeister: It's added automatically in the `create_model` method for the above reason. I think it's better to not filter it out as it might create hard-to-debug errors.

openslides_backend/action/relations.py
normanjaeckel: Hm. What about an update case where the field is set to `None` or `[]`. In this case the reverse field is set to `None`. Maybe we want to prevent this case with the OnDelete flag. But maybe this is not correct and we have to use the `required` flag instead.

tests/system/action/motion_workflow/test_delete.py
normanjaeckel: Yes. We can leave the `load_example_data` in code for such testing cases but should not use it in permanent tests because it is too slow.

tests/system/action/topic/test_delete.py
```diff
@@ -67,3 +67,11 @@ def test_create_delete(self) -> None:
         self.assert_status_code(response, 200)
         self.assert_model_deleted("topic/1")
         self.assert_model_deleted("list_of_speakers/1")
+
+    def test_example_data(self) -> None:
```
normanjaeckel, line 71: Maybe also remove or change this test because load_example_data is a really slow test case.

tests/system/base.py
normanjaeckel: Ok.

openslides_backend/action/relations.py
jsangmeister: I am not sure I understand you example. If you think we need this (the current tests do not need it, at least), feel free to add it back, but you would've to fix the test again then :D

normanjaeckel: You can leave it out but please open an issue that I have to look at the "required" case again.

tests/system/action/topic/test_delete.py
jsangmeister: Changed.

PR #304, opened 2020-11-11
Add group.set_permission action + tests.

reiterl:


openslides_backend/action/group/set_permission.py
```diff
@@ -0,0 +1,35 @@
...
+        }
+    )
+
+    def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
+        set_ = instance.pop("set")
+        permission = instance.pop("permission")
+        group = self.datastore.get(
+            FullQualifiedId(self.model.collection, instance["id"]), ["permissions"]
+        )
+        instance["permissions"] = group["permissions"]
```
jsangmeister, line 30: I would suggest to only set the `permissions` field if something was changed, so we don't create unnecessary write requests.

reiterl: Done, if the permissions are not changed, the instance is not updated.

jsangmeister (approved):

PR #306, opened 2020-11-12
Update mediafile.update action: handle children in non parent case.

reiterl:
In the case, where a mediafile doesn't have a parent, the handle_children method was not called.
Changed this, in both cases the children are handled.


jsangmeister:
Please add a test which explicitly tests this, since it seemingly fell through with the other tests

reiterl:
Add a test for mediafile update children case.

jsangmeister (approved):
you can squash & merge

PR #307, opened 2020-11-12
Fix structured relation handling

jsangmeister:
Should fix the issue in https://github.com/OpenSlides/openslides-backend-coordination/issues/156

### DIFF
--- a/openslides_backend/action/base.py
+++ b/openslides_backend/action/base.py
@@ -208,7 +208,7 @@ class Action(BaseAction, metaclass=SchemaProvider):
                 r"^"
                 + field_name[: field.index]
                 + r"\$"
-                + r"([^_]*)"
+                + r"(.*)"
                 + field_name[field.index :]
                 + r"$"
             )
--- a/openslides_backend/action/generics.py
+++ b/openslides_backend/action/generics.py
@@ -11,7 +11,7 @@ from ..models.fields import (
 from ..shared.exceptions import ActionException
 from ..shared.interfaces.event import Event
 from ..shared.interfaces.write_request_element import WriteRequestElement
-from ..shared.patterns import ID_PATTERN, FullQualifiedId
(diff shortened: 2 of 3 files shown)


openslides_backend/action/base.py
```diff
@@ -208,7 +208,7 @@ def get_structured_fields_in_instance(
                 r"^"
                 + field_name[: field.index]
                 + r"\$"
-                + r"([^_]*)"
+                + r"(.*)"
```
jsangmeister, line 211: Prefix has to be unique among all structured fields, so we can match everything here. Especially important to match replacements with underscores. @FinnStutzenstein we should check in the validator of the models file that the prefixes are unique, please remember this when we move it :D

openslides_backend/action/generics.py
```diff
@@ -73,10 +73,6 @@ def create_action_prepare_dataset(self, payload: ActionPayload) -> DataSet:
                     )
                     for instance_field, replacement in structured_fields:
                         if isinstance(field, BaseTemplateRelationField):
-                            if not ID_PATTERN.match(replacement):
-                                raise ActionException(
-                                    "Template relation fields can only use replacements which are ids."
-                                )
```
jsangmeister, line 79: This was wrong: e.g. `logo_$_id` is a template relation field which can hold non-id replacements. If we want to differentiate between this and "normal" template relation fields, we would have to add a `TemplateRelationTagField` (or something like this). @normanjaeckel if you want, feel free to do this, but I think for now this suffices. We just have to care to only write valid data to the fields in the actions, but we have to care about that anyway.

openslides_backend/action/base.py
FinnStutzenstein: Ok, I was working on the validator this morning; I added a note.

normanjaeckel (approved):

PR #310, opened 2020-11-12
Remove errors when nothing to do

### LINKED ISSUE #247: Handling of actions which have nothing to do
Came across this question in https://github.com/OpenSlides/openslides-backend/pull/245/ as well as #229. In case the actions finds out it has nothing to do, both @reiterl and @normanjaeckel throw an exception. I think we should just ignore it. Especially in case of #245, one might just click the button to be sure everyone is deleted without actually checking it, and it might be confusing to present an error then. What do you think?

jsangmeister:
fixes #247 

Do we have more places where this happens? I only found the one (and #245, which is handled there)

### DIFF
--- a/openslides_backend/action/list_of_speakers/delete_all_speakers.py
+++ b/openslides_backend/action/list_of_speakers/delete_all_speakers.py
@@ -1,7 +1,6 @@
 from typing import Iterable, Union
 
 from ...models.models import ListOfSpeakers
-from ...shared.exceptions import ActionException
 from ...shared.patterns import FullQualifiedId
 from ..action_interface import ActionResponseResultsElement
 from ..base import Action, ActionPayload, DataSet, WriteRequestElement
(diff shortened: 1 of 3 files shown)


jsangmeister:
Also added https://github.com/OpenSlides/openslides-backend-coordination/issues/29. Now sets up on #245 to be able to handle empty write requests.

openslides_backend/action/generics.py
```diff
@@ -432,6 +432,12 @@ def delete_action_prepare_dataset(self, payload: ActionPayload) -> DataSet:
 
         return {"data": data}
 
+    def update_instances(self, payload: ActionPayload) -> ActionPayload:
```
normanjaeckel, line 435: What is the difference to `get_updated_instances` which is defined in base.py? This new method is not used in this PR.

normanjaeckel (changes requested):

openslides_backend/action/generics.py
jsangmeister: Ah, probably an error from the rebase. Will fix once #245 is merged. Good catch

jsangmeister: removed.

PR #311, opened 2020-11-12
Fix action and add relation test

jsangmeister:
see https://github.com/OpenSlides/openslides-backend-coordination/issues/64

@normanjaeckel can you take a look at this? I added a test for it and it seems like structured relations are not updated automatically.

### DIFF
--- a/openslides_backend/action/actions/meeting/base_set_mediafile_action.py
+++ b/openslides_backend/action/actions/meeting/base_set_mediafile_action.py
@@ -1,0 +1,52 @@
+from typing import Any, Dict, List
+
+from ....models.models import Meeting
+from ....shared.exceptions import ActionException
+from ....shared.patterns import Collection, FullQualifiedId
+from ....shared.schema import required_id_schema
+from ...generics.update import UpdateAction
+from ...util.default_schema import DefaultSchema
+
+
+class BaseMeetingSetMediafileAction(UpdateAction):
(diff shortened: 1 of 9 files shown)


jsangmeister:
Rebased on #328. Test was not fixed, since it's another case (reverse structured field of a structured field). So please fix this @normanjaeckel

normanjaeckel:
Please rebase again on #328. I added some fixes to #328 that might help also for this test here.

jsangmeister:
Rebased. Test is not fixed.

normanjaeckel:
Fixed. Please reflect change also in #337.

openslides_backend/action/relations.py
```diff
@@ -227,7 +227,7 @@ def perform(self) -> Relations:
         #  - All other fields.
         if self.field.generic_relation:
             return self.prepare_result_to_fqid(add, remove, rels, related_name)
-        elif self.field.structured_relation:
+        elif self.field.structured_relation or self.field.structured_tag:
```
normanjaeckel, line 230: This was the fix.

jsangmeister: Well, that was easy...

jsangmeister:
> Fixed. Please reflect change also in #337.

Yes, thank you. Will merge this probably after #337 and take care of the merge conflicts here.

PR #312, opened 2020-11-12
Fix user.create_temporary

jsangmeister:
Added `vote_delegations_from_ids` and did a little cleanup


jsangmeister:
Currently it is only checked that the vote delegation users exists. Please check if more validation is needed

normanjaeckel:
No comments from me.

PR #313, opened 2020-11-12
Fixes auth-lib-dependency

GabrielInTheWorld:


jsangmeister (approved):

dev/Dockerfile-dev
```diff
@@ -23,5 +23,6 @@ EXPOSE 9002
 EXPOSE 9003
 
 ENV OPENSLIDES_BACKEND_DEBUG 1
+ENV OS4_DEVELOPMENT 1
```
normanjaeckel, line 26: Is this an env variable defined by the authlib? Please add an explaination to the README.

FinnStutzenstein: We'll add this env variable for every service, since multiple service need the info "I'm in development mode" and we just use `OS4_DEVELOPMENT` for every service

jsangmeister: It's essentially the same as previously `OPENSLIDES_BACKEND_DEBUG`. We should drop the latter and just use `OS4_DEVELOPMENT` in the backend.

PR #323, opened 2020-11-18
Committee.create

r-peschke:
committee.create #113 ready

### DIFF
--- a/dev/Dockerfile-dev
+++ b/dev/Dockerfile-dev
@@ -18,6 +18,7 @@ COPY Makefile Makefile
 COPY setup.cfg .
 
 RUN chmod 777 -R .
+RUN chmod 777 /root
 
 EXPOSE 9002
 EXPOSE 9003
--- a/setup.cfg
+++ b/setup.cfg
@@ -27,7 +27,7 @@ exclude_lines =
     pragma: no cover
     raise NotImplementedError
     return NotImplemented
-fail_under = 95
+fail_under = 97
 
 [mypy-fastjsonschema]
 ignore_missing_imports = true
--- a/tests/system/action/committee/test_create.py
+++ b/tests/system/action/committee/test_create.py
@@ -1,1 +1,137 @@
-# TODO
+from tests.system.action.base import BaseActionTestCase
+
+
+class CommitteeCreateActionTest(BaseActionTestCase):
+    def test_create(self) -> None:
+        self.create_model("organisation/1", {"name": "test_organisation1"})
+        self.create_model("user/20", {"username": "test_user20"})
+        self.create_model("user/21", {"username": "test_user21"})
+        self.create_model("user/22", {"username": "test_user22"})
+        committee_name = "test_committee1"
+        description = "<p>Test Committee</p>"
+
+        response = self.client.post(
+            "/",
+            json=[
+                {
+                    "action": "committee.create",
+                    "data": [
+                        {
+                            "name": committee_name,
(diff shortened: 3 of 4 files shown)


setup.cfg
```diff
@@ -27,7 +27,7 @@ exclude_lines =
     pragma: no cover
     raise NotImplementedError
     return NotImplemented
-fail_under = 95
+fail_under = 59
```
FinnStutzenstein, line 30: @jsangmeister Is this value not checked in the ci? Locally it gave Ralf errors, I guess it is not in sync with the GH actions. Either a `--fail-under` is missing [here](https://github.com/OpenSlides/openslides-backend/blob/master/dev/run-tests.sh#L9), or I wonder why the marked line is not respected in the first case (maybe this file is not in the container?)

jsangmeister: Yes, it is. See https://github.com/OpenSlides/openslides-backend/pull/323/checks?check_run_id=1418394701, section `Run tests`, line 2393. It is also the correct value, in my local instance (and also in GH actions, as you can see) the coverage is over 98%. Probably a configuring error...

dev/Dockerfile-dev
```diff
@@ -18,6 +18,7 @@ COPY Makefile Makefile
 COPY setup.cfg .
 
 RUN chmod 777 -R .
+RUN chmod 777 /root
```
jsangmeister, line 21: Why do you need this?

jsangmeister (approved):
Looks good besides the two remarks from Finn and me!

dev/Dockerfile-dev
r-peschke: This helps VS Code to attach and install it's the remote debugger in the container. Honor is to Finn;-). Next step could be to debug the code really in VSCode instead of using pdb

jsangmeister: Ah, ok

tests/system/action/committee/test_create.py
```diff
@@ -1 +1,120 @@
-# TODO
+from tests.system.action.base import BaseActionTestCase
+
+
+class CommitteeCreateActionTest(BaseActionTestCase):
+    def test_create(self) -> None:
```
jsangmeister, line 5: Ah, you should probably add one more test where only the required fields are given to make sure this works as intended.

jsangmeister (changes requested):

jsangmeister (approved):
Ok, you can squash and merge this.

tests/system/action/committee/test_create.py
```diff
@@ -1 +1,137 @@
...
+                            "member_ids": [20, 21],
+                            "manager_ids": [20, 22],
+                        }
+                    ],
+                }
+            ],
+        )
+        self.assert_status_code(response, 200)
+        model = self.get_model("committee/1")
+        assert model.get("name") == committee_name
```
normanjaeckel, line 32: Hello @r-peschke. Thank you for joining this team. I hope you enjoy working and co-working here ... ;-)

In such lines it might be better to use `self.assertEqual(model.get("name"), committee_name)` but you do not have to change it in this closed case.

jsangmeister: I had the same thought some while ago and talked about this with @reiterl who uses `assert ...` consistently. We came to the conclusion that it really does not make a difference and since the existing tests don't follow a common pattern with this, everyone can do what he wants ;)

PR #326, opened 2020-11-19
Make admin active

jsangmeister:


jsangmeister:
needed because of auth service update

PR #327, opened 2020-11-20
Committee.delete

r-peschke:
committee.delete including tests

### DIFF
--- a/tests/system/action/committee/test_delete.py
+++ b/tests/system/action/committee/test_delete.py
@@ -1,0 +1,75 @@
+from tests.system.action.base import BaseActionTestCase
+
+
+class CommitteeDeleteActionTest(BaseActionTestCase):
+    COMMITTEE_ID = 1
+    COMMITTEE_FQID = "committee/1"
+
+    def create_data(self) -> None:
+        self.create_model("organisation/1", {"name": "test_organisation1"})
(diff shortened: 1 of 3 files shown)


tests/system/action/committee/test_delete.py
```diff
@@ -0,0 +1,71 @@
...
+        self.create_data()
+        response = self.client.post(
+            "/",
+            json=[{"action": "committee.delete", "data": [{"id": self.COMMITTEE_ID}]}],
+        )
+
+        self.assert_status_code(response, 200)
+        self.assert_model_deleted(self.COMMITTEE_FQID)
+        self.assert_model_exists("user/20")
+        self.assert_model_exists("user/21")
```
jsangmeister, line 41: Rather than just checking that the users still exists, you can check that the relations are updated correctly. You can just provide a dict as the second argument which is checked against the model, e.g. in this case `{"member_in_meeting_ids": [], "manager_in_meeting_ids": []}` (I think these are the names of the reverse fields).

jsangmeister (commented):
Looks good!

tests/system/action/committee/test_delete.py
r-peschke: Done

normanjaeckel:
You also have to rebase because of dependency changes.

jsangmeister (approved):

PR #328, opened 2020-11-20
Test the tests & fix them

jsangmeister:
Added the `validate_fields` method in the base test case to validate all models that are created and THERE WERE SO MANY ERRORS. Especially related to structured fields. And @reiterl: not every model has a name, besides the point that it is almost never needed for the tests :D

@normanjaeckel: I fixed the structured relation/template field errors, but the way I did it seems kinda hacky and not the most elegant. PLEASE add some comments to your relation handling code, I spend hours to fix these errors and still do not understand what exactly is going on there. Also check if you find a more elegant and uniform way of handling these fields.

### DIFF
--- a/openslides_backend/action/base.py
+++ b/openslides_backend/action/base.py
@@ -208,15 +208,7 @@ class Action(BaseAction, metaclass=SchemaProvider):
         """
         structured_fields: List[Tuple[str, str]] = []
         for instance_field in instance.keys():
-            regex = (
-                r"^"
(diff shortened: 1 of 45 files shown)


normanjaeckel:
Resolved conflicts added some more fixes and restructured it a little bit. Also added some comments.

normanjaeckel:
Ready to merge.

jsangmeister:
Thanks for your cleanup! Looks good.

PR #332, opened 2020-11-24
Update motion.update_metadata

reiterl:
Include a set_recommendation_extension_reference_ids method and use it.
Also add some tests/update tests.

### DIFF
--- a/tests/system/action/motion/test_update_metadata.py
+++ b/tests/system/action/motion/test_update_metadata.py
@@ -4,12 +4,25 @@ from tests.system.action.base import BaseActionTestCase
 class MotionUpdateMetadataActionTest(BaseActionTestCase):
     def test_update_correct(self) -> None:
         self.create_model("meeting/2538", {"name": "name_jkPIYjFz"})
-        self.create_model("motion/111", {"meeting_id": 2538})
         self.create_model(
-            "motion_category/4", {"meeting_id": 2538, "name": "name_GdPzDztT"}
+            "motion/111",
+            {
+                "meeting_id": 2538,
+            },
         )
         self.create_model(
-            "motion_block/51", {"meeting_id": 2538, "title": "title_ddyvpXch"}
+            "motion_category/4",
+            {"meeting_id": 2538, "name": "name_GdPzDztT", "motion_ids": []},
+        )
+        self.create_model(
+            "motion_block/51",
+            {"meeting_id": 2538, "title": "title_ddyvpXch", "motion_ids": []},
+        )
+        self.create_model(
+            "motion/112",
+            {
+                "meeting_id": 2538,
+            },
         )
 
         response = self.client.post(
@@ -21,7 +34,7 @@ class MotionUpdateMetadataActionTest(BaseActionTestCase):
                         {
                             "id": 111,
                             "state_extension": "test_blablab_noon",
-                            "recommendation_extension": "ext_sldennt",
+                            "recommendation_extension": "ext_sldennt [motion/112]",
                             "category_id": 4,
                             "block_id": 51,
                             "supporter_ids": [],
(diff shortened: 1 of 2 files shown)


FinnStutzenstein:
We have to modify the behaviour of non-existant motions a bit:

You currently raise an ActionException if the motion does not exist. This is problematic if the referenced motion is deleted. Than the deleted motion is removed from `recommendation_extension_reference_ids` (this should be tested), but not from the actual `recommendation_extension`. Since the `recommendation_extension` is a handwritten userinput, we should not modify it but accept, that there are non-existing references (The client shows them as `<undefined model motion/X>` or so to hint the user, that the data is not valid). So I prefer just ignoring non-exitent models and do not put them into `recommendation_extension_reference_ids`. But still raise an error, if the collection is not supported (like `assignment/3`), or the referenced model is not from the same meeting.

jsangmeister:
> But still raise an error, if the collection is not supported (like `assignment/3`), or the referenced model is not from the same meeting.

Please do that and also add a test for this.

Also add a test that the motions must all be of the same meeting, even if this is handled by the equal_fields check.

reiterl:
Added two checks:
* same meeting_ids in motions.
* collections must be motions

Added some tests.

openslides_backend/action/motion/update_metadata.py
```diff
@@ -61,15 +62,31 @@ def set_recommendation_extension_reference_ids(
...
             exists_res = self.datastore.exists(
                 collection=Collection(collection),
                 filter=FilterOperator("id", "=", int(id_)),
             )
             if exists_res.get("exists"):
+                meeting_id = self.get_meeting_id(id_)
+                if orig_meeting_id != meeting_id:
+                    raise ActionException(
+                        f"Meeting_ids of motion {instance['id']} and {id_} don't match."
+                    )
```
jsangmeister, line 80: You misunderstood me, the check itself is already handled by the `equal_fields`. Only a test for this was needed, so you should be able to remove this again with the test still working (apart from a different error message).

jsangmeister (changes requested):
Also a rebase is necessary

openslides_backend/action/motion/update_metadata.py
reiterl: This is not handled by equal_fields. I tested it. So my check is correct.

jsangmeister: You are right, but it should be. I fixed it in https://github.com/OpenSlides/openslides-backend/pull/354. Please rebase this branch and remove your manual check

reiterl: Rebased and removed the check. I needed to update the generic field handling too.

tests/system/action/motion/test_update_metadata.py
```diff
@@ -4,7 +4,7 @@
 class MotionUpdateMetadataActionTest(BaseActionTestCase):
     def test_update_correct(self) -> None:
         self.create_model("meeting/2538", {"name": "name_jkPIYjFz"})
-        self.create_model("motion/111", {"meeting_id": 2538})
+        self.create_model("motion/111", {"meeting_id": 2538, "referenced_in_motion_recommendation_extension_ids": []})
```
jsangmeister: This should not be necessary since `[] == None` for the relation handling. Does it have a specific reason you added it?

jsangmeister (approved):
Can be squashed and merged

tests/system/action/motion/test_update_metadata.py
reiterl: No, I just want to make sure, that this is not the reason of an error.

jsangmeister: ok

reiterl: I have removed the `referenced_in_motion_recommendation_extension_ids` again.

PR #334, opened 2020-11-25
Implement actions for role including tests

r-peschke:
Implements the actions for role-object

### DIFF
--- a/openslides_backend/action/role/deduplicate_permissions_mixin.py
+++ b/openslides_backend/action/role/deduplicate_permissions_mixin.py
@@ -1,0 +1,14 @@
+from typing import Any, Dict
+
+from ..base import Action
+
+
+class DeduplicatePermissionsMixin(Action):
+    def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
+        """
+        Deduplicate permissions
+        """
+        instance = super().update_instance(instance)
+        if "permissions" in instance:
+            instance["permissions"] = list(set(instance["permissions"]))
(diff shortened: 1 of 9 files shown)


FinnStutzenstein:
As discussed: Plase make `permissions` unique. E.g. if the user sends `["a", "a", "b"]` the result should be `["a", "b"]` (the ordering is irrelevant, so `["b", "a"]` is also fine).

openslides_backend/action/role/deduplicate_permissions_mixin.py
```diff
@@ -0,0 +1,15 @@
...
+from ..base import Action
+
+
+class DeduplicatePermissionsMixin(Action):
+    def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
+        """
+        Deduplicate permissions
+        """
+        instance = super().update_instance(instance)
+        if instance.get("permissions") is None:
```
FinnStutzenstein, line 12:
```suggestion
        if "permissions" in instance:
            instance["permissions"] = list(set(instance["permissions"]))
```

FinnStutzenstein (approved):
One little change suggested. @jsangmeister or @normanjaeckel may approve and merge, if everything is fine.

jsangmeister (approved):

openslides_backend/action/role/deduplicate_permissions_mixin.py
r-peschke: Done

PR #335, opened 2020-11-26
sequence_number per meeting with checking race condition ready

r-peschke:
… tested

Still WIP, missing handling and testing of race condition

### DIFF
--- a/openslides_backend/action/action.py
+++ b/openslides_backend/action/action.py
@@ -77,7 +77,7 @@ class Action(BaseAction, metaclass=SchemaProvider):
     ) -> None:
         self.services = services
         self.permission = services.permission()
-        self.datastore = services.datastore()
+        self.datastore = relation_manager.datastore
         self.auth = services.authentication()
         self.media = services.media()
         self.relation_manager = relation_manager
--- a/openslides_backend/action/action_handler.py
+++ b/openslides_backend/action/action_handler.py
@@ -1,8 +1,9 @@
+from copy import deepcopy
 from typing import Dict, Iterable, List, Optional, Tuple, Union
 
 import fastjsonschema
 
-from ..shared.exceptions import ActionException, EventStoreException
+from ..shared.exceptions import ActionException, DatastoreException, EventStoreException
 from ..shared.handlers.base_handler import BaseHandler
 from ..shared.interfaces.write_request_element import WriteRequestElement
 from ..shared.schema import schema_version
@@ -48,6 +49,8 @@ class ActionHandler(BaseHandler):
     Action handler. It is the concret implementation of Action interface.
     """
 
+    MAX_RETRY = 3
+
     @classmethod
     def get_actions_dev_status(cls) -> Iterable[Tuple[str, Union[str, Dict]]]:
         """
@@ -72,19 +75,25 @@ class ActionHandler(BaseHandler):
         except fastjsonschema.JsonSchemaException as exception:
             raise ActionException(exception.message)
 
-        # TODO: Start a loop here and retry parsing actions and writing to event
-        # store for some time if event store sends ModelLocked Exception
+        retried = 0
+        payload_copy = deepcopy(payload)
+        while True:
+            # Parse actions and creates events
+            write_request_element, results = self.parse_actions(payload)
 
-        # Parse actions and creates events
-        write_request_element, results = self.parse_actions(payload)
-
-        # Send events to datastore
-        if write_request_element:
-            try:
-                self.datastore.write(write_request_element)
-            except EventStoreException as exception:
-                raise ActionException(exception.message)
-
+            # Send events to datastore
+            if write_request_element:
+                try:
+                    self.datastore.write(write_request_element)
+                except DatastoreException as exception:
+                    retried += 1
+                    payload = deepcopy(payload_copy)
+                    if retried > self.MAX_RETRY:
+                        raise ActionException(exception.message)
+                    continue
+                except EventStoreException as exception:
+                    raise ActionException(exception.message)
+            break
         # Return action result
         # TODO: This is a fake result because in this place all actions were
         # always successful.
--- a/openslides_backend/services/datastore/adapter.py
+++ b/openslides_backend/services/datastore/adapter.py
@@ -7,7 +7,12 @@ from ...shared.exceptions import DatastoreException
 from ...shared.filters import And, Filter, FilterOperator
 from ...shared.interfaces.logging import LoggingModule
 from ...shared.interfaces.write_request_element import WriteRequestElement
-from ...shared.patterns import Collection, FullQualifiedField, FullQualifiedId
+from ...shared.patterns import (
+    Collection,
+    CollectionField,
+    FullQualifiedField,
+    FullQualifiedId,
+)
 from . import commands
 from .deleted_models_behaviour import DeletedModelsBehaviour
 from .http_engine import HTTPEngine as Engine
@@ -22,7 +27,7 @@ class DatastoreAdapter(DatastoreService):
     Adapter to connect to readable and writeable datastore.
     """
 
-    # The key of this dictionary is a stringified FullQualifiedId or FullQualifiedField
+    # The key of this dictionary is a stringified FullQualifiedId or FullQualifiedField or CollectionField
     locked_fields: Dict[str, int]
 
     def __init__(self, engine: Engine, logging: LoggingModule) -> None:
@@ -268,8 +273,16 @@ class DatastoreAdapter(DatastoreService):
         return response
 
     def max(
-        self, collection: Collection, filter: Filter, field: str, type: str = None
+        self,
+        collection: Collection,
+        filter: Filter,
+        field: str,
+        type: str = None,
+        lock_result: bool = False,
     ) -> Aggregate:
+        """
+        Per default records, marked as deleted, are counted
+        """
         # TODO: This method does not reflect the position of the fetched objects.
         command = commands.Max(
             collection=collection, filter=filter, field=field, type=type
@@ -278,23 +291,22 @@ class DatastoreAdapter(DatastoreService):
             f"Start MAX request to datastore with the following data: {command.data}"
         )
         response = self.retrieve(command)
+        if lock_result:
+            self.update_locked_fields(
+                CollectionField(collection, field), response.get("position")
+            )
         return response
 
     def update_locked_fields(
         self,
-        key: Union[FullQualifiedId, FullQualifiedField],
+        key: Union[FullQualifiedId, FullQualifiedField, CollectionField],
         position: int,
     ) -> None:
         """
         Updates the locked_fields map by adding the new value for the given FQId or
-        FQField. If there is an existing value we take the smaller one.
+        FQField. To work properly in case of retry/reread we have to accept the new value always.
         """
-        current_position = self.locked_fields.get(str(key))
-        if current_position is None:
-            new_position = position
-        else:
-            new_position = min(position, current_position)
-        self.locked_fields[str(key)] = new_position
+        self.locked_fields[str(key)] = position
 
     def reserve_ids(self, collection: Collection, amount: int) -> Sequence[int]:
         command = commands.ReserveIds(collection=collection, amount=amount)
--- a/tests/system/action/lock.py
+++ b/tests/system/action/lock.py
@@ -1,0 +1,42 @@
+import threading
+from contextlib import contextmanager
+from typing import Iterator
+
+from openslides_backend.services.datastore.adapter import DatastoreAdapter
+from openslides_backend.shared.interfaces.write_request_element import (
+    WriteRequestElement,
+)
+
+pytest_thread_local = threading.local()
+
+
+@contextmanager
+def monkeypatch_datastore_adapter_write() -> Iterator[None]:
+    """
+    Use a patched DatastoreAdapter.write in this context,
+    which wraps the write with an optional lock from pytest_thread_local.testlock
+    """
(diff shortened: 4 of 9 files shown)


jsangmeister (approved):
Looks good so far

openslides_backend/shared/lock.py
```diff
@@ -0,0 +1,3 @@
+import threading
```
FinnStutzenstein, line 1: Please add a comment, why this code is here? What does the lock solve?

r-peschke: With last commit there is more stuff in this file.

1. There is the instance of the thread.local-store to share by import: pytest_thread_local
2. There is the Context-Manager to apply and restore the MonkeyPatch for the write-method of the datastore-adapter
3. There is the write-Method to patch the original from adapter. It wraps the original write to inject the threading.lock to stop one thread and let another thread pass the first one.

The lock itself is created and controlled by the using test. An example to use this in conjunction see in tests_create_sequence_number.py Test: test_create_sequence_numbers_race_condition with race condition

openslides_backend/action/action_handler.py
```diff
@@ -75,16 +78,25 @@ def handle_request(self, payload: Payload, user_id: int) -> ActionResponse:
...
+        payload_copy = deepcopy(payload)
+        while True:
+            # Parse actions and creates events
+            write_request_element, results = self.parse_actions(payload)
+
+            # Send events to datastore
+            if write_request_element:
+                try:
+                    self.datastore.write(write_request_element)
+                except DatastoreException as exception:
```
FinnStutzenstein, line 91: What is the difference between `DatastoreException` and `EventStoreException`?

openslides_backend/shared/lock.py
```diff
@@ -0,0 +1,49 @@
...
+    """
+    Wraps the write of the datastore.adapter and stops a thread, if the testlock attribute is set thread-local.
+    See example in test_create_sequence_numbers_race_condition
+
+    Hint:
+    You could use the commented wre-Variable as blueprint to see/print the data of write_request_element
+    """
+    # wre = ", ".join([f"{str(e['type'])} {str(e['fqid'])}" for e in write_request_element.events])
+    threadname = (
+        pytest_thread_local.name if hasattr(pytest_thread_local, "name") else "kein TN"
```
FinnStutzenstein, line 38: please translate `kein TN`.

openslides_backend/shared/lock.py
```diff
@@ -0,0 +1,49 @@
...
+
+def write(self, write_request_element: WriteRequestElement) -> None:  # type: ignore
+    """
+    Wraps the write of the datastore.adapter and stops a thread, if the testlock attribute is set thread-local.
+    See example in test_create_sequence_numbers_race_condition
+
+    Hint:
+    You could use the commented wre-Variable as blueprint to see/print the data of write_request_element
+    """
+    # wre = ", ".join([f"{str(e['type'])} {str(e['fqid'])}" for e in write_request_element.events])
```
FinnStutzenstein, line 36: remove, if not needed.

tests/system/action/motion/test_create_sequence_number.py
```diff
@@ -0,0 +1,256 @@
...
+            },
+        )
+        self.create_model(
+            "motion_state/" + state_str,
+            {"name": "name_state" + state_str, "meeting_id": meeting_id},
+        )
+
+    def test_create_sequence_numbers(self) -> None:
+        self.create_model("meeting/222", {"name": "meeting222"})
+        self.create_workflow(workflow_id=12, meeting_id=222)
```
FinnStutzenstein, line 31: You can use your defaults here.

openslides_backend/shared/lock.py
FinnStutzenstein: if the file is only used by tests, move it to a test-folder.

openslides_backend/shared/lock.py
```diff
@@ -0,0 +1,49 @@
...
+from openslides_backend.services.datastore.adapter import DatastoreAdapter
+from openslides_backend.shared.interfaces.write_request_element import (
+    WriteRequestElement,
+)
+
+pytest_thread_local = threading.local()
+
+
+@contextmanager
+def monkeyPatch_DatastoreAdapter_write() -> Iterator[None]:
```
FinnStutzenstein, line 14: Please use snak_case for the name

tests/system/action/motion/test_create_sequence_number.py
```diff
@@ -0,0 +1,256 @@
...
+        response = self.client.post(
+            "/",
+            json=[
+                {
+                    "action": "motion.create",
+                    "data": [
+                        {
+                            "title": "motion_title2",
+                            "meeting_id": 222,
+                            "workflow_id": 13,
```
FinnStutzenstein, line 63: why are you creating a different workflow here? It should not make a difference, if you use workflow 12.

tests/system/action/motion/test_create_sequence_number.py
```diff
@@ -0,0 +1,256 @@
...
+                    "meeting_id": 222,
+                    "workflow_id": 13,
+                    "testlock": None,
+                    "name": "Passing Thread",
+                },
+            )
+
+            testlock.acquire()
+            thread1.start()
+            sleep(1)
```
FinnStutzenstein, line 216: Am I correct, that you need the sleep, so you are sure, that thread 1 reads first, before theard 2 writes new data?
Maybe you can use an [event](https://docs.python.org/3/library/threading.html#event-objects) here: The event is set in the mocked datastore write method just before acquiring the lock. The main thread can wait for the event and start thread 2.

This would be the execution order:
- main thread: acquire the lock, so thread 1 is blocked
- main thread: start thread 1 and wait for the event
- thread 1: fetches data, set the event, wait for the lock
- main thread: get the event, so thread 2 can be started
- main thread: wait for thread 2 to finish, release lock
- thread 1 finishes with some error

Or with the retires: no error. Maybe it should be asserted, that a retry happened.

This would require to add another element next to the lock, but since you have the structure already it should be fairly simple.

tests/system/action/motion/test_create_sequence_number.py
```diff
@@ -0,0 +1,256 @@
...
+            thread2.start()
+            thread2.join()
+            testlock.release()
+            thread1.join()
+
+        self.assert_model_not_exists("motion/1")
+        model2 = self.get_model("motion/2")
+        model3 = self.get_model("motion/3")
+        self.assertNotEqual(model3["sequence_number"], model2["sequence_number"])
```
FinnStutzenstein, line 225: Please be specific. I await that the thread 2 produces 1 and the halted thread 2.

FinnStutzenstein (changes requested):

openslides_backend/shared/lock.py
r-peschke: Done

openslides_backend/action/action_handler.py
r-peschke: DatastoreException is thrown in case of Datastore Problems like locked fields/models. EventStoreException is defined and imported and caught once, but seems to be never raised up to now.

openslides_backend/shared/lock.py
r-peschke: done

openslides_backend/shared/lock.py
r-peschke: done

tests/system/action/motion/test_create_sequence_number.py
r-peschke: Done

openslides_backend/shared/lock.py
r-peschke: Done

tests/system/action/motion/test_create_sequence_number.py
r-peschke: Done

tests/system/action/motion/test_create_sequence_number.py
r-peschke: This is specific:
Thread-1 gets the next id=1, but has to wait
Thread-2 gets and stores its motion with id=2
Thread-1 continues after acquiring the lock, but fails (id=1 is lost) and retries. On retry it gets the id=3

tests/system/action/motion/test_create_sequence_number.py
r-peschke: Done

openslides_backend/action/action_handler.py
FinnStutzenstein: Can you refactor this in another PR? I'll create an issue...

tests/system/action/motion/test_create_sequence_number.py
FinnStutzenstein:
```suggestion
        self.assertEqual(model2["sequence_number"], 2)
        self.assertEqual(model3["sequence_number"], 3)
```

tests/system/action/lock.py
```diff
@@ -0,0 +1,53 @@
...
+    Wraps the write of the datastore.adapter and stops a thread, if the testlock attribute is set thread-local.
+    See example in test_create_sequence_numbers_race_condition
+    """
+    threadname = (
+        pytest_thread_local.name
+        if hasattr(pytest_thread_local, "name")
+        else "No threadname"
+    )
+    if hasattr(pytest_thread_local, "testlock") and pytest_thread_local.testlock:
+        print(f"{threadname}: Acquiring")
```
FinnStutzenstein, line 39: These prints are nice for debugging. But I think they should be removed, if this should be merged.

tests/system/action/motion/test_create_sequence_number.py
r-peschke: Done
except the sequence_numbers are 1 and 2 for the id's 2 and 3 ;-)

FinnStutzenstein: Ah yes

FinnStutzenstein (approved):

tests/system/action/lock.py
r-peschke: Done

tests/system/action/motion/test_create_sequence_number.py
```diff
@@ -0,0 +1,262 @@
...
+                        }
+                    ],
+                }
+            ],
+        )
+        self.assert_status_code(response, 200)
+        model = self.get_model("motion/2")
+        self.assertEqual(model.get("sequence_number"), 2)
+
+    def test_create_sequence_numbers_race_condition(self) -> None:
```
jsangmeister, line 171: Please move the race condition tests to a seperate file

tests/system/action/motion/test_create_sequence_number.py
```diff
@@ -0,0 +1,262 @@
...
+
+        The lock-object will be shared in threading.local(), instance created in lock.py.
+        If possible you should pass as an argument to the thread function(s).
+        """
+        pytest_thread_local.name = "MainThread_RC"
+        self.create_model("meeting/222", {"name": "meeting222"})
+        self.create_workflow(workflow_id=12, meeting_id=222)
+        self.create_workflow(workflow_id=13, meeting_id=222)
+
+        # with MonkeyPatch_DatastoreAdapter_write():
```
jsangmeister, line 186: remove

jsangmeister (commented):
Sorry I'm late to the party :)

openslides_backend/action/action.py
```diff
@@ -77,7 +77,7 @@ def __init__(
     ) -> None:
         self.services = services
         self.permission = services.permission()
-        self.datastore = services.datastore()
+        self.datastore = relation_manager.datastore
```
normanjaeckel, line 80: Formerly we initiate the datastore object like we do with permission object, auth object and media object. Now this is not done (in the Relation Manager it is not done). Why?

r-peschke: The instance of the datastore tracks the locked_fields. Therefore it is necessary to use only one instance during the handling of a request. This instance is created in the request-handler, explicitly in BaseHandler and should be "distributed" to the dependent classes like the Action class or RelationManager.

normanjaeckel: Ok, I understand. In #358 I propose a more explicit way.

openslides_backend/services/datastore/adapter.py
```diff
@@ -278,23 +291,22 @@ def max(
...
-        FQField. If there is an existing value we take the smaller one.
+        FQField. To work properly in case of retry/reread we have to accept the new value always.
         """
-        current_position = self.locked_fields.get(str(key))
-        if current_position is None:
-            new_position = position
-        else:
-            new_position = min(position, current_position)
-        self.locked_fields[str(key)] = new_position
+        self.locked_fields[str(key)] = position
```
normanjaeckel, line 309: Please check again: We have two cases: We always have to take the smallest position of a field to be locked if we are inside  an action (e. g. an action fetches the same model multiple times. The first position should be locked not the last.). If we are outside and want to retry in the retry-loop introduced with this PR of course the locked_fields property has to be reset.

Am I right?

r-peschke: Yes, I'm really sorry, but I think you are right! If there are 2 values, for whatever reason, we have to take the smaller one. On retry we have to reset all locked_fields. I'll implement this in the continuing issue #131.

normanjaeckel: Do you mean #351 ?

r-peschke: yes, # 131 was the formerly number, before Finn reorganized the system.

PR #336, opened 2020-11-26
Add set_number_mixin and update motion.create.

reiterl:
Add some tests to motion create.
Uses a new model. Needs the model to be updated to fully pass the tests.

### DIFF
--- a/tests/system/action/motion/test_set_number_mixin.py
+++ b/tests/system/action/motion/test_set_number_mixin.py
@@ -1,0 +1,1225 @@
+from typing import Any, Dict
+
+from werkzeug.wrappers import Response
+
+from tests.system.action.base import BaseActionTestCase
+
+
+class MotionSetNumberMixinTest(BaseActionTestCase):
+    def test_create_set_number_return_because_number_preset(self) -> None:
+        self.create_model(
+            "meeting/222", {"name": "name_SNLGsvIV", "motions_number_type": "manually"}
+        )
+        self.create_model(
+            "motion_workflow/12",
+            {
+                "name": "name_workflow1",
+                "first_state_id": 34,
+                "state_ids": [34],
+                "meeting_id": 222,
+            },
+        )
+        self.create_model(
+            "motion_state/34",
+            {"name": "name_state34", "meeting_id": 222, "set_number": True},
+        )
+        response = self.client.post(
+            "/",
+            json=[
+                {
+                    "action": "motion.create",
+                    "data": [
+                        {
+                            "title": "test_Xcdfgee",
+                            "meeting_id": 222,
+                            "workflow_id": 12,
+                            "text": "test",
+                            "number": "A003",
+                        }
+                    ],
+                }
+            ],
+        )
+        self.assert_status_code(response, 200)
+        model = self.get_model("motion/1")
+        assert model.get("title") == "test_Xcdfgee"
+        assert model.get("meeting_id") == 222
+        assert model.get("number") == "A003"
+
+    def test_create_set_number_return_because_number_type_manually(self) -> None:
+        self.create_model(
+            "meeting/222", {"name": "name_SNLGsvIV", "motions_number_type": "manually"}
+        )
+        self.create_model(
+            "motion_workflow/12",
+            {
+                "name": "name_workflow1",
+                "first_state_id": 34,
+                "state_ids": [34],
+                "meeting_id": 222,
+            },
+        )
+        self.create_model(
+            "motion_state/34",
+            {"name": "name_state34", "meeting_id": 222, "set_number": True},
+        )
+        response = self.client.post(
+            "/",
+            json=[
+                {
+                    "action": "motion.create",
+                    "data": [
+                        {
+                            "title": "test_Xcdfgee",
+                            "meeting_id": 222,
+                            "workflow_id": 12,
+                            "agenda_create": True,
+                            "text": "test",
+                        }
+                    ],
+                }
+            ],
+        )
+        self.assert_status_code(response, 200)
+        model = self.get_model("motion/1")
+        assert model.get("title") == "test_Xcdfgee"
+        assert model.get("meeting_id") == 222
+        assert model.get("number") is None
+        agenda_item = self.get_model("agenda_item/1")
+        self.assertEqual(agenda_item.get("meeting_id"), 222)
+        self.assertEqual(agenda_item.get("content_object_id"), "motion/1")
+
+    def test_create_set_number_good(self) -> None:
+        self.create_model("meeting/222", {"name": "name_SNLGsvIV"})
+        self.create_model(
+            "motion_workflow/12",
+            {"name": "name_workflow1", "first_state_id": 34, "state_ids": [34]},
+        )
+        self.create_model(
+            "motion_state/34",
+            {"name": "name_state34", "meeting_id": 222, "set_number": True},
+        )
+        response = self.client.post(
+            "/",
+            json=[
+                {
+                    "action": "motion.create",
+                    "data": [
+                        {
+                            "title": "test_Xcdfgee",
+                            "meeting_id": 222,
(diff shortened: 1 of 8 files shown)


jsangmeister:
@FinnStutzenstein can you check the tests? I think you are more into the topic than me

openslides_backend/action/motion/set_number_mixin.py
```diff
@@ -0,0 +1,124 @@
...
+        )
+        if existing_number_value:
+            return existing_number_value
+
+        if lead_motion_id:
+            filter: Union[And, FilterOperator] = FilterOperator(
+                "lead_motion_id", "=", lead_motion_id
+            )
+        elif meeting.get("motions_number_type") == "per_category":
+            filter = FilterOperator("category_id", "=", category_id)
```
FinnStutzenstein, line 106: here, a filter for the meeting is missing: If the category id is `None`, you want only motions from the current meeting.

openslides_backend/action/motion/set_number_mixin.py
```diff
@@ -0,0 +1,124 @@
...
+        else:
+            filter = FilterOperator("meeting_id", "=", meeting_id)
```
FinnStutzenstein, line 108: Here, we want the max+1 from all **lead motions**, so you need to also filter for `lead_motion_id==None`.

tests/system/action/motion/test_create.py
```diff
@@ -409,3 +409,1358 @@ def test_create_with_submitters(self) -> None:
...
+        self.assert_status_code(response, 200)
+        model = self.get_model("motion/1")
+        assert model.get("title") == "test_Xcdfgee"
+        assert model.get("meeting_id") == 222
+        assert model.get("number") is None
+        agenda_item = self.get_model("agenda_item/1")
+        self.assertEqual(agenda_item.get("meeting_id"), 222)
+        self.assertEqual(agenda_item.get("content_object_id"), "motion/1")
+
+    def test_create_set_number_return_because_no_set_number(self) -> None:
```
FinnStutzenstein, line 494: this test does the same as the test above, but does not create an agenda item.

tests/system/action/motion/test_create.py
```diff
@@ -409,3 +409,1358 @@ def test_create_with_submitters(self) -> None:
...
+            "motion_state/34",
+            {"name": "name_state34", "meeting_id": 222, "set_number": True},
+        )
+        self.create_model(
+            "motion/11",
+            {
+                "title": "title_FcnPUXJB",
+                "meeting_id": 222,
+                "state_id": 34,
+                "number": "001",
```
FinnStutzenstein, line 677: "003"

tests/system/action/motion/test_create.py
```diff
@@ -409,3 +409,1358 @@ def test_create_with_submitters(self) -> None:
...
+                "first_state_id": 34,
+                "state_ids": [34],
+            },
+        )
+        self.create_model(
+            "motion_state/34",
+            {"name": "name_state34", "meeting_id": 222, "set_number": True},
+        )
+
+    def test_issue24_manually_1(self) -> None:
```
FinnStutzenstein, line 859: please choose another description than `issue24`.

tests/system/action/motion/test_create.py
```diff
@@ -409,3 +409,1358 @@ def test_create_with_submitters(self) -> None:
         assert submitter_2.get("meeting_id") == 222
         assert submitter_2.get("user_id") == 57
         assert submitter_2.get("motion_id") == 1
+
+    def test_create_set_number_return_because_number_preset(self) -> None:
```
FinnStutzenstein, line 413: Please put all numbering tests into an own file. There, you can create more structure specific for the numbering tests: Make some functions, that creates models. Make a small wrapper for `self.client.post`, so that not everything must always explicitly written. Maybe do multiple classes for different scenarios like manually, serially, per category. Choose better names for te tests (in a gooed named test class, you can just drop `issue24` and `motion_create`). And please merge your own tests into the ones described in the issue.

As far as I looked, I'm missing tests for `state.set_number=False` and maybe 1-2 tests for the `motion.set_state` action.

FinnStutzenstein (changes requested):

openslides_backend/action/motion/set_number_mixin.py
```diff
@@ -0,0 +1,124 @@
...
+            return max_result["max"] + 1
+        return 1
+
+    def check_if_unique(self, number: str, meeting_id: int) -> bool:
+        filter = And(
+            FilterOperator("meeting_id", "=", meeting_id),
+            FilterOperator("number", "=", number),
+            FilterOperator("meta_deleted", "=", False),
+        )
+        exits_result = self.datastore.exists(
```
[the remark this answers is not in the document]
reiterl: Fixed.

openslides_backend/action/motion/set_number_mixin.py
reiterl: Done.

openslides_backend/action/motion/set_number_mixin.py
reiterl: Done. Added the filter.

tests/system/action/motion/test_create.py
reiterl: Removed the double test.

tests/system/action/motion/test_create.py
reiterl: Done.

tests/system/action/motion/test_create.py
reiterl: Done.

reiterl: Renamed it into "complex_example" tests.

tests/system/action/motion/test_create.py
reiterl: Moved the mixin tests into an own file.

reiterl: Add a test for set_number False and motion. set_state.

tests/system/action/motion/test_set_number_mixin.py
```diff
@@ -0,0 +1,1391 @@
+from tests.system.action.base import BaseActionTestCase
+
+
+class MotionSetNumberMixinTest(BaseActionTestCase):
```
FinnStutzenstein, line 4: @jsangmeister I do not know the style in the backend: I would like to see more setup-functions and wrappers, e.g. for `self.client.post`. If it is always motion create this can be put inside an post-wrapper. The testcases itself are very large. Maybe - depending on the style - you can give Ludwig more hints about organizing the test code better.

FinnStutzenstein (approved):
Looks good, the logic seems to be correct. The only point are the huge size of each individual testcase. Maybe add helperfunctions, so each test itself is smaller and more descriptive. Make multiple classes for different sets of cases (e.g. numbering manually, per category, serially).

You can also create all models for all tests but only work on a subset. The structure is generated in one funciton and i equal for all tests. This helps understand every test since they are all on the same data basis.

tests/system/action/motion/test_create.py
FinnStutzenstein: Not fully done yet. See the new comment below.

FinnStutzenstein:
@reiterl The `motion.follow_recommendation` is missing (at least in tests. I do not know, if this calles `set_state` internally or so).

reiterl:
`motion.follow_recommendation` is missing.

tests/system/action/motion/test_set_number_mixin.py
jsangmeister: Yes, so far most tests were short enough that it didn't really matter. But for bigger tests (like this) this would be a good idea.

reiterl:
The `follow_recommendation` was handled correctly, by calling the set_state code. I added a test for it.

Issue #325, opened 2020-11-19
Restructure action handling and implement transaction models


jsangmeister:
I would like to restructure the general workflow of an action to make it more clear what is happening at each point in the code. The split in `prepare_dataset` and `create_write_request` is unintuitive and not necessary, IMO. Also, it is not clear what a `dataset` even is or what fields it must/should have, since the type declaration contains an `Any` and every time I tried to explain it to someone (@reiterl, @r-peschke), I failed to formulate the point of it.

Also, the second point: I talked with Finn and as previously (loosly) discussed in #27, we need some kind of transaction flag for our actions. In case of an import, we want to execute as many payloads as possible and return errors for the rest. In most other cases, it might be fine to raise an error if anything failed and execute nothing. The easiest way to handle this is to include this in the restructure and make the action code independent from the chosen model. Then, this can easily be done in the base action and transmitted via a simple flag in the payload.

So, here is my suggestion of the workflow in the base action (to be more precise, the `perform` function in `openslides_backend/action/base.py`). The current code looks like this:
```
self.user_id = user_id
self.validate(payload)
self.check_permissions(payload)
dataset = self.prepare_dataset(payload)
yield from self.create_write_request_elements(dataset)
```
My suggestion would be something among the lines of:
```
self.user_id = user_id
self.check_permissions(payload)
# validate now only checks the general form of the payload: that it is an array of objects.
self.validate(payload)
instances = self.get_updated_instances(payload)
for instance in instances:
    self.validate_instance(instance)
    self.update_instance(instance)  # can be overwritten by CreateAction etc. to set defaults, validate relations etc.
    yield self.create_write_request_element(instance)
```
Since we use `update_instance` etc. everywhere, it should be better to just include it in the base action. We get rid of the ominous `DataSet` and the code becomes more cleaner, readable and understandable and it becomes easier to implement the two different transaction methods.

It may become necessary for the actions to be state aware and save previous results in some field. Example: When importing users, the successful imports (which are not yet written to the datastore) have to be checked to make sure a username is not added more than once. The schema could easily be amended for this.

What do you think, whom it may concern?

normanjaeckel:
This approach is ok for me. At the beginning of the life of this code I thought we should differenciate between two phases: first get all stuff from db which is necessary to calculate the new model, second the method to create the events that should be send out. At this point the `DataSet` was born. We can let it die now if you like ... ;-)

FinnStutzenstein:
[...]

PR #337, opened 2020-11-26
Restructure Action handling & fix relation handling with multiple actions

### LINKED ISSUE #321: Creating multiple topics result in inconsistent data
Action request:
```
[{'action': 'topic.create', 'data': [{'meeting_id': 1,
'title': 'A', 'text': '', 'agenda_type': 1, 'agenda_weight': 1000},
{'meeting_id': 1, 'title': 'B', 'text': '', 'agenda_type': 1,
'agenda_weight': 1001}]}]
```
topic 12 and 12 with agenda items 18 and 19 and LOS 21 and 22 are created. Write request:
```
WriteRequestElement(events=[{'type': 'create', 'fqid': FQId topic/11, 'fields': {'meeting_id': 1, 'title': 'A', 'text': '', 'id': 11}}, {'type': 'update', 'fqid': FQId meeting/1, 'fields': {'topic_ids': [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12]}}, {'type': 'create', 'fqid': FQId topic/12, 'fields': {'meeting_id': 1, 'title': 'B', 'text': '', 'id': 12}}, {'type': 'update', 'fqid': FQId meeting/1, 'fields': {'topic_ids': [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12]}}, {'type': 'create', 'fqid': FQId agenda_item/18, 'fields': {'content_object_id': FQId topic/11, 'type': 1, 'weight': 1000, 'level': 0, 'is_hidden': False, 'is_internal': False, 'id': 18, 'meeting_id': 1}}, {'type': 'update', 'fqid': FQId topic/11, 'fields': {'agenda_item_id': 18}}, {'type': 'update', 'fqid': FQId meeting/1, 'fields': {'agenda_item_ids': [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18]}}, {'type': 'create', 'fqid': FQId list_of_speakers/21, 'fields': {'content_object_id': FQId topic/11, 'id': 21, 'meeting_id': 1}}, {'type': 'update', 'fqid': FQId topic/11, 'fields': {'list_of_speakers_id': 21}}, {'type': 'update', 'fqid': FQId meeting/1, 'fields': {'list_of_speakers_ids': [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21]}}, {'type': 'create', 'fqid': FQId agenda_item/19, 'fields': {'content_object_id': FQId topic/12, 'type': 1, 'weight': 1001, 'level': 0, 'is_hidden': False, 'is_internal': False, 'id': 19, 'meeting_id': 1}}, {'type': 'update', 'fqid': FQId topic/12, 'fields': {'agenda_item_id': 19}}, {'type': 'update', 'fqid': FQId meeting/1, 'fields': {'agenda_item_ids': [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 19]}}, {'type': 'create', 'fqid': FQId list_of_speakers/22, 'fields': {'content_object_id': FQId topic/12, 'id': 22, 'meeting_id': 1}}, {'type': 'update', 'fqid': FQId topic/12, 'fields': {'list_of_speakers_id': 22}}, {'type': 'update', 'fqid': FQId meeting/1, 'fields': {'list_of_speakers_ids': [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 22]}}], information={FQId topic/11: ['Object created', 'Object attached to agenda item', 'Object attached to list of speakers'], FQId meeting/1: ['Object attached to topic', 'Object attached to topic', 'Object attached to agenda item', 'Object attached to list of speakers', 'Object attached to agenda item', 'Object attached to list of speakers'], FQId topic/12: ['Object created', 'Object attached to agenda item', 'Object attached to list of speakers'], FQId agenda_item/18: ['Object created'], FQId list_of_speakers/21: ['Object created'], FQId agenda_item/19: ['Object created'], FQId list_of_speakers/22: ['Object created']}, user_id=1)
```
The important part is adding of both agenda items to the `meeting/agenda_item_ids`:
```
{'type': 'update', 'fqid': FQId meeting/1, 'fields': {'agenda_item_ids': [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18]}}
...
{'type': 'update', 'fqid': FQId meeting/1, 'fields': {'agenda_item_ids': [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 19]}}
```
The last one is the one written into the DS, so there is a lost update (id 18 is not there). The same is true for `meeting/list_of_speakers_ids` but not for `meeting/topic_ids` (this worked).

(@GabrielInTheWorld This was the issue you reported to me via email)

jsangmeister:
fixes #321


dev/requirements_development.txt
```diff
@@ -7,5 +7,6 @@ pytest==6.0.2
 pytest-cov==2.10.1
 pyyaml==5.3.1
 pip-check==2.6
+autoflake==1.4
```
normanjaeckel, line 10: What does this tool do and where is it used?

openslides_backend/action/util/typing.py
```diff
@@ -20,14 +18,3 @@
 )
 
 ActionError = Any
-
-
-class Action(Protocol):  # pragma: no cover
```
normanjaeckel, line 25: You have to re-add this protocol (and also rename it). It is used in http/views.py.

openslides_backend/action/action.py
```diff
@@ -1,139 +1,408 @@
...
+    def base_update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
+        """
+        Updates one instance of the payload. This can be overridden by custom
+        action classes.
+        """
+        return instance
 
-        # Parse actions and creates events
-        write_request_element, results = self.parse_actions(payload)
+    def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
```
normanjaeckel, line 154: I think this method is not useful here. Where do we need it? The code can also be in base_update_instance and if some mixins want to create generic code they can define `special_update_instance` method.

dev/requirements_development.txt
jsangmeister: It removes unneeded import which are reported by flake. It was really annoying to remove them all by hand... It's used in the Makefile

openslides_backend/action/util/typing.py
jsangmeister: I don't see why it's needed there, it was only used for one cast

openslides_backend/action/action.py
jsangmeister: Yes, I generally agree, but all three generic actions (create, update, delete) use such a method and it would therefore be defined in all three files. I found that unnecessary and moved it one level up. We could write `return self.update_instance(instance)` in the `base_update_instance` to symbolize the usage, if you want. Or make a better suggestion, I'm not so sure about this either.

jsangmeister:
Should be done now and simplify the general handling and implementation of future changes.

jsangmeister:
Since the PR got pretty big, here a summary of the changes:
- Relation handling was moved to a `RelationManager` which is single-handedly responsible for handling all relation correctly. It uses the old `RelationsHandler` (now `SingleRelationHandler`) as a helper class and takes care fo repeated updates of the same field. It is instanciated once for each call to the backend.
- Restructured the `action` folder (this is probably where most of the changes come from). To not mix the general action classes with the specific actions, the latter were moved into the folder `actions`. I'm not too happy with the subpath `action/actions/.../` but didn't have a better name for it. All "base" classes for the actions were moved into appropriate subfolders and sometimes renamed.
- Added `autoflake` ro remove unneeded imports.
- Restructured the `perform` method of the actions. It is now clearer than before (I hope), what is done when and where one would need to intervene for specific changes. It also takes care of all relation handling now since this is shared by all actions.
- As a first step towards the transaction model described in #325, each payload item is now validated alone and not together with the other payload. This lead to the implementation fo the `SingularPayloadMixin` to still be able to have actions which take exactly one item.
- Extended the usage of the Model meta class. This enables direct field access, so one does not have to iterate all fields each time a specific field is accessed. Also implemented this as a prefix map to unify the handling with structured fields (which have unique prefixes). Replaced most tuples `(field_name, field)` with just the field; the name can be accessed via `field.own_field_name`.

jsangmeister:
Merging this since multiple changes depend on this. If you have any remarks, you can still add them

jsangmeister:
Link #325

openslides_backend/action/action.py
normanjaeckel: Ok, I follow your suggestion in #358

openslides_backend/models/base.py
```diff
@@ -21,11 +22,18 @@ def __new__(metaclass, class_name, class_parents, class_attributes):  # type: ig
             metaclass, class_name, class_parents, class_attributes
         )
         if class_name != "Model":
+            new_class.field_prefix_map = {}
```
normanjaeckel, line 25: Where do we use this new field_prefix_map? I can not find it in the code.

Also: Why do we want to save the template fields twice, once with the pythonic attribute name (e. g. `group__ids` for `User` and once with a shortend name (e. g. `group_`)?

jsangmeister: I would not call it magic, but yes, thank you for the docs :D my bad for that.

PR #338, opened 2020-11-26
Update motion.update action.

reiterl:
Add rules and add tests.

### DIFF
--- a/openslides_backend/action/actions/motion/update.py
+++ b/openslides_backend/action/actions/motion/update.py
@@ -2,6 +2,8 @@ import time
 from typing import Any, Dict
 
 from ....models.models import Motion
+from ....shared.exceptions import ActionException
+from ....shared.patterns import Collection, FullQualifiedId
 from ...generics.update import UpdateAction
 from ...util.default_schema import DefaultSchema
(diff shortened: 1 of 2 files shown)


openslides_backend/action/motion/update.py
```diff
@@ -33,5 +35,33 @@ class MotionUpdate(UpdateAction, AmendmentParagraphsMixin):
 
     def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
         instance["last_modified"] = round(time.time())
+        if instance.get("text"):
+            motion = self.datastore.get(
+                FullQualifiedId(self.model.collection, instance["id"]), ["text"]
+            )
```
jsangmeister, line 41: Please bundle all requests into one that only gets executed if any of the conditions apply.

jsangmeister (changes requested):

openslides_backend/action/motion/update.py
reiterl: Bundle the motion get calls into one. Add a condition so, it is only called, if needed.

jsangmeister (approved):
Rebase

PR #339, opened 2020-12-03
Generate models

r-peschke:
WIP: To discuss, modell.py generated with errors, no black possible, see line 617 (Known error: WIP)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,7 +26,7 @@ test-unit-integration:
 	pytest tests/unit tests/integration
 
 generate-models:
-	PYTHONPATH=. python cli/generate_models.py
+	PYTHONPATH=. python3 cli/generate_models.py
 	black openslides_backend/models/models.py
 
 check-models:
--- a/cli/generate_models.py
+++ b/cli/generate_models.py
@@ -2,6 +2,7 @@ import hashlib
 import os
 import string
 import sys
+from collections import ChainMap
 from textwrap import dedent, indent
 from typing import Any, Dict, List, Optional, Union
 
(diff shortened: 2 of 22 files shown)


r-peschke:
Pull request is ready to merge. Errors in single_relation_handler.py are expected and in consequence failing tests, because of changed structures in models.py.

FinnStutzenstein:
@jsangmeister Do you do the adaption of the relations?

jsangmeister:
yes

jsangmeister:
Adjusted the relation handling and unified it. Hopefully, future additions will be easier with this. Instead of handling different cases in different methods, everything is now transformed into lists of fqids and handled on a per-collection basis. I removed the nested structured relations since we don't need it currently. We can add it back if we need it again some day. 

Depends on https://github.com/OpenSlides/OpenSlides/pull/5769 for the merge and new model definitions.

normanjaeckel:
Hm. I wonder why should use a dictionary of collection and field name for non generic-relation case where this dict can only have one entry. For me the fundamental difference between a generic relation field and a non generic relation field is the content if the field (fqid or just an id). ...

jsangmeister:
> Hm. I wonder why should use a dictionary of collection and field name for non generic-relation case where this dict can only have one entry. For me the fundamental difference between a generic relation field and a non generic relation field is the content if the field (fqid or just an id). ...

Simple: For unified code. This way, every relation can be handled the same way, except for the transformation to ids/fqids at the end. The differentiation is done via the class.

normanjaeckel:
Ok. But do we checksomewhere that we do not have multiple entries in this dict for non generic fields? Maybe in models validator?

r-peschke:
Closed, because solved by @jsangmeister togtehter with fixing the field-relations

jsangmeister:
> Ok. But do we check somewhere that we do not have multiple entries in this dict for non generic fields? Maybe in models validator?

Yes, this would be good. Will include in the respective issue.

PR #341, opened 2020-12-03
Make checksum check optional

jsangmeister:


FinnStutzenstein:
Why did you closed it?

jsangmeister:
This did not work as the check was just shown as successful, even when it failed. I instead set the check in the repository settings to not required, so one can merge a PR even if this check fails. Unfortunately, GH actions does not provide any functionality to set the status to "warning" or something alike...

PR #347, opened 2020-12-08
Adjusts adapter to authlib

GabrielInTheWorld:
This correlates with [openslides-auth-service#55](https://github.com/OpenSlides/openslides-auth-service/pull/55)

### DIFF
--- a/openslides_backend/services/auth/adapter.py
+++ b/openslides_backend/services/auth/adapter.py
@@ -1,8 +1,14 @@
 from typing import Dict, Optional, Tuple
+from urllib import parse
 
-from authlib import AuthHandler
-from authlib.constants import ANONYMOUS_USER, AUTHENTICATION_HEADER  # noqa
-from authlib.exceptions import AuthenticateException, InvalidCredentialsException
+from authlib import (
+    ANONYMOUS_USER,
+    COOKIE_NAME,
+    HEADER_NAME,
+    AuthenticateException,
+    AuthHandler,
+    InvalidCredentialsException,
+)
 
 from ...shared.exceptions import AuthenticationException as BackendAuthException
 from ...shared.interfaces.logging import LoggingModule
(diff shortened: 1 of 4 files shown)


openslides_backend/services/auth/adapter.py
```diff
@@ -32,7 +38,9 @@ def authenticate(
             f"Start request to authentication service with the following data: {headers}"
         )
         try:
-            return self.auth_handler.authenticate(headers, cookies)
+            access_token = headers.get(HEADER_NAME, None)
+            cookie = cookies.get(COOKIE_NAME, "")
```
FinnStutzenstein, line 42: Is the type correct (str not bytes)?

requirements.txt
```diff
@@ -9,4 +9,4 @@ Werkzeug==1.0.1
 gunicorn==20.0.4
 bleach==3.2.1
 PyPDF2==1.26.0
-git+git://github.com/OpenSlides/openslides-auth-service@6e3a2a3b5d71334a273cb2fc39cc376e8e7290d6#egg=authlib&subdirectory=auth/libraries/pip-auth # authlib
+git+git://github.com/OpenSlides/openslides-auth-service@4f4dae3b0ed85e17086c5d5fccca8341c28d92c5#egg=authlib&subdirectory=auth/libraries/pip-auth # authlib
```
FinnStutzenstein, line 12: This must be changed, if https://github.com/OpenSlides/openslides-auth-service/pull/55 is merged (the merge commit will have a different hash), so we'öl wait for the auth service first.

openslides_backend/services/auth/adapter.py
GabrielInTheWorld: I think so. The [requests-docs](https://requests.readthedocs.io/en/master/user/quickstart/#cookies) suggests so, I think.

```
Cookies
If a response contains some Cookies, you can quickly access them:

>>> url = 'http://example.com/some/cookie/setting/url'
>>> r = requests.get(url)

>>> r.cookies['example_cookie_name']
'example_cookie_value'
To send your own cookies to the server, you can use the cookies parameter:

>>> url = 'https://httpbin.org/cookies'
>>> cookies = dict(cookies_are='working')

>>> r = requests.get(url, cookies=cookies)
>>> r.text
'{"cookies": {"cookies_are": "working"}}'
```

requirements.txt
GabrielInTheWorld: Yeah, you're right, that the merge-commit has another hash, but the hash of the PR-commit is the same. 
Or do you want, that the commit-hash is the one from the merge-commit?

FinnStutzenstein: > one from the merge-commit?

Definitly, since this commit belongs to the commits from `master`, the PR-commit is not (or not direct) coupled with master.

GabrielInTheWorld: Okay

PR #348, opened 2020-12-08
Added created and last_modified to motion.create.

reiterl:
In motion.create these automatic generated values were missing. Included them.
Timestamp is an integer, so use round(time.time()) for it.
Added some asserts to check it.


jsangmeister (approved):
Rebase needed

PR #349, opened 2020-12-08
Add motion.set_support_self + tests.

reiterl:
Added checks.
Added support add/remove code.
Only use instance if something changed.
Added tests.

### DIFF
--- a/openslides_backend/action/actions/motion/set_support_self.py
+++ b/openslides_backend/action/actions/motion/set_support_self.py
@@ -1,6 +1,7 @@
 from ....models.models import Motion
+from ....services.datastore.commands import GetManyRequest
 from ....shared.exceptions import ActionException
-from ....shared.patterns import Collection, FullQualifiedId
+from ....shared.patterns import Collection
 from ....shared.schema import required_id_schema
 from ...generics.update import UpdateAction
 from ...util.default_schema import DefaultSchema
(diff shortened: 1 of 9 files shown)


tests/system/action/motion/test_set_support_self.py
```diff
@@ -0,0 +1,193 @@
...
+            json=[
+                {
+                    "action": "motion.set_support_self",
+                    "data": [{"motion_id": 1, "support": True}],
+                }
+            ],
+        )
+        self.assert_status_code(response, 200)
+        model = self.get_model("motion/1")
+        assert model.get("supporter_ids") == [1]
```
jsangmeister, line 109: Please assert that the reverse relation is set correctly.

tests/system/action/motion/test_set_support_self.py
```diff
@@ -0,0 +1,193 @@
...
+            json=[
+                {
+                    "action": "motion.set_support_self",
+                    "data": [{"motion_id": 1, "support": False}],
+                }
+            ],
+        )
+        self.assert_status_code(response, 200)
+        model = self.get_model("motion/1")
+        assert model.get("supporter_ids") == []
```
jsangmeister, line 153: Please assert that the reverse relation is set correctly.

openslides_backend/action/actions/motion/set_support_self.py
```diff
@@ -0,0 +1,66 @@
...
+                ["motions_supporters_min_amount"],
+            )
+            if meeting.get("motions_supporters_min_amount") == 0:
+                raise ActionException("Motion supporters system deactivated.")
+            if not motion.get("state_id"):
+                raise ActionException("Motion is missing state_id.")
+            state = self.datastore.get(
+                FullQualifiedId(Collection("motion_state"), motion["state_id"]),
+                ["allow_support"],
+            )
```
jsangmeister, line 46: Please bundle these in as few datastore calls as possible: One to fetch all motions and one to fetch all meetings and states should be enough (so 2 calls total).

jsangmeister (changes requested):

tests/system/action/motion/test_set_support_self.py
reiterl: Added a reverse relation check.

tests/system/action/motion/test_set_support_self.py
reiterl: Added a reverse relation check.

openslides_backend/action/actions/motion/set_support_self.py
reiterl: Use Get Many Requests to reduce to only two datastore calls.

tests/system/action/motion/test_set_support_self.py
jsangmeister: Also check that the template field is set correctly (`supported_motion_$_ids == ["1"]`)

tests/system/action/motion/test_set_support_self.py
jsangmeister: same as above

jsangmeister (changes requested):

reiterl:
Add checks for the template field.

jsangmeister (commented):
looks good. Some tests are failing which is loosely related to https://github.com/OpenSlides/openslides-backend/issues/365 (Background: `equal_fields` does not work for users since the y do not necessarily have a `meeting_id`). I will take care of this.

PR #354, opened 2020-12-10
Fix relation validation for update actions

jsangmeister:


jsangmeister:
Dumb typo. I wonder why it hasn't been noticed until now, but anyway...

PR #358, opened 2020-12-11
Used initialized datastore directly instead from relation manager.

normanjaeckel:


jsangmeister:
I don't like handling the datastore separately like this. Can't we just make a singleton from it? Or at least, if that's (whyever) not possible, we can add the instance instead of the class to the `Services` object. The intention behind this class was to bundle the services and prevent exactly this situation, handling the services separately.

jsangmeister:
Ah ok, since we need a new object for each request, a singleton is probably not the correct idea, since we would have to reset the locked_fields on each request... I guess it's fine this way then. Maybe we should extract things like the datastore and the relation manager to a container class (like `Services`) which is created at the start of the request and distributed to all needed classes. This would reduce the signature size of all using classes

PR #359, opened 2020-12-12
Re-added correct schemas to health path.

normanjaeckel:


openslides_backend/action/mixins/singular_action_mixin.py
```diff
@@ -2,6 +2,7 @@
 
 from ...shared.exceptions import ActionException
 from ...shared.schema import schema_version
+from ..action import Action
 from ..util.typing import ActionPayload
 
 singular_schema = fastjsonschema.compile(
```
jsangmeister, line 8: This schema is obsolet now, you can just use the `action_payload_schema`

openslides_backend/action/action_handler.py
```diff
@@ -17,11 +17,20 @@
     Payload,
 )
 
+action_payload_schema = {
+    "$schema": schema_version,
+    "title": "Action payload",
+    "type": "array",
+    "items": {"type": "object"},
+    "minItems": 1,
+    "uniqueItems": True,
```
jsangmeister, line 26: This is not needed. I removed it in a previous PR from the other schemas. It should be possible to just give an empty list of payloads - then nothing happens. Also, the items don't have to be unique. Imagine a case where you want for some reason to create the same object (with the same content) multiple times. It doesn't matter if this makes sense but we should not forbid it.

jsangmeister (changes requested):

openslides_backend/action/mixins/singular_action_mixin.py
normanjaeckel: I do not understand you objection. The SingularActionMixin ensures that one can only send one item inside specific action payloads.

normanjaeckel: Please open new PR if something is still open.

openslides_backend/action/action_handler.py
normanjaeckel: Fixed.

PR #360, opened 2020-12-14
Race Condition Tests for sort motion_category and motions in category, assert retry

r-peschke:
Wait for pull request #361, because of failing test.


r-peschke:
Included in #364

PR #361, opened 2020-12-14
Reset locked fields before getting them (again)

r-peschke:
Choose smallest position value in locking fields, but reset locked_fields, before getting them (again)


r-peschke:
Included in Pull Request #364

PR #370, opened 2020-12-19
Action list for permission service

normanjaeckel:


ostcar:
"list_of_speakers.delete":                  "TODO",

ostcar:
There are some implicit rules. We have to document them somewhere.

1. The action name of an "Generic permission check" action consists of two parts separated by a dot. The first part has to be a collection-name.
2. if a "Generic permission check" action has a field "meeting_id" in the payload, it has to be the meeting with the relevant permissions.
3. If a "Generic permission check" has a field "id" ind the payload, it has to be the id of the collection, that gave the action its name.
4. If the payload does not have the fields "id" or "meeting_id", it can not be a "Generic permission check" action

normanjaeckel:
These rules should be documented in permission repo.

ostcar:
Could you merge this? I use the new form of the health route and would like to checkout the master

tests/system/action/list_of_speakers/test_delete.py
```diff
@@ -1,30 +1,31 @@
-from tests.system.action.base import BaseActionTestCase
+# from tests.system.action.base import BaseActionTestCase
 
+# list_of_speakers.delete is an internal action where this test does not work.
```
jsangmeister, line 3: Why does this not work? The test should be executed either way, if something changed which makes this test fail, the test should be adjusted accordingly to make it work again...

jsangmeister:
I am not sure I understand why this is needed for the permission service. Does the service call the backend on each request to see which permissions are needed for the action?

ostcar:
No. Not at runtime. The developer calls it at development time to autogenerates values out of it.

[This script](https://github.com/OpenSlides/openslides-permission-service/blob/master/internal/collection/build_autogen_def/main.go) generates [this file](https://github.com/OpenSlides/openslides-permission-service/blob/master/internal/collection/autogen_def.go)

To do this, [just call this](https://github.com/OpenSlides/openslides-permission-service#autogen-actions)

jsangmeister:
Ah, okay, I see. Why did you not want to put this directly into the permission service? Seems like the easier way instead of fetching it from the backend

ostcar:
Would also be possible. But in this case the backend has to be started before the permission service. When we integrate it into the permission service, they are independent.

tests/system/action/list_of_speakers/test_delete.py
normanjaeckel: We have no usecase at the moment to delete list of speakers via such a call. I had to disable the test to get the permission thing work. See #379

jsangmeister: Yes, we have. `list_of_speakers.delete` is called on almost every model when it is deleted via the cascade deletion. So we need tests for it. Please add them back ASAP.

Issue #351, opened 2020-12-08
Check actions for locked_fields/race-conditions with relations and parallel actions in payloads


r-peschke:
Im Prinzip geht es um das Überprüfen des Field-Lockings, das bisher weitgehend nicht benutzt wurde.
Konkret geht es 

1. um das Verhalten bei Relationen und
1. bei mehreren parallelen Aktionen innerhalb eines Payloads

Ziel des Issues ist es dabei, auszuloten, welche Logik und welche Tests fehlen und die, je nach Aufwand, möglichst generisch zu erstellen.

### Bei jedem Create muss ein collection/max als Feld gelocked werden
So könnte über das Retry eine neue, gültige id vergeben werden. >Checken, wie die id-Vergabe jetzt aussieht bei mehreren Creates in einer Action oder mehreren Paralleln Actions

## Done
- Retry im Action_Handler für DatastoreModelLockedException
- In einem Testthread aufgetretene Assertions können mit spezifischen thread-Assert-Methoden behandelt werden. Grundsätzliches Raisen im thread.join nicht möglich, da dann laufende Threads unerreichbar sind 
- Adapter filter-Methode: Zusätzliches locked_field auf Collection/max, um neu hinzugefügte Werte zuerkennen
- Adapter-Methode get_lock_position_for_collection_id liest per max-Command den Position Wert einer Collection/id. Übergangsweise beim Filter eingesetzt.
 
### Tests
- Test für motion.sequence_number mit Race Condition
- Test für motion_category Sortierung Motion innerhalb einer Kategorie
- Test für motion_category Baumsortierung motion-Kategorien
- ListOfSpeakersReAddLastActionTest.test_correct
Test läuft durch mit modifizierter Ermittlung des position-Wertes für das Kollektionfeld speaker/id mittels der Methode get_lock_position_for_collection_id.
- Generische Test von RelationField und RelationListField

r-peschke:
I stop this issue for the moment, because we have to resolve some problems.
The handling of relations, using more than 1 action in a payload ist not working as expexted, see failing tests:
The pattern of failing is the following:
- Action 1 change an object 1
- Action 2 changes in any way data of object 1 again, but without knowing and respecting the changes of action 1. As a result there are remaining field values of the backward relation.

Some possible solutions (without deep knowledge of datastore):

[...]

PR #374, opened 2020-12-22
Race condition tests and base work, see related Issue #351

r-peschke:
- Reset locked_fields on retry in handle_request
- adapter:update_locked_fields takes again min of positions
- adapter:filter uses temporary get_lock_position_for_collection_id
- Race Condition Tests for motion_category sort and sort motion in category
- General assert for retry exception in race conditions
- Test linear and tree sort deleting element in race condition
- Tests for adapter method reset_locked_fields
- Issue #351 Some relation tests and thread-test-mixin: Stop work, see related Issue


r-peschke:
Close this, because it has been continued and finished with #430

PR #376, opened 2021-01-02
Removed DummyAction

normanjaeckel:


jsangmeister:
Why do you want to remove them? They are needed for the cascade deletion. Without them for example deleting an assignment with an associated poll (like in the exaple data) fails because it does not find a delete action for the poll and options. I would suggest to keep it in until all actions are implemented.

normanjaeckel:
Ready to merge.

jsangmeister:
Yes, it is a dummy action. You removed it in this PR. And it does work, I implemented the dummy actions specially for this. Only thing missing is a test for this, but I didn't think it was necessary to test dummy actions. -- And you are right, we should implement these, but until they are implemented, we NEED these dummy actions.

FinnStutzenstein:
@jsangmeister Please provide a failing test proving your statement.

jsangmeister (approved):

jsangmeister:
Mentioned issues should be fixed with #387. Must rebase

PR #380, opened 2021-01-06
Changed some perms.

normanjaeckel:


FinnStutzenstein:
Sorry, but why do we have permissions here???? The Backend should not know about them

PR #391, opened 2021-01-08
Add user.set_present action + tests.

reiterl:
Add the action and some tests. Uses get_updated_instances.

### DIFF
--- a/openslides_backend/action/actions/user/set_present.py
+++ b/openslides_backend/action/actions/user/set_present.py
@@ -1,0 +1,58 @@
+from ....models.models import User
+from ....shared.exceptions import ActionException
+from ....shared.patterns import Collection, FullQualifiedId
+from ....shared.schema import required_id_schema
+from ...generics.update import UpdateAction
+from ...util.default_schema import DefaultSchema
+from ...util.register import register_action
+from ...util.typing import ActionPayload
+
+
+@register_action("user.set_present")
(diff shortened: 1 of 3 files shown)


openslides_backend/action/actions/user/set_present.py
```diff
@@ -0,0 +1,58 @@
...
+        add meeting_id if present is True.
+        remove meeting_id if present is False.
+        """
+        for instance in payload:
+            if self.user_id == instance["id"]:
+                meeting = self.datastore.get(
+                    FullQualifiedId(Collection("meeting"), instance["meeting_id"]),
+                    ["users_allow_self_set_present"],
+                )
+                if not meeting.get("users_allow_self_set_present"):
```
jsangmeister, line 37: This does only apply if the user does not have the  `user.can_manage` permission. @FinnStutzenstein what is the preferred way for the backend to check permissions?

jsangmeister (approved):
Looks good. The comment is probably a point for later discussion

openslides_backend/action/actions/user/set_present.py
FinnStutzenstein: Please create an issue in the backend coordination repo describing the payload. I'm confused. the action is called `user.set_present`, but it does not work for arbirtary users. Also a check for temporary users are completly missing. What does this action do and **should** do?

PR #396, opened 2021-01-13
Add poll.create, option.create and vote.create actions + tests

reiterl:


openslides_backend/action/actions/poll/create.py
```diff
@@ -0,0 +1,150 @@
...
+from ...util.register import register_action
+from ..option.create import OptionCreateAction
+
+options_schema = {
+    "description": "A option inside a poll create schema",
+    "type": "object",
+    "properties": {
+        "text": {"type": "string", "description": "the text of an option"},
+        "content_object_id": optional_fqid_schema,
+        "Y": {"type": "string", "pattern": r"^-?(\d|[1-9]\d+)\.\d{6}$"},
```
jsangmeister, line 17: Please extract this regex as a constant and import it from somewhere it makes most sense

tests/system/action/poll/test_create.py
```diff
@@ -0,0 +1,104 @@
+from tests.system.action.base import BaseActionTestCase
```
jsangmeister, line 1: We need more tests here, e.g. creating without options, giving global options when it is disabled, giving options with votes which do not fit the pollmethod,...

openslides_backend/action/actions/poll/create.py
```diff
@@ -0,0 +1,150 @@
...
+                    if instance["pollmethod"] == "N":
+                        data["no"] = self.parse_vote_value(option, "N")
+                    else:
+                        data["yes"] = self.parse_vote_value(option, "Y")
+                        if instance["pollmethod"] in ("YN", "YNA"):
+                            data["no"] = self.parse_vote_value(option, "N")
+                        if instance["pollmethod"] == "YNA":
+                            data["abstain"] = self.parse_vote_value(option, "A")
+
+            payload.append(data)
```
jsangmeister, line 97: I think a check is missing here: If I create a poll with `pollmethod == "Y"` and give an option with `N == 1.000000`, the value will not be set on the option, but as far as I see it, it will just be passed through to `option.create`, so there will be a vote created with value `N` for a yes-poll.

jsangmeister (changes requested):

openslides_backend/action/actions/poll/create.py
reiterl: Extracted the decimal schema and moved it into the schema.py file.

openslides_backend/action/actions/poll/create.py
reiterl: If pollmethod is "Y", data["no"]  is not set. and it is not transferred to option.create.

jsangmeister: Ah yes, my bad.

tests/system/action/poll/test_create.py
jsangmeister: Ok. Please create an issue for this

jsangmeister (approved):
Minor thing, rest looks good. Can be squashed and merged after that

openslides_backend/shared/schema.py
```diff
@@ -25,3 +25,5 @@
 }
 id_list_schema: Schema = {**base_list_schema, "items": required_id_schema}
 fqid_list_schema: Schema = {**base_list_schema, "items": required_fqid_schema}
+
+decimal_schema: Schema = {"type": "string", "pattern": r"^-?(\d|[1-9]\d+)\.\d{6}$"}
```
jsangmeister, line 29: Seems like the comment here wasn't saved :( Well anyway, please use this schema also in the `fields.py` for the `DecimalField`

reiterl: Done.

PR #398, opened 2021-01-14
Activate reader dev mode & update models

jsangmeister:
To have access to the `get_everything` route


jsangmeister:
also updated models (just some changes to the settings)

PR #400, opened 2021-01-14
Update adapter and interface to new definition

jsangmeister:
see https://github.com/OpenSlides/OpenSlides/pull/5768


jsangmeister:
Merging to fix tests

PR #403, opened 2021-01-15
Add option.update action and tests

reiterl:


openslides_backend/action/actions/option/update.py
```diff
@@ -0,0 +1,173 @@
...
+from ...util.default_schema import DefaultSchema
+from ...util.register import register_action
+from ..vote.create import VoteCreate
+from ..vote.update import VoteUpdate
+
+
+@register_action("option.update")
+class OptionUpdateAction(UpdateAction):
+    """
+    Action to update a user.
```
jsangmeister, line 17: c&p typo

openslides_backend/action/actions/option/update.py
```diff
@@ -0,0 +1,173 @@
...
+    """
+
+    model = Option()
+    schema = DefaultSchema(Option()).get_update_schema(
+        additional_optional_fields={
+            "Y": {"type": "string", "pattern": r"^-?(\d|[1-9]\d+)\.\d{6}$"},
+            "N": {"type": "string", "pattern": r"^-?(\d|[1-9]\d+)\.\d{6}$"},
+            "A": {"type": "string", "pattern": r"^-?(\d|[1-9]\d+)\.\d{6}$"},
```
jsangmeister, line 25: Please use the decimal schema here that you created in another PR.

openslides_backend/action/actions/option/update.py
```diff
@@ -0,0 +1,173 @@
...
+from ....shared.exceptions import ActionException
+from ....shared.patterns import Collection, FullQualifiedId
+from ...generics.update import UpdateAction
+from ...util.default_schema import DefaultSchema
+from ...util.register import register_action
+from ..vote.create import VoteCreate
+from ..vote.update import VoteUpdate
+
+
+@register_action("option.update")
```
jsangmeister, line 14: I guess this action should also be internal? I don't see why the client would or should ever set the values on an option directly

jsangmeister: If that's the case, you can probably fuse this with the `set_auto_fields` action from the other PR. We don't need 2 internal actions on the same fields.

openslides_backend/action/actions/option/update.py
```diff
@@ -0,0 +1,173 @@
...
+                payload_create.append(
+                    {
+                        "option_id": instance["id"],
+                        "value": "A",
+                        "weight": instance["abstain"],
+                        "meeting_id": option["meeting_id"],
+                    }
+                )
+            else:
+                payload_update.append({"id": vote_id, "weight": instance["abstain"]})
```
jsangmeister, line 80: You can reduce the amount of redundant code by iterating over the needed values (`yes`, `no`, `abstain`)

openslides_backend/action/actions/vote/create.py
```diff
@@ -0,0 +1,17 @@
+from ....models.models import Vote
+from ...generics.create import CreateAction
+from ...util.default_schema import DefaultSchema
+from ...util.register import register_action
+
+
+# TODO should be internal
```
jsangmeister, line 7: Please set it to internal then

jsangmeister (changes requested):
I do not see the need for this action, please explain it to me. Also, basic tests for `vote.create` and `vote.update` are missing.

reiterl:
This action should be used in the analog type to update the options directly.
The poll.vote action should not be used for for update in analog case.

openslides_backend/action/actions/option/update.py
reiterl: Relabeled.

openslides_backend/action/actions/option/update.py
reiterl: Use a loop to simplify the code here.

openslides_backend/action/actions/vote/create.py
reiterl: Set to internal.

openslides_backend/action/actions/option/update.py
reiterl: In the analog type we want to update options directly. So it is not internal.
The set_auto_fields does something different then this action.

openslides_backend/action/actions/option/update.py
reiterl: Include the decimal schema in schema and use it here.

reiterl:
Add a tests for vote create and vote update.

jsangmeister (changes requested):
Test for `vote.create` is still missing...

jsangmeister (approved):
My bad, `vote.create` is already tested. Only rebase needed

PR #407, opened 2021-01-22
Add poll.anonymize action and a test.

reiterl:
Add the internal vote.remove_user_id for this.

### DIFF
--- a/openslides_backend/action/actions/poll/anonymize.py
+++ b/openslides_backend/action/actions/poll/anonymize.py
@@ -1,0 +1,56 @@
+from typing import Any, Dict, List
+
+from ....models.models import Poll
+from ....services.datastore.commands import GetManyRequest
+from ....shared.patterns import Collection, FullQualifiedId
+from ...generics.update import UpdateAction
+from ...util.default_schema import DefaultSchema
+from ...util.register import register_action
+from ...util.typing import ActionPayload
+from ..vote.remove_user_id import VoteRemoveUserId
+
+
+@register_action("poll.anonymize")
+class PollAnonymize(UpdateAction):
+    """
+    Action to anonymize a poll.
+    """
+
+    model = Poll()
+    schema = DefaultSchema(Poll()).get_update_schema()
+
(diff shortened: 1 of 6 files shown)


openslides_backend/action/actions/vote/remove_user_id.py
```diff
@@ -0,0 +1,21 @@
+from typing import Any, Dict
+
+from ....models.models import Vote
+from ...generics.update import UpdateAction
+from ...util.default_schema import DefaultSchema
+from ...util.register import register_action
+
+
+@register_action("vote.remove_user_id", internal=True)
+class VoteRemoveUserId(UpdateAction):
```
jsangmeister, line 10: Please rename this to `VoteAnonymize` for consistency

openslides_backend/action/actions/poll/anonymize.py
```diff
@@ -0,0 +1,56 @@
...
+        for instance in payload:
+            option_ids = self._get_option_ids(instance["id"])
+            options = self._get_options(option_ids)
+
+            for option_id in options:
+                option = options[option_id]
+                if option.get("vote_ids"):
+                    self._remove_user_id_from(option["vote_ids"])
+            if False:
+                yield instance
```
jsangmeister, line 32: This looks just wrong. If this is because mypy wants you to yield something, I would suggest to `return []` at the very end of the function

jsangmeister (changes requested):

openslides_backend/action/actions/vote/remove_user_id.py
reiterl: Renamed into `VoteAnonymize`.

openslides_backend/action/actions/poll/anonymize.py
reiterl: Okay using `return []` is better. mypy wants a return value, so I used that hack. But with this return it works too.

jsangmeister (approved):

PR #421, opened 2021-01-27
Updated models.py

### LINKED ISSUE #408: Fix typo and update models.
https://github.com/OpenSlides/openslides-backend/blob/master/openslides_backend/action/actions/meeting/create_update_delete.py#L93

After https://github.com/OpenSlides/OpenSlides/pull/5802 is merged we have to update models and fix this typo.

normanjaeckel:
Fixes #408


tests/system/action/user/test_update_temporary.py
```diff
@@ -85,77 +75,78 @@ def test_update_all_fields(self) -> None:
...
-                {
-                    "action": "user.update_temporary",
-                    "data": [{"id": 111, "vote_weight": "1.500000"}],
-                }
-            ],
-        )
-        self.assert_status_code(response, 200)
-        model = self.get_model("user/111")
-        assert model.get("vote_weight") == "1.500000"
+    # TODO: Re-enabled these tests after vote weight is reimplemented
```
jsangmeister, line 78: Please stop commenting out tests with TODO's. They tend to stick around and noone cares about it. So, what has to be done here? If possible, please fix it in this PR since it belongs here since this apparently broke the tests.

jsangmeister (changes requested):
I don't know about the rest of the changes, seems like it went above my head that we get rid of the `role` model and the like. But after looking at the model changes it looks right on first glance. Maybe @FinnStutzenstein wants to check this?

FinnStutzenstein:
This looks good so far regarding the models. Yes, we got rid of `role`, since Emanuel rethought his design choices from last year at the Entwicklertreffen. Regarding the test: Yes, it must be fixed here since this PR breaked it so far.

tests/system/action/user/test_update_temporary.py
normanjaeckel: I understand the problem but in the current case it should be an extra task. It is not just a small fix but an extension of fields that were not there before (i. e. non relational template fields for vote_weight). I can open an issue for this and, if you like, remove the tests completely. But please let us solve this thing in an extra step.

normanjaeckel:
PR rebased and ready to merge.

tests/system/action/user/test_update_temporary.py
normanjaeckel: See https://github.com/OpenSlides/openslides-backend/issues/422

jsangmeister (approved):
ok

PR #424, opened 2021-01-29
Issue397 Adjust WR to per-action base with locked-fields

r-peschke:
Fails with most of tests, because of changed interface to datastore.
This should be adjusted first, then I should test this one against datastore

### DIFF
--- a/tests/system/action/test_action_command_format.py
+++ b/tests/system/action/test_action_command_format.py
@@ -1,0 +1,291 @@
+from unittest.mock import Mock
+
+from openslides_backend.action.action_handler import ActionHandler
+from openslides_backend.action.util.typing import Payload
+
+from .base import BaseActionTestCase
+
+# TODO: parse_action Tests als UnitTests zum Action Handler, die anderen, die das schreib format testen mal schauen
+
+
+class GeneralActionCommandFormat(BaseActionTestCase):
+    """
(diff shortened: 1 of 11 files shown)


tests/system/action/test_action_command_format.py
```diff
@@ -0,0 +1,291 @@
+from unittest.mock import Mock
+
+from openslides_backend.action.action_handler import ActionHandler
+from openslides_backend.action.util.typing import Payload
+
+from .base import BaseActionTestCase
+
+# TODO: parse_action Tests als UnitTests zum Action Handler, die anderen, die das schreib format testen mal schauen
```
jsangmeister, line 8: Please either resolve this or remove it and open an issue for this.

jsangmeister (approved):
Looks good except the TODO. Will implement the datastore changes today.

tests/system/action/test_action_command_format.py
r-peschke: I think we should leave it as system test, because it needs to call at minimum the DatastoreService

FinnStutzenstein (approved):
Looks good, the CI isn't happy though.

tests/system/action/test_action_command_format.py
jsangmeister: I meant the `TODO` ;) the test itself is fine.

PR #427, opened 2021-02-01
issue350: Refactored DatastoreExceptions

r-peschke:


openslides_backend/services/datastore/adapter.py
```diff
@@ -57,10 +57,17 @@ def retrieve(self, command: commands.Command) -> DatastoreResponse:
...
+                if type_verbose == "MODEL_LOCKED":
+                    raise DatastoreLockedException(
+                        " ".join(
+                            (
+                                error_message,
+                                f"Model '{additional_error_message.get('key')}' raises {type_verbose} error.",
+                            )
+                        )
+                    )
+                elif type_verbose == "MODEL_DOES_NOT_EXIST":
```
FinnStutzenstein, line 70: Maybe add a ModelDoesNotExist exception?

openslides_backend/action/action_handler.py
```diff
@@ -88,22 +92,24 @@ def handle_request(self, payload: Payload, user_id: int) -> ActionResponse:
...
-                except EventStoreException as exception:
+            except DatastoreLockedException as exception:
+                retried += 1
+                payload = deepcopy(payload_copy)
+                if retried > self.MAX_RETRY:
                     raise ActionException(exception.message)
+                continue
+            except ServiceException as exception:
+                raise ActionException(exception.message)
+            except Exception as exception:
```
FinnStutzenstein, line 109: Did you had this catch-all in there previously? I can understand ist, but this is too high in the structure. This catch-all should be done in the http request handling method as far down as possible, so it really catches unwanted errors.

Here, you can be more specific. What exceptions do you expect to catch here? Please explicitly name and handle them. I would also propose splitting this method if possible to seperate error handling from other logic.

FinnStutzenstein (changes requested):

openslides_backend/services/datastore/adapter.py
r-peschke: Why would you do that? There is only one more step in handling (this kind of) actions: catching and transfering them to the ("unified") ActionException. And this is done on the most down level as possible (ServiceException), because they are all handled the same.
A diversification is in my opinion only necessary, where we need to distinguish them like in the locked case with the DatastoreLockedException. See also my other comment.

openslides_backend/action/action_handler.py
r-peschke: Respecting the catch-all I agree with you and removed it.
But for the rest we have only 2 kind of exceptions:
1. Special ones, we want to retry the action. This kind has to be a special expression type and has to be listed together with the DatastoreLockedException as a tuple. If we need another reaction, we would catch the specific Exception-classes in another case.
2. The rest of our foreseeable Exception are the ServiceException, which should be the base class of all services raising exceptions. They are all converted to ActionExceptions, that is the common reaction for all of them.

I included the parse_actions-method into the try-block, because there could be at minimum a MediaServiceException, which is also a ServiceException with the same handling.

jsangmeister (approved):

FinnStutzenstein (approved):
Please rebase to fix the CI

PR #429, opened 2021-02-03
Add get_users presenter. Add tests.

reiterl:
Update presenter base.

### DIFF
--- a/openslides_backend/presenter/get_users.py
+++ b/openslides_backend/presenter/get_users.py
@@ -67,7 +67,7 @@ class GetUsers(BasePresenter):
         return users
 
     def _check_name(self, user: Dict[str, Any], name: str) -> bool:
-        return user.get(name) is not None and user[name] in self.data["filter"]
+        return user.get(name) is not None and self.data["filter"] in user[name]
 
     def filter_keyword(self, users: List[Dict[str, Any]]) -> List[Dict[str, Any]]:
         if self.data.get("filter"):
--- a/openslides_backend/presenter/get_users.py
+++ b/openslides_backend/presenter/get_users.py
@@ -1,0 +1,106 @@
+from operator import itemgetter
+from typing import Any, Dict, List
+
+import fastjsonschema
+
+from ..services.datastore.deleted_models_behaviour import DeletedModelsBehaviour
+from ..shared.patterns import Collection
+from ..shared.schema import schema_version
+from .base import BasePresenter
+from .presenter import register_presenter
+
+get_users_schema = fastjsonschema.compile(
+    {
+        "$schema": schema_version,
+        "type": "object",
+        "title": "get_users",
+        "description": "get user ids data",
+        "properties": {
+            "start_index": {"type": "integer"},
+            "entries": {"type": "integer"},
+            "sort_criteria": {
+                "type": "array",
+                "item": {
+                    "type": "string",
+                    "enum": ["username", "first_name", "last_name"],
+                },
+            },
+            "reverse": {"type": "boolean"},
+            "include_temporary": {"type": "boolean"},
+            "filter": {"type": ["string", "null"]},
+        },
+        "required": [
+            "start_index",
+            "entries",
+            "sort_criteria",
+            "reverse",
+            "include_temporary",
+        ],
+        "additionalProperties": False,
+    }
+)
+
+
+@register_presenter("get_users")
+class GetUsers(BasePresenter):
+    """
+    Gets all users and return some user_ids.
+    """
+
+    schema = get_users_schema
+
+    def get_result(self) -> Any:
+        users = self.get_all_users()
+        users = self.filter_temp_users(users)
+        users = self.filter_keyword(users)
+
+        # Sort users
+        users = [
+            {
+                "username": user.get("username", ""),
+                "id": user["id"],
+                "first_name": user.get("first_name", ""),
+                "last_name": user.get("last_name", ""),
+            }
+            for user in users
+        ]
+        users.sort(key=itemgetter(*self.data.get("sort_criteria")), reverse=self.data["reverse"])
+
+        print("1",users)
+        # Paginating
+        start_index = self.data["start_index"]
+        end_index = self.data["start_index"] + self.data["entries"]
+        users = users[start_index:end_index]
+
+        print("2", users)
+        # Returning
+        return {"users": [user["id"] for user in users]}
+
+    def get_all_users(self) -> List[Dict[str, Any]]:
+        return list(
+            self.datastore.get_all(
+                Collection("user"),
+                ["id", "username", "first_name", "last_name", "meeting_id"],
+                DeletedModelsBehaviour.NO_DELETED,
+            ).values()
+        )
+
+    def filter_temp_users(self, users: List[Dict[str, Any]]) -> List[Dict[str, Any]]:
+
+        if not self.data.get("include_temporary", False):
+            users = [user for user in users if user.get("meeting_id") is None]
+        return users
+
+    def _check_name(self, user: Dict[str, Any], name: str) -> bool:
+        return user.get(name) is not None and user[name] in self.data["filter"]
+
+    def filter_keyword(self, users: List[Dict[str, Any]]) -> List[Dict[str, Any]]:
+        if self.data.get("filter"):
+            users = [
+                user
+                for user in users
+                if self._check_name(user, "username")
+                or self._check_name(user, "first_name")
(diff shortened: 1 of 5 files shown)


openslides_backend/presenter/get_users.py
```diff
@@ -0,0 +1,108 @@
...
+        users = self.paginate_users(users)
+        return {"users": [user["id"] for user in users]}
+
+    def get_all_users(self) -> List[Dict[str, Any]]:
+        return list(
+            self.datastore.get_all(
+                Collection("user"),
+                ["id", "username", "first_name", "last_name", "meeting_id"],
+                DeletedModelsBehaviour.NO_DELETED,
+            ).values()
```
jsangmeister, line 60: mmh. When designing the datastore, we specificially warned about such a use case since a `get_all` on users might return a HUGE amount of data. If this is okay in this instance depends on just how temporary this presenter should be. @FinnStutzenstein What do you think?

FinnStutzenstein: Ok for now.

openslides_backend/presenter/get_users.py
```diff
@@ -0,0 +1,108 @@
...
+        )
+
+    def filter_temp_users(self, users: List[Dict[str, Any]]) -> List[Dict[str, Any]]:
+
+        if not self.data.get("include_temporary", False):
+            users = [user for user in users if user.get("meeting_id") is None]
+        return users
+
+    def _check_name(self, user: Dict[str, Any], name: str) -> bool:
+        return user.get(name) is not None and user[name] in self.data["filter"]
```
FinnStutzenstein, line 70: I think the last part should be `self.data["filter"] in user[name]`

If i have the query `john` I do want user with the username `johndoe` to be included.

openslides_backend/presenter/get_users.py
```diff
@@ -0,0 +1,108 @@
...
+                "username": self._str_if_None(user.get("username")),
+                "id": user["id"],
+                "first_name": self._str_if_None(user.get("first_name")),
+                "last_name": self._str_if_None(user.get("last_name")),
+            }
+            for user in users
+        ]
+        default_criteria = ["last_name", "first_name", "username"]
+        criteria = self.data.get("sort_criteria", default_criteria)
+        users.sort(key=itemgetter(*criteria), reverse=self.data.get("reverse", False))
```
FinnStutzenstein, line 101: This does not work if I gave you `["about_me", "default_structure_level"]` as sorting criteria.

You need to sort the users **stable** by the last criterion first, than the second-last, ..., and the last sort operator sorts them by the first criterion.

Given a criterion, you can use `sort(users, key=lambda user: user[criterion])` to sort a list of dicts by one key. (<-- I wrote the piece of code by mind, not 100% sure, if it works this way)

FinnStutzenstein: (I also assumed that the builtin `sort` is stable - this needs to be checked)

FinnStutzenstein (changes requested):

FinnStutzenstein:
@reiterl please link the issue.

openslides_backend/presenter/get_users.py
reiterl: Okay, fixed.

openslides_backend/presenter/get_users.py
reiterl: Giving `["about_me", "default_structure_level"]` is not allowed. Allowed are combinations of last_name, first_name, username.
I 'm not sure what you mean with stable and last criterion first.
Sorting is the first key first, if the first keys are equal, then the second key is used ,if first and second key are equal, the third key. is used. This is done using tuples.

`sort(users, key=lambda user: user[criterion])`
Do you mean the `sorted` function?

openslides_backend/presenter/get_users.py
```diff
@@ -0,0 +1,133 @@
...
+
+ALLOWED = {
+    "first_name": "",
+    "last_name": "",
+    "username": "",
+    "title": "",
+    "is_active": False,
+    "is_physical_person": False,
+    "email": "",
+    "last_email_send": "",
```
FinnStutzenstein, line 46: I think that this is a number - use 0 as a default. The client often wants to know about not-sended emails, so sorting them first sounds good.

openslides_backend/presenter/get_users.py
```diff
@@ -0,0 +1,133 @@
...
+        criteria = self.data.get("sort_criteria", default_criteria)
+
+        not_allowed = [crit for crit in criteria if crit not in ALLOWED]
+        if not_allowed:
+            raise PresenterException(f"Sort criteria '{not_allowed}' are not allowed")
+        return criteria
+
+    def get_all_users(self, criteria: List[str]) -> List[Dict[str, Any]]:
+        fields = criteria[:]
+        for name in ("username", "first_name", "last_name"):
```
FinnStutzenstein, line 82: maybe add `meeting_id` and `id` here, so you can just pass `fields` to `get_all`.

openslides_backend/presenter/get_users.py
```diff
@@ -0,0 +1,133 @@
...
+        return list(
+            self.datastore.get_all(
+                Collection("user"),
+                ["id", *fields, "meeting_id"],
+                DeletedModelsBehaviour.NO_DELETED,
+            ).values()
+        )
+
+    def filter_temp_users(self, users: List[Dict[str, Any]]) -> List[Dict[str, Any]]:
+
```
FinnStutzenstein, line 95: please remove this empty line

openslides_backend/presenter/get_users.py
```diff
@@ -0,0 +1,133 @@
...
+            "reverse": {"type": "boolean"},
+            "include_temporary": {"type": "boolean"},
+            "filter": {"type": ["string", "null"]},
+        },
+        "required": [],
+        "additionalProperties": False,
+    }
+)
+
+ALLOWED = {
```
FinnStutzenstein, line 38: Please add a comment that the values are default values for `None`.

openslides_backend/presenter/get_users.py
```diff
@@ -0,0 +1,133 @@
...
+
+    def sort_users(
+        self, users: List[Dict[str, Any]], criteria: List[str]
+    ) -> List[Dict[str, Any]]:
+        users_new = []
+        for user in users:
+            for crit in criteria:
+                if user.get(crit) is None:
+                    user[crit] = ALLOWED[crit]
+            users_new.append(user)
```
FinnStutzenstein, line 122: I think you must not create a new list there - remove this line and use `users` down below.

FinnStutzenstein (changes requested):
This looks good now - I did not look to close on the tests, they seem right at a first glance. 

Just some small changes left

openslides_backend/presenter/get_users.py
FinnStutzenstein: @jsangmeister I wonder why the CI did not catch this? Black or flake8 should complain, right?

jsangmeister: Seems like it is not configured to catch this. I have never looked at the possible settings of these tools, feel free to modify it ;)

openslides_backend/presenter/get_users.py
reiterl: No, in the models.py stands: it is a CharField.

openslides_backend/presenter/get_users.py
reiterl: Okay.

openslides_backend/presenter/get_users.py
reiterl: Removed.

openslides_backend/presenter/get_users.py
reiterl: Done.

openslides_backend/presenter/get_users.py
reiterl: Done.

openslides_backend/presenter/get_users.py
FinnStutzenstein: Ok, this is a mistake. I fixed it here: https://github.com/OpenSlides/OpenSlides/pull/5912

FinnStutzenstein (approved):
One minor change needed, than it is ready for merge! :+1:

openslides_backend/presenter/get_users.py
reiterl: Updated this.

PR #442, opened 2021-02-05
isssue441: Add final slash to mediafile-upload url

r-peschke:
Without  final slash in url the mediaservice can't upload the file, but also didn't return a correct error to caller


jsangmeister (approved):

jsangmeister:
For such small issues which you can directly fix it is not necessarily needed to create an issue, these can also just stand on it's own ;)

Issue #454, opened 2021-02-11
Rename payload to action_data


FinnStutzenstein:
Splitup from #261

Issue #261, opened 2020-10-29
Return Array in error case


jsangmeister:
See https://github.com/OpenSlides-action-crew/openslides-backend-coordination/issues/44#issuecomment-718486353

Like we return an array of objects in the success case, we should return an array in the error case. Also, in the mixed case, we need to:

1. make sure to return an error with success and error responses
2. document how the following actions are handled in case of an error and add a test for that. Example: An array with actions `[A, B]` is sent. `A` throws an error. Does `B` still get executed?

normanjaeckel:
We do not have a mixed case so it would be useless to provide an array in error case. If an array of actions is send and one of them causes an error, the result of all of them is error.

This handling is still a point of discussing: https://github.com/OpenSlides/openslides-backend/issues/27

FinnStutzenstein:
> If an array of actions is send and one of them causes an error, the result of all of them is error.

Please provide an index in this case, so the client knows, which action was the one that failed.

jsangmeister:
Also, I would suggest to then not return an array in the success case, since it's not uniform. Especially since we just multiply the default success message as many times as actions are sent ;)

normanjaeckel:
Index in error case is ok for me.

In success case I would like to stay with an array because sometimes the client likes to get extra information (like a new id in create case).

jsangmeister:
This should be solved then

normanjaeckel:
It is not. We do not send an index in any error case at the moment.

jsangmeister:
I thought it was done with https://github.com/OpenSlides/openslides-backend/pull/286 as well, my bad. As far as I understood it, the index should be returned whenever an ActionException is thrown inside an action to know which action failed.

FinnStutzenstein:
Yes, the interface in https://github.com/OpenSlides/OpenSlides/pull/5696 must be implemented.

r-peschke:
What is to be done in client, backend and in the response?
What is action and payload? With #397 we send actions with events to the datastore-service, but even this actions could consist of more than one backend-action like motion.create.

r-peschke:
See the definition of the interface for ActionService: https://github.com/OpenSlides/OpenSlides/blob/openslides4-dev/docs/interfaces/action-service.txt. It defines, how to call the ActionService in backend and it's respond format and isn't yet implemented as described, especially missing the atomic-parameter in calling handle_request method and the resulting actions of this parameter. But it defines the names of the different entities, that we should follow in general.

[...]

Issue #27, opened 2020-02-04
Let actions run independently


normanjaeckel:

normanjaeckel:
Or perhaps not?

normanjaeckel:
@FinnStutzenstein Please let us discuss this: I recommend that the actions do not run independently so the client can decide: Independently: Send different requests. Dependently: Send one request with different actions. Because we use HTTP2 this might not be a network problem.

The advantage is that I do not have to implement extra logic to decide whether some actions belong together or not.

FinnStutzenstein:
What is the dependent behavior: If some actions fail, do all actions fail and nothing is written to the datastore? Or do all actions prior to the first fail succeed?

normanjaeckel:
If some actions fail, nothing should be written to the datastore. Even in import case.

If the client wants to import indepentently it should send extra "requests".

Only sending emails is a special case: "Failing" here means that we can not validate data correctly. But if all data are valid and then the backend tries to handle the email to the mail transfer agent then the action was successful even if we can not guarantee that the email will go out or come in.

FinnStutzenstein:
Ok

FinnStutzenstein:
Ah not ok, we should discuss this in a telco with @ThomasJunk

normanjaeckel:
Actions can not and should not run independently. Client should send independent request if he wants independent actions.

FinnStutzenstein:
> can not
Well they can, if someone would implemented it

> should not
Why should a creation of a topic and a motion update should depend on each other?

Ok for now since I cannot think of a case where it is strongly needed. But: If there is such a case, I will never send multiple requests (Learned from OS3...) and it should be implemented.

normanjaeckel:
We can keep this issue open and let the service as is at the moment. Please let us discuss when such a case came up.

PR #461, opened 2021-02-12
Add atomic flag

jsangmeister:
fixes https://github.com/OpenSlides/openslides-backend/issues/261
I *think* this closes https://github.com/OpenSlides/openslides-backend/issues/27

Tests are still missing. Code should be ready. If someone can tell me how to get mypy to work correctly with generics, I would be very happy :D

partly fixes https://github.com/OpenSlides/openslides-backend/issues/454, rest (mostly in `get_updated_instances`) will follow if this is fine.

### DIFF
--- a/openslides_backend/action/action.py
+++ b/openslides_backend/action/action.py
@@ -23,7 +23,7 @@ from ..shared.patterns import FullQualifiedField, FullQualifiedId
 from ..shared.typing import ModelMap
 from .relations.relation_manager import RelationManager
 from .relations.typing import FieldUpdateElement, ListUpdateElement
-from .util.typing import ActionData, ActionResultElement
+from .util.typing import ActionData, ActionResultElement, ActionResults
 
 
 class SchemaProvider(type):
@@ -36,6 +36,11 @@ class SchemaProvider(type):
         if schema is not None:
             attrs["schema_validator"] = fastjsonschema.compile(schema)
         return super().__new__(cls, name, bases, attrs)
+
+
+def native(method: Callable) -> Callable:
+    setattr(method, "_native", True)
+    return method
 
 
 class BaseAction:  # pragma: no cover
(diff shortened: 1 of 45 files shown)


jsangmeister:
Added tests.

r-peschke (commented):
It looks quite good to me!
Just a few things:

1.  We should explicitly explain, what "atomic" true or false means. It seems to me, we had he opposite explanation: True means everything is done or with an error not done, whereas False means, not atomic, some could be done, other not.
2. Errors in execute_other_actions are indexed with that of the outer action. Perhaps we should store the Collection-Class or something similar with it, because this could be really misleading.
3. Shouldn't we store the indices on the write_requests objects. Advantages : 
     - The check for required fields will made on the write_results and their events. It's to late for the indexing, i.e. maybe they get all the last error_index
     - In future we could send the indices to other services for use
     - We could address the execute_other_action with a more specific index

jsangmeister:
> It looks quite good to me!
> Just a few things:
> 
> 1. We should explicitly explain, what "atomic" true or false means. It seems to me, we had he opposite explanation: True means everything is done or with an error not done, whereas False means, not atomic, some could be done, other not.
Yes, you are right, I switched it up. Will change it to work according to the documentation.
> 2. Errors in execute_other_actions are indexed with that of the outer action. Perhaps we should store the Collection-Class or something similar with it, because this could be really misleading.
This is correct: Any delegated actions have their origin in the outer action, so the outer action index is the one sent to the client. Any further indexication (?) does not make sense since the client doesn't know about the internal structure of the backend.
> 3. Shouldn't we store the indices on the write_requests objects. Advantages :
>    
>    * The check for required fields will made on the write_results and their events. It's to late for the indexing, i.e. maybe they get all the last error_index
>    * In future we could send the indices to other services for use
>    * We could address the execute_other_action with a more specific index
I think we do not need this for now. We already have multiple cases where the payload gets changed in such a way that index can't be correctly determined. It would mean a lot of work and overhead to change the code to always keep the correct index. Since this is only an indicator for the client, this is not necessary at the moment. We can talk about it if it comes up at a later point again.

Issue #447, opened 2021-02-06
Add get_deleted_models flag for aggregate functions


GabrielInTheWorld:
After requesting `motion_submitters.delete` with payload `{ id: 16 }` (`user_id: 3`) and readding the same user with the following payload:
```
{ 
action: "motion_submitter.create"
data: [{ motion_id: 12, user_id: 3 }]
}
```

throws the following error:
```
Fehler: (user_id, motion_id) must be unique
```

Is there anything wrong with the payloads? Do you need more information?

jsangmeister:
This is a bug in the backend. It is similar to https://github.com/OpenSlides/openslides-backend/issues/207, which was fixed with https://github.com/OpenSlides/openslides-backend/pull/233: The `exists` method returns results of all models, even the deleted ones. Analogous to https://github.com/OpenSlides/openslides-backend/pull/233, we should implemented a flag which adds the filter for `meta_deleted` to the given filter if needed. The same needs to be done for `count`, `min`, `max`, before we run into the same problem again ;)

PR #480, opened 2021-02-18
Add DeletedModelsBehaviour to all aggregate methods

jsangmeister:
Unified the usage with the `filter` method.
Fixes https://github.com/OpenSlides/openslides-backend/issues/447

### DIFF
--- a/openslides_backend/services/datastore/adapter.py
+++ b/openslides_backend/services/datastore/adapter.py
@@ -199,14 +199,9 @@ class DatastoreAdapter(DatastoreService):
         get_deleted_models: DeletedModelsBehaviour = None,
         lock_result: bool = False,
     ) -> Dict[int, PartialModel]:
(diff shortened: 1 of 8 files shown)


openslides_backend/services/datastore/adapter.py
```diff
@@ -199,14 +199,9 @@ def filter(
         get_deleted_models: DeletedModelsBehaviour = None,
```
r-peschke, line 199: I would prefer to set the get_delete_models-parameter to a valid value corresponding it's type, i.e. DeletedModelsBehaviour.NO_DELETED, here and also in the following method-definitions

r-peschke (changes requested):

openslides_backend/services/datastore/adapter.py
jsangmeister: ok, done.

PR #481, opened 2021-02-19
Add the projector.delete action. Add some tests.

reiterl:
Skip meeting update code, because the relevant field is missing.

### DIFF
--- a/openslides_backend/action/action.py
+++ b/openslides_backend/action/action.py
@@ -129,6 +129,8 @@ class Action(BaseAction, metaclass=SchemaProvider):
             relation_updates = self.handle_relation_updates(instance)
             self.write_requests.extend(relation_updates)
 
+            instance = self.final_update_instance(instance)
+
             write_request = self.create_write_requests(instance)
             self.write_requests.extend(write_request)
 
@@ -178,6 +180,12 @@ class Action(BaseAction, metaclass=SchemaProvider):
         """
         Updates one instance of the payload. This can be overridden by custom
         action classes. Meant to be called inside base_update_instance.
+        """
+        return instance
+
+    def final_update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
+        """
+        Updates one instance of the payload or executes other actions.
         """
         return instance
 
--- a/openslides_backend/action/actions/meeting/replace_projector_id.py
+++ b/openslides_backend/action/actions/meeting/replace_projector_id.py
@@ -1,0 +1,54 @@
+from ....models.models import Meeting
+from ....shared.patterns import FullQualifiedId
+from ....shared.schema import required_id_schema
+from ...generics.update import UpdateAction
+from ...util.default_schema import DefaultSchema
+from ...util.register import register_action
+from ...util.typing import ActionPayload
+
+
+@register_action("meeting.replace_projector_id", internal=True)
+class MeetingReplaceProjectorId(UpdateAction):
+    """
+    Internal action to replace default projector id with reference id.
+    """
+
+    model = Meeting()
+    schema = DefaultSchema(Meeting()).get_update_schema(
+        additional_optional_fields={"projector_id": required_id_schema}
+    )
+
+    def get_updated_instances(self, payload: ActionPayload) -> ActionPayload:
+        for instance in payload:
+            projector_id = instance.pop("projector_id")
+            fields = [
+                "default_projector_${}_id".format(replacement)
+                for replacement in (
+                    "agenda_all_items",
+                    "topics",
+                    "list_of_speakers",
+                    "current_list_of_speakers",
+                    "motion",
+                    "amendment",
+                    "motion_block",
+                    "assignment",
+                    "user",
+                    "mediafile",
+                    "projector_message",
+                    "projector_countdowns",
+                    "assignment_poll",
+                    "motion_poll",
+                    "poll",
+                )
+            ]
+            meeting = self.datastore.get(
+                FullQualifiedId(self.model.collection, instance["id"]),
+                fields + ["reference_projector_id"],
+            )
+            changed = False
+            for field in fields:
+                if meeting.get(field) == projector_id:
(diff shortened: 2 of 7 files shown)


jsangmeister (approved):
Please wait for https://github.com/OpenSlides/openslides-backend/pull/482 to also test default projectors

reiterl:
A test is failing. Joshua, can you look at this?

jsangmeister:
This will be solved with the updated datastore adapter in #486.

reiterl:
Merged Ralf's branch into this.
Update a test to better fit intention.
The tests runs thru and the logic seems okay.

openslides_backend/action/action.py
```diff
@@ -181,6 +183,12 @@ def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
         """
         return instance
 
+    def final_update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
```
FinnStutzenstein, line 186: use augment/modify instead of update since it is expected to return the instance.

openslides_backend/action/actions/meeting/replace_projector_id.py
```diff
@@ -0,0 +1,39 @@
+from ....models.models import Meeting
+from ....shared.patterns import FullQualifiedId
+from ....shared.schema import required_id_schema
+from ...generics.update import UpdateAction
+from ...util.default_schema import DefaultSchema
+from ...util.register import register_action
+from ...util.typing import ActionData
+from .shared_meeting import meeting_projector_default_object_list
```
FinnStutzenstein, line 8: rename to `meeting_projector_default_replacements`

openslides_backend/action/actions/projector/delete.py
```diff
@@ -0,0 +1,37 @@
...
+    def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
+        projector = self.datastore.get(
+            FullQualifiedId(self.model.collection, instance["id"]),
+            ["used_as_reference_projector_meeting_id", "meeting_id"],
+        )
+        if projector.get("used_as_reference_projector_meeting_id"):
+            raise ActionException(
+                "A used as reference projector is not allowed to delete."
+            )
+        instance["remember_meeting_id"] = projector.get("meeting_id")
```
FinnStutzenstein, line 30: can you save this as a class member?

FinnStutzenstein (commented):
Looks good, only small changes required from my side.

openslides_backend/action/action.py
```diff
@@ -181,6 +183,12 @@ def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
...
+        """
+        Updates one instance of the payload or executes other actions.
+        """
+        return instance
```
jsangmeister, line 190: I'm not a fan of this. It seems hacky and bypasses (by intent, but still) the relation logic. With the automatic inclusion of all updated models in the `additional_relation_models`, the old version should have worked. What was the error there when just executing `MeetingReplaceProjectorId` as a dependency?

jsangmeister (changes requested):

openslides_backend/action/action.py
FinnStutzenstein: (Note: I think @r-peschke did this part of the code)

jsangmeister: Ah, I see. @r-peschke did you change this after #486 was merged? If so, what exactly did not work with the old setup?

openslides_backend/action/action.py
r-peschke: It is named (final_)update_instance, because `update_instance` is the name of the method working the same way, which is called at the beginning  of the action process.

openslides_backend/action/actions/projector/delete.py
r-peschke: class member is no good idea, because that would impact all instances of class.
class-instance member is less risky, but could fail in actions with more projectors (and different meeting's to create) to act with.
To be secure, we could use an instance-dict-object. But in IMO the instance-dict is the simplest and most intuitive method. But ugly! ;-)

openslides_backend/action/action.py
r-peschke: The problem is the update-sequence of the renaming-action. There is no way of putting them behind the deletions of the relations, generated by the basic delete action. The only thing we could do, is prevent the delete logic of resetting relations, changed in between. But IMO to do this correctly with all kind of relations is a harder way to go.

jsangmeister: I think with #518 this should work. Example: Let's say `meeting/1/default_projector_$motion_id=1`, `meeting/1/reference_projector_id=42`  and we delete `projector/1`. `ProjectorDelete` calls `MeetingReplaceProjectorId` __before__ any write requests are created and, more importantly, before any changes are written to the cache layer. `MeetingReplaceProjectorId` sets `meeting/1/default_projector_$motion_id=42`, which is saved to the cache and the write request is created. Then, control is returned to `ProjectorDelete`, which proceeds to delete `projector/1`, but since the relation in the meeting is already updated to another projector, it does **not** set this to `null`. Everything should work as intended.

This should probably already work with our current setup, we would just have to use the `fetch_model` method instead of `get` in the `MeetingReplaceProjectorId` action. This line seems much cleaner than adding another helper method ;)

openslides_backend/action/actions/meeting/replace_projector_id.py
r-peschke: Done

r-peschke:
To be resolved with [cache_layer_datastore (from projector delete)](https://github.com/OpenSlides/openslides-backend/pull/551)

jsangmeister:
I think this can be closed since it was fully ported to #551, right @r-peschke?

r-peschke:
Close in favor of https://github.com/OpenSlides/openslides-backend/pull/5551

Issue #493, opened 2021-02-23
Backend sets value '0' for undefined options


GabrielInTheWorld:
I sent the following request to the backend:
```
[{'action': 'poll.create', 'data': [{'meeting_id': 1, 'title': 'Abstimmung', 'majority_method': 'simple', 'onehundred_percent_base': 'YNA', 'pollmethod': 'YNA', 'type': 'analog', 'options': [{'content_object_id': 'motion/3'}], 'content_object_id': 'motion/3'}]}]
```

In the client I received then the following data:
![Initial_0_Werte](https://user-images.githubusercontent.com/10206688/108817823-c5ece100-75b8-11eb-80fd-8efd45751c08.png)

![Initial_0_Werte_Autoupdate](https://user-images.githubusercontent.com/10206688/108818014-12382100-75b9-11eb-98e2-9ee2212ffe5d.png)

Please leave undefined values as undefined. :)

FinnStutzenstein:
I would expect the values to be set to `-2` (undocumented).

GabrielInTheWorld:
> I would expect the values to be set to `-2` (undocumented).

This is also a possible value.

GabrielInTheWorld:
But I need here another value than '0'.

FinnStutzenstein:
Ok, the default for analog votes should be -2. This has to be changed for all places special logic for analog polls exists.

FinnStutzenstein:
@GabrielInTheWorld Please add the behaviour here: https://github.com/OpenSlides/OpenSlides/wiki/Voting#analog-polls

GabrielInTheWorld:
> @GabrielInTheWorld Please add the behaviour here: https://github.com/OpenSlides/OpenSlides/wiki/Voting#analog-polls

Done. I've updated the wiki.

PR #502, opened 2021-02-24
Check new defined uniqueness of poll.options

r-peschke:
[Checks the options in poll for uniqueness](https://github.com/OpenSlides/openslides-backend/issues/504) and [sets the values for absent options to -2](https://github.com/OpenSlides/openslides-backend/issues/493)

### DIFF
--- a/openslides_backend/action/actions/poll/create.py
+++ b/openslides_backend/action/actions/poll/create.py
@@ -79,7 +79,19 @@ class PollCreateAction(CreateAction):
 
         # handle non-global options
         weight = 1
+        unique_set = set()
         for option in instance.get("options", []):
+            c_letter = "T" if "text" in option else "C"
+            content = (
+                option.get("content_object_id")
+                if c_letter == "C"
+                else option.get("text")
+            )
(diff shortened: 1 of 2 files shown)


FinnStutzenstein (approved):

openslides_backend/action/actions/poll/create.py
```diff
@@ -36,7 +37,6 @@ class PollCreateAction(CreateAction):
                 "type": "array",
                 "items": options_schema,
                 "minItems": 1,
-                "uniqueItems": True,
```
jsangmeister, line 39: Isn't this check still right? Why did you remove it?

r-peschke: It is still (or again) right, but not sufficient. The new check is stricter and therefore misleading. This check active we have 2 different error-messages for the same error from user/developer view.

jsangmeister: Ah okay, I see. So for consistent error messages we remove it here.

jsangmeister (approved):

PR #515, opened 2021-03-04
Rename payload to action_data

jsangmeister:
closes #454 

Forgot to add the renaming to #461, so adding that with this PR.

@all: Please consequently use the new wording:
- Payload: Everything that is sent from the client.
- PayloadElement: A single Element of this list which contains the name of the action and the ActionData.
- ActionData: The list of data that is supplied for a single action.
- Instance (Action data element, ...): May have different names, depending on the context. Most of the time it reflects an instance of a model (e.g. for CUD actions)

PR #516, opened 2021-03-04
Implement backend permission check

### LINKED ISSUE #459: Enhance action interface for permissions
Discussion thread for the integration of the permissions in the backend.

We can probably build it similarily to OS3: The base case is easy. Each action can define a class variable `permission`. The `check_permissions` function, instead of calling the permission service, can be enhanced to do a general check if the permission is granted and the action can be executed. For more complex/additional permission checks, the method can be overridden.

Thoughts?

jsangmeister:
closes #459

If this looks good to you, I can enhance the test setup to make it easy to test permissions.


jsangmeister:
based on #515 to keep the naming consistent

openslides_backend/action/action.py
```diff
@@ -134,29 +136,94 @@ def perform(
...
+        # switch between internal and external permission service
+        if self.permission:
+            meeting_id = self.get_meeting_id(instance)
+            if self.has_perm(self.permission, meeting_id):
+                return
+        else:
+            if self.permission_service.is_allowed(self.name, self.user_id, [instance]):
+                return
+
+        raise PermissionDenied(f"You are not allowed to perform action {self.name}.")
```
FinnStutzenstein, line 152: For debugging: include the required permission.

openslides_backend/action/action.py
```diff
@@ -134,29 +136,94 @@ def perform(
...
+
+    def get_meeting_id(self, instance: Dict[str, Any]) -> int:
+        """
+        To be overridden in subclasses. Must return the meeting_id that belongs to this instance.
+        """
+        raise NotImplementedError()
+
+    def has_perm(self, permission: str, meeting_id: int) -> bool:
```
FinnStutzenstein, line 160: I think the logic is right. IMO you should ass a new auth module (or permission module) with these kinds of functions. One additional thought. You can include `enable_anonymous` and `default_group_id` into the first call to the datastore.

openslides_backend/action/action.py
```diff
@@ -134,29 +136,94 @@ def perform(
...
             )
+        else:
+            user = {}
+
+        # superadmins have all permissions
+        if user.get("organisation_management_level") == "superadmin":
+            return True
+
+        # get correct group ids for this user
+        if not user.get(f"group_${meeting_id}_ids"):
```
FinnStutzenstein, line 179: remove the not and switch the logic around

openslides_backend/action/generics/get_meeting_id_mixin.py
```diff
@@ -0,0 +1,19 @@
...
+class GetMeetingIdMixin(Action):
+    """
+    Mixin to easily return the meeting id for update and delete actions.
+    """
+
+    def get_meeting_id(self, instance: Dict[str, Any]) -> int:
+        if instance.get("meeting_id"):
+            return instance["meeting_id"]
+        else:
+            db_instance = self.fetch_model(
```
FinnStutzenstein, line 16: This will fail for organisation, committee, users, ...

Maybe `get_meeting_id` should return `Optional[int]`, so the generic code calling `get_meeting_id` can return an error. This method must be overwritten by custom logic for non-meeting models.

For users: The temporary-user thing is dangerous since you would get a meeting id from the user model! I'm unsure if this could be benefitial or fatal ;).

FinnStutzenstein (commented):
Ok for the first implementation. You should do these things next:

- Add tests for the generic logic (maybe use some simple model for it; not tags since they do not have a permissions structure; or create a dummy action for testing if this is possible).
- We do not want to test anonymous, etc. for every action, so this one should be complete:
  * anonymous
  * user as guest
  * user not related to meeting
  * user in some groups
  * user in admin group
  * user has perm
  * user has some parent perm
  * user has not the right perm
  * user is superadmin
- this includes adding helper functions to add new users, changing groups, assigning groups, enabling/disabling the anonymous and use a different user than the admin to make requests.

openslides_backend/action/generics/get_meeting_id_mixin.py
r-peschke: > For users: The temporary-user thing is dangerous since you would get a meeting id from the user model! I'm unsure if this could be benefitial or fatal ;).
I suspect it is more fatal than benefitial: It's really not that what you expect to get (my yesterdays experience!)

r-peschke: Why do you implement this as mixin? Wouldn't it be easier to implement a method in Action class and overwrite it, where needed? Returning `Optional[int]` is a good idea.

openslides_backend/permissions/__init__.py
```diff
@@ -0,0 +1,16 @@
+from typing import Optional
+
+from .permissions import permissions
+
+
+def is_child_permission(child: str, parent: str) -> bool:
```
r-peschke, line 6: It seems to me that this algorithm in conjunction with permissions.py will not work correctly: A child can have more than one parent, that suffices the permission requirement, in the permission.yml. For instance to gain access for permission  `list_of_speakers.can_see` having permission `list_of_speakers.can_be_speaker
` is sufficient. `list_of_speakers.can_manage` instead should be, but isn't. It is not marked as `parent` of `list_of_speakers.can_see`.

tests/system/action/tag/test_create.py
```diff
@@ -35,3 +35,12 @@ def test_create_wrong_field(self) -> None:
...
         )
+
+    def test_create_no_permissions(self) -> None:
+        self.set_management_level("can_manage_users")
+        self.create_model("meeting/577", {"name": "name_YBEqrXqz"})
+        response = self.request(
+            "tag.create", {"name": "test_Xcdfgee", "meeting_id": 577}
+        )
+        self.assert_status_code(response, 403)
+        self.assert_model_not_exists("tag/1")
```
r-peschke, line 46: I would like to have in general an self.assertIn with the message to learn and understand and notice, if there is another kind of error as expected. In this case it is "You do not belong to meeting 577".

r-peschke (changes requested):
See the comments on different points. IMO the is_child functionality don't work as expected.

openslides_backend/action/generics/get_meeting_id_mixin.py
jsangmeister: > For users: The temporary-user thing is dangerous since you would get a meeting id from the user model! I'm unsure if this could be benefitial or fatal ;).

This is fine: The use case of this method is to check to which meeting this models belongs to. In case of temporary users, this is the `meeting_id`.


> This will fail for organisation, committee, users, ...

Yes, this is okay. All models/action which are not inside a meeting must override the permission check anyway, so this method never gets called.

> Why do you implement this as mixin? Wouldn't it be easier to implement a method in Action class and overwrite it, where needed?

I don't like putting code into the base class which only works for a part of the subclasses. It seems cleaner to add it as an abstract method and override it via a mixin, if multiple subclasses need the same logic. But ultimately this comes down to personal preference I guess, or do you have any arguments for one or the other?

openslides_backend/permissions/__init__.py
jsangmeister: You are right! I did miss the double entry there. Well, this makes my nice algorithm a lot more complex...

openslides_backend/action/generics/get_meeting_id_mixin.py
FinnStutzenstein: > so this method never gets called.

You should fail hard if it gets called accidentally. So if someone forgets it he should get a good error.

jsangmeister: It currently does: if this is called on a model which does not have a `meeting_id`, it will fail in line 19 with an `AttributeError`. Which is okay in this case, I think, since it indicates the the programmer fucked up and it should never happen in any code that the client calls (or our CI did not work properly XOR there were absolutely no tests for this action, which are both improbable enough for me).

openslides_backend/action/action.py
jsangmeister: ok

openslides_backend/action/action.py
jsangmeister: done

tests/system/action/tag/test_create.py
jsangmeister: done.

openslides_backend/permissions/__init__.py
jsangmeister: Fixed it. Thanks for the catch!

jsangmeister:
Implemented notes. More tests will follow tomorrow.

openslides_backend/action/generics/get_meeting_id_mixin.py
r-peschke: > I don't like putting code into the base class which only works for a part of the subclasses. It seems cleaner to add it as an abstract method and override it via a mixin, if multiple subclasses need the same logic. But ultimately this comes down to personal preference I guess, or do you have any arguments for one or the other?
May be it's only personal preference, but with KISS it's more pythonic: A method to overwrite instead of a mixin class is IMO simper and more stupid;-)

openslides_backend/permissions/permissions.py
```diff
@@ -1,38 +1,54 @@
...
-    "assignment.can_nominate_other": "assignment.can_manage",
-    "assignment.can_manage": None,
-    "assignment.can_nominate_self": None,
-    "mediafile.can_see": "mediafile.can_manage",
-    "mediafile.can_manage": None,
-    "user.can_see": "user.can_see_extra_data",
-    "user.can_see_extra_data": "user.can_manage",
-    "user.can_manage": None,
-    "user.can_change_own_password": None,
+permissions: Dict[str, List[str]] = {
```
r-peschke, line 6: What do you think of putting all parents in the dict for each permission? We don't have a really deep structure and then the check would be easy again.

jsangmeister: Yes, I thought about that. Would have also been possible, but this way also works and we don't have extra data in this file ;)

openslides_backend/action/generics/get_meeting_id_mixin.py
jsangmeister: Sure, I can change it if you like it better

jsangmeister:
Added tests and helper methods. Please check again.

Regarding the changed motion_state test: When a field has an enum, it currently is not allowed to set this to None (since it's not part of the enum). E.g. for the `organisation_management_level` (and probably most other fields that are not required) this is wrong, since it must be settable to None. This currently results in that the `css_class` of a motion state can be None now, too. If this is not desired, it should be set as required.

FinnStutzenstein:
See #547

tests/system/action/base.py
```diff
@@ -11,23 +14,124 @@ class BaseActionTestCase(BaseSystemTestCase):
...
+            "meeting.create",
+            {
+                "committee_id": 1,
+                "name": "test",
+                "welcome_title": "title",
+            },
+        )
+        self.assert_status_code(response, 200)
+
+    def set_anonymous(self, enable: bool = True, meeting_id: int = 1) -> None:
```
FinnStutzenstein, line 56: Maybe two functions: `enable_anonymous` and `disable_anonymous`.

FinnStutzenstein (approved):
Looks good. This should be a good base to work on. I like the fake model examples, so anyone can see how to test permissions.

One proposal: We had spelling mistakes of permissions in the past (server and client): I would like to have an enum-like class with constants for permissions, so it is compiler checked, that the permissions are spelled correctly. It might be also auto generated with your script.

jsangmeister:
> One proposal: We had spelling mistakes of permissions in the past (server and client): I would like to have an enum-like class with constants for permissions, so it is compiler checked, that the permissions are spelled correctly. It might be also auto generated with your script.

Yes, I remember thinking about that... :D will implement.

tests/system/action/base.py
jsangmeister: Mmh, I think it's fine this way

jsangmeister:
Done.

jsangmeister:
And btw: I'm calling other actions now in the tests. This drastically slows the tests down (about 50% compared to directly inserting the data into the db)...

FinnStutzenstein:
> I'm calling other actions now in the tests. 

What do you mean by this? :D

jsangmeister:
> > I'm calling other actions now in the tests.
> 
> What do you mean by this? :D

I called actions in the setup code for the tests. Replaced this now with `set_models` calls.

FinnStutzenstein (approved):
OK for me.

PR #527, opened 2021-03-08
Motion workflow schema without first state

r-peschke:


FinnStutzenstein:
Ok, whats about `first_state_of_workflow_id` in [motion_state.create](https://github.com/OpenSlides/OpenSlides/wiki/motion_state.create)?

r-peschke:
> Ok, whats about `first_state_of_workflow_id` in [motion_state.create](https://github.com/OpenSlides/OpenSlides/wiki/motion_state.create)?

With current functionality of motion_state.create it is useless, because it never change the field `first_state_of_workflow_id`, when it is set. And it is always set!
On creation of motion_workflow a first motion_state is always created and set. You could change everything, except being the first_state, on motion_state object, following the payload.
We have to decide, whether it should be possible to change physically the first_state or to customize the defined first_state.

tests/system/action/motion_state/test_create.py
```diff
@@ -23,6 +23,42 @@ def test_create(self) -> None:
...
+        )
+        response = self.request_json(
+            [
+                {
+                    "action": "motion_state.create",
+                    "data": [
+                        {
+                            "name": "first state two",
+                            "workflow_id": 42,
+                            "first_state_of_workflow_id": 42,
```
FinnStutzenstein, line 50: Also add a test with
```
"workflow_id": 42,
"first_state_of_workflow_id": 43,
```
With both workflows existing. This must fail, too.

FinnStutzenstein (commented):
Looks good!

tests/system/action/motion_state/test_create.py
r-peschke: done

FinnStutzenstein (approved):

r-peschke:
> > Ok, whats about `first_state_of_workflow_id` in [motion_state.create](https://github.com/OpenSlides/OpenSlides/wiki/motion_state.create)?
> 
> With current functionality of motion_state.create it is useless, because it never change the field `first_state_of_workflow_id`, when it is set. And it is always set!
> On creation of motion_workflow a first motion_state is always created and set. You could change everything, except being the first_state, on motion_state object, following the payload.
> We have to decide, whether it should be possible to change physically the first_state or to customize the defined first_state.

You cannot change the first_state_id, you have to customize the first state.

PR #539, opened 2021-03-11
Use exec format

FinnStutzenstein:
@r-peschke Does this work for you? The server starts up correctly.


r-peschke (approved):
I don't know what I did, but thank you, it works like a charm

PR #540, opened 2021-03-11
Fix action results

### LINKED ISSUE #522: Multiple return values for one action with one instance
I'm sending one action with one instance to the action service:
```
[{action: "speaker.sort", data: [{list_of_speakers_id: 3, speaker_ids: [1, 15, 14]}]}]
```
In response I get `[[null, null, null]]`. These are too many return values! There should be one return value per instance, so I expect `[[null]]`.

jsangmeister:
fixes #522

Some explanation: By default, one `ActionResult` element is created per instance in the action data (e.g. normal create action: 3 instances in action data -> 3 ids must be returned). For actions which change the action data (meaning they overwrite `get_updated_instance`, this no longer holds true - they have to manually set the results to return. For actions which override the method BUT do not change the list of instances, but only the instances itself, it can now just mark the `get_updated_instances` method with `original_instances` and everything should work (like `resource.upload`).

### DIFF
--- a/openslides_backend/action/action.py
+++ b/openslides_backend/action/action.py
@@ -42,8 +42,8 @@ class SchemaProvider(type):
         return super().__new__(cls, name, bases, attrs)
 
 
-def native(method: Callable) -> Callable:
-    setattr(method, "_native", True)
+def original_instances(method: Callable) -> Callable:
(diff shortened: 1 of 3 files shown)


FinnStutzenstein (approved):
Looks good, not tested.

PR #544, opened 2021-03-12
Add projector_countdown create/update/delete and tests

reiterl:
Update models.py


tests/system/action/projector_countdown/test_create.py
```diff
@@ -0,0 +1,59 @@
...
+                "title": "test2",
+                "description": "good description",
+                "default_time": None,
+            },
+        )
+        self.assert_status_code(response, 200)
+        model = self.get_model("projector_countdown/2")
+        assert model.get("title") == "test2"
+        assert model.get("default_time") == 11
+        assert model.get("countdown_time") == 11
```
jsangmeister, line 59: Please add a test where no default time is given (not even `None`). This should also lead to a default_time of 11

jsangmeister (changes requested):

openslides_backend/action/actions/projector_countdown/create.py
```diff
@@ -0,0 +1,43 @@
...
+    schema = DefaultSchema(ProjectorCountdown()).get_create_schema(
+        required_properties=["meeting_id", "title"],
+        optional_properties=["description", "default_time"],
+    )
+
+    def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
+        self.check_title_unique(instance)
+
+        # set default_time if needed and countdown_time
+        if not instance.get("default_time"):
```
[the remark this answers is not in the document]
reiterl: I aware of this problem. Perhaps we should remove the default.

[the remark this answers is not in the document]
jsangmeister: Yes, I think we should. Defaults in the `models.yml` are useful in cases where no manual handling is applied. When we adjust the field in the action anyway, the default is unnecessary (or in this case even harmful).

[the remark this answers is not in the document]
reiterl: The default is removed. Now I will update the tests.

tests/system/action/projector_countdown/test_create.py
reiterl: Updated the test: Remove the default_time: None and generate a new model without default of default_time, so the default can be generated from the meeting.

openslides_backend/action/actions/projector_countdown/create.py
```diff
@@ -0,0 +1,43 @@
...
+            meeting = self.datastore.get(
+                FullQualifiedId(Collection("meeting"), instance["meeting_id"]),
+                ["projector_countdown_default_time"],
+            )
+            instance["default_time"] = meeting.get("projector_countdown_default_time")
```
jsangmeister, line 33: What happens if `projector_countdown_default_time` is not set? Or should this field be required in the meeting?

FinnStutzenstein: is should be required

jsangmeister: ok, I will open an issue for it.

jsangmeister (approved):

PR #551, opened 2021-03-16
cache_layer_datastore (from projector delete)

r-peschke:
WIP: Originally the https://github.com/OpenSlides/openslides-backend/pull/481 for projector_delete.

### DIFF
--- a/openslides_backend/action/action.py
+++ b/openslides_backend/action/action.py
@@ -261,10 +261,7 @@ class Action(BaseAction, metaclass=SchemaProvider):
         )
         if fields:
             event["fields"] = fields
-            if fqid in self.datastore.additional_relation_models:
-                self.datastore.additional_relation_models[fqid].update(fields)
-            else:
-                self.datastore.additional_relation_models[fqid] = fields
+            self.datastore.update_additional_models(fqid, fields)
         if list_fields:
             event["list_fields"] = list_fields
         return WriteRequest(
--- a/openslides_backend/action/action.py
+++ b/openslides_backend/action/action.py
@@ -455,11 +455,13 @@ class Action(BaseAction, metaclass=SchemaProvider):
             )  # non-generic fields can only have one target collection
             value = [FullQualifiedId(field.get_target_collection(), id) for id in value]
         return value
-    
-    def apply_instance(self, instance: Dict[str, Any], fqid: Optional[FullQualifiedId] = None) -> None:
+
+    def apply_instance(
+        self, instance: Dict[str, Any], fqid: Optional[FullQualifiedId] = None
+    ) -> None:
         if not fqid:
             fqid = FullQualifiedId(self.model.collection, instance["id"])
-        self.datastore.update_additional_models(fqid, instance)
+        self.datastore.additional_relation_models[fqid].update(instance)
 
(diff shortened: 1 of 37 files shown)


openslides_backend/action/relations/relation_manager.py
```diff
@@ -163,13 +162,18 @@ def set_structured_field(
...
-            for replacement, value in field_value.items():
-                set_structured_field(field, str(replacement), value)
+            if isinstance(field_value, dict):
+                additional_instance_fields[field_name] = get_template_field_db_value(
+                    field_name
+                )
+                for replacement, value in field_value.items():
+                    set_structured_field(field, str(replacement), value)
+            elif isinstance(field_value, list):
+                pass  # Todo: check this one: for default_projector_$_id there is the list of possible replacements
```
jsangmeister, line 172: I think this should never happen - do you have a case where a list gets passed here?

openslides_backend/action/relations/single_relation_handler.py
```diff
@@ -314,16 +296,17 @@ def prepare_result(
...
                     type="add", value=new_value, modified_element=own_fqid
                 )
             else:
                 assert fqid in remove
                 new_value = rel[related_name]
-                assert (
-                    own_fqid in new_value
-                ), f"Invalid relation update: {own_fqid} is not in {new_value} (Collectionfield {self.model.collection}/{self.field_name})"
+                if own_fqid not in new_value:
+                    continue  # maybe replaced by other action
```
jsangmeister, line 309: when does this happen?

openslides_backend/action/action.py
```diff
@@ -393,7 +392,12 @@ def validate_relation_fields(self, instance: Dict[str, Any]) -> None:
             for equal_field in field.equal_fields:
                 if not (own_equal_field_value := instance.get(equal_field)):
                     fqid = FullQualifiedId(self.model.collection, instance["id"])
-                    db_instance = self.fetch_model(fqid, [equal_field])
+                    db_instance = self.datastore.fetch_model(
+                        fqid,
+                        [equal_field],
+                        db_additional_relevance=InstanceAdditionalBehaviour.ADDITIONAL_BEFORE_DBINST,
```
jsangmeister, line 398: since `ADDITIONAL_BEFORE_DBINST` is the default, I would suggest to remove it here and everywhere else where it isn't needed. Makes the code more readable and less confusing since you don't have to bother with any db-instance-handling if it's the default case.

jsangmeister (commented):
Looks very good and more importantly, it doesn't look like you needed "hacks" for specific action, so it all works with the generic code! Let's talk about the few remarks later

openslides_backend/action/relations/relation_manager.py
r-peschke: > I think this should never happen - do you have a case where a list gets passed here?

Yes, for default_projector_$_id there is the list of possible replacements

openslides_backend/action/action.py
r-peschke: There would be more defaults to change this way. But if we want to replace the get-method by the fetch.model-method we should use here the defaults from get.

PR #552, opened 2021-03-16
Add projector.add_to_preview and tests.

reiterl:

PR #554, opened 2021-03-18
Update models & some cleanup

### LINKED ISSUE #541: poll global fields
The fields `poll/global_yes`, `poll/global_` and `poll/global_abstain` must have `False` as the default value.

jsangmeister:
closes #541


jsangmeister:
Fixes also a test which fell through when merging both #552 and #543 without rebasing on each other, so merging this now. Any remarks can be done in another PR

PR #559, opened 2021-03-19
Issue520 Return value of execute_other_action without write_request

r-peschke:


openslides_backend/action/actions/meeting/create_update_delete.py
```diff
@@ -200,29 +199,28 @@ def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
...
+        )
+        instance["default_group_id"] = fqid_default_group.id
+        instance["admin_group_id"] = fqid_admin_group.id
 
         # Add user to admin group
         action_data = [
             {
                 "id": self.user_id,
-                "group_$_ids": {str(instance["id"]): [used_groups_dict["Admin"]]},
+                "group_$_ids": {str(instance["id"]): [fqid_admin_group.id]},
```
jsangmeister, line 223: You don't have to cast it to a string here, int should work as well. Just as a side note ;)

jsangmeister (approved):

openslides_backend/action/actions/meeting/create_update_delete.py
r-peschke: The schema-validator don't accept it as int;-):
```
data = {'group_$_ids': {1: [3]}, 'id': 1}

>   ???
E   TypeError: expected string or bytes-like object

<string>:191: TypeError
```

Issue #558, opened 2021-03-18
Make meeting.projector_countdown_default_time required


jsangmeister:
see https://github.com/OpenSlides/openslides-backend/pull/544#discussion_r596918075

FinnStutzenstein:
also projector_countdown_warning_time

PR #563, opened 2021-03-22
Add creation of two default projector countdowns.

reiterl:
Update projector_countdown.
Update tests.

### DIFF
--- a/openslides_backend/action/actions/meeting/create_update_delete.py
+++ b/openslides_backend/action/actions/meeting/create_update_delete.py
@@ -225,16 +225,15 @@ class MeetingCreate(CreateActionWithDependencies):
             }
         ]
         self.execute_other_action(UserUpdate, action_data)
+        self.apply_instance(instance)
         action_data_countdowns = [
             {
                 "title": "List of speakers countdown",
                 "meeting_id": instance["id"],
-                "default_time": instance["projector_countdown_default_time"],
             },
             {
                 "title": "Voting countdown",
                 "meeting_id": instance["id"],
-                "default_time": instance["projector_countdown_default_time"],
             },
         ]
         action_results = self.execute_other_action(
--- a/openslides_backend/action/actions/meeting/create_update_delete.py
+++ b/openslides_backend/action/actions/meeting/create_update_delete.py
@@ -11,6 +11,7 @@ from ...util.register import register_action_set
 from ..group.create import GroupCreate
 from ..motion_workflow.create import MotionWorkflowCreateSimpleWorkflowAction
 from ..projector.create import ProjectorCreateAction
+from ..projector_countdown.create import ProjectorCountdownCreate
 from ..user.update import UserUpdate
 from .shared_meeting import meeting_projector_default_replacements
 
@@ -224,6 +225,24 @@ class MeetingCreate(CreateActionWithDependencies):
             }
         ]
         self.execute_other_action(UserUpdate, action_data)
+        action_data_countdowns = [
+            {
+                "title": "List of speakers countdown",
+                "meeting_id": instance["id"],
+                "used_as_list_of_speaker_countdown_meeting_id": instance["id"],
+                "default_time": instance["projector_countdown_default_time"],
+            },
+            {
+                "title": "Voting countdown",
+                "meeting_id": instance["id"],
+                "used_as_poll_countdown_meeting_id": instance["id"],
+                "default_time": instance["projector_countdown_default_time"],
+            },
+        ]
+        self.execute_other_action(
+            ProjectorCountdownCreate,
+            action_data_countdowns,
+        )
         return instance
 
     def get_dependent_action_data(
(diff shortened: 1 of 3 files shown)


openslides_backend/action/actions/meeting/create_update_delete.py
```diff
@@ -224,6 +225,24 @@ def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
             }
         ]
         self.execute_other_action(UserUpdate, action_data)
+        action_data_countdowns = [
+            {
+                "title": "List of speakers countdown",
+                "meeting_id": instance["id"],
+                "used_as_list_of_speaker_countdown_meeting_id": instance["id"],
+                "default_time": instance["projector_countdown_default_time"],
```
FinnStutzenstein, line 233: Do you need to give the default time? `ProjectorCountdownCreate` should handle it I think. Maybe (@jsangmeister) this is only possible with the DS cache?

openslides_backend/action/actions/projector_countdown/create.py
```diff
@@ -18,7 +18,12 @@ class ProjectorCountdownCreate(CreateAction):
     model = ProjectorCountdown()
     schema = DefaultSchema(ProjectorCountdown()).get_create_schema(
         required_properties=["meeting_id", "title"],
-        optional_properties=["description", "default_time"],
+        optional_properties=[
+            "description",
+            "default_time",
+            "used_as_poll_countdown_meeting_id",
```
FinnStutzenstein, line 24: Is it possible to give these values from outside (Can a client pass data in there)?

FinnStutzenstein (changes requested):

openslides_backend/action/actions/meeting/create_update_delete.py
reiterl: The meeting does not exist yet. Leads to an error.

openslides_backend/action/actions/projector_countdown/create.py
reiterl: So it is more clear, what happens.

openslides_backend/action/actions/meeting/create_update_delete.py
reiterl: And it is the same.

openslides_backend/action/actions/projector_countdown/create.py
FinnStutzenstein: A client should not be able to set these fields. What happens, if a client sends these fields now? Please add tests to verify that a client cannot set these fields.

If this is not possible: You must create a new internal action just for this purpose. @r-peschke did this, too, if I remember correctly.

openslides_backend/action/actions/meeting/create_update_delete.py
jsangmeister: > this is only possible with the DS cache?

Yes, it should be, but `projector_countdown_default_time` is currently not required and therefore not set on meeting creation (see https://github.com/OpenSlides/openslides-backend/issues/558). I would assume it should work once that issue is resolved.

openslides_backend/action/actions/projector_countdown/create.py
jsangmeister: Yes, you'll need a seperate internal action for this or else these values can be set from outside.

jsangmeister: But this shouldn't even be necessary: description and default_time are correct (see https://github.com/OpenSlides/OpenSlides/wiki/projector_countdown.create) and the relation fields can just be set on the meeting instead of on the countdown. If you need an example, the group relations (`admin_group_id` and `default_group_id`) are set in the same fashion.

openslides_backend/action/actions/meeting/create_update_delete.py
FinnStutzenstein: It is a question of order: Are the defaults set earlier than this statement?

jsangmeister: > Are the defaults set earlier than this statement?

Yes.

> Leads to an error.

You probably have to update the cache (`self.apply_instance(instance)`) before executing the other action. Please implement #558 first, then check here that the default is actually set and write it to the cache and then try again to leave the field out.

openslides_backend/action/actions/projector_countdown/create.py
reiterl: Set the meeting *_countdown_id s with action_results in the instance.

openslides_backend/action/actions/meeting/create_update_delete.py
reiterl: The apply_instance(instance) set a pre version of the meeting into the cache. Then the default_time is not needed in the action_data_create.

FinnStutzenstein (approved):
LGTM

jsangmeister (approved):

PR #597, opened 2021-03-23
Add permision check to assignment.

reiterl:
Add base_permission_test method to the base test class.

### DIFF
--- a/tests/system/action/assignment/test_delete.py
+++ b/tests/system/action/assignment/test_delete.py
@@ -3,7 +3,10 @@ from tests.system.action.base import BaseActionTestCase
 
 class AssignmentDeleteActionTest(BaseActionTestCase):
     def test_delete_correct(self) -> None:
-        self.create_model("assignment/111", {"title": "title_srtgb123"})
+        self.create_model("meeting/110", {})
+        self.create_model(
+            "assignment/111", {"meeting_id": 110, "title": "title_srtgb123"}
+        )
         response = self.request("assignment.delete", {"id": 111})
         self.assert_status_code(response, 200)
         self.assert_model_deleted("assignment/111")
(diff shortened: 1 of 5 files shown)


tests/system/action/assignment/test_delete.py
```diff
@@ -3,25 +3,32 @@
 
 class AssignmentDeleteActionTest(BaseActionTestCase):
     def test_delete_correct(self) -> None:
-        self.create_model("assignment/111", {"title": "title_srtgb123"})
+        self.create_model("meeting/110", {})
+        self.create_model(
+            "assignment/111", {"meeting_id": 110, "title": "title_srtgb123"}
+        )
```
r-peschke, line 9: Maybe it looks better to use set_models if there are more than one model actions

tests/system/action/base.py
```diff
@@ -196,3 +196,12 @@ def _fetch_groups(self, group_ids: List[int]) -> Dict[int, List[Dict[str, Any]]]
...
         return partitioned_groups
+
+    def base_permission_test(
+        self, models: Dict[str, Any], action: str, action_data: Dict[str, Any]
+    ) -> None:
+        self.set_management_level(None)
+        self.set_models(models)
+        response = self.request(action, action_data)
+        self.assert_status_code(response, 403)
+        assert "You do not belong to meeting" in response.json["message"]
```
r-peschke, line 207: I think it would be better to have a really permission error, i.e. `You are not allowed to perform action {self.name}`, see tests in openslide/openslides-backend/tests/system/action/test_permissions.py.
You could use the helper functions from openslide/openslides-backend/tests/system/action/base.py like create_meeting for a simple setup of base test

r-peschke (changes requested):

tests/system/action/base.py
reiterl: A 403 is a real permission error. It is in the same way done, as Joshua did by tags.

tests/system/action/assignment/test_delete.py
reiterl: Done, use now set_models in test_delete.

tests/system/action/base.py
r-peschke: okay

r-peschke (approved):

tests/system/action/base.py
jsangmeister: I agree with @r-peschke, see #609

PR #601, opened 2021-03-23
Issue445 validate vote delegations

r-peschke:
Ready incl. the missing tests

### DIFF
--- a/openslides_backend/action/actions/user/user_mixin.py
+++ b/openslides_backend/action/actions/user/user_mixin.py
@@ -1,6 +1,8 @@
 from collections import defaultdict
+from functools import reduce
 from typing import Any, Dict, List
 
+from ....services.datastore.commands import GetManyRequest
 from ....shared.exceptions import ActionException
 from ....shared.filters import FilterOperator
 from ....shared.patterns import Collection, FullQualifiedId
@@ -22,6 +24,7 @@ class UserMixin(Action):
                 raise ActionException(
                     f"A user with the username {instance['username']} already exists."
                 )
+        self.check_existence_of_from_users(instance)
         self.check_meeting_and_users(instance, user_fqid)
         if "vote_delegated_$_to_id" in instance:
             self.check_vote_delegated__to_id(instance, user_fqid)
@@ -34,9 +37,7 @@ class UserMixin(Action):
     ) -> None:
         mapped_fields = [
             f"vote_delegations_${meeting_id}_from_ids"
-            for meeting_id, delegated_to in instance[
-                "vote_delegated_$_to_id"
-            ].items()
+            for meeting_id, delegated_to in instance["vote_delegated_$_to_id"].items()
             if delegated_to
         ]
         if not mapped_fields:
@@ -52,9 +53,7 @@ class UserMixin(Action):
                 ].items()
             }
             user_self.update(update_dict)
-        for meeting_id, delegated_to_id in instance[
-            "vote_delegated_$_to_id"
-        ].items():
+        for meeting_id, delegated_to_id in instance["vote_delegated_$_to_id"].items():
             if user_fqid.id == delegated_to_id:
                 raise ActionException(
                     f"User {delegated_to_id} can't delegate the vote to himself."
@@ -122,6 +121,25 @@ class UserMixin(Action):
                 raise ActionException(
                     f"User(s) {error_user_ids} can't delegate their votes , because they receive vote delegations."
                 )
+
+    def check_existence_of_from_users(self, instance: Dict[str, Any]) -> None:
+        if "vote_delegations_$_from_ids" in instance:
+            user_ids = set(
+                reduce(
+                    (lambda x, y: x + y),  # type: ignore
+                    instance["vote_delegations_$_from_ids"].values(),
+                )
+            )
+            get_many_request = GetManyRequest(
+                self.model.collection, list(user_ids), ["id"]
+            )
+            gm_result = self.datastore.get_many([get_many_request])
+            users = gm_result.get(self.model.collection, {})
+
+            set_action_data = user_ids
+            diff = set_action_data.difference(users.keys())
+            if len(diff):
+                raise ActionException(f"The following users were not found: {diff}")
 
     def check_meeting_and_users(
         self, instance: Dict[str, Any], user_fqid: FullQualifiedId
--- a/openslides_backend/action/actions/user/user_mixin.py
+++ b/openslides_backend/action/actions/user/user_mixin.py
@@ -1,14 +1,17 @@
-from typing import Any, Dict
+from collections import defaultdict
+from typing import Any, Dict, List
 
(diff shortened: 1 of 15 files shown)


FinnStutzenstein:
> Have a look at projector/test_update.py

This is great! It allows the client for easier updates (The projection defaults will likely be set during an update of a projector - now I just have to issue a single update action and must not update the meeting too).

The test name does not fit anymore. And can you add a test trying to set a different meeting where the projector does not belong to?

Please add the `used_as_default_$_in_meeting_id` field to the docs https://github.com/OpenSlides/OpenSlides/wiki/projection#projectorupdate

Also adjust the docs for https://github.com/OpenSlides/OpenSlides/wiki/user.update (and/or https://github.com/OpenSlides/OpenSlides/wiki/user.update_temporary) regarding the vote delegation.

openslides_backend/action/relations/single_relation_handler.py
```diff
@@ -179,26 +180,44 @@ def perform(self) -> RelationFieldUpdates:
...
                     else:
-                        # We added a value to the *:1 field which was already populated.
-                        # The relation handling does currently not support this kind of
-                        # relation cascading.
-                        message = (
-                            f"You can not set {fqfield} to a new value because this "
-                            "field is not empty."
-                        )
-                        raise ActionException(message)
+                        rel_update["value"] = current_value[0]
```
jsangmeister, line 184: This means that the length of `current_value` must alway be 1 at this point, right? Then I would like to add an `assert` here

jsangmeister (changes requested):
Looks good except one remark!

openslides_backend/action/relations/single_relation_handler.py
```diff
@@ -297,7 +316,17 @@ def prepare_result(
             if fqid in add:
                 if own_fqid in rel[related_name]:
                     continue
-                new_value = rel[related_name] + [own_fqid]
+                if rel[related_name] and self.type in ("1:1", "m:1"):
+                    assert len(rel[related_name]) == 1
```
r-peschke, line 320: Here is the "assert". It needs to be of len one at this preparation point.

openslides_backend/action/relations/single_relation_handler.py
r-peschke: The assert is [here](https://github.com/OpenSlides/openslides-backend/pull/601/files#r601538005)

jsangmeister: Ok

openslides_backend/action/actions/user/user_mixin.py
```diff
@@ -19,4 +24,158 @@ def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
...
+            user_fqid, mapped_fields, lock_result=True, exception=False
+        )
+        if "vote_delegations_$_from_ids" in instance:
+            update_dict = {
+                f"vote_delegations_${meeting_id}_from_ids": delegated_from
+                for meeting_id, delegated_from in instance[
+                    "vote_delegations_$_from_ids"
+                ].items()
+            }
+            user_self.update(update_dict)
```
jsangmeister, line 55: I think this would override any given `vote_delegated__to_id` from the same meeting, right?

openslides_backend/action/actions/user/user_mixin.py
```diff
@@ -19,4 +24,158 @@ def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
...
+                    lock_result=True,
+                )
+                if user.get(mapped_field):
+                    error_user_ids.append(user_id)
+            if error_user_ids:
+                raise ActionException(
+                    f"User(s) {error_user_ids} can't delegate their votes , because they receive vote delegations."
+                )
+
+    def check_existence_of_to_and_from_users(self, instance: Dict[str, Any]) -> None:
```
jsangmeister, line 125: I _think_ this method is unecessary because the relation handling should throw an error if a related model does not exist. Did you test without it?

openslides_backend/action/actions/user/user_mixin.py
jsangmeister: Please add a test where you give e.g. `vote_delegated_$1_to_id` as well as `vote_delegated_$1_from_ids` and check that this fails (if you didn't already do that, I did not find it among the tests).

jsangmeister (changes requested):
Wow! Good test coverage!

r-peschke:
> > Have a look at projector/test_update.py
> 
> This is great! It allows the client for easier updates (The projection defaults will likely be set during an update of a projector - now I just have to issue a single update action and must not update the meeting too).
> 
> The test name does not fit anymore. And can you add a test trying to set a different meeting where the projector does not belong to?
> 
done
> Please add the `used_as_default_$_in_meeting_id` field to the docs https://github.com/OpenSlides/OpenSlides/wiki/projection#projectorupdate
> 
done
> Also adjust the docs for https://github.com/OpenSlides/OpenSlides/wiki/user.update (and/or https://github.com/OpenSlides/OpenSlides/wiki/user.update_temporary) regarding the vote delegation.
done

openslides_backend/action/actions/user/user_mixin.py
r-peschke: To check for that error, the method is not necessary. The method exist, because of the code from [here](https://github.com/OpenSlides/openslides-backend/blob/e55d7804507859e49f9bd4eb62860336c03b2f57/openslides_backend/action/actions/user/temporary_user_mixin.py#L36). It gives the error message more specific ("The following users were not found: {...}"). Without this method it would be "The following models do not belong to meeting 222: ['user/7']", because it would be detected in asserts_belong to meeting.py. I think there has been an issue to make the error message more understandable.

openslides_backend/action/actions/user/user_mixin.py
r-peschke: > I think this would override any given `vote_delegated__to_id` from the same meeting, right?
Yes, if it is changed it should overwrite it

r-peschke: > Please add a test where you give e.g. `vote_delegated_$1_to_id` as well as `vote_delegated_$1_from_ids` and check that this fails (if you didn't already do that, I did not find it among the tests).

There have been 3 tests, but I added one more test, checking for error. IMO it works quite well

openslides_backend/action/actions/user/user_mixin.py
jsangmeister: Ah okay, I see. Since we have it now, we may just keep it, although I would say the latter error messsage would also suffice

openslides_backend/action/actions/user/user_mixin.py
jsangmeister: ok

jsangmeister (approved):

PR #604, opened 2021-03-24
Projection of a meeting fails

r-peschke:


FinnStutzenstein (approved):
tested with the client. works

openslides_backend/action/util/assert_belongs_to_meeting.py
```diff
@@ -15,6 +15,10 @@ def assert_belongs_to_meeting(
 
     errors: List[FullQualifiedId] = []
     for fqid in fqids:
+        if fqid.collection.collection == "meeting":
+            if fqid.id == meeting_id:
+                continue
+            errors.append(fqid)
```
jsangmeister, line 21: A `continue` or `else` is missing here, for meetings we do not need to fetch anything from the datastore.

jsangmeister (changes requested):

jsangmeister (approved):

openslides_backend/action/util/assert_belongs_to_meeting.py
r-peschke: done

PR #607, opened 2021-03-24
issue 594 define schema for JSONField

r-peschke:
Defined a specific schema for JSON field and changed some tests


FinnStutzenstein (approved):
Tested with the client. Works

openslides_backend/models/fields.py
```diff
@@ -100,7 +100,12 @@ def get_schema(self) -> Schema:
 
 
 class JSONField(TextField):
```
jsangmeister, line 102: I think this no longer should inherit from `TextField`, right? I think the initial thought was that it should represent the json-encoded string, but this is obsolet now.

jsangmeister (changes requested):

openslides_backend/models/fields.py
r-peschke: Yes and done

tests/system/action/projector/test_project.py
```diff
@@ -135,49 +135,10 @@ def test_project_not_unique_ids(self) -> None:
...
-                "options": "",
+                "options": {},
                 "stable": False,
                 "type": "test",
             },
         )
         self.assert_status_code(response, 400)
         assert "data.ids must contain unique items" in response.data.decode()
-
-    def test_meeting_as_content_object_ok(self) -> None:
```
jsangmeister, line 146: I don't think you meant to remove these tests, right? :D

jsangmeister (approved):

PR #603, opened 2021-03-24
Add permissions to list_of_speakers actions. Update tests.

reiterl:

PR #608, opened 2021-03-24
Add permissions to motion_submitter actions. Update tests.

reiterl:


openslides_backend/action/actions/motion_submitter/sort.py
```diff
@@ -27,3 +32,12 @@ def get_updated_instances(self, action_data: ActionData) -> ActionData:
             filter_id=instance["motion_id"],
             filter_str="motion_id",
         )
+
+    def get_meeting_id(self, instance: Dict[str, Any]) -> int:
```
FinnStutzenstein, line 36: Doesn't the submitter have a `meeting_id`?

FinnStutzenstein (commented):
OK, but please see https://github.com/OpenSlides/openslides-backend/issues/609

openslides_backend/action/actions/motion_submitter/sort.py
reiterl: A submitter has a meeting_id, what is the point?

FinnStutzenstein: nevermind

jsangmeister (changes requested):
Looks good except same notes as in #603 regarding the duplication of the models

jsangmeister (changes requested):
Please use the enums here (and in all other places where you used explicit string) as @FinnStutzenstein remarked in #603. I missed it in some already merged PRs, too, I guess, so please revisit these as well.

FinnStutzenstein:
> Please use the enums here (and in all other places where you used explicit string)

Should be done. These are action names, not permission names

jsangmeister (approved):

PR #616, opened 2021-03-25
Refine base_permission_test and update the permission tests.

reiterl:
Add another way to call the base_permission_test.
Two tests should be done:
1. user with group (belong to meeting) but no permission and check 403 and Missing permission in msg.
2. user with group with permission and check 200.


jsangmeister (approved):

jsangmeister:
You should probably merge this first and then adjust the other PR's accordingly

PR #620, opened 2021-03-26
Add permissions to motion_category. Update tests.

reiterl:


openslides_backend/action/actions/motion_category/number_motions.py
```diff
@@ -203,3 +205,10 @@ def get_number(
             )
             number = f"{prefix}{blank}{number_value_str}"
         return number, number_value_map[motion_id]
+
+    def get_meeting_id(self, instance: Dict[str, Any]) -> int:
+        motion_category = self.datastore.get(
+            FullQualifiedId(Collection("motion_category"), instance["id"]),
+            ["meeting_id"],
+        )
+        return motion_category["meeting_id"]
```
jsangmeister, line 214: Since we seem to have mulitple places where the "action model" differs from the "permission model", if you so want, I think it would be good to add another field to the base action class, something like `permission_model`, and change the base `get_meeting_id`. If it's empty, the method does the same as before, if not, it uses that model for the datastore call instead of the `model`.

jsangmeister (changes requested):

openslides_backend/action/actions/motion_category/number_motions.py
reiterl: That sounds good. I am working on it.

reiterl: Done.

jsangmeister (changes requested):
Looks good! Please also change all other places where you overwrote the `get_meeting_id`

reiterl:
Checked, get_meeting_id is used in list_of_speakers, replaced it there.

jsangmeister:
Please also replace the usage in `MotionSubmitterSort`, then this can be merged.

jsangmeister (approved):

PR #622, opened 2021-03-26
issue595_projection_of_non_temporary_user

r-peschke:
Please, check especially the first 5 tests. I'm not sure if they behave like they should.
See also my question from https://github.com/OpenSlides/openslides-backend/issues/595

### DIFF
--- a/tests/system/action/projector/test_project.py
+++ b/tests/system/action/projector/test_project.py
@@ -54,32 +54,119 @@ class ProjectorProject(BaseActionTestCase):
                 "ids": [23],
                 "content_object_id": "assignment/453",
                 "meeting_id": 1,
-                "options": "",
+                "options": {},
                 "stable": False,
                 "type": "test",
             },
         )
         self.assert_status_code(response, 200)
-        self.assert_model_exists("projection/105", {'content_object_id': 'assignment/452', 'history_projector_id': 23, 'stable': False, 'weight': 2})
-        self.assert_model_exists("projection/106", {'content_object_id': 'assignment/452', 'current_projector_id': 23, 'stable': True})
-        self.assert_model_exists("projection/112", {'content_object_id': 'assignment/453', 'current_projector_id': 23, "stable": False, "options": "", "type": "test"})
-        self.assert_model_exists("projection/110", {'content_object_id': 'assignment/453', 'history_projector_id': 75, 'stable': False, 'type': 'test', 'weight': 2})
-        self.assert_model_exists("projection/111", {'content_object_id': 'assignment/453', 'current_projector_id': 75, 'stable': True})
-        self.assert_model_exists("projector/23", {"current_projection_ids": [106, 112], "history_projection_ids": [105], "scroll": 0})
-        self.assert_model_exists("projector/75", {"current_projection_ids": [111], "history_projection_ids": [110]})
+        self.assert_model_exists(
+            "projection/105",
+            {
+                "content_object_id": "assignment/452",
+                "history_projector_id": 23,
+                "stable": False,
+                "weight": 2,
+            },
+        )
+        self.assert_model_exists(
+            "projection/106",
+            {
+                "content_object_id": "assignment/452",
+                "current_projector_id": 23,
+                "stable": True,
+            },
+        )
+        self.assert_model_exists(
+            "projection/112",
+            {
+                "content_object_id": "assignment/453",
+                "current_projector_id": 23,
+                "stable": False,
+                "options": {},
+                "type": "test",
+            },
+        )
+        self.assert_model_exists(
+            "projection/110",
+            {
+                "content_object_id": "assignment/453",
+                "history_projector_id": 75,
+                "stable": False,
+                "type": "test",
+                "weight": 2,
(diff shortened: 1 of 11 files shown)


tests/system/action/projector/test_project.py
```diff
@@ -52,132 +53,166 @@ def test_project(self) -> None:
...
+            "projector.project", {"ids": [], "content_object_id": "user/2", "meeting_id": 1}
         )
-        self.assert_model_exists(
-            "projection/112",
-            {
-                "content_object_id": "meeting/1",
-                "current_projector_id": 65,
-                "meeting_id": 1,
-            },
+        self.assert_status_code(response, 200)
```
FinnStutzenstein, line 151: Can you check, that there is a projection with this user? (Same for the temporary user test)

tests/system/action/projector/test_project.py
```diff
@@ -52,132 +53,166 @@ def test_project(self) -> None:
             {
                 "ids": [23],
                 "content_object_id": "assignment/453",
-                "options": {},
+                "meeting_id": 1,
```
FinnStutzenstein, line 56: I see: You've added a meeting id. Ok

tests/system/action/projector/test_project.py
```diff
@@ -52,132 +53,166 @@ def test_project(self) -> None:
...
+            response.json["message"]
+        )
+
+    def test_project_wrong_meeting_by_content_meeting(self) -> None:
+        response = self.request(
+            "projector.project",
+            {"ids": [], "content_object_id": "meeting/1", "meeting_id": 2, "stable": True},
         )
+        self.assert_status_code(response, 400)
+        self.assertIn("The following models do not belong to meeting 2: [FullQualifiedId('meeting/1')]", response.json["message"])
```
FinnStutzenstein, line 203: Is it possible to add a `__str__` or `__repr__` to `FullQualifiedId` so it just prints `meeting/1`?

tests/system/action/projector/test_project.py
```diff
@@ -52,132 +53,166 @@ def test_project(self) -> None:
...
-            {
-                "content_object_id": "meeting/1",
-                "current_projector_id": 65,
-                "meeting_id": 1,
-            },
+        self.assert_status_code(response, 200)
+
+    def test_temporary_user_as_content_object_okay(self) -> None:
+        """
+        Although the temporaray user has to have a group, this test checks the temporary user without group set
```
FinnStutzenstein, line 155: So you are relying on `assert_belongs_to_meeting` to handle this, right? I would propose to add a group - maybe the behavior of it or some permission checks will require the group in some time.

FinnStutzenstein (commented):
Looks good

tests/system/action/projector/test_project.py
jsangmeister: Yes, this was needed to be able to unproject normal users - if no projector id is given, there is no way to determine the meeting which wanted to unproject them.

jsangmeister (approved):
LGTM. Did not check the test logic since @FinnStutzenstein is the expert on this one ;)

tests/system/action/projector/test_project.py
r-peschke: done for both

tests/system/action/projector/test_project.py
r-peschke: done

tests/system/action/projector/test_project.py
r-peschke: All checks of objects belonging to meeting rely on `assert_belongs_to_meeting`. Following the wiki it is obligatory for a temporary user to belong to a group. Having a group a temporary user would be always accepted to the meeting (of this group),  and there would be no difference to the normal user. But I wanted the user to be accepted with it's speciality having a meeting_id as temporary user.

tests/system/action/projector/test_project.py
```diff
@@ -52,132 +53,384 @@ def test_project(self) -> None:
...
+        )
+        self.assert_model_exists(
+            "projection/111",
+            {"current_projector_id": 75, "history_projector_id": None, "stable": True},
+        )
 
-    def test_project_wrong_meeting(self) -> None:
+    def test_try_to_store_second_unstable_projection(self) -> None:
+        """
+        Didn't succeed to store a second unstable projection
```
FinnStutzenstein, line 203: Since the status code is 200 the test name and docstring do not fit.

FinnStutzenstein (approved):
Two minor remarks. Looks good!

r-peschke:
I think we should restrict the number of payloadelements/actions to 1 for these action.
To implement this action completly we would need a filtered cache, for instance to prevent more than one unstable projection.
Additional the action is build with projector as model and yields as much projector-creates/updates as we have projectors in the ids list. Our testbase checks this number against that of the results, which is always 1.

FinnStutzenstein:
> I think we should restrict the number of payloadelements/actions to 1 for these action.

I cannot tell, whether this is necessary.

jsangmeister (approved):

PR #626, opened 2021-03-31
Add explicit test for group meeting_ids for temporary users. Update t…

reiterl:
…he tests.


jsangmeister (approved):
Ah right, forgot that the equal fields check already caught this one. But double is better ;)

reiterl:
Ralf has changed the place of the group_ids, so this pr is no longer useful.

jsangmeister:
> Ralf has changed the place of the group_ids, so this pr is no longer useful.

What do you mean?

reiterl:
Ah, talked with Ralf about it. I will merge this, after a rebase.

PR #634, opened 2021-04-07
Add permissions to resource action and update tests.

reiterl:


FinnStutzenstein (approved):
Logic is good. @jsangmeister Please check for code style. Regarding the `user_id>0`: maybe a generic `assert_not_anonymous` would be good. Whats your opinion?

jsangmeister:
> Regarding the `user_id>0`: maybe a generic `assert_not_anonymous` would be good. Whats your opinion?

Yes, that sounds very reasonable.

jsangmeister (changes requested):
Looks good. Please add an `assert_not_anonymous` like @FinnStutzenstein suggested and also use it in all other places where it is checked for anonymous.

reiterl:
Add a `check_not_anonymous()` function and use it instead of `user_id >0`.

jsangmeister:
This is not what we (at least I) meant: `assert_not_anonymous` would directly throw an error if the user is anonymous. An example can be found [here](https://github.com/OpenSlides/openslides-backend/blob/0fd7405993ea5f42dbe514cdfa9058f1559d3b34/openslides_backend/action/actions/user/set_password_self.py#L44). You can make the error message more general like "You cannot do this as anonymous". You can then use that method here and in other places where `auth.is_anonymous` or `user_id (==|>) 0` is used.

reiterl:
Now assert in resource mixins that the user is not anonymous.

jsangmeister (approved):
Not exactly what I meant, but ok.

PR #635, opened 2021-04-07
Add permissions to speaker, update tests.

reiterl:


FinnStutzenstein (approved):
LGTM

openslides_backend/action/actions/speaker/create_update_delete.py
```diff
@@ -179,14 +180,31 @@ def validate_fields(self, instance: Dict[str, Any]) -> Dict[str, Any]:
                 )
         return super().validate_fields(instance)
 
+    def check_permissions(self, instance: Dict[str, Any]) -> None:
+        if instance.get("user_id") == self.user_id:
+            self.permission = Permissions.ListOfSpeakers.CAN_BE_SPEAKER
```
jsangmeister, line 185: Mmh, I don't like the changing of class variables here. How about you extend the base `check_permissions` to accept an optional `permission` argument which, if given, is used instead of `self.permission`?

jsangmeister (changes requested):

openslides_backend/action/actions/speaker/create_update_delete.py
reiterl: Mmh, I don't like to change the signature of check_permissions, so I use has-perm to do it without changing of class variables.

jsangmeister: ok

jsangmeister (approved):

PR #638, opened 2021-04-08
Issue473 add projector relations to meeting.update

r-peschke:


openslides_backend/action/actions/meeting/update.py
```diff
@@ -0,0 +1,149 @@
...
+        if "reference_projector_id" in instance:
+            if reference_projector_id := instance["reference_projector_id"]:
+                meeting_check.append(
+                    FullQualifiedId(Collection("projector"), reference_projector_id)
+                )
+            else:
+                raise ActionException("The reference projector can't be set to null.")
+        if "default_projector_$_id" in instance:
+            if any((id is None for id in instance["default_projector_$_id"].values())):
+                raise ActionException("The default projecor can't be set to null.")
```
jsangmeister, line 139: I would suggest to set this relation to `required` in the models file, so you can save these checks.

jsangmeister (changes requested):
Good split of files. If you'd want to you could even go ahead and put the settings keys in their own file.

openslides_backend/action/actions/meeting/update.py
r-peschke: If we set them to required:
- the reference_projector_id required will be caught by the schema validator
- the default_projector_$_id required can be caught also by the schema validator. We have to create a required validation schema
- To catch these errors in the validate_required fields method would be more difficult:
-- it's difficult to bypass the schema and other validations
-- Template fields are not implemented on required. We had to implement this with regex for only this case

I think we should realize the schema validation variant

r-peschke:
> Good split of files. If you'd want to you could even go ahead and put the settings keys in their own file.
The meeting_setting_keys are only used one in the update.py. Then IMO it would be better to put them directly into the schema definition.

jsangmeister (approved):

jsangmeister (approved):

PR #648, opened 2021-04-13
Add permissions to most poll und option actions.

reiterl:
Update and add tests.
Use a mixin for the poll actions.
Full tests are in poll.create tests. Use less tests in the rest of the actions, because it is the some mixin.

### DIFF
--- a/openslides_backend/action/actions/option/update.py
+++ b/openslides_backend/action/actions/option/update.py
@@ -1,15 +1,14 @@
 from typing import Any, Dict, List, Optional, Tuple
 
 from ....models.models import Option, Poll
-from ....permissions.permission_helper import has_perm
-from ....permissions.permissions import Permission, Permissions
 from ....services.datastore.commands import GetManyRequest
-from ....shared.exceptions import ActionException, PermissionDenied
-from ....shared.patterns import KEYSEPARATOR, Collection, FullQualifiedId
+from ....shared.exceptions import ActionException
+from ....shared.patterns import Collection, FullQualifiedId
 from ....shared.schema import decimal_schema
 from ...generics.update import UpdateAction
 from ...util.default_schema import DefaultSchema
 from ...util.register import register_action
+from ..poll.mixins import check_poll_or_option_perms
 from ..poll.set_state import PollSetState
 from ..vote.create import VoteCreate
 from ..vote.update import VoteUpdate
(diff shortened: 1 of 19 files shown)


openslides_backend/action/actions/option/update.py
```diff
@@ -191,3 +195,21 @@ def check_state_change(
...
+        elif content_object_id.startswith("assignment" + KEYSEPARATOR):
+            self._check_perm(Permissions.Assignment.CAN_MANAGE, meeting_id)
+        else:
+            self._check_perm(Permissions.Poll.CAN_MANAGE, meeting_id)
+
+    def _check_perm(self, perm: Permission, meeting_id: int) -> None:
+        if not has_perm(self.datastore, self.user_id, perm, meeting_id):
+            msg = f"You are not allowed to perform action {self.name}."
+            msg += f" Missing permission: {perm}"
+            raise PermissionDenied(msg)
```
jsangmeister, line 215: Please deduplicate this code with the poll permission mixin. Either extract this into its own seperate function which is then called in the mixin, split the mixin into two (one which holds the actual code and one which overrides `check_permissions`), etc.

tests/system/action/poll/test_create.py
```diff
@@ -596,3 +595,118 @@ def test_create_user_option_invalid(self) -> None:
...
+                "global_yes": True,
+                "global_no": True,
+                "global_abstain": True,
+                "onehundred_percent_base": "Y",
+                "majority_method": "simple",
+            },
+            Permissions.Poll.CAN_MANAGE,
+        )
+
+    def test_create__no_permissions(self) -> None:
```
jsangmeister, line 696: please remove double underscore

jsangmeister (changes requested):
rest looks good!

openslides_backend/action/actions/option/update.py
reiterl: Reduced the dublication.
I added check_poll_or_option_permission and change _check_perm into an function.
Use these functions to reduce the code dublication.

tests/system/action/poll/test_create.py
reiterl: Removed.

openslides_backend/action/actions/poll/mixins.py
```diff
@@ -19,16 +20,49 @@ def check_permissions(self, instance: Dict[str, Any]) -> None:
...
+    datastore: DatastoreService,
+    user_id: int,
+    perm: Permission,
+    meeting_id: int,
+    action_name: str,
+) -> None:
+    if not has_perm(datastore, user_id, perm, meeting_id):
+        msg = f"You are not allowed to perform action {action_name}."
+        msg += f" Missing permission: {perm}"
+        raise PermissionDenied(msg)
```
jsangmeister, line 68: This does not need to be a separate method, you can just determine the needed permission in the if-else-block above and then move this code up and call `has_perm` with the determined permission.

jsangmeister (changes requested):

openslides_backend/action/actions/poll/mixins.py
reiterl: Refactor check_poll_or_option_perm. Remove _check_perm and move perm calculation in the beginning.

jsangmeister (approved):

Issue #652, opened 2021-04-15
Use better crypto library


jsangmeister:
Replace the random library used for password (and in the future user token) creation with the `secrets` library. See https://github.com/django/django/blob/main/django/utils/crypto.py and https://docs.python.org/3/library/secrets.html.

PR #659, opened 2021-04-15
Generate default_password and/or store hashed_password.

r-peschke:
Unify password generation

### DIFF
--- a/openslides_backend/action/actions/user/password_mixin.py
+++ b/openslides_backend/action/actions/user/password_mixin.py
@@ -13,11 +13,8 @@ class PasswordCreateMixin(BaseAction):
         return instance
 
     def generate_and_set_password(self, instance: Dict[str, Any]) -> Dict[str, Any]:
-        password = self.generate_password()
-        instance["default_password"] = password
-        hashed_password = self.auth.hash(password)
-        instance["password"] = hashed_password
-        return instance
+        instance["default_password"] = self.generate_password()
+        return self.set_password(instance)
 
     @staticmethod
     def generate_password() -> str:
--- a/openslides_backend/action/actions/user/password_mixin.py
+++ b/openslides_backend/action/actions/user/password_mixin.py
@@ -1,0 +1,27 @@
+import random
+from typing import Any, Dict
+
+from ...action import BaseAction
+
+
+class PasswordCreateMixin(BaseAction):
+    def set_password(self, instance: Dict[str, Any]) -> Dict[str, Any]:
+        password = instance.get("default_password", "")
+        instance["default_password"] = password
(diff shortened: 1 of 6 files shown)


openslides_backend/action/actions/user/password_mixin.py
```diff
@@ -0,0 +1,27 @@
+import random
+from typing import Any, Dict
+
+from ...action import BaseAction
+
+
+class PasswordCreateMixin(BaseAction):
+    def set_password(self, instance: Dict[str, Any]) -> Dict[str, Any]:
+        password = instance.get("default_password", "")
+        instance["default_password"] = password
```
jsangmeister, line 10: This is redundant

openslides_backend/action/actions/user/password_mixin.py
```diff
@@ -0,0 +1,27 @@
...
+        hashed_password = self.auth.hash(password)
+        instance["password"] = hashed_password
+        return instance
+
+    def generate_and_set_password(self, instance: Dict[str, Any]) -> Dict[str, Any]:
+        password = self.generate_password()
+        instance["default_password"] = password
+        hashed_password = self.auth.hash(password)
+        instance["password"] = hashed_password
+        return instance
```
jsangmeister, line 20: And this can be unified with the method above

jsangmeister (changes requested):
While you're at it , you could go ahead and do #652 in this PR as well. The user token creation can then be adjusted accordingly after that.

r-peschke:
> While you're at it , you could go ahead and do #652 in this PR as well. The user token creation can then be adjusted accordingly after that.
This is a bug issue and Gabriel asked me yesterday to resolve it urgently. Because of that IMO it is better to resolve #652 separate.

jsangmeister:
> This is a bug issue and Gabriel asked me yesterday to resolve it urgently. Because of that IMO it is better to resolve #652 separate.

ok

openslides_backend/action/actions/user/password_mixin.py
r-peschke: Done

openslides_backend/action/actions/user/password_mixin.py
r-peschke: Done

openslides_backend/action/actions/user/password_mixin.py
jsangmeister: No, I meant specifically the two marked lines, because you are retrieving the field from `instance`and then directly writing it back :D

jsangmeister: Ah, I see you already resolve it yourself

PR #662, opened 2021-04-16
validate_required_fields checks required dependent from field-type

r-peschke:


openslides_backend/models/base.py
```diff
@@ -138,12 +138,34 @@ def get_required_fields(self) -> Iterable[fields.Field]:
...
+        It returns True if the related value in instance doesn't fulfill the require-requirement
+        """
+        if field.own_field_name not in instance:
+            return False
+        if isinstance(
+            field, (fields.IntegerField, fields.BooleanField, fields.FloatField)
+        ):
+            return instance[field.own_field_name] is None
+        else:
+            return not instance[field.own_field_name]
```
jsangmeister, line 171: This method belongs to the `Field`, not the `Model`, and it should suffice to just pass the value of the field instead of the full instance. I would suggest to move this method to the base `Field` class and overwrite it (either directly or via a mixin) in `IntegerField`, `BooleanField` and `FloatField`. That way, it is also easily adjustable on a per-field basis.

jsangmeister (changes requested):

openslides_backend/models/base.py
r-peschke: Good idea, also it needs instance and create/update information => Done

jsangmeister (approved):
This looks better. You could now go ahead and create a Mixin like `NullableField` so you don't have the same method copy-pasted thrice, but I guess this is also fine.

PR #669, opened 2021-04-20
Add get_forwarding_meetings presenter.

reiterl:
Update tests.
Modify the infrastructure to allow the permission checks (user_id passed
thru) and allow to test the permissions.

### DIFF
--- a/openslides_backend/presenter/get_forwarding_meetings.py
+++ b/openslides_backend/presenter/get_forwarding_meetings.py
@@ -39,7 +39,7 @@ class GetForwardingMeetings(BasePresenter):
             Permissions.Motion.CAN_MANAGE,
             self.data["meeting_id"],
         ):
-            msg = "You are not allowed to perform persenter get_forwarding_meetings"
+            msg = "You are not allowed to perform presenter get_forwarding_meetings"
             msg += f" Missing permission: {Permissions.Motion.CAN_MANAGE}"
             raise PermissionDenied(msg)
 
--- a/openslides_backend/presenter/get_forwarding_meetings.py
+++ b/openslides_backend/presenter/get_forwarding_meetings.py
@@ -1,0 +1,87 @@
+from typing import Any
+
+import fastjsonschema
+
+from ..permissions.permission_helper import has_perm
+from ..permissions.permissions import Permissions
+from ..shared.exceptions import PermissionDenied
+from ..shared.patterns import Collection, FullQualifiedId
+from ..shared.schema import required_id_schema, schema_version
+from .base import BasePresenter
+from .presenter import register_presenter
+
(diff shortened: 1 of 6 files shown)


openslides_backend/presenter/get_forwarding_meetings.py
```diff
@@ -0,0 +1,87 @@
...
+
+    def get_result(self) -> Any:
+        # check permission
+        if not has_perm(
+            self.datastore,
+            self.user_id,
+            Permissions.Motion.CAN_MANAGE,
+            self.data["meeting_id"],
+        ):
+            msg = "You are not allowed to perform persenter get_forwarding_meetings"
```
jsangmeister, line 42: typo

tests/system/presenter/test_get_forwarding_meetings.py
```diff
@@ -0,0 +1,107 @@
...
+        self.assertEqual(
+            data,
+            [
+                {
+                    "id": 3,
+                    "name": "com3",
+                    "meetings": [{"id": 2, "name": "meeting2"}],
+                }
+            ],
+        )
```
jsangmeister, line 107: Please add a new test (or add it to this test) where a meeting is excluded from the result because the user does not have the CAN_CREATE permission in any of its meetings.

jsangmeister (changes requested):
Looks good!

jsangmeister:
If we need more permissions checks for presenters in the future, we could think about unifying that like for actions, but I think that would currently be overkill.

tests/system/presenter/test_get_forwarding_meetings.py
reiterl: This is checked with meeting/3 asaik. It is excluded, because of the missing permission.

openslides_backend/presenter/get_forwarding_meetings.py
reiterl: Fixed.

tests/system/presenter/test_get_forwarding_meetings.py
jsangmeister: No, meeting 3 belongs to committee 3 which is included in the result. What I meant is adding a fourth committee with it's own meeting in which the user does not have the `motion.can_create` permission and then asserting that the additional committee is not included in the result.

reiterl:
Rebased.

jsangmeister (changes requested):
see open comment

tests/system/presenter/test_get_forwarding_meetings.py
reiterl: Updated the complex. Add another meeting / committee without perms. The additional committee is not in the results.

jsangmeister (approved):

PR #670, opened 2021-04-21
Add draft motion.create_forwarded and tests.

reiterl:
Add a draft version of a create_forwarded action and add some tests.

It is similar to the create action, but a bit shorted. I think it is good enough for discussions.
I think I shouldn't do another mixin, because this would enlarge the complexity.


openslides_backend/action/actions/motion/create_forwarded.py
```diff
@@ -0,0 +1,113 @@
...
+from ..list_of_speakers.list_of_speakers_creation import (
+    CreateActionWithListOfSpeakersMixin,
+)
+from ..motion_submitter.create import MotionSubmitterCreateAction
+from .sequential_numbers_mixin import SequentialNumbersMixin
+from .set_number_mixin import SetNumberMixin
+
+
+@register_action("motion.create_forwarded")
+class MotionCreate(
```
jsangmeister, line 22: Please rename

jsangmeister (changes requested):
You should deduplicate this with `motion.create`. I agree that a single mixin is not the way to go here, since it would not have a strictly defined task, but rather mutliple. So, multiple approaches:
- Make this a subclass of `MotionCreate`. Create a method in that class for each single task (e.g. `calculate_state_id`, `check_origin_id`, etc.) and call them accordingly. In this class, overwrite `update_instance` and also call them accordingly, but only the ones you need. The benefit of this approach would be that you don't have to explicitly inherit from all mixins and base classes again and do not have to set some class variables like `model` (I think, I'm not quite sure how python handles class variables here).
- Create a common base class for `MotionCreate` and `MotionCreateForwarded` which implements the methods mentioned above. Inherit both classes from the base class and call the methods accordingly. This has the same benefits as the first suggestion and may conceptionally be even clearer - like @r-peschke discovered, randomly inheriting from existing actions is not always the best idea ;)
- Create a seperate mixin for each task like setting the state id and checking the origin id. Probably my least favourite approach, since it complicates the class hierarchy even more and creates even more super classes.

So the best way to go would probably be approach #2. Please talk to me if you have any questions or other suggestions regarding this. But copy-pasting the code should not be the way to go here.

Also, the permission checks are missing completely. I suppose we also need the checks regarding the origin id in the normal `motion.create` @FinnStutzenstein?

jsangmeister:
> You should deduplicate this with `motion.create`. I agree that a single mixin is not the way to go here, since it would not have a strictly defined task, but rather mutliple. So, multiple approaches:
> 
> * Make this a subclass of `MotionCreate`. Create a method in that class for each single task (e.g. `calculate_state_id`, `check_origin_id`, etc.) and call them accordingly. In this class, overwrite `update_instance` and also call them accordingly, but only the ones you need. The benefit of this approach would be that you don't have to explicitly inherit from all mixins and base classes again and do not have to set some class variables like `model` (I think, I'm not quite sure how python handles class variables here).
> * Create a common base class for `MotionCreate` and `MotionCreateForwarded` which implements the methods mentioned above. Inherit both classes from the base class and call the methods accordingly. This has the same benefits as the first suggestion and may conceptionally be even clearer - like @r-peschke discovered, randomly inheriting from existing actions is not always the best idea ;)
> * Create a seperate mixin for each task like setting the state id and checking the origin id. Probably my least favourite approach, since it complicates the class hierarchy even more and creates even more super classes.
> 
> So the best way to go would probably be approach #2. Please talk to me if you have any questions or other suggestions regarding this. But copy-pasting the code should not be the way to go here.
> 
> Also, the permission checks are missing completely. I suppose we also need the checks regarding the origin id in the normal `motion.create` @FinnStutzenstein?

openslides_backend/action/actions/motion/create_forwarded.py
reiterl: Done. And renamed this and the test class into MotionCreateForwarded.

reiterl:
Use Approach 2 to deduplicate the code.
Add a MotionCreateBase class and include there the code from both actions in functions.
Created functions: set_state_from_workflow, check_for_origin_id,create_submitters,set_created_last_modified_and_number
and use them in both actions (create/create_forwarded).

openslides_backend/action/actions/motion/create_base.py
```diff
@@ -0,0 +1,105 @@
...
+            submitter_ids = [self.user_id]
+        self.apply_instance(instance)
+        action_data = []
+        for user_id in submitter_ids:
+            action_data.append({"motion_id": instance["id"], "user_id": user_id})
+        self.execute_other_action(MotionSubmitterCreateAction, action_data)
+
+        instance["sequential_number"] = self.get_sequential_number(
+            instance["meeting_id"]
+        )
```
jsangmeister, line 93: This does not belong in this function. You can probably add it to `set_created_last_modified_and_number` and rename that to something like `set_additional_motion_fields` (and add a docstring explaining which fields are set)

jsangmeister (changes requested):
Looks good! Permission checks and tests for them are still missing.

openslides_backend/action/actions/motion/create_base.py
reiterl: Add set_sequential_number() function and set this number there.

reiterl:
Add permissions to this action.
Add three tests. Two missing permissions and one with all permissions test.

jsangmeister (approved):

openslides_backend/action/actions/motion/create.py
```diff
@@ -103,78 +83,9 @@ def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
...
-        self.set_number(
-            instance,
-            instance["meeting_id"],
-            instance["state_id"],
-            instance.get("lead_motion_id"),
-            instance.get("category_id"),
-        )
-
+        self.set_state_from_workflow(instance, meeting)
+        self.check_for_origin_id(instance)
```
FinnStutzenstein, line 87: There is not `origin_id` settable for motion.create. This can be just implemented in the create_forwarded class

jsangmeister: Ah ok. I will remove this from [the wiki](https://github.com/OpenSlides/OpenSlides/wiki/motion.create) then.

jsangmeister: @reiterl please remove from the action. Then this can be merged

reiterl: Remove origin_id from motion.create and update tests.

jsangmeister (approved):

openslides_backend/action/actions/motion/create_base.py
```diff
@@ -0,0 +1,106 @@
...
+                FullQualifiedId(Collection("motion_workflow"), workflow_id),
+                ["first_state_id"],
+            )
+            instance["state_id"] = workflow.get("first_state_id")
+        else:
+            raise ActionException(
+                "No matching default workflow defined on this meeting"
+            )
+
+    def check_for_origin_id(self, instance: Dict[str, Any]) -> None:
```
FinnStutzenstein, line 54: Move this directly to `create_forwarded.py` since it is only needed there and a special thing for forwarded motions. the regular `motion.create` should not know about it.

tests/system/action/motion/test_create.py
```diff
@@ -175,87 +175,6 @@ def test_create_missing_state(self) -> None:
             response.json["message"]
         )
 
-    def test_correct_origin_id_set(self) -> None:
```
FinnStutzenstein, line 178: Add a small test with an `origin_id` set and assert that is rejected due to the `origin_id`.

tests/system/action/motion/test_create_forwarded.py
```diff
@@ -0,0 +1,174 @@
...
+            },
+            "committee/52": {"name": "name_EeKbwxpa"},
+            "committee/53": {
+                "name": "name_auSwgfJC",
+                "forward_to_committee_ids": [52],
+            },
+        }
+
+    def test_correct_origin_id_set(self) -> None:
+        self.set_models(
```
FinnStutzenstein, line 36: these looks very copy&pasted. Please clean them up for every test, so only relevant things are set.

FinnStutzenstein (changes requested):
Some remarks are left

openslides_backend/action/actions/motion/create_base.py
reiterl: Moved `check_for_origin_id()` to motion.create_forwarded.

tests/system/action/motion/test_create.py
reiterl: Add a missing origin test. It is a short test, where the origin_id is set and the motion is missing.

reiterl: Is that you are asking for? If not, please explain in more detail the test, which is missing.

tests/system/action/motion/test_create_forwarded.py
reiterl: Add a test_model and use it in most contexts. This reduce the set_model code.

tests/system/action/motion/test_create.py
FinnStutzenstein: Yes, this sounds good. But where is this test?

reiterl: The test should be `test_missing_origin`

GabrielInTheWorld:
I have the following structure:
```
Orga
|
|--- Default Committee (id: 1)
|    |- Meeting A (id: 1)
|    |- Meeting U (id: 2)
|
|--- Committee B (id: 2)
|    |- Meeting B (id: 3)
|
|--- Committee C (id: 3)
     |- Meeting C (id: 4)
     |- Meeting D (id: 5)
     |- Meeting E (id: 6)
```

Motions can be forwarded from `committee/1` to `committee/2` and `committee/3`.
The `user/1` is a member of all meetings (not of the committees). If I signed in as `user/1`, I can forward motions from `meeting/1` to `meeting/4`, `meeting/5` and `meeting/6`, but not to `meeting/3`. If I try to forward motions from `meeting/1` to `meeting/3`, then the following error appears: `Fehler: Committee id 1 not in [2, 3, 4]`.
Are there any other restrictions I have to consider? Otherwise, I think this is a bug...

GabrielInTheWorld:
Successfully tested!

GabrielInTheWorld (approved):

reiterl:
Finn, I added a test which checks, if the origin_id is not in create.

FinnStutzenstein (approved):

PR #640, opened 2021-04-09
Issue587 permissions user  see restructuring and integration OrganisationManagementLevel

r-peschke:

Issue #639, opened 2021-04-08
Implement organisation_tag actions


FinnStutzenstein:
- https://github.com/OpenSlides/OpenSlides/wiki/organisation_tag.create
- https://github.com/OpenSlides/OpenSlides/wiki/organisation_tag.update
- https://github.com/OpenSlides/OpenSlides/wiki/organisation_tag.delete

Note, that it needs an updated models.py (See https://github.com/OpenSlides/OpenSlides/pull/6000)

GabrielInTheWorld:
Please keep in mind, that https://github.com/OpenSlides/OpenSlides/wiki/committee.update also allows adding organisation_tags.

FinnStutzenstein:
Right, the same goes for committee.create, meeting.create, meeting.update

PR #675, opened 2021-04-23
Add organisation tag actions and tests.

reiterl:
Don't include permissions. Wait for https://github.com/OpenSlides/openslides-backend/pull/640 for this.
Orga level permissions should be easy then.

### DIFF
--- a/tests/system/action/organisation_tag/test_create.py
+++ b/tests/system/action/organisation_tag/test_create.py
@@ -3,8 +3,10 @@ from tests.system.action.base import BaseActionTestCase
 
 class OrganisationTagCreate(BaseActionTestCase):
     def test_create(self) -> None:
+        self.set_models({"organisation/1": {}})
         response = self.request(
-            "organisation_tag.create", {"name": "wSvQHymN", "color": "#eeeeee"}
+            "organisation_tag.create",
+            {"name": "wSvQHymN", "color": "#eeeeee", "organisation_id": 1},
         )
         self.assert_status_code(response, 200)
         organisation_tag = self.get_model("organisation_tag/1")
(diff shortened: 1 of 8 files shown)


jsangmeister (approved):
Please note the actions mentioned in #639 that can set organisation tags. If you do not want to implement that in this PR, then please open a new issue for that. Same goes for the permissions.

FinnStutzenstein (commented):
On create I've added `organisation_id: Id;` to the payload. The orga tags weren't assigned to the organisation beforehand.

jsangmeister:
> On create I've added `organisation_id: Id;` to the payload. The orga tags weren't assigned to the organisation beforehand.

Since we only allow one organisation, can't we add that implicitely in the backend? Seems unnecessary that the client always sends it if the only allowed value is 1 anyway.

FinnStutzenstein:
If we add more orgas, this is one place less to change and one special case less to program into since just adding `organisation_id` should work out of the box. The current limitation is the client, not the backend. Yes, there are no `organisation.create` actions, but technically, this would work.

jsangmeister:
Ok, if you mean

reiterl:
Add the permissions and test.

reiterl:
Add organisation_id to organisation_tag create.

tests/system/action/organisation_tag/test_create.py
```diff
@@ -0,0 +1,54 @@
...
+    def test_create(self) -> None:
+        self.set_models({"organisation/1": {}})
+        response = self.request(
+            "organisation_tag.create",
+            {"name": "wSvQHymN", "color": "#eeeeee", "organisation_id": 1},
+        )
+        self.assert_status_code(response, 200)
+        organisation_tag = self.get_model("organisation_tag/1")
+        assert organisation_tag.get("name") == "wSvQHymN"
+        assert organisation_tag.get("color") == "#eeeeee"
```
FinnStutzenstein, line 14: assert, that it is linked to the organisation. check if `organisation/1/organisation_tag_ids` is `[]`.

FinnStutzenstein (approved):

reiterl:
Add issue https://github.com/OpenSlides/openslides-backend/issues/699 for the set organisation_tag code.

PR #686, opened 2021-04-27
Update committee.update: Add receive_forwardings_from_committee_ids. 

reiterl:
Add a test.

### DIFF
--- a/tests/system/action/committee/test_update.py
+++ b/tests/system/action/committee/test_update.py
@@ -6,6 +6,7 @@ class CommitteeUpdateActionTest(BaseActionTestCase):
     COMMITTEE_FQID = "committee/1"
     COMMITTEE_NAME = "committee_testname"
     COMMITTEE_ID_FORWARD = 2
+    COMMITTEE_FQID_FORWARD = "committee/2"
 
     def create_data(self) -> None:
         self.set_models(
@@ -78,6 +79,25 @@ class CommitteeUpdateActionTest(BaseActionTestCase):
         )
         self.assertEqual(model.get("template_meeting_id"), 200)
         self.assertEqual(model.get("default_meeting_id"), 201)
+
+    def test_update_receive_forwardings(self) -> None:
+        self.create_data()
+        response = self.request(
+            "committee.update",
+            {
+                "id": self.COMMITTEE_ID_FORWARD,
+                "receive_forwardings_from_committee_ids": [self.COMMITTEE_ID],
+            },
+        )
+        self.assert_status_code(response, 200)
+        self.assert_model_exists(
+            self.COMMITTEE_FQID,
+            {"forward_to_committee_ids": [self.COMMITTEE_ID_FORWARD]},
+        )
+        self.assert_model_exists(
+            self.COMMITTEE_FQID_FORWARD,
+            {"receive_forwardings_from_committee_ids": [self.COMMITTEE_ID]},
+        )
 
     def test_update_wrong_member_ids(self) -> None:
         self.create_data()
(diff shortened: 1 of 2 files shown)


jsangmeister (changes requested):
Please add more tests to make sure the relation handling works correctly in all cases:
- updating both `forward_to_committee_ids` and `receive_forwardings_from_committee_ids` in one request
- A->B exists, test that the request for B with B->C works and sets the reverse relations on A and C correctly
- A->C and B->C exist, test that the request for C with {B, D}->C works and sets the reverse relations on A and D correctly
- C->A and C->B exists, test that the request for C with C->{B,D} works and sets the reverse relations on A and D correctly
- C->A and C->B exists, test that the request for C with C->{} works and sets the reverse relations on A and B correctly

I hope it is clear what the notation means, else feel free to ask.

reiterl:
I'm not sure, why we should do this tests. It uses just the normal relation handling. I don't want to check the normal relation handling here.

jsangmeister:
> I'm not sure, why we should do these tests. It uses just the normal relation handling. I don't want to check the normal relation handling here.

We are using system tests for all our tests. We do not have many n:m relations which can be updated from both sides, expecially not at the same time. So please add the tests to ensure that the relation handling works correctly.

jsangmeister:
Also, more tests for simoultanious updates of both fields are needed, similar to tha other ones described, e.g. deleting one and setting the other, removing only one entry off of multiple, ...

reiterl:
Add some more tests.

tests/system/action/committee/test_update.py
```diff
@@ -132,6 +132,190 @@ def test_update_both_forwarded_and_received(self) -> None:
         )
         self.assert_model_exists("committee/3", {"forward_to_committee_ids": [1]})
 
+    def test_update_complex_1(self) -> None:
+        """A->B exists, test that the request for B with B->C works and sets the reverse relations on A and C correctly"""
```
jsangmeister, line 136: This test is unnecessary since the previous relations are not influenced by the request. I don't know what I was thinking there :D anyway, you can delete this

tests/system/action/committee/test_update.py
```diff
@@ -316,6 +316,96 @@ def test_update_complex_4(self) -> None:
...
+                "id": 1,
+                "forward_to_committee_ids": [],
+                "receive_forwardings_from_committee_ids": [],
+            },
+        )
+        self.assert_status_code(response, 200)
+        self.assert_model_exists(
+            "committee/1",
+            {
+                "receive_forwardings_from_committee_ids": [],
```
jsangmeister, line 347: Also assert `forward_to_committee_ids == []`

tests/system/action/committee/test_update.py
```diff
@@ -316,6 +316,96 @@ def test_update_complex_4(self) -> None:
...
+            },
+        )
+        self.assert_model_exists(
+            "committee/2",
+            {
+                "forward_to_committee_ids": [4],
+                "receive_forwardings_from_committee_ids": [3],
+            },
+        )
+        self.assert_model_exists("committee/3", {"forward_to_committee_ids": [2]})
```
jsangmeister, line 404: Also assert `receive_forwardings_from_committee_ids == []`

jsangmeister (changes requested):
Looks good!

tests/system/action/committee/test_update.py
reiterl: Done.

tests/system/action/committee/test_update.py
reiterl: Done.

tests/system/action/committee/test_update.py
reiterl: Done.

jsangmeister (approved):

PR #687, opened 2021-04-27
Update way to generate PermissionDenied message. Refactoring.

reiterl:


openslides_backend/action/action.py
```diff
@@ -107,7 +107,13 @@ def perform(
             self.validate_instance(instance)
             # perform permission check not for internal actions
             if not internal:
-                self.check_permissions(instance)
+                try:
+                    self.check_permissions(instance)
+                except PermissionDenied as e:
+                    if e.message.startswith("Missing"):
```
jsangmeister, line 113: Why do you check the start of the error message? I think it should be added to all messages.

openslides_backend/action/actions/assignment_candidate/mixins.py
```diff
@@ -29,8 +29,7 @@ def check_permissions(self, instance: Dict[str, Any]) -> None:
         if assignment.get("phase") == "voting":
             permission = Permissions.Assignment.CAN_MANAGE
             if not has_perm(self.datastore, self.user_id, permission, meeting_id):
-                msg = f"You are not allowed to perform action {self.name}."
-                msg += f" Missing permission: {permission}"
+                msg = f"Missing permission: {permission}"
                 raise PermissionDenied(msg)
```
jsangmeister, line 33: You can generalize it even more by adding a constructor to `PermissionDenied` which takes a permission (or Management Level) and the calculate the errro message based on that directly. If a string is given, it is just set directly. Note that `Permission` inherits from `str`, so you have to check for `Permission` first in the constructor.

jsangmeister: Then, most of the error raises become one-liners.

jsangmeister (changes requested):

openslides_backend/action/actions/assignment_candidate/mixins.py
reiterl: There are also "Forbidden fields" permission denies or OrganisationLevel permission denies.
This would result in a complex __init__ to save one line. Too complex. Wouldn't do.

openslides_backend/action/action.py
reiterl: No, it should not added to all messages. There are as example anonymous permission denied and special permission messages.

FinnStutzenstein: Than use inheritance and create special child classes of `PermissionDenied`.

FinnStutzenstein: Then use inheritance and create special child classes of `PermissionDenied`.

jsangmeister: Yes, that would have also been my next suggestion, didn't think about the possible other error causes. Maybe something like `MissingPermission`?

openslides_backend/action/actions/assignment_candidate/mixins.py
jsangmeister: With inheritance, you also solve this. `MissingPermission` gets the constructor described above and `PermissionDenied` can be used in all other cases.

reiterl: Use a `MissingPermission` exception to calculate the message in the missing case.
It inherits from PermissionDenied and needs a permission and the action name.

openslides_backend/action/action.py
reiterl: Add `MissingPermission `exception and use it.

openslides_backend/action/action.py
```diff
@@ -107,13 +108,7 @@ def perform(
             self.validate_instance(instance)
             # perform permission check not for internal actions
             if not internal:
-                try:
-                    self.check_permissions(instance)
-                except PermissionDenied as e:
-                    if e.message.startswith("Missing"):
-                        msg = f"You are not allowed to perform action {self.name}."
-                        e.message = msg + " " + e.message
-                    raise e
```
jsangmeister: You can keep this and add the action name here, this way you don't have to pass `self.name` for every exception. You only need to catch `MissingPermission` then.

openslides_backend/action/actions/resource/mixins.py
```diff
@@ -21,5 +21,6 @@ def check_permissions(self, instance: Dict[str, Any]) -> None:
         ):
             return
 
-        msg = "Missing can_manage_organisation"
+        msg = f"You are not allowed to perform action {self.name}."
+        msg += "Missing can_manage_organisation"
         raise PermissionDenied(msg)
```
jsangmeister: Please enhance `MissingPermission` so that it also takes OML/CML as input

jsangmeister (changes requested):

openslides_backend/action/actions/resource/mixins.py
reiterl: What is OML and CML?

jsangmeister: Organisation management level and committee management level - see https://github.com/OpenSlides/OpenSlides/wiki/Permission-System#organisation-management-level-oml

openslides_backend/action/action.py
reiterl: I dislike to update the message here, if it is not necessary.

jsangmeister: The whole point of this PR was to remove the duplication from the permission exceptions, and that would do exactly that.

reiterl: The code is in again with the `MissingPermission` and without the `if e.message.startswith("Missing"):`

openslides_backend/action/actions/resource/mixins.py
reiterl: We should also allow string and other stuff here. So I use Any as type.

FinnStutzenstein (approved):
LGTM. @jsangmeister Do you know, if there are more places that needs the new exception? From the actions modified it looks like all currently implemented special cases are changed.

openslides_backend/action/actions/resource/mixins.py
jsangmeister: Where do you want to give strings? I would suggest that `Union[Permission, OrganisationManagementLevel, CommitteeManagementLevel]` should be what we want

jsangmeister:
> LGTM. @jsangmeister Do you know, if there are more places that needs the new exception? From the actions modified it looks like all currently implemented special cases are changed.

The `check_permissions` of the base action could also be adjusted. That's the only other place I found.

openslides_backend/action/actions/resource/mixins.py
reiterl: Done.

reiterl:
How should the `check_permissions` be modified?

openslides_backend/action/actions/resource/mixins.py
reiterl: Oh, not done, I run into circular imports with it.
* permissions.management_level import exceptions.PermissionException
* exceptions import permissions.management_level.
I think `Any `is a good solution to this.

reiterl:
Adjusted check_permissions.

openslides_backend/action/actions/resource/mixins.py
jsangmeister: Using `Any` is the lazy solution when deaaling with circular imports :D I think the mistake here is that the management levels throw a `PermissionDenied`, this should be a simple `TypeError`.

openslides_backend/action/action.py
```diff
@@ -195,7 +195,7 @@ def check_permissions(self, instance: Dict[str, Any]) -> None:
                     meeting_id,
                 ):
                     return
-                msg_appendix = f" Missing permission: {self.permission}"
+                raise MissingPermission(self.permission)
```
jsangmeister, line 198: You can also throw a `MissingPermission` in the above case for OML.

jsangmeister (changes requested):

openslides_backend/action/actions/resource/mixins.py
reiterl: Changed.

openslides_backend/action/action.py
reiterl: Changed.

openslides_backend/shared/exceptions.py
```diff
@@ -96,6 +101,8 @@ class PermissionDenied(ViewException):
 class MissingPermission(PermissionDenied):
     def __init__(
         self,
-        permission: Any,
+        permission: Union[
+            Permission, OrganisationManagementLevel, CommitteeManagementLevel
+        ],
     ) -> None:
         self.message = f"Missing permission: {permission}"
```
jsangmeister, line 108: Maybe we should change the message to include the type name of the permission, so that instead of `Missing permission: can_manage_users` it says `Missing OrganisationManagementLevel: can_manage_users`

jsangmeister (changes requested):
Please see my last comment, it should be easy to do with `type(...).__name__`. Then this can be merged and used in #701.

openslides_backend/shared/exceptions.py
reiterl: `type(..).__name__` doesn't give the right text. I use `isinstance` to get the permission category and set a string then.

jsangmeister: Why, what does it return? I'm not a fan of case distinctions there.

reiterl: It should return `Permission`. It returns `_Motion` as example.

jsangmeister: Ah, right. Well, I guess we can leave it like this then for now

jsangmeister (approved):

PR #688, opened 2021-04-27
Add handling of lead_motion_id for motion.create. Add a test.

reiterl:


tests/system/action/motion/test_set_number_mixin.py
```diff
@@ -1082,8 +1082,8 @@ def test_complex_example_per_category_2_4(self) -> None:
         )
         self.assert_status_code(response, 200)
         motion4 = self.get_model("motion/4")
-        assert motion4.get("number") == "A 002"
-        assert motion4.get("number_value") == 2
+        assert motion4.get("number") == "A 003"
+        assert motion4.get("number_value") == 3
```
jsangmeister, line 1086: Why did this change?

jsangmeister (changes requested):
I just noticed that the wiki said to add `motion_block_id` to the whitelist in the permission check. The correct field name is `block_id`. Can you add this change to this PR?

openslides_backend/action/actions/motion/create.py
```diff
@@ -90,6 +90,14 @@ def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
                 raise ActionException(
                     "You can't give amendment_paragraph_$ in this context"
                 )
+        # if lead_motion use category_id and block_id from the lead_motion
+        if instance.get("lead_motion_id"):
+            lead_motion = self.datastore.get(
```
FinnStutzenstein, line 95: what happens, if the lead motion id is invalid, e.g. a non existing lead motion? Maybe add a test

openslides_backend/action/actions/motion/create.py
```diff
@@ -90,6 +90,14 @@ def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
                 raise ActionException(
                     "You can't give amendment_paragraph_$ in this context"
                 )
+        # if lead_motion use category_id and block_id from the lead_motion
+        if instance.get("lead_motion_id"):
```
FinnStutzenstein, line 94: In general, this should only be done if the request user does not have `motion.can_manage`. Please add a test creating an amendment as an admin (having `motion.can_manage`) and setting `category_id` and `block_id` to something different than the lead motion. this must succeed.

Change the test slightly, that the request user does not have `motion.can_manage`: It must still succeed, but ignore the provided `block_id` and `category_id` and use the data from the lead motion.

FinnStutzenstein (changes requested):

openslides_backend/action/actions/motion/create.py
jsangmeister: `datastore.get` will throw an error. This should be no problem

FinnStutzenstein: But is it brought back to the client properly or does it do a 500er?

jsangmeister: It is delivered to the client

reiterl:
I have changed `motion_block_id` into `block_id`.
Add a `has_perm` check before setting `block_id` and `category_id.`

openslides_backend/action/actions/motion/create.py
reiterl: Changed, if the user doesn't have `motion.can_manage` both values are changed. And updated a test for it.

tests/system/action/motion/test_set_number_mixin.py
reiterl: Because the category_id handling was changed. So the category was used more times.
In the newest implementation, this is reverted again.

reiterl:
Rebased.

openslides_backend/action/actions/motion/create.py
jsangmeister: > It must still succeed, but ignore the provided `block_id` and `category_id` and use the data from the lead motion.

Shouldn't the request be denied because of missing permissions in this case? You try to set fields which you are not allowed to set. We said we didn't want to silently ignore parts of the payload.

FinnStutzenstein: Yes, this is right.

reiterl: I don't understand this discussion, what should be done here?

jsangmeister: Add a test where the request user does not have `motion.can_manage` and tries to create a motion providing `lead_motion_id`, `category_id` and `block_id`. The request should fail because the user does not have permission to update the two latter fields. This has to be adjusted in the code accordingly, currently it is just silently ignored (I think).

reiterl: No, that is not right. The requirement says: If the user doesn't have `motion.can_manage`, then a whitelist should be checked and `category_id` and `block_id` should be in the wl, because of the `lead_motion_id`. Else it doesn't make sense.

jsangmeister: Yes, as @FinnStutzenstein and I discussed above, this requirement was wrong. We want to throw errors in OS4 if the user sends any invalid input instead of silently ignoring it. I will change the wiki entry.

reiterl:
Changed the category_id code.

tests/system/action/motion/test_create.py
```diff
@@ -290,6 +290,33 @@ def test_create_missing_origin_id(self) -> None:
             "data must not contain {'origin_id'} properties" in response.json["message"]
         )
 
+    def test_create_lead_motion(self) -> None:
```
FinnStutzenstein, line 293: Name is wrong: `create_amendment` since the motion that has a lead motion is an amendment.

tests/system/action/motion/test_create.py
```diff
@@ -290,6 +290,33 @@ def test_create_missing_origin_id(self) -> None:
...
+        )
+        response = self.request(
+            "motion.create",
+            {
+                "title": "test_Xcdfgee",
+                "meeting_id": 1,
+                "workflow_id": 12,
+                "text": "test",
+                "lead_motion_id": 3,
+                "block_id": 13,
```
FinnStutzenstein, line 311: Please make two tests from this single one:
- The request user have not motion.can_manage. Do not give `block_id` and `category_id`. They must be taken from the lead motion.
- The request user has motion.can_manage. Do not give `block_id` and `category_id`. They must not be taken from the lead motion and be empty.

tests/system/action/motion/test_create.py
```diff
@@ -401,3 +434,34 @@ def test_create_permission_lead_motion(self) -> None:
             },
         )
         self.assert_status_code(response, 200)
+        self.assert_model_exists("motion/4", {"block_id": 13, "category_id": 12})
+
+    def test_create_lead_motion_extra_category_id(self) -> None:
```
FinnStutzenstein, line 439: name proposal: `test_create_amendment_no_perms_category_id`.

Maybe also test the `block_id` in a second test.

FinnStutzenstein (changes requested):

tests/system/action/motion/test_create.py
reiterl: Renamed.

tests/system/action/motion/test_create.py
reiterl: The first point is the `test_create_permission_amendment`
Add another test for the second point: `test_create_permission_amendment_2`

tests/system/action/motion/test_create.py
reiterl: Add a test and renamed the test.

tests/system/action/motion/test_create.py
```diff
@@ -290,6 +290,33 @@ def test_create_missing_origin_id(self) -> None:
...
+                "workflow_id": 12,
+                "text": "test",
+                "lead_motion_id": 3,
+                "block_id": 13,
+                "category_id": 12,
+            },
+        )
+        self.assert_status_code(response, 200)
+        self.assert_model_exists(
+            "motion/4", {"lead_motion_id": 3, "block_id": 13, "category_id": 12}
```
FinnStutzenstein, line 317: Since the same ids were used on create and the lead motion (12 and 13) it is not clear, where the final ids are taken from. Please use a different block/category in the lead motion and assert, that the ids from the request are taken.

tests/system/action/motion/test_create.py
```diff
@@ -401,3 +434,98 @@ def test_create_permission_lead_motion(self) -> None:
             },
         )
         self.assert_status_code(response, 200)
+        self.assert_model_exists("motion/4", {"block_id": 13, "category_id": 12})
+
+    def test_create_permission_amendment_2(self) -> None:
```
FinnStutzenstein, line 439: Rename: `test_create_amendment_non_admin` or so.

FinnStutzenstein: Please add a block and category to the lead motion, so it can be asserted, that these block/category are also taken for the amendment.

FinnStutzenstein (changes requested):

openslides_backend/action/actions/motion/create.py
```diff
@@ -69,6 +69,15 @@ def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
                 raise ActionException(
                     "You can't give amendment_paragraph_$ in this context"
                 )
+        # if lead_motion and not has perm motion_can_manage
+        # use category_id and block_id from the lead_motion
+        if instance.get("lead_motion_id"):
```
FinnStutzenstein, line 74: The comment above is right: This should only happens if there is a lead motion *and the user has not manage rights*. The second part is missing, so the test below is wrong.

tests/system/action/motion/test_create.py
```diff
@@ -290,6 +290,35 @@ def test_create_missing_origin_id(self) -> None:
...
+                "workflow_id": 12,
+                "text": "test",
+                "lead_motion_id": 3,
+                "block_id": 13,
+                "category_id": 12,
+            },
+        )
+        self.assert_status_code(response, 200)
+        self.assert_model_exists(
+            "motion/4", {"lead_motion_id": 3, "block_id": 123, "category_id": 114}
```
FinnStutzenstein, line 319: expected: The block id and category id form the request is taken, so ids 12 and 13 since the request user has motion.can_manage

openslides_backend/action/actions/motion/create.py
```diff
@@ -69,6 +69,15 @@ def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
                 raise ActionException(
                     "You can't give amendment_paragraph_$ in this context"
                 )
+        # if lead_motion and not has perm motion_can_manage
```
FinnStutzenstein, line 72:
```suggestion
        # if lead_motion and not has perm motion.can_manage
```

FinnStutzenstein (changes requested):

openslides_backend/action/actions/motion/create.py
reiterl: I use the perm check again and update the tests. 
If lead_motion_id is set and the user has _not_ the motion.can_manage right, then category_id and block_id are set from the lead motion.

tests/system/action/motion/test_create.py
reiterl: Update the test_create and the test_set_number_mixin  tests.

FinnStutzenstein (approved):

jsangmeister (approved):

PR #689, opened 2021-04-28
Update projector.project

reiterl:
Update logic to move equal fields to history. If stable is true unset
the projections. Update the tests.

So adapt the code to the new use case.


reiterl:
Updated method name and comments.

FinnStutzenstein (approved):

PR #690, opened 2021-04-28
Use a better random lib: secrets.

reiterl:
Change the two `random` uses to `secrets` uses. This is a better lib for generating crypto tokens or passwords.

### DIFF
--- a/openslides_backend/action/util/crypto.py
+++ b/openslides_backend/action/util/crypto.py
@@ -1,0 +1,15 @@
+import secrets
+
+RANDOM_STRING_CHARS = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"
+
+
+def get_random_string(length: int, allowed_chars: str = RANDOM_STRING_CHARS) -> str:
+    """
+    Return a securely generated random string.
+    The bit length of the returned value can be calculated with the formula:
+        log_2(len(allowed_chars)^length)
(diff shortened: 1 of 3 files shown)


jsangmeister (changes requested):
Please move the actual random generation to a new file in the util folder, to a method like `get_random_string`, which is then called in these two files. This way we have the used library in a central place and changes are easier. You can probably just copy the method from https://github.com/django/django/blob/4ab3ef238edfa7c7b7c2c4420c7ae5b722bde121/django/utils/crypto.py#L50

openslides_backend/action/util/crypto.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+def get_random_string(length: int, allowed_chars: str = RANDOM_STRING_CHARS) -> str:
+    """
+    Return a securely generated random string.
+    The bit length of the returned value can be calculated with the formula:
+        log_2(len(allowed_chars)^length)
+    For example, with default `allowed_chars` (26+26+10), this gives:
+      * length: 12, bit length =~ 71 bits
+      * length: 22, bit length =~ 131 bits
```
jsangmeister, line 13: You can remove this part of the docstring, we don't care about the bit length

jsangmeister (changes requested):

openslides_backend/action/util/crypto.py
reiterl: I deleted this comment lines.

jsangmeister (approved):

PR #703, opened 2021-05-03
Install reader pip package

jsangmeister:
Package implemented in https://github.com/OpenSlides/openslides-datastore-service/pull/126. Branch has to be changed before merged.

The test time reduction is not as high as I hoped, but still ok: without these changes, the tests needed 253.50s (0:04:13). With these changes, it is reduced to 202.81s (0:03:22), which is a reduction of almost exactly 20%.


dev/Dockerfile-dev
```diff
@@ -1,6 +1,6 @@
 FROM python:3.8.5-slim-buster
 
-RUN apt-get update && apt-get install --yes make git curl ncat vim bash-completion
+RUN apt-get update && apt-get install --yes make git curl ncat vim bash-completion gcc libpq-dev
```
jsangmeister, line 3: New dependencies needed for psycopg2 (installed by the readerlib)

dev/Dockerfile-dev
```diff
@@ -20,6 +20,7 @@ COPY setup.cfg .
 
 RUN chmod 777 -R .
 RUN chmod 777 /root
+RUN chmod 777 -R /usr/local/lib/
```
jsangmeister, line 23: I added this to be able to install pip packages from inside the docker container. @FinnStutzenstein do know a better way/does this have unwanted side effects?

FinnStutzenstein: This is fine

openslides_backend/action/action.py
```diff
@@ -42,6 +42,7 @@
 from .relations.typing import FieldUpdateElement, ListUpdateElement
 from .util.assert_belongs_to_meeting import assert_belongs_to_meeting
 from .util.typing import ActionData, ActionResultElement, ActionResults
+from ..services.datastore.with_database_context import with_database_context
```
FinnStutzenstein, line 45: Imported but not used?

FinnStutzenstein (commented):
As a quick evaluation: The changes in the DB adapter seems very natural and the fact that more code is removed than added looks very promising.

openslides_backend/action/action.py
jsangmeister: Yes, at the beginning i had though about creating the context here, probably a leftover. It's still WIP after all :D

jsangmeister:
Changes are done and the new time is 212.38s (0:03:32), so even slower than before, which could be because we got more tests in the mean time, but overall it did not pay off to move the transaction handling to the backend.

r-peschke (approved):
Hard to review, but I trust in you:-)

PR #705, opened 2021-05-05
Add permissions for personal notes

reiterl:
Add meeting_ids to models. Generate models is not working at the moment, so just pick only the needed field.
Add some tests for no permissions, the permission allowed test is the default create/update/delete test.

### DIFF
--- a/tests/system/action/personal_note/test_create.py
+++ b/tests/system/action/personal_note/test_create.py
@@ -3,6 +3,7 @@ from tests.system.action.base import BaseActionTestCase
 
 class PersonalNoteCreateActionTest(BaseActionTestCase):
     def test_create(self) -> None:
+        # checks permissions too.
         self.set_models(
             {
                 "meeting/110": {"name": "name_meeting_110"},
@@ -69,4 +70,37 @@ class PersonalNoteCreateActionTest(BaseActionTestCase):
         self.assertIn(
             "(user_id, content_object_id) must be unique.",
             response.json["message"],
+        )
+
+    def test_create_no_permission_user_not_in_meeting(self) -> None:
+        self.set_models(
+            {
+                "meeting/110": {"name": "name_meeting_110"},
+                "motion/23": {"meeting_id": 110},
+                "user/1": {"meeting_ids": []},
+            }
+        )
+        response = self.request(
+            "personal_note.create", {"content_object_id": "motion/23", "star": True}
+        )
+        self.assert_status_code(response, 403)
+        assert "User not associated with meeting." in response.json["message"]
+
+    def test_create_no_permission_anon_user(self) -> None:
+        self.set_models(
+            {
+                "meeting/110": {"name": "name_meeting_110"},
+                "motion/23": {"meeting_id": 110},
+                "user/1": {"meeting_ids": [110]},
+            }
+        )
+        self.set_anonymous(meeting_id=110)
+        response = self.request(
+            "personal_note.create",
+            {"content_object_id": "motion/23", "star": True},
+            anonymous=True,
+        )
+        self.assert_status_code(response, 403)
+        assert (
+            "Anonymous user cannot do personal_note.create." in response.json["message"]
         )
--- a/tests/system/action/personal_note/test_create.py
+++ b/tests/system/action/personal_note/test_create.py
@@ -43,9 +43,9 @@ class PersonalNoteCreateActionTest(BaseActionTestCase):
         )
 
     def test_create_not_unique(self) -> None:
-        self.set_models(self.test_models)
         self.set_models(
             {
+                **self.test_models,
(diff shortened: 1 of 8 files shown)


tests/system/action/personal_note/test_create.py
```diff
@@ -1,14 +1,20 @@
...
+        }
+
     def test_create(self) -> None:
-        self.set_models(
-            {
-                "meeting/110": {"name": "name_meeting_110"},
-                "motion/23": {"meeting_id": 110},
-            }
-        )
+        # checks permissions too.
```
jsangmeister, line 16: What do you mean with this comment?

tests/system/action/personal_note/test_create.py
```diff
@@ -1,14 +1,20 @@
+from typing import Any, Dict
+
 from tests.system.action.base import BaseActionTestCase
 
 
 class PersonalNoteCreateActionTest(BaseActionTestCase):
+    def setUp(self) -> None:
+        super().setUp()
+        self.test_model: Dict[str, Dict[str, Any]] = {
```
jsangmeister, line 9:
```suggestion
        self.test_models: Dict[str, Dict[str, Any]] = {
```

tests/system/action/personal_note/test_create.py
```diff
@@ -43,18 +44,13 @@ def test_create_no_star_and_no_html(self) -> None:
...
-                },
-            }
-        )
+        self.test_model["personal_note/1"] = {
+            "star": True,
+            "note": "blablabla",
+            "user_id": 1,
+            "content_object_id": "motion/23",
+        }
+        self.set_models(self.test_model)
```
jsangmeister, line 53:
```suggestion
        self.set_models({
            **self.test_model,
            "personal_note/1": {
                "star": True,
                "note": "blablabla",
                "user_id": 1,
                "content_object_id": "motion/23",
            }
        })
```
I think this would be more clear, else it seems like you overwrite something in the test models

tests/system/action/personal_note/test_delete.py
```diff
@@ -1,23 +1,29 @@
+from typing import Any, Dict
+
 from tests.system.action.base import BaseActionTestCase
 
 
 class PersonalNoteDeleteActionTest(BaseActionTestCase):
+    def setUp(self) -> None:
+        super().setUp()
+        self.test_model: Dict[str, Dict[str, Any]] = {
```
jsangmeister, line 9:
```suggestion
        self.test_models: Dict[str, Dict[str, Any]] = {
```

tests/system/action/personal_note/test_update.py
```diff
@@ -1,11 +1,25 @@
+from typing import Any, Dict
+
 from tests.system.action.base import BaseActionTestCase
 
 
 class PersonalNoteUpdateActionTest(BaseActionTestCase):
+    def setUp(self) -> None:
+        super().setUp()
+        self.test_model: Dict[str, Dict[str, Any]] = {
```
jsangmeister, line 9:
```suggestion
        self.test_models: Dict[str, Dict[str, Any]] = {
```

reiterl: Renamed into `test_models`

tests/system/action/personal_note/test_delete.py
reiterl: Renamed.

tests/system/action/personal_note/test_create.py
reiterl: Okay, there are more locations of this idiom. Changed them in the tests.

tests/system/action/personal_note/test_create.py
reiterl: Never mind, it was a reminder, that this test also checks permissions.

tests/system/action/personal_note/test_create.py
reiterl: I'll try to use a double `set_models` instead of patching `test_models`. Perhaps that works.

jsangmeister: But `personal_note/1` is not in the test models? If it is, you previous approach is ok, because then you actually overwrite it, whihc is also done in other places.

reiterl:
Moved the user checks from update_instance into check_permissions, update the exception type, and update the tests.

tests/system/action/personal_note/test_create.py
reiterl: Ah, you just mean this special case, where the `personal_note/1` is not in the model.

jsangmeister: Yes, but it's not that important. You can leave it as-is if it works.

jsangmeister (approved):

PR #709, opened 2021-05-10
Update models & fix tests

jsangmeister:
Most changes are deleted files since temporary users where removed. Had to adjust some logic and tests, but we still have the other issues open in case I missed something.


jsangmeister:
I don't know how mush sense a review makes here: There are some changes, but they are buried beneath all the removed files etc. If one of you wants to check it, you can; if not, I would say we can merge it anyway and find any errors along the way ;)

tests/system/action/resource/test_upload.py
```diff
@@ -179,8 +179,8 @@ def test_upload_and_mixed_sep_actions(self) -> None:
         self.assert_model_not_exists("resource/2")
         self.assert_model_not_exists("resource/3")
 
-        # TODO: When the retry-problem (issue440) is solved, the 9 has to be substituted with a 3
-        self.media.upload_resource.assert_called_with(file_content2, 9, used_mimetype)
+        # TODO: When the retry-problem (issue440) is solved, the 7 has to be substituted with a 3
```
FinnStutzenstein, line 182: please create an issue for this.

tests/system/action/speaker/test_update.py
```diff
@@ -24,10 +24,8 @@ def test_update_correct(self) -> None:
                 },
             }
         )
-        response = self.request("speaker.update", {"id": 890, "marked": True})
+        response = self.request("speaker.update", {"id": 890, "speech_state": "pro"})
         self.assert_status_code(response, 200)
-        model = self.get_model("speaker/890")
```
FinnStutzenstein, line 29: Why did you delete these asserts?

tests/system/action/user/test_create.py
```diff
@@ -40,36 +42,27 @@ def test_create_some_more_fields(self) -> None:
                 "username": "test_Xcdfgee",
                 "default_vote_weight": "1.500000",
                 "organisation_management_level": "can_manage_users",
-                "guest_meeting_ids": [110, 111],
-                "committee_as_member_ids": [78],
-                "committee_as_manager_ids": [79],
+                "committee_ids": [78, 79],
```
FinnStutzenstein, line 45: Is the CML already implemented?

FinnStutzenstein: (there is an extra issue, but does the field exists already?)

openslides_backend/action/action_handler.py
```diff
@@ -125,7 +125,7 @@ def execute_write_requests(
         get_write_requests: Callable[..., Tuple[List[WriteRequest], T]],
         *args: Any,
     ) -> T:
-        retried = 0
+        retried = 1
```
FinnStutzenstein, line 128: `retries`

FinnStutzenstein (approved):
Just small remarks

tests/system/action/user/test_create.py
jsangmeister: The general structure was implemented by @r-peschke, but currently the permission check would throw a `NotImplemented`. The field exists because I updated the models in this PR, but no logic for it is implemented yet.

tests/system/action/speaker/test_update.py
jsangmeister: Ah, I removed them because originally I just deleted the `marked` field from the payload, so there was nothing to assert. But the validation complains if the request contains no fields, so I already added the `speech_state` field and did not think about adding these back. Will do though

PR #711, opened 2021-05-11
Add permissions to committee, based on pr_709

reiterl:

PR #713, opened 2021-05-12
Add permissions to the meeting.update action.

reiterl:
Add some tests.
The jitsi fields could not be included in the update payload, because they are read_only.


openslides_backend/action/actions/meeting/update.py
```diff
@@ -144,3 +153,79 @@ def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
...
+        if "present_user_ids" in instance and not has_perm(
+            self.datastore, self.user_id, Permissions.User.CAN_MANAGE, instance["id"]
+        ):
+            raise MissingPermission(Permissions.User.CAN_MANAGE)
+
+        # group C check
+        if (
+            "reference_projector_id" in instance
+            or any(
+                [field for field in instance if field.startswith("default_projector")]
```
jsangmeister, line 177: I don't like this; if we add a new field with the same prefix this can lead to unwanted results. Maybe solve it with a regex or validate again afterwards that only wanted fields were matched

openslides_backend/action/actions/meeting/update.py
```diff
@@ -144,3 +153,79 @@ def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
...
+        # group E check
+        if "organisation_tag_ids" in instance:
+            meeting = self.datastore.get(
+                FullQualifiedId(self.model.collection, instance["id"]), ["committee_id"]
+            )
+            cml_field = f"committee_${meeting['committee_id']}_management_level"
+            user = self.datastore.get(
+                FullQualifiedId(Collection("user"), self.user_id), [cml_field]
+            )
+            has_cml_permission = user.get(cml_field) == "can_manage"
```
jsangmeister, line 222: see https://github.com/OpenSlides/openslides-backend/pull/711#discussion_r633275788

jsangmeister (changes requested):
@FinnStutzenstein how should the jitsi fields be updated? Or is the `read_only` obsolet?

openslides_backend/action/actions/meeting/update.py
reiterl: It is not needed, the field in the instance is always "default_projector_$_id". So a single check is enough.

openslides_backend/action/actions/meeting/update.py
reiterl: Use the enum here.

openslides_backend/action/actions/meeting/update.py
jsangmeister: Ah, right.

FinnStutzenstein:
> is the read_only obsolet?

Yes I think so.

jsangmeister:
@reiterl please create a PR for the models to remove the `read_only` tag. Then, add these fields with the permissions described in https://github.com/OpenSlides/OpenSlides/wiki/meeting.update

jsangmeister (changes requested):
Permissions for jitsi fields are still missing

reiterl:
Add the new permission check for the jitsi fields (group F).

jsangmeister (approved):

PR #715, opened 2021-05-14
Autogenerate username from first/last name if missing. Add some tests.

reiterl:


openslides_backend/action/actions/user/create.py
```diff
@@ -47,8 +50,37 @@ class UserCreate(
...
             instance = self.generate_and_set_password(instance)
         else:
             instance = self.set_password(instance)
         return super().update_instance(instance)
+
+    def generate_username(self, instance: Dict[str, Any]) -> str:
+        count = 0
+
+        while True:
+            new_username = instance.get("first_name", "") + instance.get(
```
FinnStutzenstein, line 71: Please add a space there, if both parts are given.

openslides_backend/action/actions/user/create.py
```diff
@@ -47,8 +50,37 @@ class UserCreate(
...
+            )
+            if count > 0:
+                new_username += str(count)
+
+            result = self.datastore.filter(
+                Collection("user"),
+                FilterOperator("username", "=", new_username),
+                ["id"],
+            )
+            if result and instance["id"] not in result.keys():
```
FinnStutzenstein, line 82: I wonder how `instance["id"] not in result.keys()` should be possible, if the user wasn't saved yet.

FinnStutzenstein (changes requested):
It looks promising, just two remarks.

openslides_backend/action/actions/user/create.py
reiterl: Add a space if both are set. If first_name is "Jon" and last_name "Smith" and it is the third one, the username is "Jon Smith2".

openslides_backend/action/actions/user/create.py
reiterl: Removed that part.

FinnStutzenstein (approved):

openslides_backend/action/actions/user/create.py
```diff
@@ -47,8 +50,42 @@ class UserCreate(
...
+            possible_space = (
+                " " if instance.get("first_name") and instance.get("last_name") else ""
+            )
+            new_username = (
+                instance.get("first_name", "")
+                + possible_space
+                + instance.get("last_name", "")
+            )
+            if count > 0:
+                new_username += str(count)
```
jsangmeister, line 80: Please also add a space before the number to stay consistent with OS3

openslides_backend/action/actions/user/create.py
```diff
@@ -47,8 +50,42 @@ class UserCreate(
...
+
+        while True:
+            possible_space = (
+                " " if instance.get("first_name") and instance.get("last_name") else ""
+            )
+            new_username = (
+                instance.get("first_name", "")
+                + possible_space
+                + instance.get("last_name", "")
+            )
```
jsangmeister, line 78: According to [the wiki](https://github.com/OpenSlides/OpenSlides/wiki/user.create#generate-a-username), `title` should also be included in the generated name. @FinnStutzenstein is that correct? That could potentially lead to usernames like "Prof. 2" etc.

jsangmeister: Also, if that is correct, @reiterl you can just join the instance parts with a space between them, like here: https://github.com/OpenSlides/OpenSlides/blob/bdbb5839cc127393f2ff32687b5470d1630bc53e/server/openslides/users/models.py#L90 (instead of manually inserting the space)

jsangmeister (changes requested):
Please also add a test where username, first and last name are given and the username already exists. It should throw an error (instead of generating a username).

openslides_backend/action/actions/user/create.py
jsangmeister: Title removed from wiki.

openslides_backend/action/actions/user/create.py
reiterl: Done.

reiterl:
Added a test. 
It tests: username already exists and username, first_name and last_name are given in request.

jsangmeister (approved):

FinnStutzenstein (approved):

Issue #679, opened 2021-04-26
User restructure: Delete temporary actions


FinnStutzenstein:

jsangmeister:
While updating the models, I had to delete them all because some needed fields were deleted. However, neither did I check exhaustively whether some mixins etc. can also be deleted nor did I check if some code may be recycled for the new permission checks. Someone should Ctrl+F through the whole repo for all instances of the deleted fields (e.g. `guest_ids`, `member_ids`, etc.) and delete all remaining instances as soon as the models update is finished.

reiterl:
Checked: `member_ids, manager_ids, guest_ids, guest_meeting_ids, temporary_user_ids, marked`.
Found one surplus guest_meeting_ids, and one guest_ids in one comment.

PR #717, opened 2021-05-17
Update user-altering actions with user scope permission check

jsangmeister:


jsangmeister:
Only `user.delete` is updated yet, other actions still missing

openslides_backend/action/actions/user/delete.py
```diff
@@ -9,17 +8,22 @@
 from ...util.default_schema import DefaultSchema
 from ...util.register import register_action
 from .check_temporary_mixin import CheckTemporaryNoForInstanceMixin
+from .user_scope_permission_check_mixin import UserScopePermissionCheckMixin
 
 
 @register_action("user.delete")
-class UserDelete(CheckTemporaryNoForInstanceMixin, DeleteAction):
+class UserDelete(
+    CheckTemporaryNoForInstanceMixin, UserScopePermissionCheckMixin, DeleteAction
```
r-peschke, line 16: CheckTemporaryNoForInstanceMixin: This has been needed for temporary users

openslides_backend/action/actions/user/user_scope_permission_check_mixin.py
```diff
@@ -0,0 +1,94 @@
...
+                    {
+                        OrganisationManagementLevel.CAN_MANAGE_USERS: 1,
+                        CommitteeManagementLevel.CAN_MANAGE: meeting["committee_id"],
+                        Permissions.User.CAN_CHANGE_OWN_PASSWORD: scope_id,
+                    }
+                )
+        else:
+            raise MissingPermission({OrganisationManagementLevel.CAN_MANAGE_USERS: 1})
+
+    def get_user_scope(self, id: int) -> Tuple[UserScope, int]:
```
r-peschke, line 72: I don't understand, what this is for. What's about a user belonging to 2 committees? Committe permissions is not valid, without Organisation Level he has no permission at all?!

tests/unit/test_user_scopes.py
```diff
@@ -0,0 +1,80 @@
...
+        self.set_user_data({})
+        assert self.get_scope() == UserScope.Organisation
+
+    def test_single_meeting(self) -> None:
+        self.set_user_data({"meeting_ids": [1]})
+        assert self.get_scope() == UserScope.Meeting
+
+    def test_multiple_meetings(self) -> None:
+        self.set_user_data({"meeting_ids": [1, 2]})
+        assert self.get_scope() == UserScope.Organisation
```
r-peschke, line 42: Why can't this be Committee Scope? 2 Meetings in 2 Committee?

r-peschke (commented):
The scoped user is the user to act on and **not** the request user

openslides_backend/action/actions/user/user_scope_permission_check_mixin.py
```diff
@@ -0,0 +1,94 @@
...
+                CommitteeManagementLevel.CAN_MANAGE,
+                meeting["committee_id"],
+            ) and not has_perm(
+                self.datastore, self.user_id, Permissions.User.CAN_MANAGE, scope_id
+            ):
+                raise MissingPermission(
+                    {
+                        OrganisationManagementLevel.CAN_MANAGE_USERS: 1,
+                        CommitteeManagementLevel.CAN_MANAGE: meeting["committee_id"],
+                        Permissions.User.CAN_CHANGE_OWN_PASSWORD: scope_id,
```
r-peschke, line 66: May be this should be Permissions.User.CAN_MANAGE

openslides_backend/action/actions/user/user_scope_permission_check_mixin.py
jsangmeister: No, as soon as the user is in 2 meetings, he is either in committee scope (if those belong to the same committee and the user is also part of that) or else in organisation scope. The use case is that if you can create a user for your scope, you should be able to edit it as well. As soon as the user is moved up in his scope by an admin with more permissions, you do not have permissions to edit it anymore. A user has never multiple scopes or, in this case, mutiple scope ids.

openslides_backend/action/actions/user/delete.py
jsangmeister: No, but that work is part of #679 and I did not want to mix it in here

openslides_backend/action/actions/user/user_scope_permission_check_mixin.py
jsangmeister: Yes, you are of course correct, will fix it

openslides_backend/action/actions/user/user_scope_permission_check_mixin.py
r-peschke: > No, as soon as the user is in 2 meetings, he is either in committee scope (if those belong to the same committee and the user is also part of that) or else in organisation scope.
Imagine the request user has permission user.can_manage in meeting 1 and 2. The field group_$_ids belongs to field group C. With this permissions the request user is able to create a new user or update an existing user with groups for both meetings. See group permission [here](https://github.com/OpenSlides/OpenSlides/wiki/user.create#permissions).

I'm not sure about the idea of scope: Do you want to prevent the request user, even with permission for both meetings, to create such a user? Same for committees, one level above?
This would facilitate the permission handling for users a lot!

> The use case is that if you can create a user for your scope, you should be able to edit it as well. As soon as the user is moved up in his scope by an admin with more permissions, you do not have permissions to edit it anymore. A user has never multiple scopes or, in this case, multiple scope ids.

FinnStutzenstein: For creating a user: If you have user.can_manage in two meetings of the same committe, but you cannot manage the committee, you must create two user instances for the same person. Since a user with two meetings would be in committee scope in this case. This is wanted - if you are a manager of multiple meetings in a committee, you should actually be a committee manager.

The more interesting thing is updating a user, especially groups B and C:

Currently, there should be [the basic check](https://github.com/OpenSlides/OpenSlides/wiki/Users#Permissions-for-altering-a-user) (I think it is the `check_permissions_for_scope` function, right?). For altering fields of group B it does not make any sense. If you have `user.can_manage` in a meeting and the updated user is part of the meeting, you should be able to change meeting-specific fields **regardless** of the scope. This is logical since mostly only the meeting adminscare about meeting-specific attributes. (Users with CML/OML can also update these fields, but must follow the scope guidelines)

Group C is even more tricky: The same argument as for group B holds: If there you have `user.can_manage` in a meeting, you can change the groups from users in your meeting. This has a speciality: You cannot set the groups for a user, that didn't had groups before (if you do not have CML/OML for managing users). In other words: You cannot put new users into the meeting. The same goes for setting the groups to `[]`, which must be set to `[<default_group_id>]` if the user only is a meeting admin.

All other fields, especially group A are protected by the basic check.

jsangmeister:
Should be done now. @r-peschke if you want, you can check again, but I added no new changes, only copied the changes from user.delete to the other actions

openslides_backend/action/actions/user/user_scope_permission_check_mixin.py
jsangmeister: | The same goes for setting the groups to [], which must be set to [<default_group_id>] if the user only is a meeting admin.

So the meeting admin should not be able to remove users from the meeting? This is still missing in https://github.com/OpenSlides/OpenSlides/wiki/user.update

FinnStutzenstein (approved):

PR #719, opened 2021-05-18
User Restructure: Remove temp_filter from get_users and update the tests

reiterl:


FinnStutzenstein (approved):
Ok for me. Maybe we need other things later (like scope, ...), but removing temporary users is enough for now

Issue #722, opened 2021-05-20
Add is_demo_user to user.update


jsangmeister:
see https://github.com/OpenSlides/OpenSlides/wiki/user.update

Issue #685, opened 2021-04-26
User restructure: Alter user-committee relation and `committee_$_management_level`


FinnStutzenstein:
See models.yml (or https://github.com/OpenSlides/OpenSlides/pull/6029) for new fields.

jsangmeister:
Add tests for CML to ensure the template field works correctly with committees. E.g. add a structured field with a replacement committee id which doesn't exist and make sure it is rejected.

jsangmeister:
@r-peschke please add this in #723

PR #723, opened 2021-05-21
Issue680/1 user.create/update

r-peschke:
All the logic is implemented, reworked tests

Rebased on master with organiZation. Intent or not: On OrganizationManagementLevel the item CAN_MANAGE_ORGANISATION is still with S

### DIFF
--- a/openslides_backend/action/actions/user/create.py
+++ b/openslides_backend/action/actions/user/create.py
@@ -67,6 +67,15 @@ class UserCreate(
             raise ActionException(
                 "To create a user you need to add him to a permission-group, add him to a committee or give him an Organisation Management Level of at least 'can manage users'."
             )
+
+        if instance.get("committee_$_management_level"):
+            if diff := set(
+                instance.get("committee_$_management_level", {}).keys()
+            ) - set(map(str, instance.get("committee_ids", []))):
+                raise ActionException(
+                    f"You must add the user to the committee(s) '{', '.join(diff)}', because you want to give him committee management level permissions."
+                )
+
         if not instance.get("username"):
             instance["username"] = self.generate_username(instance)
         if not instance.get("default_password"):
--- a/openslides_backend/action/actions/user/create.py
+++ b/openslides_backend/action/actions/user/create.py
@@ -1,9 +1,7 @@
 from typing import Any, Dict
-from ....permissions.management_levels import (
-    OrganisationManagementLevel,
-)
 
 from ....models.models import User
+from ....permissions.management_levels import OrganisationManagementLevel
 from ....shared.exceptions import ActionException
 from ....shared.filters import FilterOperator
 from ....shared.patterns import Collection
@@ -60,8 +58,17 @@ class UserCreate(
             or instance.get("last_name")
         ):
             raise ActionException("Need username or first_name or last_name")
-        if OrganisationManagementLevel(instance.get("organisation_management_level", "no_right")) < OrganisationManagementLevel.CAN_MANAGE_USERS and len(instance.get("group_$_ids", [])) == 0 and len(instance.get("committee_ids", [])) == 0:
-            raise ActionException("To create a user you need to add him to a permission-group, add him to a committee or give him an Organisation Management Level of at least 'can manage users'.")
+        if (
+            OrganisationManagementLevel(
+                instance.get("organisation_management_level", "")
+            )
+            < OrganisationManagementLevel.CAN_MANAGE_USERS
+            and len(instance.get("group_$_ids", [])) == 0
+            and len(instance.get("committee_ids", [])) == 0
+        ):
+            raise ActionException(
+                "To create a user you need to add him to a permission-group, add him to a committee or give him an Organisation Management Level of at least 'can manage users'."
+            )
         if not instance.get("username"):
             instance["username"] = self.generate_username(instance)
         if not instance.get("default_password"):
--- a/openslides_backend/action/actions/user/create.py
+++ b/openslides_backend/action/actions/user/create.py
@@ -1,4 +1,7 @@
 from typing import Any, Dict
+from ....permissions.management_levels import (
+    OrganisationManagementLevel,
+)
 
 from ....models.models import User
 from ....shared.exceptions import ActionException
@@ -38,6 +41,7 @@ class UserCreate(
             "organization_management_level",
             "is_present_in_meeting_ids",
             "committee_ids",
+            "committee_$_management_level",
             "group_$_ids",
             "vote_delegations_$_from_ids",
             "vote_delegated_$_to_id",
@@ -56,6 +60,8 @@ class UserCreate(
             or instance.get("last_name")
         ):
             raise ActionException("Need username or first_name or last_name")
+        if OrganisationManagementLevel(instance.get("organisation_management_level", "no_right")) < OrganisationManagementLevel.CAN_MANAGE_USERS and len(instance.get("group_$_ids", [])) == 0 and len(instance.get("committee_ids", [])) == 0:
+            raise ActionException("To create a user you need to add him to a permission-group, add him to a committee or give him an Organisation Management Level of at least 'can manage users'.")
         if not instance.get("username"):
             instance["username"] = self.generate_username(instance)
         if not instance.get("default_password"):
--- a/openslides_backend/action/actions/user/create.py
+++ b/openslides_backend/action/actions/user/create.py
@@ -59,9 +59,7 @@ class UserCreate(
         ):
             raise ActionException("Need username or first_name or last_name")
         if (
-            OrganisationManagementLevel(
-                instance.get("organisation_management_level", "")
-            )
+            OrganisationManagementLevel(instance.get("organisation_management_level"))
             < OrganisationManagementLevel.CAN_MANAGE_USERS
             and len(instance.get("group_$_ids", [])) == 0
             and len(instance.get("committee_ids", [])) == 0
--- a/openslides_backend/action/actions/user/create.py
+++ b/openslides_backend/action/actions/user/create.py
@@ -1,7 +1,7 @@
 from typing import Any, Dict
 
 from ....models.models import User
-from ....permissions.management_levels import OrganisationManagementLevel
+from ....permissions.management_levels import OrganizationManagementLevel
 from ....shared.exceptions import ActionException
 from ....shared.filters import FilterOperator
 from ....shared.patterns import Collection
@@ -59,13 +59,13 @@ class UserCreate(
         ):
             raise ActionException("Need username or first_name or last_name")
         if (
-            OrganisationManagementLevel(instance.get("organisation_management_level"))
-            < OrganisationManagementLevel.CAN_MANAGE_USERS
+            OrganizationManagementLevel(instance.get("organization_management_level"))
+            < OrganizationManagementLevel.CAN_MANAGE_USERS
             and len(instance.get("group_$_ids", [])) == 0
             and len(instance.get("committee_ids", [])) == 0
         ):
             raise ActionException(
-                "To create a user you need to add him to a permission-group, add him to a committee or give him an Organisation Management Level of at least 'can manage users'."
+                "To create a user you need to add him to a permission-group, add him to a committee or give him an Organization Management Level of at least 'can manage users'."
             )
 
         if instance.get("committee_$_management_level"):
--- a/openslides_backend/action/actions/user/create_update_permissions_mixin.py
+++ b/openslides_backend/action/actions/user/create_update_permissions_mixin.py
@@ -1,5 +1,5 @@
 from collections import defaultdict
-from typing import Any, Dict, List, Set, Tuple
+from typing import Any, Dict, List, Optional, Set, Tuple
 
 from ....permissions.management_levels import (
     CommitteeManagementLevel,
@@ -23,61 +23,70 @@ class PermissionVarStore:
                 "organization_management_level",
                 "committee_$_management_level",
                 "group_$_ids",
+                "committee_ids",
             ],
         )
-        self.user_oml = OrganizationManagementLevel(  # type: ignore
-            self.user.get("organization_management_level", "no_right")
+        self.user_oml = OrganizationManagementLevel(
+            self.user.get("organization_management_level")
         )
-        self.called_get_user_committees_and_meetings = False
-        self.called_get_user_meetings_with_user_can_manage = False
+        self._user_committees: Optional[Set[int]] = None
+        self._user_meetings: Optional[Set[int]] = None
 
-    def user_committees(self, committee_ids: List[str] = []) -> Set[int]:
+    @property
+    def user_committees(self) -> Set[int]:
         """ Set of committee-ids where the request user has manage rights """
-        if not self.called_get_user_committees_and_meetings:
+        if self._user_committees is None:
             (
                 self._user_committees,
                 self._user_committees_meetings,
-            ) = self._get_user_committees_and_meetings(committee_ids)
-            self.called_get_user_committees_and_meetings = True
+            ) = self._get_user_committees_and_meetings(
+                self.user.get("committee_$_management_level", [])
+            )
         return self._user_committees
 
-    def user_committees_meetings(self, committee_ids: List[str] = []) -> Set[int]:
+    @property
+    def user_committees_meetings(self) -> Set[int]:
         """ Set of meetings where the request user has manage rights from committee """
-        if not self.called_get_user_committees_and_meetings:
+        if self._user_committees is None:
             (
                 self._user_committees,
                 self._user_committees_meetings,
-            ) = self._get_user_committees_and_meetings(committee_ids)
-            self.called_get_user_committees_and_meetings = True
+            ) = self._get_user_committees_and_meetings(
+                self.user.get("committee_$_management_level", [])
+            )
         return self._user_committees_meetings
 
-    def user_meetings(self, meeting_ids: List[str] = []) -> Set[int]:
+    @property
+    def user_meetings(self) -> Set[int]:
         """ Set of meetings where the request user has user.can_manage permissions """
-        if not self.called_get_user_meetings_with_user_can_manage:
+        if self._user_meetings is None:
             self._user_meetings = self._get_user_meetings_with_user_can_manage(
-                meeting_ids
+                self.user.get("group_$_ids", [])
             )
-            self.called_get_user_meetings_with_user_can_manage = True
         return self._user_meetings
 
     def _get_user_committees_and_meetings(
-        self, committee_ids: List[str]
+        self, committee_manage_ids: List[str]
     ) -> Tuple[Set[int], Set[int]]:
         """
-        Returns a set of committees, where the request user has minimum
-        CommitteeManagementLevel.CAN_MANAGE and a set of meetings belonging to those committees
+        Returns a set of committees and a set of meetings
+        belonging to those committees, where the request user has minimum
+        CommitteeManagementLevel.CAN_MANAGE and is member of committee_id,
         """
-        user_committees = set()
+        user_committees = set(map(int, committee_manage_ids))
         user_meetings = set()
-        if committee_ids:
+        user_committees = user_committees & set(self.user.get("committee_ids", set()))
+        if user_committees:
+            committees = tuple(user_committees)
+            user_committees = set()
             user = self.datastore.get(
                 FullQualifiedId(Collection("user"), self.user_id),
                 [
                     f"committee_${committee_id}_management_level"
-                    for committee_id in committee_ids
+                    for committee_id in committees
                 ],
             )
-            for committee_id in committee_ids:
+            for committee_id in committees:
                 value = user.get(f"committee_${committee_id}_management_level")
                 if (
                     CommitteeManagementLevel(value)
@@ -85,7 +94,7 @@ class PermissionVarStore:
                 ):
                     user_committees.add(int(committee_id))
 
-            committees = (
+            committees_d = (
                 self.datastore.get_many(
                     [
                         GetManyRequest(
@@ -98,7 +107,7 @@ class PermissionVarStore:
                 .get(Collection("committee"), {})
                 .values()
             )
-            for committee in committees:
+            for committee in committees_d:
                 user_meetings.update(committee.get("meeting_ids", []))
 
         return user_committees, user_meetings
@@ -134,9 +143,9 @@ class PermissionVarStore:
             )
 
             for group in groups:
-                if "user.can_manage" in group.get("permissions", []) or group.get(
-                    "admin_group_for_meeting_id"
-                ):
+                if Permissions.User.CAN_MANAGE in group.get(
+                    "permissions", []
+                ) or group.get("admin_group_for_meeting_id"):
                     user_meetings.add(group.get("meeting_id"))
 
         return user_meetings
@@ -188,6 +197,7 @@ class CreateUpdatePermissionsMixin(UserScopePermissionCheckMixin):
         self._check_OML_in_instance(permstore, instance)
         actual_group_fields = self._get_actual_grouping_from_instance(instance)
 
+        # Ordered by supposed velocity advantages. Changing order only can effect the sequence of detected errors for tests
         self.check_group_E(permstore, actual_group_fields["E"])
         self.check_group_D(permstore, actual_group_fields["D"], instance)
         self.check_group_C(permstore, actual_group_fields["C"], instance)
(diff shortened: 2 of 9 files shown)


jsangmeister (changes requested):
Not-complete review. It seems very (maybe overly?) complex, so I did not check through everything, but if it works, it's fine by me.

jsangmeister:
Fixed one remark in the test and cleaned up code. The other remark can be fixed when #724 is merged.

jsangmeister:
Also you are correct, some tests are missing, especially for groups D-F. Special tests for the new CML can either be also done here or (probably better) as soon as this is merged with #685.

jsangmeister:
> Fixed one remark in the test and cleaned up code. The other remark can be fixed when #724 is merged.

Fixed that one, too. My other remarks are still open as well as more tests.

openslides_backend/action/actions/user/create_update_permissions_mixin.py
```diff
@@ -1,151 +1,422 @@
...
+        if not self.called_get_user_committees_and_meetings:
+            (
+                self._user_committees,
+                self._user_committees_meetings,
+            ) = self._get_user_committees_and_meetings(committee_ids)
+            self.called_get_user_committees_and_meetings = True
+        return self._user_committees_meetings
+
+    def user_meetings(self, meeting_ids: List[str] = []) -> Set[int]:
+        """ Set of meetings where the request user has user.can_message permissions """
```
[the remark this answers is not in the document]
r-peschke: done

openslides_backend/action/actions/user/create_update_permissions_mixin.py
```diff
@@ -1,151 +1,422 @@
...
+        if committee_ids:
+            user = self.datastore.get(
+                FullQualifiedId(Collection("user"), self.user_id),
+                [
+                    f"committee_${committee_id}_management_level"
+                    for committee_id in committee_ids
+                ],
+            )
+            for key, value in user.items():
+                r = re.search("\\$[0-9]+", key)
```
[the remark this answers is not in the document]
r-peschke: Done

openslides_backend/action/actions/user/create_update_permissions_mixin.py
```diff
@@ -1,151 +1,422 @@
...
+            or permstore.user_oml >= OrganisationManagementLevel.CAN_MANAGE_USERS
+        ):
+            return
+
+        if (uid := instance.get("id", -1) == -1) :
+            self.datastore.update_additional_models(
+                FullQualifiedId(Collection("user"), uid), instance
+            )
+
+        scope, scope_id = self.get_user_scope(uid)
```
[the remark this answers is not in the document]
r-peschke: Done

openslides_backend/action/actions/user/create.py
```diff
@@ -60,8 +58,17 @@ def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
             or instance.get("last_name")
         ):
             raise ActionException("Need username or first_name or last_name")
-        if OrganisationManagementLevel(instance.get("organisation_management_level", "no_right")) < OrganisationManagementLevel.CAN_MANAGE_USERS and len(instance.get("group_$_ids", [])) == 0 and len(instance.get("committee_ids", [])) == 0:
-            raise ActionException("To create a user you need to add him to a permission-group, add him to a committee or give him an Organisation Management Level of at least 'can manage users'.")
+        if (
+            OrganisationManagementLevel(
+                instance.get("organisation_management_level", "")
```
jsangmeister, line 63: You should be able to remove the default value, OML can handler `None`. If mypy complains, then you could try to change the definition here: https://github.com/OpenSlides/openslides-backend/blob/26e286c3216ae4cb6e4982373c1db56d413fcfb4/openslides_backend/permissions/management_levels.py#L9 to `value: Any` - I favor shorter code over mypy appeasement ;)

openslides_backend/action/actions/user/create_update_permissions_mixin.py
```diff
@@ -396,16 +397,18 @@ def _get_all_committees_from_instance(self, instance: Dict[str, Any]) -> Set[int
...
-                FullQualifiedId(Collection("user"), instance_user_id), ["committee_ids"]
+                FullQualifiedId(Collection("user"), instance_user_id),
+                ["committee_ids", "committee_$_management_level"],
             )
             committees.update(user.get("committee_ids", []))
-
-        if committee_str := user.get("committee_$_management_level", []):
-            committees.update(map(int, committee_str))
+            if committee_str := user.get("committee_$_management_level", []):
+                committees.update(map(int, committee_str))
```
jsangmeister, line 411: I think this is wrong: You also return all existent committee management levels here, right? Suppose user A has some rights in committee C1. I update him so that he has also rights in C2, where I have rights, but I don't have any in C1. This would currently fail, since C1 is returned here, but it should succeed, because I don't touch C1.

If this is correct, please add an explicit test for it.

openslides_backend/action/actions/user/update.py
```diff
@@ -42,5 +42,6 @@ class UserUpdate(
             "vote_delegated_$_to_id",
             "vote_delegations_$_from_ids",
             "group_$_ids",
+            "is_demo_user",
```
jsangmeister, line 45: I see you are also doing https://github.com/OpenSlides/openslides-backend/issues/722 in this PR ;) please assign yourself there and link this issue, I assume you also included it in the permission check then

openslides_backend/action/actions/user/user_scope_permission_check_mixin.py
```diff
@@ -69,15 +69,25 @@ def check_permissions_for_scope(self, instance: Dict[str, Any]) -> None:
         else:
             raise MissingPermission({OrganisationManagementLevel.CAN_MANAGE_USERS: 1})
 
-    def get_user_scope(self, id: int) -> Tuple[UserScope, int]:
+    def get_user_scope(
+        self, id: int, instance: Optional[Dict[str, Any]] = None
```
jsangmeister, line 73: If user is given, `id` is always `None`. It would be cleaner to make both arguments optional, so that in this case it is enough to call `get_user_scope(user=instance)`.

tests/system/action/user/test_update.py
```diff
@@ -76,10 +82,24 @@ def test_update_template_fields(self) -> None:
...
+                    2: CommitteeManagementLevel.CAN_MANAGE,
+                },
+                "committee_ids": [2],
             },
         )
         self.assert_status_code(response, 200)
         user = self.get_model("user/223")
+        assert (
+            CommitteeManagementLevel(user.get("committee_$1_management_level"))
+            == CommitteeManagementLevel.NO_RIGHT
```
jsangmeister, line 96: Please assert explicitly that `user.get("committee_$1_management_level") is None` since the constructor will convert everything not matching any level to `NO_RIGHT`

jsangmeister (changes requested):

openslides_backend/action/actions/user/create.py
jsangmeister: but I just noticed that below for CML mypy doesn't complain, so maybe it just works without the change

openslides_backend/action/actions/user/update.py
jsangmeister: saw that you did, I will go ahead and assign you then.

jsangmeister:
Also you can remove the first 2 commits from this PR, they are rebase relics, I think. A `git rebase --onto master HEAD~3` (replacing 3 with the number of commits you want to keep) should fix this

openslides_backend/action/actions/user/create.py
r-peschke: You are  right, works without further changes :-)

openslides_backend/action/actions/user/create_update_permissions_mixin.py
r-peschke: You are right, I removed getting these committee_ids. Nevertheless I added a test for it

openslides_backend/action/actions/user/create_update_permissions_mixin.py
```diff
@@ -1,150 +1,424 @@
...
+        if uid := instance.get("id"):
+            self.datastore.update_additional_models(
+                FullQualifiedId(Collection("user"), uid), instance
+            )
+
+        scope, scope_id = self.get_user_scope(uid, None if uid else instance)
+        if scope == UserScope.Organisation:
+            raise MissingPermission({OrganisationManagementLevel.CAN_MANAGE_USERS: 1})
+        elif scope == UserScope.Committee:
+            if scope_id not in permstore.user_committees(
```
FinnStutzenstein, line 217: This is also affected by the comment above: As far as I know does the CML can be set, but the user may not be part of the committee anymore.

openslides_backend/action/actions/user/user_scope_permission_check_mixin.py
r-peschke: Done

tests/system/action/user/test_update.py
r-peschke: Done

tests/system/action/user/test_create.py
```diff
@@ -290,68 +346,236 @@ def test_create_permission_superadmin(self) -> None:
...
+                "committee_ids": [60],
+                "group_$_ids": {"4": [4]},
+            },
+        )
+        self.assert_status_code(response, 403)
+        self.assertIn(
+            "You are not allowed to perform action user.create. Missing permission: OrganisationManagementLevel can_manage_users in organisation 1",
+            response.json["message"],
+        )
+
```
FinnStutzenstein, line 422: For group A I would like to have 2 small success tests (not many fields) for a create with the right CML and meeting perms.

tests/system/action/user/test_create.py
```diff
@@ -290,68 +346,236 @@ def test_create_permission_superadmin(self) -> None:
...
         self.permission_setup()
         self.update_model(
             f"user/{self.user_id}",
             {
-                "committee_ids": [60],
-                "organisation_management_level": OrganisationManagementLevel.CAN_MANAGE_USERS,
-                "group_$_ids": ["1", "4"],
-                "group_$1_ids": [2],  # admin group of meeting/1
-                "group_$4_ids": [4],  # default group of meeting/4
+                "committee_$60_management_level": CommitteeManagementLevel.CAN_MANAGE,
```
FinnStutzenstein, line 518: Here again: The user is not in the `committee/user_ids`, but has a CML set. IMO this must be catched.

tests/system/action/user/test_create.py
```diff
@@ -290,68 +346,236 @@ def test_create_permission_superadmin(self) -> None:
...
+                "group_$_ids": {1: [1]},
+            },
+        )
+        self.assert_status_code(response, 200)
+        self.assert_model_exists(
+            "user/3",
+            {"group_$_ids": ["1"], "group_$1_ids": [1], "username": "usersname"},
+        )
+
+    def test_create_permission_group_C_committee_manager(self) -> None:
```
FinnStutzenstein, line 512: please write a similar success test trying to add the user to the committee instead of the meeting.

tests/system/action/user/test_create.py
```diff
@@ -290,68 +346,236 @@ def test_create_permission_superadmin(self) -> None:
             },
         )
```
FinnStutzenstein, line 348: Please add a test with `"organisation_management_level": null,` in the payload and see, that it works (maybe in the group D section)

tests/system/action/user/test_create.py
```diff
@@ -361,123 +585,95 @@ def test_create_permission_manage_user(self) -> None:
...
-                "vote_weight_$": {"1": "12.002345"},
-                "about_me_$": {"1": "about me 1"},
-                "comment_$": {"1": "comment zu meeting/1"},
-                "vote_delegated_$_to_id": {"1": self.user_id},
-                "vote_delegations_$_from_ids": {"4": [3, 4]},
-                "group_$_ids": {"1": [2, 3], "4": [5]},
+                "username": "usersname",
+                "committee_$_management_level": {
+                    "60": CommitteeManagementLevel.CAN_MANAGE,
+                    "63": CommitteeManagementLevel.CAN_MANAGE,
```
FinnStutzenstein, line 630: Shouldn't it be forbidden to set invalid replacements here?

tests/system/action/user/test_create.py
```diff
@@ -361,123 +585,95 @@ def test_create_permission_manage_user(self) -> None:
...
-        self.set_user_groups(self.user_id, [3, 6])
-        self.set_group_permissions(3, [Permissions.User.CAN_MANAGE])
 
         response = self.request(
             "user.create",
             {
-                "username": "username",
-                "number_$": {"1": "number1", "4": "number1 in 4"},
+                "username": "usersname",
+                "committee_$_management_level": {
```
FinnStutzenstein, line 663: Same here. I would expect, that `committee_ids` must also be present.

tests/system/action/user/test_update.py
```diff
@@ -75,10 +82,24 @@ def test_update_template_fields(self) -> None:
                 "structure_level_$": {1: "level_1", 2: "level_2"},
                 "about_me_$": {1: "<p>about</p><iframe></iframe>"},
                 "vote_weight_$": {1: "1.000000", 2: "2.333333"},
+                "committee_$_management_level": {
+                    1: None,
```
FinnStutzenstein, line 86: Just a clairification: When this line were missing, the user would have CML for committees 1 and 2, but `user/committee_ids` would be `[2]` after the update?

tests/system/action/user/test_update.py
```diff
@@ -302,22 +273,46 @@ def test_update_permission_manage_user(self) -> None:
...
                 "default_structure_level": "new default_structure_level",
                 "default_vote_weight": "1.234000",
-                "organisation_management_level": OrganisationManagementLevel.CAN_MANAGE_USERS,
-                "committee_ids": [60, 63],
             },
         )
 
-    def test_update_permission_user_can_manage(self) -> None:
-        """ May update group B and C fields """
+    def test_update_permission_group_A_no_permission(self) -> None:
```
FinnStutzenstein, line 279: please add a short version of this for a user in committee scope and meeting scope.

tests/system/action/user/test_update.py
```diff
@@ -400,121 +392,350 @@ def test_update_permission_OML_not_high_enough(self) -> None:
...
+        )
+
+    def test_update_permission_group_C_committee_manager(self) -> None:
+        """ May update group C group_$_ids by committee permission """
+        self.permission_setup()
+        self.update_model(
+            f"user/{self.user_id}",
+            {
+                "committee_$60_management_level": CommitteeManagementLevel.CAN_MANAGE,
+                "committee_$_management_level": ["60"],
```
FinnStutzenstein, line 411: Same here: I would expect it to fail since the request user is not part of the committee

tests/system/action/user/test_update.py
```diff
@@ -400,121 +392,350 @@ def test_update_permission_OML_not_high_enough(self) -> None:
...
+            {
+                "id": 111,
+                "group_$_ids": {1: [2], 4: [5]},
+            },
+        )
+        self.assert_status_code(response, 403)
+        self.assertIn(
+            "You need CommitteeManagementLevel.can_manage permission for committee 60, because you try to add or remove meetings in Committee-scope!",
+            response.json["message"],
+        )
```
FinnStutzenstein, line 561: Please write a simple test, that setting `group_$<some id>_ids` to `[]` correctly removed the `<some id>` from `group_$_ids` and the field `group_$<some id>_ids` is set to none.

r-peschke:
> Also you can remove the first 2 commits from this PR, they are rebase relics, I think. A `git rebase --onto master HEAD~3` (replacing 3 with the number of commits you want to keep) should fix this

Done

FinnStutzenstein (changes requested):
The code looks fine - it a bit logic-heavy, but so are the requirements. I would like to have a few more (mostly success) tests. I wrote comments at the places. Mainly tests are needed to check, that the permission checks also works for other scopes. E.g. one time, you only tested the OML, but group A can also be updated in lower scopes with the request user not having any OML.

Another large thing (I wrote many comments about it) is the check, that the user actually is part of a committee. Currently, you only check the `committee_$_management_level`, but not the `user/committee_ids` (or the reverse relation). This leads to inconsistent situation where a user can do something within a committee while not even being part of it. This needs to be adjusted; I've wrote some thoughts into the comments while reviewing.

openslides_backend/action/actions/user/create.py
jsangmeister: Exactly, I wanted to write this as a suggestion, but somehow this was deactived this morning...

openslides_backend/action/actions/user/create.py
```diff
@@ -38,6 +39,7 @@ class UserCreate(
             "organisation_management_level",
             "is_present_in_meeting_ids",
             "committee_ids",
+            "committee_$_management_level",
```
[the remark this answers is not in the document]
jsangmeister: Yes, I think the latter possibility is easier.

openslides_backend/action/actions/user/create.py
jsangmeister: Ah, it wasn't pushed yet ;)

jsangmeister:
Another thing: In https://github.com/r-peschke/openslides-backend/blob/ee903bd1c4a3bb6a28777ae1a4a5bbb633b9148b/openslides_backend/action/relations/relation_manager.py#L153 an error is thrown if the `replacement_collection` is not `meeting`. I wonder why this doesn't happen here, can you look into this?

openslides_backend/action/actions/user/create.py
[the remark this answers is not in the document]
r-peschke: What's the problem about the implementation? If the user has CML permission for C1, he can manage a meeting within C1, whether the user is member of C1 or not! May be he's the manager of C1, without being member of C1 as Ortsverband, but member of Bundesverband C2.
If you don't agree:
Solution 2 is easier, but hard to understand for the user! Therefore I would prefer number 1, setting automatically the membership, when he gots the CML permission. To remove it has to be done on both manually.

openslides_backend/action/actions/user/create_update_permissions_mixin.py
```diff
@@ -1,150 +1,424 @@
...
+        self.user = self.datastore.get(
+            FullQualifiedId(Collection("user"), self.user_id),
+            [
+                "organisation_management_level",
+                "committee_$_management_level",
+                "group_$_ids",
+            ],
+        )
+        self.user_oml = OrganisationManagementLevel(  # type: ignore
+            self.user.get("organisation_management_level", "no_right")
```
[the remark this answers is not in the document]
r-peschke: Done

openslides_backend/action/actions/user/create.py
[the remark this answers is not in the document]
jsangmeister: Ah ok, that's how you interpreted it, but not how it was intended: Analogous to a meeting, where you can only have permissions (by being part of a group) if you are part of the meeting, one can only have a CML in a committee if he is part of that committee. I guess this wasn't clear enough from the wiki...

openslides_backend/action/actions/user/create_update_permissions_mixin.py
```diff
@@ -1,150 +1,424 @@
...
+        )
+        self.called_get_user_committees_and_meetings = False
```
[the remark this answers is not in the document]
r-peschke: Done

openslides_backend/action/actions/user/create.py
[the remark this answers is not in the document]
FinnStutzenstein: Actually I didn't thought about that, but as I said: it is inconsistent if you can manage a committee where you are not a member. The actual user of the UI won't ever see this since he cannot access meeting where he is not assigned, so this case does not happen in the UI, but it leaves a security issue open: When you remove a user from a committee, he currently still can manage it.

openslides_backend/action/actions/user/create_update_permissions_mixin.py
```diff
@@ -1,150 +1,424 @@
...
+                            ["meeting_id", "permissions", "admin_group_for_meeting_id"],
+                        )
+                    ]
+                )
+                .get(Collection("group"), {})
+                .values()
+            )
+
+            for group in groups:
+                if "user.can_manage" in group.get("permissions", []) or group.get(
```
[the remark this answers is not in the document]
r-peschke: Done

openslides_backend/action/actions/user/create_update_permissions_mixin.py
```diff
@@ -1,150 +1,424 @@
...
-        )
-        if user_oml == OrganisationManagementLevel.SUPERADMIN:
+        permstore = PermissionVarStore(self.datastore, self.user_id)
+        if permstore.user_oml == OrganisationManagementLevel.SUPERADMIN:
+            return
+
+        self._check_OML_in_instance(permstore, instance)
+        actual_group_fields = self._get_actual_grouping_from_instance(instance)
+
+        self.check_group_E(permstore, actual_group_fields["E"])
```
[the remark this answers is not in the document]
r-peschke: The order is made by supposed velocity advantages. It's now documented in source.
Changing the order only effects the sequence, permissions errors are detected. Therefore some tests checking the error message may fail.

tests/system/action/user/test_create.py
r-peschke: done

openslides_backend/action/actions/user/create.py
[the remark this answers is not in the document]
r-peschke: Today I talked with Emanuel and we will stay with committee manage permission implying being part of a committee. 
I think then we should do it analogous to meeting: remove the field committee_ids (or use it as calculated field like meeting_ids) and use the committee_$_management_level analogue to group_$_ids to prove being part of committee. For not beeing manager of a committee we would use something like "can_see". 
IMO this would be consistent, simple to check (see PR with checking necessary dependencies) and easier to understand.  
Perhaps we could talk about on Monday.

openslides_backend/action/actions/user/create_update_permissions_mixin.py
```diff
@@ -1,150 +1,424 @@
...
+                "group_$_ids",
+            ],
+        )
+        self.user_oml = OrganisationManagementLevel(  # type: ignore
+            self.user.get("organisation_management_level", "no_right")
+        )
+        self.called_get_user_committees_and_meetings = False
+        self.called_get_user_meetings_with_user_can_manage = False
+
+    def user_committees(self, committee_ids: List[str] = []) -> Set[int]:
```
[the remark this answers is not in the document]
r-peschke: All the 3 of them are now properties. All params they need, are values from the request users instance.

openslides_backend/action/actions/user/create_update_permissions_mixin.py
r-peschke: user_committee is now a property with all committees of the request user, where the user has manage permission, which now includes beeing part of the committee.

tests/system/action/user/test_create.py
r-peschke: done in test_create.py: test_create_permission_group_D_permission_with_OML

tests/system/action/user/test_create.py
r-peschke: see test_create_permission_group_D_permission_with_OML

tests/system/action/user/test_create.py
r-peschke: Is catched now explicitly in update.py respectively create.py. But IMO would be better to remove committee_ids field

tests/system/action/user/test_create.py
r-peschke: They are valid. self.create_meeting creates a committee with meeting + offset 59.

tests/system/action/user/test_create.py
r-peschke: done

tests/system/action/user/test_update.py
r-peschke: yes

tests/system/action/user/test_update.py
r-peschke: done

tests/system/action/user/test_update.py
r-peschke: now with committee_ids

tests/system/action/user/test_update.py
r-peschke: done, see test_update_remove_group_from_user

FinnStutzenstein:
> Intent or not: On OrganizationManagementLevel the item CAN_MANAGE_ORGANISATION is still with S

Not intended. Fixed with https://github.com/OpenSlides/openslides-backend/pull/746

FinnStutzenstein (approved):

PR #724, opened 2021-05-26
Use a CalculatedFieldHandler to calculate user.meeting_ids. 

reiterl:
Update tests.

### DIFF
--- a/openslides_backend/action/relations/user_meeting_ids_handler.py
+++ b/openslides_backend/action/relations/user_meeting_ids_handler.py
@@ -1,14 +1,9 @@
 from typing import Any, Dict
 
-from openslides_backend.services.datastore.deleted_models_behaviour import (
-    DeletedModelsBehaviour,
-    InstanceAdditionalBehaviour,
-)
-
 from ...models.fields import Field
-from ...shared.patterns import Collection, FullQualifiedField, FullQualifiedId
+from ...shared.patterns import Collection, FullQualifiedField
 from .calculated_field_handler import CalculatedFieldHandler
-from .typing import ListUpdateElement, RelationUpdates
+from .typing import FieldUpdateElement, RelationUpdates
 
 
 class UserMeetingIdsHandler(CalculatedFieldHandler):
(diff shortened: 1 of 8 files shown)


openslides_backend/action/relations/user_meeting_ids_handler.py
```diff
@@ -0,0 +1,45 @@
...
+
+class UserMeetingIdsHandler(CalculatedFieldHandler):
+    """
+    CalculatedFieldHandler to fill the user.meeting_ids.
+    """
+
+    def process_field(
+        self, field: Field, field_name: str, instance: Dict[str, Any], action: str
+    ) -> RelationUpdates:
+        if field_name != "group_$_ids":
```
jsangmeister, line 22: Is this needed? When does it get called with other field names?

openslides_backend/action/relations/user_meeting_ids_handler.py
```diff
@@ -0,0 +1,45 @@
...
+            return {}
+        ids_set = set([int(x) for x in instance.get(field_name, []) or []])
```
jsangmeister, line 24:
```suggestion
        ids_set = set(int(x) for x in instance.get(field_name, []))
```

openslides_backend/action/relations/user_meeting_ids_handler.py
```diff
@@ -0,0 +1,45 @@
...
+        fqid = FullQualifiedId(field.own_collection, instance["id"])
+        db_instance = self.datastore.fetch_model(
+            fqid,
+            [field_name],
+            get_deleted_models=DeletedModelsBehaviour.NO_DELETED,
+            db_additional_relevance=InstanceAdditionalBehaviour.ONLY_DBINST,
+            exception=False,
+        )
+        db_ids_set = set([int(x) for x in db_instance.get(field_name, []) or []])
```
jsangmeister, line 33:
```suggestion
        db_ids_set = set(int(x) for x in db_instance.get(field_name, []))
```

jsangmeister (changes requested):
I did not intent that partial list updates are used here, but if it works that's ok. But tests are missing for the handler -  write a few tests where you change the group ids of a user and assert that the meeting ids are also changed. And/Or extend existing tests where the groups are changed and assert that the meeting ids are changed accordingly there.

openslides_backend/action/relations/user_meeting_ids_handler.py
reiterl: It will be called with `group_$<meeting_id>_ids`, too. I don't want to use these calls.

openslides_backend/action/relations/user_meeting_ids_handler.py
reiterl: Done

openslides_backend/action/relations/user_meeting_ids_handler.py
reiterl: Done.

openslides_backend/action/relations/user_meeting_ids_handler.py
jsangmeister: Ah ok, thats fine then

PR #725, opened 2021-05-26
Check for deleted model fields in the code.

reiterl:
Update one comment.
Remove a surplus guest_meeting_ids.


reiterl:
I found some uses of `member_ids, manager_ids`, but only in user.create and user.update code. I haven't fixed that.

jsangmeister:
Please do a serach for "temporary", there are more things I believe that can be removed, like the `CheckForTemporaryYes/NoMixins`. I'm assuming all the code there is obsolet now and can just be deleted, right @r-peschke?

openslides_backend/permissions/permission_helper.py
```diff
@@ -17,7 +17,6 @@ def has_perm(
             FullQualifiedId(Collection("user"), user_id),
             [
                 f"group_${meeting_id}_ids",
-                "guest_meeting_ids",
```
FinnStutzenstein, line 20: Is this function adapted to the new user modelling?

jsangmeister:
Also, the user tests can be adjusted - some tests and helper methods became obsolet

r-peschke:
> Please do a serach for "temporary", there are more things I believe that can be removed, like the `CheckForTemporaryYes/NoMixins`. I'm assuming all the code there is obsolet now and can just be deleted, right @r-peschke?

Yes, all these mixins can be deleted, sometimes the whole overload can be deleted

openslides_backend/permissions/permission_helper.py
r-peschke: I think we have to remove the "guest_meeting_ids"

r-peschke (changes requested):
I think we should remove the method is_temporary in permission_helper.py

openslides_backend/permissions/permission_helper.py
jsangmeister: Yes, I adapted it in my initial PR because all kinds of tests failed without it.

reiterl:
Rebased and rm is_temporary and check_temporary_mixin.

reiterl:
Removed the `temporary_user_mixin.py`

FinnStutzenstein (approved):

jsangmeister (changes requested):
There are still some instances where `temporary` is mentioned:
- `motion_submitter.create` in the docstring
- `user.merge_to_non_temporary` can be renamed and the comment updated as well as its tests
- `test_update_delegation.py` still has a function `setup_vote_delegation_temporary` which needs to be renamed or removed, depending on its usage
- In `test_update_delegation.py` as well as `test_create_delegation.py` are many helper functions which were previously used exactly twice, once for temporary users and once for normal once. You can go over them (they all start with `t_` as far as I see it) and check their usage and move them to the test where they are used, if they are used only once.

r-peschke (approved):
Apart from the changes @jsangmeister request: LGTM

jsangmeister (changes requested):
Also, `guest_meeting_ids` is still present in some places and should be removed.

reiterl:
Working on the different places of "temporary". Update the delegation tests.

openslides_backend/action/actions/motion_submitter/create.py
```diff
@@ -31,9 +31,6 @@ class MotionSubmitterCreateAction(CreateActionWithInferredMeetingMixin, CreateAc
     relation_field_for_meeting = "motion_id"
 
     def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
-        """
-        Check if motion and user belong to the same meeting if the user is a temporary user.
```
jsangmeister, line 35: The docstring can stay, since it now applies to all users instead of only temporary ones.

jsangmeister (changes requested):
`guest_meeting_ids` is still missing, otherwise it looks good

openslides_backend/action/actions/motion_submitter/create.py
reiterl: Rm the guest_meeting_ids.

reiterl: Add a comment: _Check if motion and user belong to the same meeting._

jsangmeister (approved):

PR #730, opened 2021-05-27
Add _missing_ method to ManagementLevels and remove no_right from code

jsangmeister:
I discovered that enums in python offer a `_missing_` method and I added it to the `CompareRightLevel` class for a little cleaner access and creation.

### DIFF
--- a/openslides_backend/permissions/management_levels.py
+++ b/openslides_backend/permissions/management_levels.py
@@ -1,14 +1,24 @@
 from enum import Enum
+from typing import Optional
 
 from .base_classes import VerbosePermission
 
 
 class CompareRightLevel(str, VerbosePermission, Enum):
-    def __new__(cls, value: str, weight: int):  # type: ignore
+    def __new__(
+        cls, value: Optional[str], weight: Optional[int] = None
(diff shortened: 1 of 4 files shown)


r-peschke (approved):
Looks really good!

openslides_backend/permissions/management_levels.py
```diff
@@ -1,15 +1,25 @@
...
         obj._value_ = value
         obj.weight = weight
         return obj
 
+    @classmethod
+    def _missing_(cls, value: object) -> "CompareRightLevel":
+        """
+        Always return the first enum item if no matching one was found. -> NO_RIGHT must always be listed first.
+        """
+        return next(iter(cls))
```
normanjaeckel, line 21: Proposal: Define NO_RIGHT in this parent class with default value ("no_right") and use it here instead of implicit catch of first enum entry (to prevent security issues if a further developer does not keep this in mind.

jsangmeister: Unfortunately, python does not allow subclassing of enums which define members: https://docs.python.org/3/library/enum.html#restricted-enum-subclassing

PR #731, opened 2021-05-28
Remove permission service

reiterl:


openslides_backend/action/action.py
```diff
@@ -193,9 +191,8 @@ def check_permissions(self, instance: Dict[str, Any]) -> None:
                 ):
                     return
                 raise MissingPermission(self.permission)
-        else:
-            if self.permission_service.is_allowed(self.name, self.user_id, [instance]):
-                return
+        elif is_dev_mode():
```
jsangmeister, line 194: Please remove this, ignoring permission errors in dev mode might lead to unnoticed errors.

jsangmeister (changes requested):

jsangmeister:
waiting for #735, #677, #734

FinnStutzenstein:
Why waiting for #677? it might take a while. Since it is not productively used, I would favor removing the perms service and implement some dummy thing for voting and create an issue, if needed and not resolved with #677

jsangmeister:
Also ok for me

openslides_backend/action/action.py
```diff
@@ -194,9 +191,6 @@ def check_permissions(self, instance: Dict[str, Any]) -> None:
                 ):
                     return
                 raise MissingPermission(self.permission)
-        else:
```
FinnStutzenstein, line 197: Can the `if` be removed where this `else` belongs to? What does happen, if `self.permission` is not defined otherwise? If this may happen I would suggest using a `NotImplementedError` to catch errors, where a developer did not provide a permission and neither overwrote the `check_permissions` method.

FinnStutzenstein (approved):

openslides_backend/action/action.py
jsangmeister: Well, in that case it would just throw a 403 on every call, which would be okay I guess (and would happen during tests, too, so the developer would catch it). But we can also change it if you want

reiterl: It will throw an PermissionDenied, so the developer will notice it.

reiterl:
Rebased, and switched poll.vote to internal.

openslides_backend/action/actions/poll/vote.py
```diff
@@ -14,7 +14,7 @@
 from ..vote.user_token_helper import get_user_token
 
 
-@register_action("poll.vote")
+@register_action("poll.vote", internal=True)
```
jsangmeister, line 17: This is not possible, since the client can then no longer call this action and this is used in the staging server and full setup. Just overwrite the `check_permissions` method here to return directly, without any checks.

jsangmeister (changes requested):

openslides_backend/action/actions/poll/vote.py
FinnStutzenstein: ...and create an issue

jsangmeister: Actually, an issue is not needed, since the action will be removed once #677 is done

reiterl: Rm `internal` and add an empty `check_permissions` to `poll.vote`.

jsangmeister (approved):

PR #735, opened 2021-05-31
Add permissions to organisation.update.

reiterl:
Add some tests for group a and group b


reiterl:
The organisation field `enable_electronic_voting` is ro in the model. Exclude it from the optional_properties.

jsangmeister (approved):

PR #741, opened 2021-06-03
Updated dependencies and adjusted code accordingly

jsangmeister:


openslides_backend/http/http_exceptions.py
```diff
@@ -1,5 +1,5 @@
 import json
-from typing import Any, Dict, List, Optional, Tuple
+from typing import Any, Dict, List, Optional, Text, Tuple
```
FinnStutzenstein, line 2: looking at https://docs.python.org/3/library/typing.html#typing.Text, it seems that we should use `str` since we did not migrate from python2, or did werkzeug changed its type annotations?

FinnStutzenstein (approved):
Looks like it weren't to much breaking changes.

r-peschke (commented):
The tests for user update/create are more or less useless, because most of them are written newly for the new permissions. Perhaps you should merge firs and I will rebase on this

openslides_backend/http/http_exceptions.py
jsangmeister: Yes, the latter. The correct type would be WSGIEnvironment, but since I do not seem to have access to that, I will stick to their analog typedef.

jsangmeister:
> The tests for user update/create are more or less useless, because most of them are written newly for the new permissions. Perhaps you should merge firs and I will rebase on this

Sure, this were just automatic changes from black anyway.

r-peschke (commented):
LGTM

PR #744, opened 2021-06-04
Add user_ids to meeting.create. Add two tests.

reiterl:


FinnStutzenstein (commented):
Adding users looks good. This is missing:

> Additionally, the request user is added to the admin group

reiterl:
Adding self.user_id is already done, by this code.

 # Add user to admin group
        action_data = [
            {
                "id": self.user_id,
                "group_$_ids": {str(instance["id"]): [fqid_admin_group.id]},
            }
        ]
        self.execute_other_action(UserUpdate, action_data)

openslides_backend/action/actions/meeting/create.py
```diff
@@ -136,7 +139,28 @@ def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
...
         self.execute_other_action(UserUpdate, action_data)
+        # Add users to default group
+        if instance.get("user_ids"):
+            user_ids = instance.pop("user_ids")
+            committee = self.datastore.get(
+                FullQualifiedId(Collection("committee"), instance["committee_id"]),
+                ["user_ids"],
+            )
+            if not all(
+                user_id in committee.get("user_ids", []) for user_id in user_ids
```
jsangmeister, line 150: Probably more performant to create two sets from these and compare them

reiterl: I don't think so and the code would be more compilcate.

jsangmeister: Well, this way it's O(n^2) since you iterate both lists. With sets, it would be O(2*n*log(n))=O(n*log(n)) for the construction and then again O(n*log(n)) for the searching. So, it would be more performant ;) But I don't think it's that important, that's why I only created a comment.

jsangmeister (approved):

FinnStutzenstein (approved):

openslides_backend/action/actions/meeting/create.py
reiterl: I think it is O(2*n).

reiterl: Transfer the committee user_ids into a dict/set could save time. If the "in" operator is O(n) on lists.

PR #763, opened 2021-06-10
Fix action response

jsangmeister:


tests/system/action/mediafile/test_delete.py
```diff
@@ -14,7 +14,7 @@ def setUp(self) -> None:
             },
         }
 
-    def test_delete_correct(self) -> None:
+    def test_delete_correctx(self) -> None:
```
FinnStutzenstein, line 17: remove the `x`

FinnStutzenstein (commented):
Looks good, not tested. If I read correctly: You have chosen the sort variant of `[None]`, since this action never returns a result, so there is no need to return `None` for every payload?

jsangmeister:
> Looks good, not tested. If I read correctly: You have chosen the sort variant of `[None]`, since this action never returns a result, so there is no need to return `None` for every payload?

Correct, at least for the generic case. Individual action my of course overwrite this, but only they know what and how many elements were actually created in `get_updated_instances`.

FinnStutzenstein (approved):

PR #776, opened 2021-06-14
Fix model locked error

### LINKED ISSUE #747: model locked on poll update
The client sends this request to update an analog poll:
```
"[{"action":"poll.update","data":[{"id":6,"majority_method":"simple","onehundred_percent_base":"valid","pollmethod":"YNA","title":"Ballot"}]},{"action":"option.update","data":[{"id":14,"Y":"1.000000","A":"3.000000","N":"2.000000"},{"id":15,"Y":"4.000000","N":"-1.000000"}]}]"
```
which results in `Error: Datastore service sends HTTP 400. Model 'poll/6' raises MODEL_LOCKED error.` I would expect this to work. Since I do not know what exactly is locked, I would suppose that too much is locked.

jsangmeister:
fixes #747

### DIFF
--- a/openslides_backend/action/action_handler.py
+++ b/openslides_backend/action/action_handler.py
@@ -205,6 +205,9 @@ class ActionHandler(BaseHandler):
 
                 # add locked_fields to request
                 write_request.locked_fields = self.datastore.locked_fields
+                # reset locked fields, but not addtional relation models - these might be needed
+                # by another action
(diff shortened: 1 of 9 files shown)


r-peschke (approved):
I didn't reviewed all the logic, but it seems okay for me

FinnStutzenstein (approved):

PR #781, opened 2021-06-15
Add mimtypes ttf, otf, 7z, odt, ods for mediafile and resource upload

r-peschke:
Installs debian packet mime-support to use offered mimetype/extension mapping


tsiegleauq (changes requested):
Uploading a otf works now, yet setting it as a font
`{action: "meeting.set_font", data: [{id: 1, mediafile_id: 17, place: "regular"}]}`

fails with 

```json
{
  "success": false,
  "message": "Invalid mimetype: font/otf, allowed are ['font/ttf', 'font/woff', 'application/font-woff', 'application/font-sfnt']",
  "action_error_index": 0,
  "action_data_error_index": 0
}
apparently, other methods besides to the mere upload require adjustment as well
```

jsangmeister:
> Uploading a otf works now, yet setting it as a font
> `{action: "meeting.set_font", data: [{id: 1, mediafile_id: 17, place: "regular"}]}`
> 
> fails 

`set_font` and `set_logo` have their own whitelists, see https://github.com/OpenSlides/openslides-backend/blob/e0e0bfab995fc303e426b51ab71baf840b7d8607/openslides_backend/action/actions/meeting/set_font.py#L12

openslides_backend/action/actions/mediafile/upload.py
```diff
@@ -51,6 +52,7 @@ class MediafileUploadAction(
     dependencies = [ListOfSpeakersCreate]
 
     def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
+        add_mimetypes()
```
jsangmeister, line 55: These should be added automatically, an action should not need to know about the installed mimetypes. Maybe you can move this somewhere to the setup of the server?

openslides_backend/action/actions/mediafile/add_custom_mimetypes.py
```diff
@@ -0,0 +1,13 @@
...
+def add_mimetypes() -> None:
+    """
+    for mimetypes we use the mimetypes of the current python version.
+    If we miss some, we could add them here.
+    """
+    mimetypes.add_type("font/ttf", ".ttf")
+    mimetypes.add_type("font/otf", ".otf")
+    mimetypes.add_type("application/x-7z-compressed", ".7z")
+    mimetypes.add_type("application/vnd.oasis.opendocument.text", ".odt")
+    mimetypes.add_type("application/vnd.oasis.opendocument.spreadsheet", ".ods")
```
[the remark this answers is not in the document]
r-peschke: The backend works with filenames and their extension to guess the mimetypes. Therefore we don't need for instance 'font/sfnt', because the extension is `.otf`. The other way round a filename `xyz.otf` always get the mimetype `font/ttf`. The mimetype/extension combination `woff2` will be recognized in mimetypes guesing, but I appended it as only needed type in the allowed_mimetypes list for font setting.

openslides_backend/action/actions/mediafile/upload.py
r-peschke: This is done by installing the debian packet `mime-support`. Python mimetypes looks for it, namely the file /etc/mime.types, and uses it for guessing types/extensions. All missed extensions exist with installed mime-support.

tsiegleauq (approved):
Tested. Works for me.

jsangmeister (approved):

openslides_backend/action/actions/mediafile/upload.py
jsangmeister: Perfect, that's the solution I was looking for :D

Issue #772, opened 2021-06-14
Add committee.update test


jsangmeister:
Add a test with the following payload:
```
{
    id: 1,
    forward_to_committee_ids: [1],
    receive_forwardings_from_committee_ids: [],
}
```
The resulting relations should be either be consistent or the request should fail since one of the payload fields will necessarily be ignored.

PR #788, opened 2021-06-17
Issue778 772 committee fields and permissions

r-peschke:
* committee.update with new permissions for group A and B fields
* committee.create with forward and receive forward fields
* from 772: tests for update and create: forwarding or receiving from self is not possible

### DIFF
--- a/openslides_backend/action/actions/committee/committee_common_mixin.py
+++ b/openslides_backend/action/actions/committee/committee_common_mixin.py
@@ -1,0 +1,17 @@
+from typing import Any, Dict
+
+from ....shared.exceptions import ActionException
+from ...action import Action
+
+
+class CommitteeCommonCreateUpdateMixin(Action):
+    def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
+        instance = super().update_instance(instance)
+        id = instance.get("id")
+        if id in instance.get("forward_to_committee_ids", []) or id in instance.get(
+            "receive_forwardings_from_committee_ids", []
+        ):
+            raise ActionException(
+                "Forwarding or receiving from own committee is not possible!"
+            )
+        return instance
--- a/openslides_backend/action/actions/committee/committee_common_mixin.py
+++ b/openslides_backend/action/actions/committee/committee_common_mixin.py
@@ -6,12 +6,20 @@ from ...action import Action
 
 class CommitteeCommonCreateUpdateMixin(Action):
     def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
+        """Check if own committee is forwarded or received explicitly,
+        it may not be excluded by the opposite setting
+        """
         instance = super().update_instance(instance)
+        if (
+            instance.get("forward_to_committee_ids") is None
+            or instance.get("receive_forwardings_from_committee_ids") is None
+        ):
+            return instance
         id = instance.get("id")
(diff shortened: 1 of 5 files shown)


openslides_backend/action/actions/committee/committee_common_mixin.py
```diff
@@ -0,0 +1,17 @@
...
+
+class CommitteeCommonCreateUpdateMixin(Action):
+    def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
+        instance = super().update_instance(instance)
+        id = instance.get("id")
+        if id in instance.get("forward_to_committee_ids", []) or id in instance.get(
+            "receive_forwardings_from_committee_ids", []
+        ):
+            raise ActionException(
+                "Forwarding or receiving from own committee is not possible!"
```
jsangmeister, line 15: I think this is wrong: AFAIK, forwarding is possible to the same committee. I guess you misunderstood https://github.com/OpenSlides/openslides-backend/issues/772: The point was that the "forward" and "receive" relations indicated different result, so this must be resolved. One option would be to just set the "forward" field (therefore overwriting the "receive" field), another one to raise an exception because of inconsistent data. This, however, would just move the work to the client, since the user can select its own committee in only one field, which woud have to be caught there, so I'm probably more leaning towards the first solution (if our relation handling is capable of that)

openslides_backend/action/actions/committee/update.py
```diff
@@ -36,49 +37,33 @@ class CommitteeUpdateAction(UpdateAction):
...
+            self.datastore,
+            self.user_id,
+            CommitteeManagementLevel.CAN_MANAGE,
+            instance["id"],
         ):
-            raise PermissionDenied("Missing can_manage_organization and not manager.")
+            return
+
+        raise PermissionDenied(
+            "Missing can_manage_organization or can_manage_committee."
```
jsangmeister, line 68: You can throw a `MissingPermission` here and just provide multiple missing permissions in a map.

jsangmeister (changes requested):

openslides_backend/action/actions/committee/committee_common_mixin.py
r-peschke: Talked with Emanuel and you are right. Changed

FinnStutzenstein (approved):
Apart from @jsangmeister remarks it looks good

openslides_backend/action/actions/committee/committee_common_mixin.py
```diff
@@ -0,0 +1,25 @@
...
+            instance.get("forward_to_committee_ids") is None
+            or instance.get("receive_forwardings_from_committee_ids") is None
+        ):
+            return instance
+        id = instance.get("id")
+        if (id in instance.get("forward_to_committee_ids", [])) != (
+            id in instance.get("receive_forwardings_from_committee_ids", [])
+        ):
+            raise ActionException(
+                "Forwarding or receiving to/from own must be configured in both directions!"
```
jsangmeister, line 23: Just out of curiosity: What happens if it's only defined in one direction? Does our relation handling break?

jsangmeister (approved):

openslides_backend/action/actions/committee/committee_common_mixin.py
r-peschke: No, it handles both a consecutive actions

PR #792, opened 2021-06-21
Update min/max check of poll.vote.

reiterl:
Update the tests.
I think the vote service should fix this anyway, so just a short fix.


jsangmeister (changes requested):
Please add a test nevertheless, because we will keep the tests for the vote service.

tsiegleauq (approved):
Tested. Using YN und YNA works now.

jsangmeister (approved):

PR #799, opened 2021-06-24
Fix KeyError on meeting.create

jsangmeister:


tests/system/action/user/test_create.py
```diff
@@ -147,7 +147,7 @@ def test_create_template_fields(self) -> None:
         assert user.get("vote_weight_$1") == "1.000000"
         assert user.get("vote_weight_$2") == "2.333333"
         assert set(user.get("vote_weight_$", [])) == {"1", "2"}
-        assert user.get("meeting_ids") == [1, 2]
+        assert set(user.get("meeting_ids", [])) == {1, 2}
```
r-peschke, line 150: To check lists IMO the best is self.assertCountEqual(list1, list2), which compares lists without consideration of sequence. `set` changes lists by removing dublicates

tests/system/action/user/test_update.py
```diff
@@ -118,7 +118,7 @@ def test_update_template_fields(self) -> None:
         assert user.get("vote_weight_$1") == "1.000000"
         assert user.get("vote_weight_$2") == "2.333333"
         assert set(user.get("vote_weight_$", [])) == {"1", "2"}
-        assert user.get("meeting_ids") == [1, 2]
+        assert set(user.get("meeting_ids", [])) == {1, 2}
```
r-peschke, line 121: see above

tests/system/action/meeting/test_create.py
```diff
@@ -167,6 +166,47 @@ def test_create_check_users(self) -> None:
             "user/1", {f"group_${meeting['id']}_ids": [admin_group_id]}
         )
 
+    def test_create_with_request_user(self) -> None:
+        self.set_models(
```
r-peschke, line 170: what do you think about a comment for this test, explaining group2 as admin group for the request user?

r-peschke (changes requested):

tests/system/action/user/test_create.py
jsangmeister: Ah, good call, I did not know about this method! Will change

tests/system/action/meeting/test_create.py
jsangmeister: Sure, I can add a comment

r-peschke (approved):

PR #802, opened 2021-06-25
Clear votes* and entitled_users_at_stop fields in poll reset

reiterl:


openslides_backend/action/actions/poll/reset.py
```diff
@@ -28,6 +28,10 @@ def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
         )
         instance["is_pseudoanonymized"] = poll.get("type") == Poll.TYPE_PSEUDOANONYMOUS
         instance["voted_ids"] = []
+        instance["entitled_users_at_stop"] = []
+        instance["votesvalid"] = "0.000000"
+        instance["votesinvalid"] = "0.000000"
+        instance["votescast"] = "0.000000"
```
jsangmeister, line 34: Although only a technicality, I think it's better to set these all to `None`, as would be the case for a newly created poll

jsangmeister (changes requested):

openslides_backend/action/actions/poll/reset.py
reiterl: Set these 4 fields to None in reset and update the tests.

jsangmeister (approved):

PR #804, opened 2021-06-25
Adjust committee.update permissions.

reiterl:


reiterl:
Not much to do. I just changed the MissingPermission input. And adjust the tests.

jsangmeister:
Ok, for this action these changes suffice - we can adjust the generic CML handling when we need it in another action I guess.

jsangmeister (approved):

PR #805, opened 2021-06-25
Add speech_state to speaker.create and checks.

reiterl:


jsangmeister (changes requested):
Since these are the same requirements as in `speaker.update`, can you extract the code into a mixin to prevent duplicate code?

reiterl:
The req are different from update to create. Look at the actions in the wiki. So I cannot use a mixin.

jsangmeister:
No, the requirements are exactly the same as the second and third entries in the table for `speaker.update`: https://github.com/OpenSlides/OpenSlides/wiki/speaker.update#action (the `X` to `contribution`/`pro/contra` cases.

reiterl:
I use a mixin for the create and part of the update speech_state checks.

openslides_backend/action/actions/speaker/mixins.py
```diff
@@ -0,0 +1,38 @@
...
+
+from ....permissions.permission_helper import has_perm
+from ....permissions.permissions import Permissions
+from ....shared.exceptions import ActionException
+from ....shared.patterns import Collection, FullQualifiedId
+from ...action import Action
+
+
+class CheckSpeechState(Action):
+    def check_speech_state(
```
jsangmeister, line 11: I would suggest to move the whole validation code from `speaker.update` into the mixin. The way it currently is you still have a lot of duplicated code (fetching the settings & permissions). If you move everything in here and give the db instance of speaker as an additional param (where you give `{}`) for the create case, you can deduplicate even more code.

jsangmeister (changes requested):

reiterl:
I updated the code to reduce duplications.

openslides_backend/action/actions/speaker/mixins.py
reiterl: Add the speaker param to the check_speech_state and move the speaker.update code into the mixin.

jsangmeister (approved):

PR #823, opened 2021-07-07
Update Datastore Library

FinnStutzenstein:
Note: The hardcoded commit hashes and my repo instead of `OpenSlides` must be changed!


FinnStutzenstein:
TODO: https://github.com/OpenSlides/openslides-datastore-service/blob/master/reader/lib/pip-reader/readerlib/__init__.py#L4 this must be called when starting the Backend normally.

jsangmeister:
> TODO: https://github.com/OpenSlides/openslides-datastore-service/blob/master/reader/lib/pip-reader/readerlib/__init__.py#L4 this must be called when starting the Backend normally.

proposal: move to `openslides_backend/__init__.py`

dev/docker-compose.dev.yml
```diff
@@ -28,7 +27,7 @@ services:
             - auth
     reader:
         build:
-            context: "https://github.com/OpenSlides/openslides-datastore-service.git"
+            context: "https://github.com/OpenSlides/openslides-datastore-service.git#fbe9aca1aa7b29c3ec728876eb530308ee53c165"
```
jsangmeister, line 30: Do you really want to fix the commit? this was previously not the case

FinnStutzenstein: Right

entrypoint.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+export DATASTORE_DATABASE_HOST=${DATASTORE_DATABASE_HOST:-postgresql}
+export DATASTORE_DATABASE_PORT=${DATASTORE_DATABASE_PORT:-5432}
+export DATASTORE_DATABASE_USER=${DATASTORE_DATABASE_USER:-openslides}
+export DATASTORE_DATABASE_NAME=${DATASTORE_DATABASE_NAME:-openslides}
+export DATASTORE_DATABASE_PASSWORD=${DATASTORE_DATABASE_PASSWORD:-openslides}
+
+echo "entrypoint run"
```
jsangmeister, line 9: I think this can be removed

PR #839, opened 2021-07-13
In check unique title ignore updated countdown.

reiterl:
Add a test.

### DIFF
--- a/tests/system/action/projector_countdown/test_update.py
+++ b/tests/system/action/projector_countdown/test_update.py
@@ -50,18 +50,11 @@ class ProjectorCountdownUpdate(BaseActionTestCase):
             {
                 "id": 2,
                 "title": "test",
-                "description": "good bla",
-                "default_time": 30,
-                "countdown_time": 20,
-                "running": True,
             },
         )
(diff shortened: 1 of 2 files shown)


tests/system/action/projector_countdown/test_update.py
```diff
@@ -44,6 +44,25 @@ def test_update(self) -> None:
         assert model.get("default_time") == 30
         assert model.get("countdown_time") == 20
 
+    def test_update_same_title_in_same_id(self) -> None:
+        response = self.request(
+            "projector_countdown.update",
+            {
+                "id": 2,
+                "title": "test",
+                "description": "good bla",
```
FinnStutzenstein, line 53: Please remove the 4 following fields since they are not in the scope of this test.

FinnStutzenstein (changes requested):
Small change in the test, than it can be merged.

tests/system/action/projector_countdown/test_update.py
reiterl: I removed the 4 fields from the test. And will merge it now.

Issue #818, opened 2021-07-05
Implement forwarding tree


jsangmeister:
When `motion.create_forwarded` is called, `all_origin_ids` and `all_derived_motion_ids` must be (re-)calculated for the newly created motion as well as the forwarding motion.

Implementation note: The newly created motion always has `all_derived_motion_ids == []` and `all_origin_ids` must be set to `all_origin_ids` of its parent plus the parent id. Then, the newly created id must be added to `all_derived_motion_ids` of all motions in the new `all_origin_ids`.

Deleting motions should analogously update the fields correctly.

Also, in the first test a check is missing that `derived_motion_ids` is set correctly.

FinnStutzenstein:
>  as well as the forwarding motion.

And every motion in the tree.

jsangmeister:
> > as well as the forwarding motion.
> 
> And every motion in the tree.

Actually, only all (grand-)parents, right? Siblings do not appear in the tree, at least not according to the linked description.

jsangmeister:
Field split in two in https://github.com/OpenSlides/OpenSlides/pull/6145 - logic should be similar, except that only all parents need to be updated on a new forward.

reiterl:
How should the delete logic work?
Let motion 1 -> 2 ->3 and motion 2 is deleted. What should happen with the two fields?

jsangmeister:
Good question. Maybe @emanuelschuetze has an opinion here

jsangmeister:
The logic for creation is added here: https://github.com/OpenSlides/OpenSlides/wiki/motion.create_forwarded#forwarding-tree-fields. Please double check, but I think it is already implemented correctly. I discussed the delete case with @emanuelschuetze and for display purposes it is best to just leave the deleted motion ids in the trees, so it can still be correctly displayed in the client. So I think there should be nothing to do.

PR #843, opened 2021-07-14
Add calc fields for motion.create_forwarded.

reiterl:
Calc all_derived_motion_ids and all_origin_ids.
Update the tests.

**Note:** This is only the create case. The delete case must be discussed.
Based on the gen_models branch.

### DIFF
--- a/openslides_backend/action/actions/motion/__init__.py
+++ b/openslides_backend/action/actions/motion/__init__.py
@@ -10,4 +10,5 @@ from . import (  # noqa
     set_support_self,
     sort,
     update,
+    update_all_derived_motion_ids,
 )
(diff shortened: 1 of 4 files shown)


jsangmeister (changes requested):
Please add more tests with more complex setups (e.g. multiple levels and/or multiple derived motions from one origin)

reiterl:
Added a complex test to the test_create_forwarded, this shows a problem in the code. 
Fixed that too.

jsangmeister:
waiting for a decision in #818 here

jsangmeister (approved):
I think this is correct for the new requirements

PR #848, opened 2021-07-15
Moved datastore import, so it is not needed outside the service and t…

FinnStutzenstein:
…ests


jsangmeister (approved):
What's the purpose here?

reiterl:
This seem to be the fix to the check models error in the CI, okay.

jsangmeister:
Ahh, I see

PR #852, opened 2021-07-16
Add check for required fields and update the tests.

reiterl:
Add a test for the missing required case.

### DIFF
--- a/openslides_backend/models/checker.py
+++ b/openslides_backend/models/checker.py
@@ -26,7 +26,7 @@ from openslides_backend.models.fields import (
     RelationListField,
     TimestampField,
 )
-from openslides_backend.models.models import Model
+from openslides_backend.models.models import Meeting, Model
 from openslides_backend.shared.patterns import KEYSEPARATOR, Collection
 
 SCHEMA = fastjsonschema.compile(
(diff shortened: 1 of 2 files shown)


openslides_backend/models/checker.py
```diff
@@ -409,6 +409,13 @@ def check_types(self, model: Dict[str, Any], collection: str) -> None:
                 error = f"{collection}/{model['id']}/{field}: Type error: Type is not {field_type}"
                 self.errors.append(error)
 
+            # check if required field is not empty
+            # committee_id is a special case, because it is filled after the
+            # replacement
+            if field_type.required and not model[field] and field != "committee_id":
```
jsangmeister, line 415: Please use `check_required_not_fulfilled` here since different fields have different meanings of "required". Also, writing `"committee_id"` as a string is error prone since if we add another field with the same name somewhere, this suddenly fails for no apparent reason and might lead to unwanted (or even unnoticed) side effects. You can instead check for `field_type == Meeting.committee_id`. Please also change all the other comparisons where you check for specific fields, I did not notice before that this would be a better solution.

jsangmeister (changes requested):

openslides_backend/models/checker.py
reiterl: Use check_required_not_fullfilled for the check. 
Use the field_type != Meeting.committee for the field check.

jsangmeister (approved):

PR #860, opened 2021-07-21
Add manager_ids to committee.create.

reiterl:
Use UserUpdate to set the can_manage permissions to the managers.
A manager does not need to be part of a committee. So removed the code,
which checks this.

### DIFF
--- a/openslides_backend/action/actions/committee/create.py
+++ b/openslides_backend/action/actions/committee/create.py
@@ -1,8 +1,15 @@
+from typing import Any, Dict, List
+
 from ....models.models import Committee
-from ....permissions.management_levels import OrganizationManagementLevel
+from ....permissions.management_levels import (
+    CommitteeManagementLevel,
+    OrganizationManagementLevel,
+)
+from ....shared.schema import id_list_schema
 from ...generics.create import CreateAction
 from ...util.default_schema import DefaultSchema
 from ...util.register import register_action
+from ..user.update import UserUpdate
 from .committee_common_mixin import CommitteeCommonCreateUpdateMixin
 
 
@@ -22,5 +29,24 @@ class CommitteeCreate(CommitteeCommonCreateUpdateMixin, CreateAction):
             "forward_to_committee_ids",
             "receive_forwardings_from_committee_ids",
         ],
+        additional_optional_fields={"manager_ids": id_list_schema},
     )
     permission = OrganizationManagementLevel.CAN_MANAGE_ORGANIZATION
+
+    def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
+        instance = super().update_instance(instance)
+        if "manager_ids" in instance:
+            manager_ids = instance.pop("manager_ids")
+            action_data: List[Dict[str, Any]] = []
+            for manager_id in manager_ids:
+                action_data.append(
+                    {
+                        "id": manager_id,
+                        "committee_$_management_level": {
+                            str(instance["id"]): CommitteeManagementLevel.CAN_MANAGE,
+                        },
+                    }
+                )
+            self.apply_instance(instance)
+            self.execute_other_action(UserUpdate, action_data)
+        return instance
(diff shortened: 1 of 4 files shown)


openslides_backend/action/actions/user/update.py
```diff
@@ -49,21 +45,3 @@ class UserUpdate(
             "is_demo_user",
         ],
     )
-
```
jsangmeister, line 52: I think this is not what we discussed: Of course, we can model it this way, but we should really write it down somewhere how it's supposed to be. I understood it that all users given in `manager_ids` are just also added as members to the committee. Then, this check can stay and should work for the new code (except that you need to use `fetch_model` instead of `get`. @tsiegleauq @FinnStutzenstein Is this correct or do I remember it wrong?

jsangmeister (changes requested):

openslides_backend/action/actions/user/update.py
reiterl: Re added the user check. And add the manager to the users of a committee.

r-peschke (approved):

tsiegleauq (changes requested):
tested.

updating a committee results in
`data must not contain {'manager_ids'} properties`
It seems there is no apparent reason why a committee manager should not be altered over editing. It's quite counter-intuitive that way, I added it to the [documentation](https://github.com/OpenSlides/OpenSlides/wiki/committee.update).
Please change accordingly.

reiterl:
This pullrequest is about `committee.create`. Not about `committee.update`. We should not enlarge pullrequests in this way.
Please write a new issue about your thoughts about `committee.update`.

tests/system/action/committee/test_create.py
```diff
@@ -60,6 +61,89 @@ def test_create_only_required(self) -> None:
...
+        assert model.get("user_ids") == [13]
+        self.assert_model_exists(
+            "user/13",
+            {
+                "committee_$1_management_level": CommitteeManagementLevel.CAN_MANAGE,
+                "committee_ids": [1],
+            },
+        )
+
+    def test_create_manager_ids_2(self) -> None:
```
jsangmeister, line 85:
```suggestion
    def test_create_manager_ids_with_existing_committee(self) -> None:
```

tests/system/action/committee/test_create.py
```diff
@@ -60,6 +61,89 @@ def test_create_only_required(self) -> None:
...
+            "user/13",
+            {
+                "committee_$4_management_level": CommitteeManagementLevel.CAN_MANAGE,
+                "committee_$3_management_level": CommitteeManagementLevel.CAN_MANAGE,
+                "committee_$_management_level": ["3", "4"],
+                "committee_ids": [3, 4],
+            },
+        )
+
+    def test_create_manager_ids_3(self) -> None:
```
jsangmeister, line 121:
```suggestion
    def test_create_manager_ids_and_user_ids(self) -> None:
```

openslides_backend/action/actions/committee/create.py
```diff
@@ -22,5 +30,29 @@ class CommitteeCreate(CommitteeCommonCreateUpdateMixin, CreateAction):
...
+
+    def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
+        instance = super().update_instance(instance)
+        if "manager_ids" in instance:
+            manager_ids = instance.pop("manager_ids")
+            action_data: List[Dict[str, Any]] = []
+            for manager_id in manager_ids:
+                manager = self.datastore.get(
+                    FullQualifiedId(Collection("user"), manager_id), ["committee_ids"]
+                )
```
jsangmeister, line 45: A `get_many` would be better here, but maybe you want to change this together with https://github.com/OpenSlides/openslides-backend/pull/866#issuecomment-887277980

jsangmeister (changes requested):

tests/system/action/committee/test_create.py
reiterl: Renamed.

tests/system/action/committee/test_create.py
reiterl: Renamed.

openslides_backend/action/actions/committee/create.py
reiterl: Will later fix this.

jsangmeister (approved):

PR #861, opened 2021-07-22
Add level and weight to the assigned agenda_items.

reiterl:
Update the tests. In parent_id None, use level 0 and weight 10000+num.


jsangmeister (approved):

jsangmeister:
@GabrielInTheWorld do you want to check if this works for you?

GabrielInTheWorld (approved):
I think, it's okay. If I assign more than one item to an agenda_item A, than the 2nd + x items are aligned after the A + 1 (+ ...) agenda_item. This is because the weights of the new items and of the previous items can overlap.

PR #862, opened 2021-07-22
Add admin_ids to meeting.create, add oml check to has_cml.

reiterl:
Update tests.
Two things are done in this pr. The oml can_manage_organization is checked in the has_committee_management_level, too.
Add admins setting to meeting.create.

### DIFF
--- a/openslides_backend/action/actions/meeting/create.py
+++ b/openslides_backend/action/actions/meeting/create.py
@@ -134,6 +134,7 @@ class MeetingCreate(CreateActionWithDependencies, MeetingPermissionMixin):
         instance["admin_group_id"] = fqid_admin_group.id
 
         # Add user to admin group
+        admin_ids = []
         if instance.get("admin_ids"):
             admin_ids = instance.pop("admin_ids")
             committee = self.datastore.get(
@@ -170,6 +171,7 @@ class MeetingCreate(CreateActionWithDependencies, MeetingPermissionMixin):
                     "group_$_ids": {str(instance["id"]): [fqid_default_group.id]},
                 }
                 for user_id in user_ids
+                if user_id not in admin_ids
             ]
 
             self.execute_other_action(UserUpdate, action_data)
--- a/openslides_backend/action/actions/meeting/create.py
+++ b/openslides_backend/action/actions/meeting/create.py
@@ -31,7 +31,10 @@ class MeetingCreate(CreateActionWithDependencies, MeetingPermissionMixin):
             "url_name",
             "organization_tag_ids",
         ],
-        additional_optional_fields={"user_ids": id_list_schema},
+        additional_optional_fields={
+            "user_ids": id_list_schema,
+            "admin_ids": id_list_schema,
+        },
     )
     dependencies = [
         MotionWorkflowCreateSimpleWorkflowAction,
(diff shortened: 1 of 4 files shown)


openslides_backend/action/actions/meeting/create.py
```diff
@@ -149,14 +164,12 @@ def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
...
             ):
                 raise ActionException("Only allowed to add users from committee.")
-            # filter out request user since he was already added
             action_data = [
                 {
                     "id": user_id,
                     "group_$_ids": {str(instance["id"]): [fqid_default_group.id]},
                 }
                 for user_id in user_ids
-                if user_id != self.user_id
```
jsangmeister, line 159: You probably still have to filter out all users which you added as admins, because else you will overwrite the admin group and all admins will end up in the default group. See also comment regarding the test above

jsangmeister (changes requested):
Please also add a permission test that you can now execute `meeting.create` with the correct CML.

reiterl:
Add some tests for the permissions and admin_ids.

openslides_backend/action/actions/meeting/create.py
reiterl: Filter out all admins here. The admins should go into the admin group and not into the default group.

tests/system/action/meeting/test_create.py
```diff
@@ -156,69 +149,36 @@ def test_check_action_data_fields(self) -> None:
...
         default_group_id = meeting.get("default_group_id")
         self.assert_model_exists(
             "user/2", {f"group_${meeting['id']}_ids": [default_group_id]}
         )
-        admin_group_id = meeting.get("admin_group_id")
-        self.assert_model_exists(
-            "user/1", {f"group_${meeting['id']}_ids": [admin_group_id]}
-        )
 
-    def test_create_with_request_user(self) -> None:
```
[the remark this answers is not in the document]
reiterl: Readded a same user in users and admins test.

tests/system/action/meeting/test_create.py
```diff
@@ -244,3 +252,44 @@ def test_create_permissions(self) -> None:
...
+            }
+        )
+        meeting = self.basic_test({"admin_ids": [2]})
+        assert meeting.get("user_ids") == [2]
+        admin_group_id = meeting.get("admin_group_id")
+        self.assert_model_exists(
+            "user/2", {f"group_${meeting['id']}_ids": [admin_group_id]}
+        )
+
+    def test_create_with_admin_ids_and_no_permissions(self) -> None:
```
jsangmeister, line 272: Isn't this test redundant with `test_create_no_permissions` above?

jsangmeister (changes requested):
A test where the user has OML `can_manage_organization` is still missing

tests/system/action/meeting/test_create.py
reiterl: It has a slightly different payload, but I think we can remove it.

reiterl:
I added a test _test_create_with_admin_ids_and_permissions_oml,_ where the user 1 has oml can_manage_organization.
I removed the surplus test _test_create_with_admin_ids_and_no_permissions_ .

jsangmeister (approved):

PR #866, opened 2021-07-26
Add manager_ids to committee.update.

reiterl:
Add the code and some tests. Use the UserUpdate action, use
get_updated_instances, because payload could be only id and manager_ids.

### DIFF
--- a/openslides_backend/action/actions/committee/update.py
+++ b/openslides_backend/action/actions/committee/update.py
@@ -54,38 +54,34 @@ class CommitteeUpdateAction(CommitteeCommonCreateUpdateMixin, UpdateAction):
                     instance["default_meeting_id"], instance["id"]
                 )
             if "manager_ids" in instance:
-                new_manager_ids = instance.pop("manager_ids")
+                new_manager_ids = set(instance.pop("manager_ids"))
                 filter_ = FilterOperator(
                     f"committee_${instance['id']}_management_level",
                     "=",
                     CommitteeManagementLevel.CAN_MANAGE,
                 )
                 old_manager = self.datastore.filter(Collection("user"), filter_, ["id"])
-                old_manager_ids = [int(id_) for id_ in old_manager]
+                old_manager_ids = set(int(id_) for id_ in old_manager)
 
                 action_data = []
-                for manager_id in new_manager_ids:
-                    if manager_id not in old_manager_ids:
-                        action_data.append(
-                            {
-                                "id": manager_id,
-                                "committee_$_management_level": {
-                                    str(
-                                        instance["id"]
-                                    ): CommitteeManagementLevel.CAN_MANAGE,
-                                },
-                            }
-                        )
-                for manager_id in old_manager_ids:
-                    if manager_id not in new_manager_ids:
-                        action_data.append(
-                            {
-                                "id": manager_id,
-                                "committee_$_management_level": {
-                                    str(instance["id"]): None
-                                },
-                            }
-                        )
+                for manager_id in new_manager_ids - old_manager_ids:
+                    action_data.append(
+                        {
+                            "id": manager_id,
+                            "committee_$_management_level": {
+                                str(
+                                    instance["id"]
+                                ): CommitteeManagementLevel.CAN_MANAGE,
+                            },
+                        }
+                    )
+                for manager_id in old_manager_ids - new_manager_ids:
+                    action_data.append(
+                        {
+                            "id": manager_id,
+                            "committee_$_management_level": {str(instance["id"]): None},
+                        }
+                    )
                 if action_data:
                     self.execute_other_action(UserUpdate, action_data)
             if any(key for key in instance if key != "id"):
--- a/openslides_backend/action/actions/committee/update.py
+++ b/openslides_backend/action/actions/committee/update.py
@@ -16,7 +16,6 @@ from ....shared.schema import id_list_schema
 from ...generics.update import UpdateAction
 from ...util.default_schema import DefaultSchema
 from ...util.register import register_action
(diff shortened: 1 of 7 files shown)


openslides_backend/action/actions/committee/update.py
```diff
@@ -34,20 +38,56 @@ class CommitteeUpdateAction(CommitteeCommonCreateUpdateMixin, UpdateAction):
...
-                instance["template_meeting_id"], instance["id"]
-            )
-        if instance.get("default_meeting_id"):
-            self.check_meeting_in_committee(
-                instance["default_meeting_id"], instance["id"]
-            )
-        return instance
+    def get_updated_instances(self, action_data: ActionData) -> ActionData:
+        for instance in action_data:
+            instance = super().update_instance(instance)
```
jsangmeister, line 47: Do not call `update_instance` inside `get_updated_instances`, is called by `perform` automatically afterwards. Also, I do not understand why you can't just put all your code in `update_instance`, if you `pop` the `manager_ids` before returning, everything should work nicely.

If not, at least you can leave the previous checks for `*_meeting_id` in `update_instance` since they belong there.

openslides_backend/action/actions/committee/update.py
```diff
@@ -34,20 +38,56 @@ class CommitteeUpdateAction(CommitteeCommonCreateUpdateMixin, UpdateAction):
...
+                    instance["template_meeting_id"], instance["id"]
+                )
+            if instance.get("default_meeting_id"):
+                self.check_meeting_in_committee(
+                    instance["default_meeting_id"], instance["id"]
+                )
+            if "manager_ids" in instance:
+                new_manager_ids = instance.pop("manager_ids")
+                filter_ = FilterOperator(
+                    f"committee_${instance['id']}_management_level", "=", "can_manage"
```
jsangmeister, line 59: Please use the CML enum

openslides_backend/action/actions/committee/update.py
```diff
@@ -34,20 +38,56 @@ class CommitteeUpdateAction(CommitteeCommonCreateUpdateMixin, UpdateAction):
...
+                                "id": manager_id,
+                                "committee_$_management_level": {
+                                    str(
+                                        instance["id"]
+                                    ): CommitteeManagementLevel.CAN_MANAGE,
+                                },
+                            }
+                        )
+                for manager_id in old_manager_ids:
+                    if manager_id not in new_manager_ids:
```
jsangmeister, line 78: It is more efficient (and more readable, I think) if you use sets here

tests/system/action/committee/test_update.py
```diff
@@ -518,6 +518,50 @@ def test_update_wrong_id(self) -> None:
...
+    def test_update_manager_ids_not_in_committee(self) -> None:
+        self.create_data()
+        new_name = "committee_testname_updated"
+        response = self.request(
+            "committee.update",
+            {"id": self.COMMITTEE_ID, "name": new_name, "manager_ids": [1]},
+        )
+        self.assert_status_code(response, 400)
+        assert (
+            "You must add the user to the committee(s) '1', because you want to give him committee management level permissions."
```
jsangmeister, line 542: I would assume that we handle this case the same way as for `create`, i.e. adding the user to the committee instead of throwing an error. @tsiegleauq?

tests/system/action/committee/test_update.py
```diff
@@ -518,6 +518,50 @@ def test_update_wrong_id(self) -> None:
...
+        # important request.
+        response = self.request(
+            "committee.update",
+            {"id": self.COMMITTEE_ID, "name": "test", "manager_ids": [21]},
+        )
+        self.assert_status_code(response, 200)
+        self.assert_model_exists(
+            "user/21", {"committee_$1_management_level": "can_manage"}
+        )
+        self.assert_model_exists("user/20", {"committee_$_management_level": []})
```
jsangmeister, line 563: Also assert that the actual field (`committee_$1_management_level`) is set to `None`.

jsangmeister (changes requested):

jsangmeister:
Also, you should unify the code with the `create` case and extract it to a mixin (you can just provide an empty array as old manager ids for create and use the code from this PR)

openslides_backend/action/actions/committee/update.py
reiterl: If you got the payload `{"id": X, "manager_ids": [X, Y]}` and pop the manager_ids then the ds will return a error.

reiterl:
The create code is different vs the update code, because the create code also handles the committee_ids update. I don't think a mixin would help.

jsangmeister:
> The create code is different vs the update code, because the create code also handles the committee_ids update

See comment above: https://github.com/OpenSlides/openslides-backend/pull/866#discussion_r677186652

openslides_backend/action/actions/committee/update.py
reiterl: Use CML enum in filter,too.

openslides_backend/action/actions/committee/update.py
reiterl: I now use sets.

tests/system/action/committee/test_update.py
reiterl: Add that assertion.

reiterl:
Use the `update_models` method from the mixin for both _create_ and _update_.

tests/system/action/committee/test_update.py
jsangmeister: Implementing it this way to be consistent. Can be changed later if needed

jsangmeister (approved):

PR #874, opened 2021-07-29
Add weight to motion_submitter.

reiterl:
Use it in motion.create for setting the weight.

### DIFF
--- a/openslides_backend/action/actions/motion_submitter/create.py
+++ b/openslides_backend/action/actions/motion_submitter/create.py
@@ -57,7 +57,7 @@ class MotionSubmitterCreateAction(
         exists = self.datastore.exists(collection=self.model.collection, filter=filter)
         if exists:
             raise ActionException("(user_id, motion_id) must be unique.")
-        if instance.get("weight", 10000) == 10000:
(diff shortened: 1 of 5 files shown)


openslides_backend/action/actions/motion_submitter/create.py
```diff
@@ -53,4 +57,14 @@ def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
         exists = self.datastore.exists(collection=self.model.collection, filter=filter)
         if exists:
             raise ActionException("(user_id, motion_id) must be unique.")
+        if instance.get("weight", 10000) == 10000:
+            self.set_weight(instance)
```
jsangmeister, line 61: You can just remove the default value from the models file, since you will always either set one automatically or use the provided one

jsangmeister (changes requested):

openslides_backend/action/actions/motion_submitter/create.py
reiterl: Okay, updated the model and use a different check here.

jsangmeister (approved):

PR #877, opened 2021-07-29
Move user login for presenter

jsangmeister:


jsangmeister:
waiting for auth service update from @GabrielInTheWorld

tests/unit/test_presenter.py
```diff
@@ -16,12 +16,10 @@ def setUp(self) -> None:
         self.user_id = 0
 
     def test_with_bad_key(self) -> None:
```
FinnStutzenstein, line 18: Please add a test for two presenters with different auth modes

jsangmeister: done

GabrielInTheWorld:
> waiting for auth service update from @GabrielInTheWorld

@jsangmeister: I have updated the pip-library of the auth-service. So, the backend can validate an authentication only with a jwt-cookie for mediafiles.

jsangmeister:
Should be done now

FinnStutzenstein (approved):

PR #886, opened 2021-08-03
Update projector.toggle. Add test.

reiterl:
If unstable and not result, set scroll to 0 and move all current
unstable projections into history, too.

### DIFF
--- a/openslides_backend/action/actions/projector/toggle.py
+++ b/openslides_backend/action/actions/projector/toggle.py
@@ -53,6 +53,8 @@ class ProjectorToggle(UpdateAction):
             )
 
             for projector_id in instance["ids"]:
+                # Set scroll to 0 is needed, if unstable and no results
+                scroll_to_0 = False
                 stable = instance.get("stable", False)
                 filter_ = And(
                     FilterOperator("current_projector_id", "=", projector_id),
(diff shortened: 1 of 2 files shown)


openslides_backend/action/actions/projector/toggle.py
```diff
@@ -85,8 +87,14 @@ def get_updated_instances(self, action_data: ActionData) -> ActionData:
                         "options": instance.get("options"),
                         "meeting_id": meeting_id,
                     }
+                    if not stable:
+                        self.move_all_unstable_projections_to_history(
+                            projector_id, meeting_id
+                        )
+                        scroll_to_0 = True
```
FinnStutzenstein, line 94: Just curious: Why can't you yield the projector update here?

tests/system/action/projector/test_toggle.py
```diff
@@ -86,6 +86,27 @@ def test_correct_add_projection(self) -> None:
         projector = self.get_model("projector/23")
         assert projector.get("current_projection_ids") == [1]
 
+    def test_toggle_unstable_move_into_history(self) -> None:
+        self.setup_models(False)
+        self.set_models(
```
FinnStutzenstein, line 91: set the scoll to something different than 0, so you actually assert something below.

FinnStutzenstein (approved):
Small remarks. Works great otherwise!

tests/system/action/projector/test_toggle.py
reiterl: Set the scroll before the action to 100.

openslides_backend/action/actions/projector/toggle.py
reiterl: Ah, okay, I dont need that code, I can just use the yield here. Okay.

PR #888, opened 2021-08-04
Enhance models checking and generating

FinnStutzenstein:
- exclude calculated fields (https://github.com/OpenSlides/OpenSlides/blob/openslides4-dev/docs/models.yml#L2837)
- exclude the pseudocollection "_migration_index" (will be introduced soon)
- exclude the `restriction_mode` property from the models.


r-peschke (approved):
Finally it's not in this ticket, but I miss a description of restriction_mode in model.yml.

jsangmeister (approved):

PR #889, opened 2021-08-04
Add check_permissions to check_mediafile_id

reiterl:
Add some tests.

### DIFF
--- a/openslides_backend/presenter/check_mediafile_id.py
+++ b/openslides_backend/presenter/check_mediafile_id.py
@@ -1,9 +1,20 @@
-from typing import Any
+from typing import Any, Optional
 
 import fastjsonschema
 
 from ..models.models import Mediafile
-from ..shared.patterns import FullQualifiedId
+from ..permissions.management_levels import (
+    CommitteeManagementLevel,
+    OrganizationManagementLevel,
+)
+from ..permissions.permission_helper import (
+    has_committee_management_level,
+    has_organization_management_level,
+    has_perm,
+)
+from ..permissions.permissions import Permissions
+from ..shared.exceptions import PermissionDenied
+from ..shared.patterns import Collection, FullQualifiedId
 from ..shared.schema import required_id_schema, schema_version
 from .base import BasePresenter
 from .presenter import register_presenter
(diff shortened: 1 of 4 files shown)


openslides_backend/presenter/check_mediafile_id.py
```diff
@@ -37,7 +48,99 @@ def get_result(self) -> Any:
...
+        meeting = self.datastore.get(
+            FullQualifiedId(Collection("meeting"), meeting_id),
+            ["admin_group_id", "enable_anonymous", "user_ids", "committee_id"],
+        )
+        groups_field = f"group_${meeting_id}_ids"
+        user = self.datastore.get(
+            FullQualifiedId(Collection("user"), self.user_id), [groups_field]
+        )
+        if meeting.get("admin_group_id") in user.get(groups_field, []):
+            return
```
jsangmeister, line 86: Please move this to a helper function `is_admin` (together with the above check for superadmin) and use that here and [in this meeting.update test](https://github.com/OpenSlides/openslides-backend/blob/6db4f187b50e07900569a287b4ae541a0f6cf12c/openslides_backend/action/actions/meeting/update.py#L228) (and in any other places if you find them, on a quick search I did not find more)

openslides_backend/presenter/check_mediafile_id.py
```diff
@@ -37,7 +48,99 @@ def get_result(self) -> Any:
...
+            user_groups = set(user.get(f"group_${meeting_id}_ids", []))
+            if inherited_access_group_ids & user_groups:
+                return
+        raise PermissionDenied("You are not allowed to see this mediafile.")
+
+    def check_can_see_meeting(self, meeting: Dict[str, Any]) -> bool:
+        """needs meeting to include enable_anonymous, user_ids, committee_id."""
+        if has_organization_management_level(
+            self.datastore, self.user_id, OrganizationManagementLevel.SUPERADMIN
+        ):
```
jsangmeister, line 132: You can remove this check here since it was already checked earlier in the other function

jsangmeister (changes requested):
One test is missing for the last point: The user has mediafile.can_see and the user has groups in common with meeting/inherited_access_group_ids.

reiterl:
Rebased the branch to solve the conflict.

openslides_backend/presenter/check_mediafile_id.py
reiterl: Moved a is_admin function into the permission_helpers and use it at meeting.update.

openslides_backend/presenter/check_mediafile_id.py
reiterl: Removed.

reiterl:
Add a test for the inherited_access_group.

jsangmeister (approved):

PR #890, opened 2021-08-05
Generate models

FinnStutzenstein:


openslides_backend/models/models.py
```diff
@@ -407,7 +409,9 @@ class Meeting(Model):
     users_email_sender = fields.CharField(default="OpenSlides")
     users_email_replyto = fields.CharField()
     users_email_subject = fields.CharField(default="OpenSlides access data")
-    users_email_body = fields.CharField()
+    users_email_body = fields.CharField(
+        default="Dear {name},\n\n\nthis is your personal OpenSlides login:\n\n    {url}\n\n    username: {username}\n\n    password: {password}\n\n\n\nThis email was generated automatically."
```
jsangmeister, line 413: Interesting, the yaml docs stated that for multiline strings double linebreaks are necessary to represent a normal line break, but it seems like they are just copied as in the yml file. But this can be adjusted when we can send emails ;)

jsangmeister (approved):

Issue #883, opened 2021-07-30
orgaadmin can edit superadmin


emanuelschuetze:
Create an orgaadmin. 
Edit existing superadmin and change permission to "can_manage_organization".
Login as superadmin: You have no superadmin permissions and cannot change it back!

jsangmeister:
What should the intended behaviour be here? I would suggest that one can only downgrade another user in his OML if he would have permission to upgrade him again - meaning he needs at least the same OML. One can then still lock himself out, but that's another issue that we need to tackle.

emanuelschuetze:
> I would suggest that one can only downgrade another user in his OML if he would have permission to 
> upgrade him again - meaning he needs at least the same OML.

Ok for me.

jsangmeister:
Clarification: One should only be able to edit another user if he at least matches the user's OML. If no one has objections I will add this to the wiki.

jsangmeister:
Added to wiki: https://github.com/OpenSlides/OpenSlides/wiki/user.update#permissions

PR #892, opened 2021-08-05
Add a check for changing oml which are higher than changer got.

reiterl:
Add a test.

### DIFF
--- a/openslides_backend/action/actions/user/create_update_permissions_mixin.py
+++ b/openslides_backend/action/actions/user/create_update_permissions_mixin.py
@@ -201,7 +201,7 @@ class CreateUpdatePermissionsMixin(UserScopePermissionCheckMixin):
         actual_group_fields = self._get_actual_grouping_from_instance(instance)
 
         # Ordered by supposed velocity advantages. Changing order only can effect the sequence of detected errors for tests
-        self.check_group_E(permstore, actual_group_fields["E"], instance)
+        self.check_group_E(permstore, actual_group_fields["E"])
         self.check_group_D(permstore, actual_group_fields["D"], instance)
         self.check_group_C(permstore, actual_group_fields["C"], instance)
         self.check_group_B(permstore, actual_group_fields["B"], instance)
@@ -344,9 +344,7 @@ class CreateUpdatePermissionsMixin(UserScopePermissionCheckMixin):
                     }
                 )
 
-    def check_group_E(
-        self, permstore: PermissionVarStore, fields: List[str], instance: Dict[str, Any]
-    ) -> None:
+    def check_group_E(self, permstore: PermissionVarStore, fields: List[str]) -> None:
         """Check Group E organization_management_level: OML level necessary"""
         if fields and permstore.user_oml < OrganizationManagementLevel.CAN_MANAGE_USERS:
             raise MissingPermission(OrganizationManagementLevel.CAN_MANAGE_USERS)
@@ -369,7 +367,7 @@ class CreateUpdatePermissionsMixin(UserScopePermissionCheckMixin):
                 > permstore.user_oml
             ):
                 raise PermissionDenied(
-                    f"Your organization management level is not high enough to reset a Level of {user.get('organization_management_level')}!"
+                    f"Your organization management level is not high enough to change a user with a Level of {user.get('organization_management_level')}!"
                 )
 
     def _check_OML_in_instance(
--- a/openslides_backend/action/actions/user/create_update_permissions_mixin.py
+++ b/openslides_backend/action/actions/user/create_update_permissions_mixin.py
@@ -196,6 +196,7 @@ class CreateUpdatePermissionsMixin(UserScopePermissionCheckMixin):
         if permstore.user_oml == OrganizationManagementLevel.SUPERADMIN:
             return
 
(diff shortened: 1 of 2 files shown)


reiterl:
I prevent the change of the omls, if the changer of the oml is lower oml of the user.
So a superadmin cannot be downgraded by an admin.

openslides_backend/action/actions/user/create_update_permissions_mixin.py
```diff
@@ -343,10 +343,25 @@ def check_group_D(
...
+        """Check Group E organization_management_level: OML level necessary.
+        Also checks if user has higher oml."""
         if fields and permstore.user_oml < OrganizationManagementLevel.CAN_MANAGE_USERS:
             raise MissingPermission(OrganizationManagementLevel.CAN_MANAGE_USERS)
+        user = self.datastore.get(
+            FullQualifiedId(Collection("user"), instance["id"]),
+            ["organization_management_level"],
+        )
+        if (
+            fields
```
jsangmeister, line 358: The restriction should not only apply for changing the OML of another user, but for editing him altogether (that's what I tried to say with https://github.com/OpenSlides/openslides-backend/issues/883#issuecomment-890905647). So if I have OML can_manage_users and X has OML can_manage_organization, then I cannot call `user.update` for him, indipendent on the action data.

jsangmeister (changes requested):

openslides_backend/action/actions/user/create_update_permissions_mixin.py
reiterl: So it should be a more general check? Okay.

reiterl: Worked on this.
The permission to change a user with higher oml is fully restricted.

reiterl: I also fixed the test problems.

openslides_backend/action/actions/user/create_update_permissions_mixin.py
```diff
@@ -346,28 +347,31 @@ def check_group_D(
...
 
     def check_group_F(self, fields: List[str]) -> None:
         """Group F: OML SUPERADMIN necessary, which is checked before"""
         if fields:
             raise MissingPermission(OrganizationManagementLevel.SUPERADMIN)
 
+    def _check_for_higher_OML(
+        self, permstore: PermissionVarStore, instance: Dict[str, Any]
+    ) -> None:
+        if "id" in instance:
```
jsangmeister, line 362: You can remove this check, id must be in instance

openslides_backend/action/actions/user/create_update_permissions_mixin.py
```diff
@@ -346,28 +347,31 @@ def check_group_D(
...
+            user = self.datastore.get(
+                FullQualifiedId(Collection("user"), instance["id"]),
+                ["organization_management_level"],
+            )
+            if (
+                OrganizationManagementLevel(user.get("organization_management_level"))
+                > permstore.user_oml
+            ):
+                raise PermissionDenied(
+                    f"Your organization management level is not high enough to reset a Level of {user.get('organization_management_level')}!"
```
jsangmeister, line 372: error message does not match the error

openslides_backend/action/actions/user/create_update_permissions_mixin.py
```diff
@@ -196,11 +196,12 @@ def check_permissions(self, instance: Dict[str, Any]) -> None:
         if permstore.user_oml == OrganizationManagementLevel.SUPERADMIN:
             return
 
+        self._check_for_higher_OML(permstore, instance)
         self._check_OML_in_instance(permstore, instance)
         actual_group_fields = self._get_actual_grouping_from_instance(instance)
 
         # Ordered by supposed velocity advantages. Changing order only can effect the sequence of detected errors for tests
-        self.check_group_E(permstore, actual_group_fields["E"])
+        self.check_group_E(permstore, actual_group_fields["E"], instance)
```
jsangmeister, line 204: Please revert the change to the params since it is no longer needed

jsangmeister (changes requested):
Also add a test which tries to add a basic property from group A, but fails because the OML is higher than the one from the request user.

openslides_backend/action/actions/user/create_update_permissions_mixin.py
reiterl: Update to f`"Your organization management level is not high enough to change a user with a Level of {user.get('organization_management_level')}!"`

openslides_backend/action/actions/user/create_update_permissions_mixin.py
reiterl: Done.

openslides_backend/action/actions/user/create_update_permissions_mixin.py
reiterl: No, I need this check, see the rest of the code. There it is always checked and the test_create.py failed, if it is missing.

jsangmeister: ok

jsangmeister (approved):

PR #897, opened 2021-08-09
Assert migration index at startup

FinnStutzenstein:
Detect misconfigured setups and prevent wrong data usage.


dev/Dockerfile.dev
```diff
@@ -19,7 +19,7 @@ COPY Makefile Makefile
 COPY setup.cfg .
 COPY dev/entrypoint.sh ./
 
-COPY migrations/migrate.py migrations/.
+COPY migrations/*.py migrations/
 COPY migrations/migrations migrations/migrations/.
```
jsangmeister, line 23: Wouldn't it be better to mount these folders?

openslides_backend/models/assert_migration_index.py
```diff
@@ -0,0 +1,37 @@
+from datastore.shared.di import injector
```
jsangmeister, line 1: I don't think this belongs in the `models` folder, but rather in the `migrations` folder

openslides_backend/models/__init__.py
```diff
@@ -0,0 +1 @@
+from .assert_migration_index import assert_migration_index  # noqa
```
jsangmeister, line 1: In the backend, we do not use the `__init__.py` for such exports, so I don't think it's fitting to use it here in this way

jsangmeister (changes requested):

dev/Dockerfile.dev
FinnStutzenstein: I thought about it, but when *developing* migrations, there is a dedicated setup in the migrations folder. Until you want to change the intersection (load_modules), you have to rebuild/test-edit with vim. I had to rebuild 3 times, which is a low quote I guess to make it work. I think there is currently no need

openslides_backend/models/assert_migration_index.py
FinnStutzenstein: ok, I'll move it.

dev/Dockerfile.dev
jsangmeister: I found that mounting is almost always better than copying because it eliminates one source of programming mistakes and just makes the experience a lot smoother if you need it. But I see your point

FinnStutzenstein:
@jsangmeister The action "Build and test production docker image" does not work as is:
```
psycopg2.OperationalError: could not translate host name "postgresql" to address: No address associated with hostname
```

Which is sensible since on startup a DB query is made. I can either remove this action or use a docker-compose setup. Whats your opinion?

jsangmeister:
> @jsangmeister The action "Build and test production docker image" does not work as is:
> 
> ```
> psycopg2.OperationalError: could not translate host name "postgresql" to address: No address associated with hostname
> ```
> 
> Which is sensible since on startup a DB query is made. I can either remove this action or use a docker-compose setup. Whats your opinion?

I would prefer the latter, it's better to keep tests if possible

dev/Dockerfile.dev
FinnStutzenstein: I mounted it now ;)

dev/Dockerfile.dev
```diff
@@ -10,7 +10,7 @@ COPY dev/requirements_development.txt dev/.
 
 RUN pip install --no-cache-dir --requirement dev/requirements_development.txt
 
-COPY dev/wait.sh dev/.
+COPY wait.sh ./
```
jsangmeister, line 13: Why did you move the file? I think it belongs in `/dev/`

FinnStutzenstein: It is not used in the main Dockerfile

jsangmeister (approved):

PR #898, opened 2021-08-10
History information presenter

FinnStutzenstein:
First work. Things to improve later on (on demand when new requirements come in): Permissions and displaying of the user.

### DIFF
--- a/openslides_backend/presenter/get_history_information.py
+++ b/openslides_backend/presenter/get_history_information.py
@@ -1,0 +1,71 @@
+from typing import Any, Dict, Set
+
+import fastjsonschema
+
+from ..permissions.management_levels import OrganizationManagementLevel
+from ..permissions.permission_helper import has_organization_management_level
+from ..services.datastore.commands import GetManyRequest
+from ..shared.exceptions import PermissionDenied
+from ..shared.patterns import Collection
+from ..shared.schema import required_fqid_schema, schema_version
+from .base import BasePresenter
+from .presenter import register_presenter
+
+get_history_information_schema = fastjsonschema.compile(
+    {
+        "$schema": schema_version,
+        "type": "object",
+        "title": "get_history_information data",
+        "description": "Schema to validate the get_history_information presenter data.",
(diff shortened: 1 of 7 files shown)


openslides_backend/presenter/get_history_information.py
```diff
@@ -0,0 +1,71 @@
...
+    Return all history information for one fqid.
+    """
+
+    schema = get_history_information_schema
+
+    def get_result(self) -> Any:
+        if not has_organization_management_level(
+            self.datastore, self.user_id, OrganizationManagementLevel.SUPERADMIN
+        ):
+            raise PermissionDenied("You are not a superadmin")
```
jsangmeister, line 39: This reduces the feature set from OS3 since only we are superadmins, so the organization admins can no longer look at the history or restore it. Is this intended?

tests/system/presenter/base.py
```diff
@@ -11,7 +11,7 @@ def get_application(self) -> WSGIApplication:
 
     def request(
         self, presenter: str, data: Optional[Dict[str, Any]] = None
-    ) -> Tuple[int, Dict[str, Any]]:
+    ) -> Tuple[int, Any]:
```
jsangmeister, line 14: Why did you change this?

FinnStutzenstein: The presenter returns a list instead of an object.

openslides_backend/presenter/get_history_information.py
FinnStutzenstein: This is right.There is the meeting perm `can_see_history` but I'm unsure if this is a good permission:

IMO it should be coupled to the admin of the meeting. In OS3 you can see the history list with this permission, but can not enter history mode (you need the OS3 superadmin for it). I think it was the case since we could not restrict the histoy data, so the superadmin was needed. We need a permission concept for the history list and retrieving history data. Until then, I'll leave it for the superadmin. I think how it work has to be discussed with Emanuel.

PR #901, opened 2021-08-11
First migration for archived meetings

FinnStutzenstein:


FinnStutzenstein:
Waiting for https://github.com/OpenSlides/OpenSlides/pull/6205

.github/startup_test/docker-compose.yml
```diff
@@ -13,6 +13,11 @@ services:
             - DATASTORE_WRITER_PORT=9011
         depends_on:
             - writer
+            - reader
+        # migration at startup - must be done differently in a real prod setup
```
FinnStutzenstein, line 17: TODO: remove

.github/startup_test/migration-entrypoint.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
```
FinnStutzenstein, line 1: TODO: remove

migrations/setup.cfg
```diff
@@ -0,0 +1,60 @@
...
+
+[coverage:report]
+exclude_lines =
+    pragma: no cover
+    raise NotImplementedError
+    return NotImplemented
+fail_under = 96
+omit = tests/*
+
+[coverage:html]
```
FinnStutzenstein, line 26: TODO: truncate

r-peschke (commented):
LGTM

migrations/fetch-example-data.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
```
FinnStutzenstein, line 1: TODO: Add to readme

migrations/migrations/0001_archive_meetings.py
```diff
@@ -0,0 +1,113 @@
+from typing import List, Optional, Set
```
FinnStutzenstein, line 1: TODO: Split migration

migrations/migrations/0001_archive_meetings.py
```diff
@@ -0,0 +1,113 @@
...
+        if self.meeting_ids_to_remove:
+            events.append(
+                ListUpdateEvent(
+                    "organization/1",
+                    {
+                        "remove": {
+                            "active_meeting_ids": list(self.meeting_ids_to_remove)
+                        }
+                    },
+                )
```
jsangmeister, line 111: You can merge this into one `ListUpdateEvent`, if I'm not mistaken

migrations/migrations/0001_archive_meetings.py
jsangmeister: While doing that, I would be in favor to start creating base classes for easy migrations (in this case, adding a field with a default value, namely `limit_of_meetings`). This case will probably occur often and it would be nice to just have an `AddFieldMigration` from which you can inherit and define the attribute for the collection field and the default value.

migrations/migrations/0001_archive_meetings.py
```diff
@@ -0,0 +1,113 @@
...
+        # `organization/active_meeting_ids` in this position.
+        self.meeting_ids_to_add: Set[int] = set()
+        self.meeting_ids_to_remove: Set[int] = set()
+
+    def migrate_event(
+        self,
+        event: BaseEvent,
+        old_accessor: MigrationKeyframeAccessor,
+        new_accessor: MigrationKeyframeAccessor,
+        position_data: PositionData,
```
jsangmeister, line 66: Just saw https://github.com/OpenSlides/openslides-datastore-service/pull/146, is the datastore not up-to-date in the backend?

r-peschke (approved):

migrations/migrations/0001_archive_meetings.py
FinnStutzenstein: Ok, I adjusted mypy a bit. Now, it is checked with the datastore together.

```
Argument 2 to "ListUpdateEvent" has incompatible type "Dict[str, Dict[str, List[int]]]"; expected "Dict[str, Dict[Union[str, _Field], List[Union[str, int]]]]"
```
IDK whats wrong here, I'll use `Any` (what I know is that I do not like mypy :D)

FinnStutzenstein:
I do not have the time to finish it today. I'll continue tomorrow morning

migrations/migrations/0001_archive_meetings.py
jsangmeister: Yeah, mypy seems to have problems with nested unions sometimes... Did you try `Dict[str, ListUpdatesDict]`, like defined in the datastore?

jsangmeister (approved):

setup.cfg
```diff
@@ -11,6 +11,9 @@ extend-ignore = E203,E501
 [mypy]
 disallow_untyped_defs = true
 
+[tool:pytest]
+testpaths = tests/
```
FinnStutzenstein, line 15: It also wants to execute `migrations/tests` without this line

PR #908, opened 2021-08-17
assert_model_deleted with optional field checking

r-peschke:
Allows field checking analogue to assert_model_exist for deleted objects for help in testing relations


FinnStutzenstein:
Already implemented in #901

r-peschke:
Already implemented in #901

PR #915, opened 2021-08-20
Add migrations to CI

jsangmeister:
Notes:
- The checks are all in the same job and prefixed with `always()` because I did not want to rebuild the container for every job. This leads to unwanted executions in case one of the first two build/install commands fails, but this is unintended behaviour then, anyway.
- The commit hash of the datastore is now fixed in two separate points, which I don't like. Alternatives would include just installing all backend requirements (also not nice) or creating a separate requirements file where only the datastore is listed, which can then be included in the other requirements.
- mypy does not seem to check the typing of installed pip packages, even though I removed the flags in the `setup.cfg` (see changes in the second migration). @FinnStutzenstein do you have any idea why?


migrations/migrations/0002_archive_meetings.py
```diff
@@ -56,6 +56,7 @@ def position_init(self) -> None:
     def migrate_event(
         self,
         event: BaseEvent,
+        adsany: str,
```
FinnStutzenstein, line 59: If I go into the dev shell (`make run-bash`) in the migration folder I get this result:
```
finn@4721ece4ff20:/app$ mypy .
Success: no issues found in 5 source files
finn@4721ece4ff20:/app$ mypy .
migrations/0002_archive_meetings.py:56: error: Signature of "migrate_event" incompatible with supertype "BaseMigration"
Found 1 error in 1 file (checked 5 source files)
```

(first call without this line, second call with this change). Can you reproduce this in the dev shell?

FinnStutzenstein:
> creating a separate requirements file where only the datastore is listed, which can then be included in the other requirements.

My favorite.

jsangmeister:
This should finally be done now (it's been only 3.5 months after all). I ran into some compatibility issues with the datastore filters which I fixed in https://github.com/OpenSlides/openslides-datastore-service/pull/167. Also, since we never type checked the backend together with the datastore, I had to adjust some typings. But now, everything should be working smoothly together with the migrations.

r-peschke (approved):

PR #920, opened 2021-09-01
OS4-dev to master

tsiegleauq:
Change os4-dev ref to master


jsangmeister (approved):
You'll have to execute black for correct formatting, although I have no idea why the longer string was okay and the shorter should be broken in multiple lines...

PR #922, opened 2021-09-06
Fix problem with organization_tag.

reiterl:


reiterl:
At the moment, meeting/organization_tag_ids is not needed, so we can use this solution. In future, perhaps we need to handle organization tags like the users.

jsangmeister:
I would prefer handling it similar to the users: Leaving `meeting/organization_tag_ids` as-is and adding the new meeting id to `organization_tag/meeting_ids`. It's the way a user would expect it to work if he clones a meeting.

> At the moment, meeting/organization_tag_ids is not needed

What do you mean by this? Organization tags are used in the client already, aren't they?

jsangmeister (approved):

PR #923, opened 2021-09-06
Add the user.toggle_presence_by_number action.

reiterl:
Add some tests.
In the issue the action was renamed to toggle_presence_by_number.
The action is near the set_present action.

### DIFF
--- a/openslides_backend/action/actions/user/toggle_presence_by_number.py
+++ b/openslides_backend/action/actions/user/toggle_presence_by_number.py
@@ -75,7 +75,9 @@ class UserTogglePresenceByNumber(UpdateAction):
         result = self.datastore.filter(Collection("user"), filter_, ["id"])
         if len(result.keys()) == 1:
             return list(result.keys())[0]
-        raise ActionException("Could not find only one user with the number.")
+        elif len(result.keys()) > 1:
+            raise ActionException("Found more than one user with the default number.")
+        raise ActionException("No user with this number found.")
 
     def check_permissions(self, instance: Dict[str, Any]) -> None:
         if has_organization_management_level(
--- a/openslides_backend/action/actions/user/toggle_presence_by_number.py
+++ b/openslides_backend/action/actions/user/toggle_presence_by_number.py
@@ -1,0 +1,103 @@
+from typing import Any, Dict
+
+from ....models.models import User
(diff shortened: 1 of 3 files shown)


openslides_backend/action/actions/user/toggle_presence_by_number.py
```diff
@@ -0,0 +1,103 @@
...
+            raise ActionException("Found more than one user with the number.")
+
+        filter_ = And(
+            FilterOperator(f"number_${meeting_id}", "=", ""),
+            FilterOperator("default_number", "=", number),
+        )
+        result = self.datastore.filter(Collection("user"), filter_, ["id"])
+        if len(result.keys()) == 1:
+            return list(result.keys())[0]
+        raise ActionException("Could not find only one user with the number.")
```
jsangmeister, line 78: Differentiate here between the cases "No user found" and "Too many users found" for better error messages

tests/system/action/user/test_toggle_presence_by_number.py
```diff
@@ -0,0 +1,125 @@
+from openslides_backend.permissions.management_levels import (
+    CommitteeManagementLevel,
+    OrganizationManagementLevel,
+)
+from openslides_backend.permissions.permissions import Permissions
+from tests.system.action.base import BaseActionTestCase
+
+
+class UserSetPresentActionTest(BaseActionTestCase):
```
jsangmeister, line 9: Please also test the failure cases (too many/no users with a certain number, too many/no users with a certain default number) and edge cases (if a user with a certain number exist and another user has it as the default number, only the first one is toggled; if a user has a normal number and a default number, the default number is ignored, even if it matches)

jsangmeister (changes requested):

openslides_backend/action/actions/user/toggle_presence_by_number.py
reiterl: Diff the two cases.

tests/system/action/user/test_toggle_presence_by_number.py
reiterl: Add some more tests. 3 failure tests and 2 edge tests.

jsangmeister (approved):

PR #933, opened 2021-09-13
Issue899 archive meetings

r-peschke:
Action organization.update with limit_of_meetings
Action meeting.create: Check limit_of_meetings,  set is_active_in_organization_id to the current organization's id per default
Action meeting.archive
Action meeting.unarchiving
Importing and cloning have to respect the limit_of_meetings

### DIFF
--- a/cli/check_json.py
+++ b/cli/check_json.py
@@ -11,11 +11,19 @@ def main() -> int:
         print("No files specified.")
         return 1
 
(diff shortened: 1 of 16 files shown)


reiterl (approved):
Looks good to me. I have found the modes code. You use Py 3.8 features (":="), but it is okay.

PR #934, opened 2021-09-14
Improve the committee delete error message.

reiterl:


r-peschke (approved):
I would prefer self.AssertIn and the model_check before the message check, but if you like it more your way, it's okay for me;-)

PR #940, opened 2021-09-17
Issue939 Complex motion workflow

r-peschke:


reiterl (approved):
I don't understand the new workflow and how it is linked, but the python code looks okay.

migrations/migrations/0003_motion_state_dont_set_identifier.py
```diff
@@ -0,0 +1,14 @@
...
+    """
+    This migration adds `motion_state/dont_set_identifier` with a
+    default of False to each motion_state.
+    """
+
+    target_migration_index = 4
+
+    collection = "motion_state"
+    field = "dont_set_identifier"
+    default = False
```
jsangmeister, line 14: Migration would not have been necessary here, since the default is `False`, but it doesn't hurt

PR #949, opened 2021-09-23
user.scope and archived meetings

r-peschke:


openslides_backend/action/actions/user/user_scope_permission_check_mixin.py
```diff
@@ -76,6 +76,7 @@ def get_user_scope(
         Returns the scope of the given user id together with the relevant scope id (either meeting, committee or organization).
         """
         meetings: List[int] = []
+        meetingsd: Dict[int, int]
```
jsangmeister, line 79: please use a better name

openslides_backend/action/actions/user/user_scope_permission_check_mixin.py
```diff
@@ -88,17 +89,27 @@ def get_user_scope(
...
+                GetManyRequest(
+                    Collection("meeting"),
+                    meetings,
+                    ["committee_id", "is_active_in_organization_id"],
+                )
+            ]
+        ).get(Collection("meeting"), {})
+        meetingsd = {
+            meeting_id: odict.get("committee_id")  # type: ignore
+            for meeting_id, odict in result.items()
```
jsangmeister, line 103: same for `odict`

openslides_backend/action/actions/user/user_scope_permission_check_mixin.py
jsangmeister: Also, is the declaration here necessary?

r-peschke: > please use a better name

done, not quite sure it is better

r-peschke: > Also, is the declaration here necessary?

no, removed

openslides_backend/action/actions/user/user_scope_permission_check_mixin.py
r-peschke: done

PR #954, opened 2021-09-28
Update clone meeting action to use duplicate mediafile.

reiterl:
Needs the change of the media service which includes the duplicate_mediafile


GabrielInTheWorld (changes requested):
This fix still throws this error:
```
backend_1           | DEBUG:openslides_backend.shared.handlers.base_handler:Error occured on index 0: 	mediafile/6/create_timestamp: Type error: Type is not <openslides_backend.models.fields.TimestampField object at 0x7f4dbd919130>
```

reiterl:
Worked on the timestamp problem. Timestamp should be int not float.

GabrielInTheWorld (approved):
Works

Issue #557, opened 2021-03-18
Additional tests for relation consistency


jsangmeister:
To prevent "relation leaking" (one end of a relation is updated, but the other not, leading to inconsistent data and potentially fatal errors) like we've had before, it would be good to implement additional tests. One idea would be to automatically check after each test that the datastore content is consistent and contains no such leaks. We would only check existing relations and not all that are required since the test only create a partial data set. The question is how fast these tests would be and if we want to execute them together with the other tests or seperately.

jsangmeister:
I don't think this is necessary any more.

PR #956, opened 2021-09-29
Add theme actions and update tests.

reiterl:
Need the update of the model and the initial/example data.


reiterl:
Also add a migration to remove the organization theme field.
The other fields are added or optional. As the new collection theme is added, the required fields in theme don't need to be updated.

openslides_backend/models/models.py
```diff
@@ -202,6 +205,60 @@ class OrganizationTag(Model):
...
+    warn_800 = fields.ColorField()
+    warn_900 = fields.ColorField()
+    warn_A100 = fields.ColorField()
+    warn_A200 = fields.ColorField()
+    warn_A400 = fields.ColorField()
+    warn_A700 = fields.ColorField()
+    organization_ids = fields.RelationListField(
+        to={Collection("organization"): "default_theme_id"}
+    )
+    user_ids = fields.RelationListField(to={Collection("user"): "theme_id"})
```
GabrielInTheWorld, line 259: Remove, since we don't want to implement a relation between user <-> theme, yet.

tests/system/action/meeting/test_delete.py
```diff
@@ -47,6 +49,7 @@ def test_delete_permissions_can_manage_committee(self) -> None:
         self.assert_status_code(response, 200)
         self.assert_model_deleted("meeting/1")
 
+    @pytest.mark.skip()
```
GabrielInTheWorld, line 52: Here are two options (just for remembering):

1. Delete this line, when the example-data in the main-repo (OpenSlides/OpenSlides) is up-to-date
2. Extract this test into an own and mark it as `optional`, so a PR can be merged, even when that test fails

openslides_backend/action/actions/theme/update.py
```diff
@@ -0,0 +1,62 @@
...
+            "warn_300",
+            "warn_400",
+            "warn_600",
+            "warn_700",
+            "warn_800",
+            "warn_900",
+            "warn_A100",
+            "warn_A200",
+            "warn_A400",
+            "warn_A700",
```
jsangmeister, line 59: Please deduplicate with the create case

openslides_backend/models/models.py
```diff
@@ -4,7 +4,7 @@
 from openslides_backend.models.base import Model
 from openslides_backend.shared.patterns import Collection
 
-MODELS_YML_CHECKSUM = "83dd0d1d74658e869a797bfebfd0fcc6"
+MODELS_YML_CHECKSUM = "fd53765503552b0562bd71ef2ff437d3"
```
jsangmeister, line 7: Did you generate this from a local models.yml? Please create a PR for it

openslides_backend/models/models.py
```diff
@@ -202,6 +204,59 @@ class OrganizationTag(Model):
...
+    warn_500 = fields.ColorField(required=True)
+    warn_600 = fields.ColorField()
+    warn_700 = fields.ColorField()
+    warn_800 = fields.ColorField()
+    warn_900 = fields.ColorField()
+    warn_A100 = fields.ColorField()
+    warn_A200 = fields.ColorField()
+    warn_A400 = fields.ColorField()
+    warn_A700 = fields.ColorField()
+    organization_ids = fields.RelationListField(
```
jsangmeister, line 255: Why is this a list field? I think a simple `RelationField` suffices here, since there can only ever be one organisation

openslides_backend/models/models.py
```diff
@@ -4,7 +4,7 @@
 from openslides_backend.models.base import Model
 from openslides_backend.shared.patterns import Collection
 
-MODELS_YML_CHECKSUM = "83dd0d1d74658e869a797bfebfd0fcc6"
+MODELS_YML_CHECKSUM = "fd53765503552b0562bd71ef2ff437d3"
 
 
 class Organization(Model):
```
jsangmeister, line 10: Please add `theme_ids`, since the themes need a link to the organisation or otherwise they are "lost" (not discoverable)

tests/system/action/meeting/test_delete.py
jsangmeister: I'm strongly against merging PRs with failing tests, that is the whole meaning why we do the tests.

jsangmeister (changes requested):

openslides_backend/models/models.py
reiterl: Done.

openslides_backend/models/models.py
reiterl: The pull request is on the way.

openslides_backend/action/actions/theme/update.py
reiterl: Add `THEME_REQ_FIELDS` and `THEME_OPT_FIELDS` for use in both places.

openslides_backend/models/models.py
reiterl: Added `theme_ids`

migrations/migrations/0004_remove_organization_theme.py
```diff
@@ -0,0 +1,33 @@
+from typing import Any, Dict, List, Optional
+
+from datastore.migrations.core.base_migration import BaseMigration
+from datastore.migrations.core.events import BaseEvent, CreateEvent, UpdateEvent
+from datastore.shared.util import collection_from_fqid
+
+
+class Migration(BaseMigration):
```
jsangmeister, line 8: A test for this migration is still missing

migrations/tests/test_0006_add_required_theme_id.py
```diff
@@ -0,0 +1,53 @@
...
+    assert_model(
+        "theme/1",
+        {
+            "id": 1,
+            "name": "OpenSlides Blue",
+            "accent_500": "#2196f3",
+            "primary_500": "#317796",
+            "warn_500": "#f06400",
+            "meta_deleted": False,
+            "meta_position": 1,
```
jsangmeister, line 35: please also check for the organisation id here - I think it has to be added in the migration

jsangmeister (changes requested):
Rest looks good, should be mergeable after this fix. However, it would be fatal if such a bug in a migration would be deployed, since it could potentially crash the whole datastore. Maybe we should consider implementing something like https://github.com/OpenSlides/openslides-backend/issues/557, but only for the migration tests, since they don't have any automatic relation handling and a programmer can potentially write anything in there, (currently) without any checks. We should discuss this in person I think.

tests/system/action/theme/test_delete.py
```diff
@@ -0,0 +1,34 @@
+from tests.system.action.base import BaseActionTestCase
+
+
+class ThemeDeleteActionTest(BaseActionTestCase):
```
r-peschke, line 4: Could you please add a test, showing that you can't delete the default theme of the organization

tests/system/action/organization/test_update.py
```diff
@@ -20,6 +20,7 @@ def test_update_some_more_fields(self) -> None:
         self.create_model(
```
r-peschke, line 20: Please write tests for filling/changing theme_id and theme_ids. Show that it is not possible to remove the default theme from theme_id and theme_ids

openslides_backend/action/actions/organization/update.py
```diff
@@ -24,7 +24,7 @@ class OrganizationUpdate(UpdateAction):
             "legal_notice",
             "privacy_policy",
             "login_text",
-            "theme",
+            "theme_id",
```
r-peschke, line 27: Please check that the theme_id is one of the theme_ids

tests/system/action/organization/test_update.py
reiterl: Changed.

r-peschke (approved):
Seams okay for me.
Maybe we will revert the **required = True** cases, otherwise we should write migrations for them.

PR #959, opened 2021-10-01
Fix problem with agenda_create in datastore.

reiterl:


GabrielInTheWorld (approved):
This PR fixes the previously error (`invalid fields: agenda_create`)

PR #961, opened 2021-10-01
Fix problem with poll create. Options shouldn't be in the instance.

reiterl:


GabrielInTheWorld (approved):
Fixes the error `invalid fields options`

PR #967, opened 2021-10-06
Add weight to motion state

jsangmeister:


jsangmeister:
needs https://github.com/OpenSlides/openslides-datastore-service/pull/154

jsangmeister:
ready for review

reiterl (commented):
* What does happen with the `weight`, if two states are created in one action?

jsangmeister:
> * What does happen with the `weight`, if two states are created in one action?

They will get the same weight, since the datastore state is not updated in between. This should however not be a problem for multiple reasons:
 - if the client still sorts by id second, the results are still ordered correctly
 - AFAIK, the client never creates multiple states in one request
 - If he really wants to do that, he could use non-atomic mode and send the action multiple times

Of course, when calling the action internally (e.g. from `meeting.create`), one should always give an explicit weight since else every state will get the same weight. I missed this, thank you for the hint

PR #989, opened 2021-10-18
Issue978 Catch internal smtplib error

r-peschke:


openslides_backend/action/actions/user/send_invitation_email.py
```diff
@@ -99,6 +99,9 @@ def perform(
         except SSLCertVerificationError as e:
             result = {"sent": False, "message": f"SSLCertVerificationError: {str(e)}"}
             self.results.append(result)
+        except Exception as e:
```
reiterl, line 102: That is not a good idea. You should specify which exception types you are looking for.

r-peschke: There is no concrete error. The thrown error was thrown inside the python smtplib and is not document. I wrote this exception-catch for all to not end with the "undefined Server Error" in the context of mail connection.
The relevant and known exceptions are all caught and covered by tests.

PR #992, opened 2021-10-20
Add the action organization initial_import and tests

reiterl:


openslides_backend/action/actions/organization/initial_import.py
```diff
@@ -0,0 +1,104 @@
...
+from ....shared.interfaces.write_request import WriteRequest
+from ....shared.patterns import Collection, FullQualifiedId
+from ...action import Action
+from ...mixins.singular_action_mixin import SingularActionMixin
+from ...util.default_schema import DefaultSchema
+from ...util.register import register_action
+from ...util.typing import ActionData, ActionResultElement, ActionResults
+
+
+@register_action("organization.initial_import")
```
jsangmeister, line 19: You can for now just mark it as "internal", in #921 I will implement internal routes for the docker stack.

openslides_backend/action/actions/organization/initial_import.py
```diff
@@ -0,0 +1,104 @@
...
+        try:
+            checker.run_check()
+        except CheckException as ce:
+            raise ActionException(str(ce))
+
+        return instance
+
+    def check_empty_datastore(self) -> None:
+        filter_ = FilterOperator("id", ">=", 1)
+        check_collections = ["user", "organization", "committee", "meeting"]
```
jsangmeister, line 72: Just checking for `organization` should be enough: if no organization exists, there can be no other models.

openslides_backend/action/actions/organization/initial_import.py
```diff
@@ -0,0 +1,104 @@
...
+                        "initial import",
+                        entry,
+                    )
+                )
+        return write_requests
+
+    def create_action_result_element(
+        self, instance: Dict[str, Any]
+    ) -> Optional[ActionResultElement]:
+        return {"id": 1}
```
jsangmeister, line 101: I don't think a result is necessary here since it's alway 1. You can just delete this method and return [None] above as results

openslides_backend/action/actions/organization/initial_import.py
```diff
@@ -0,0 +1,104 @@
...
+
+    def perform(
+        self, action_data: ActionData, user_id: int, internal: bool = False
+    ) -> Tuple[Optional[WriteRequest], Optional[ActionResults]]:
+        """
+        Simplified entrypoint to perform the action.
+        """
+        self.user_id = user_id
+        self.index = 0
+        action_data = self.get_updated_instances(action_data)
```
jsangmeister, line 41: remove

openslides_backend/action/actions/organization/initial_import.py
```diff
@@ -0,0 +1,104 @@
...
+        instance = next(iter(action_data))
+        self.validate_instance(instance)
+        try:
+            self.check_permissions(instance)
+        except MissingPermission as e:
+            msg = f"You are not allowed to perform action {self.name}."
+            e.message = msg + " " + e.message
+            raise e
+        instance = self.base_update_instance(instance)
```
jsangmeister, line 50:
```suggestion
        instance = self.update_instance(instance)
```

openslides_backend/action/actions/organization/initial_import.py
```diff
@@ -0,0 +1,104 @@
...
+        self.index = 0
+        action_data = self.get_updated_instances(action_data)
+        instance = next(iter(action_data))
+        self.validate_instance(instance)
+        try:
+            self.check_permissions(instance)
+        except MissingPermission as e:
+            msg = f"You are not allowed to perform action {self.name}."
+            e.message = msg + " " + e.message
+            raise e
```
jsangmeister, line 49: remove, since `check_permissions` does nothing anyway

openslides_backend/action/actions/organization/initial_import.py
```diff
@@ -0,0 +1,104 @@
...
+                )
+        return write_requests
+
+    def create_action_result_element(
+        self, instance: Dict[str, Any]
+    ) -> Optional[ActionResultElement]:
+        return {"id": 1}
+
+    def check_permissions(self, instance: Dict[str, Any]) -> None:
+        return
```
jsangmeister, line 104: remove

openslides_backend/models/checker.py
```diff
@@ -180,11 +180,16 @@ def __init__(
         self.models: Dict[str, Type["Model"]] = {
             collection.collection: model_registry[collection]
             for collection in model_registry
+            if not collection.collection.startswith("fake_model")
```
jsangmeister, line 183: We should not add test code to the "real" code. What is the exact issue why you think you need this? If it's really necessary, you should change the tests where the fake models are added so that they are removed again afterwards.

tests/system/action/organization/test_initial_import.py
```diff
@@ -0,0 +1,531 @@
...
+                        "warn_400": None,
+                        "warn_50": None,
+                        "warn_600": None,
+                        "warn_700": None,
+                        "warn_800": None,
+                        "warn_900": None,
+                        "warn_a100": None,
+                        "warn_a200": None,
+                        "warn_a400": None,
+                        "warn_a700": None,
```
jsangmeister, line 76: I think all not-required fields should be left out. This is just a huge verbose block which contributes nothing to the actual tests. You probably have to adjust the checker for this.

jsangmeister: same for all other empty (`None`/`[]`) fields

tests/system/action/organization/test_initial_import.py
```diff
@@ -0,0 +1,531 @@
...
+            "projector_message",
+            "projector_countdown",
+            "chat_group",
+            "organization_tag",
+            "resource",
+            "theme",
+        )
+        for collection in needed_collections:
+            if collection not in data["data"].keys():
+                data["data"][collection] = {}
```
jsangmeister, line 404: same goes for empty collections, they should just be left empty

tests/system/action/organization/test_initial_import.py
```diff
@@ -0,0 +1,531 @@
+from typing import Any, Dict
+
+from tests.system.action.base import BaseActionTestCase
+
+
+class OrganizationInitialImport(BaseActionTestCase):
+    def create_request_data(self, datapart: Dict[str, Any]) -> Dict[str, Any]:
+        data: Dict[str, Any] = {
```
jsangmeister, line 8: I think it would be better to fetch the initial data here (once, for all tests), analogous to the example data we use elsewhere from a fixed commit. This would make the tests way less verbose.

tests/system/action/organization/test_initial_import.py
```diff
@@ -0,0 +1,531 @@
...
+            "personal_note_ids": [],
+            **data,
+        }
+
+    def test_initial_import_simple(self) -> None:
+        self.datastore.truncate_db()
+        request_data = self.create_request_data({})
+        response = self.request("organization.initial_import", request_data)
+        self.assert_status_code(response, 200)
+        self.assert_model_exists(
```
jsangmeister, line 506: check for all models from the request data automatically that they exist with the given content

tests/system/action/organization/test_initial_import.py
```diff
@@ -0,0 +1,531 @@
+from typing import Any, Dict
+
+from tests.system.action.base import BaseActionTestCase
+
+
+class OrganizationInitialImport(BaseActionTestCase):
+    def create_request_data(self, datapart: Dict[str, Any]) -> Dict[str, Any]:
```
jsangmeister, line 7: You can remove the `datapart` if it is not used anywhere

openslides_backend/action/actions/organization/initial_import.py
reiterl: Done.

tests/system/action/organization/test_initial_import.py
reiterl: Removed.

tests/system/action/organization/test_initial_import.py
reiterl: Removed.

openslides_backend/models/checker.py
reiterl: We need to clean the model_registry of the other tests then, because some tests add fake_models to the registry and that leads to errors in this tests.

tests/system/action/organization/test_initial_import.py
jsangmeister: As I wrote above, you'll have to adjust the checker for this to work. I don't know why we implemented it this way in the first place, at least I don't know any cases where it would lead to unwanted behaviour to leave optional fields out.

tests/system/action/organization/test_initial_import.py
jsangmeister: see other comment

openslides_backend/models/checker.py
reiterl: Add code to the setup of the tests, which cleans the registry.

PR #1009, opened 2021-11-01
Issue1008 Move models.yml and co from openslides to openslides-backend

r-peschke:


cli/modelsvalidator/validate.py
```diff
@@ -0,0 +1,374 @@
...
+_field_regex = r"[a-z][a-z0-9_]*\$?[a-z0-9_]*"
+
+collectionfield_regex = re.compile(
+    f"^({_collection_regex}){KEYSEPARATOR}({_field_regex})$"
+)
+collection_regex = re.compile(f"^{_collection_regex}$")
+field_regex = re.compile(f"^{_field_regex}$")
+
+decimal_regex = re.compile("^\d+\.\d{6}$")  # NOQA
+color_regex = re.compile("^#[0-9a-f]{6}$")
```
jsangmeister, line 24: Since this file now lives in the backend, you should be able to import some of these from the "original" files instead of duplicating the definitions here (if that's possible)

dev/docker-compose.dev.yml
```diff
@@ -16,6 +16,7 @@ services:
             - ../migrations:/app/migrations
             - ../tests:/app/tests
             - ../cli:/app/cli
+            - ../docs:/app/docs
```
jsangmeister, line 19: I would prefer a different folder name instead of "docs" (not only for the docker volume, but for the real folder name). The folder began as a doc folder in the main repo, but this was before the `models.yml` was introduced. Now, all the files that lie there a no longer any kind of "documentation", but instead relevant parts of the setup/code, so we should choose a better name. Maybe something like `data` or `assets`? Although I'm not entirely happy with them, either. Maybe you have a better idea (or someone else)?

jsangmeister (changes requested):
Only a few (mostly minor) remarks. Looks very good!

Before merging, the required github actions have to be updated.

jsangmeister:
Required actions are adjusted. I think other PRs can now not be merged until this is merged, since they don't have the new actions

Makefile
```diff
@@ -26,16 +26,30 @@ test:
 test-unit-integration:
 	pytest tests/unit tests/integration
 
+check-all: validate-models.yml check-models check-initial-data.json check-example-data.json check-permissions
+
+validate-models.yml:
```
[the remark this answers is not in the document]
r-peschke: done

Makefile
```diff
@@ -26,16 +26,30 @@ test:
...
+	PYTHONPATH=. python3 cli/modelsvalidator/validate.py
```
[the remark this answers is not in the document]
r-peschke: done

cli/generate_permissions.py
```diff
@@ -31,6 +30,8 @@
     from typing import Dict, List
 
     from .base_classes import Permission
+
+    MODELS_YML_CHECKSUM = "{}"
```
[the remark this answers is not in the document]
r-peschke: done

cli/modelsvalidator/validate.py
r-peschke: done, except last two

cli/modelsvalidator/validate.py
```diff
@@ -0,0 +1,369 @@
...
+    collection_regex,
+    collectionfield_regex,
+    field_regex,
+)
+
+DEFAULT_FILES = [
+    "./global/meta/models.yml",
+]
+
+decimal_regex = re.compile("^\d+\.\d{6}$")  # NOQA
```
jsangmeister, line 18: Defined in https://github.com/OpenSlides/openslides-backend/blob/d06365386e2b39eb001c2d56f2ea6e940099d61d/openslides_backend/shared/schema.py#L29 (slightly different even! That's why code duplication is bad ;)), please unify

cli/modelsvalidator/validate.py
```diff
@@ -0,0 +1,369 @@
...
+color_regex = re.compile("^#[0-9a-f]{6}$")
```
jsangmeister, line 19: same here https://github.com/OpenSlides/openslides-backend/blob/721e7f8650a463b0b48d790deb942e5993f33d56/openslides_backend/models/fields.py#L193

jsangmeister (changes requested):
Just noticed in https://github.com/OpenSlides/OpenSlides/pull/6270: Is there a reason why you removed the migration workflow? I think it makes sense to test the initial data together with the migrations.

https://github.com/OpenSlides/OpenSlides/blob/34eb57a31da8b801d9e3a969dc2778a9b05a7835/.github/workflows/models.yml#L80

dev/docker-compose.dev.yml
r-peschke: moved to global/meta for the *.yml and global/data for the 2 json-files.

cli/modelsvalidator/validate.py
r-peschke: It's not only different, one of them is wrong.
The one from validate.py allows only positive decimals in the models.yml.
The other one allows also negative numbers in the schema.

r-peschke:
> Just noticed in [OpenSlides/OpenSlides#6270](https://github.com/OpenSlides/OpenSlides/pull/6270): Is there a reason why you removed the migration workflow? I think it makes sense to test the initial data together with the migrations.
> 
I removed the test for following reasons:

1. Since we have the initial-data.json and example-data.json always updated, there is nothing to migrate
2. Common migrations and individual migrations are tested in backend in system test. If there is a special case you wanna test, we should implement it here
3. Because of its own setup it costs to much time

openslides_backend/models/checker.py
```diff
@@ -374,6 +374,17 @@ def check_normal_fields(self, model: Dict[str, Any], collection: str) -> bool:
             error = f"{collection}/{model['id']}: Invalid fields {', '.join(f'{field} (value: {model[field]})' for field in diff)}"
             self.errors.append(error)
             errors = True
+
+        for field in self.models[collection]().get_fields():
+            if (fieldname := field.get_own_field_name()) in model_fields:
+                try:
+
```
jsangmeister, line 381: remove empty line

tests/system/action/organization/test_initial_import.py
```diff
@@ -1,17 +1,33 @@
...
 from tests.system.action.base import BaseActionTestCase
 
 
+def get_backend_migration_index(migrations: List) -> int:
+    backend_migration_index = 1
+    for migration_class in migrations:
+        backend_migration_index = max(
+            backend_migration_index, migration_class().target_migration_index
+        )
+    return backend_migration_index
```
jsangmeister, line 17: remove and use the method in the migration module once https://github.com/OpenSlides/openslides-backend/pull/677 is merged (which should be any minute...)

jsangmeister (changes requested):
Looks good! minor changes

PR #1011, opened 2021-11-03
Add user.forget_password and user.forget_password_confirm

reiterl:
Add some tests.
Uses a new auth service. And update it.

### DIFF
--- a/openslides_backend/action/actions/user/forget_password.py
+++ b/openslides_backend/action/actions/user/forget_password.py
@@ -1,12 +1,4 @@
 from collections import defaultdict
-from smtplib import (
-    SMTPAuthenticationError,
-    SMTPDataError,
-    SMTPRecipientsRefused,
-    SMTPSenderRefused,
-    SMTPServerDisconnected,
-)
-from ssl import SSLCertVerificationError
 from time import time
 from typing import Any, Dict
 from urllib.parse import quote
@@ -65,12 +57,14 @@ class UserForgetPassword(EmailMixin, UpdateAction):
                 self.model.collection, filter_, ["id", "username"]
             )
 
-            # try to send the mail.
+            # try to send the mails.
             try:
                 with EmailMixin.get_mail_connection() as mail_client:
                     for user in results.values():
-                        if self.sendmail(
+                        ok, _ = self.send_email_safe(
                             mail_client,
+                            self.logger,
+                            EmailSettings.default_from_email,
                             email,
                             PW_FORGET_EMAIL_SUBJECT,
                             self.get_email_body(
@@ -78,17 +72,14 @@ class UserForgetPassword(EmailMixin, UpdateAction):
                                 self.get_token(user["id"], email),
                                 user["username"],
                             ),
-                        ):
+                            html=False,
+                        )
+                        if ok:
                             yield {"id": user["id"], "last_email_send": round(time())}
-            except SMTPAuthenticationError as e:
-                raise ActionException(f"SMTPAuthenticationError: {str(e)}")
-            except SMTPSenderRefused as e:
-                raise ActionException(f"SMTPSenderRefused: {str(e)}")
-            except SSLCertVerificationError as e:
-                raise ActionException(f"SSLCertVerificationError: {str(e)}")
             except Exception as e:
+                self.logger.error(f"General send mail exception: {str(e)}")
                 raise ActionException(
-                    f"Unspecified mail connection exception on sending invitation email to server {EmailSettings.host}, port {EmailSettings.port}: {str(e)}"
+                    "The server was configured improperly. Please contact your administrator."
                 )
 
     def get_token(self, user_id: int, email: str) -> str:
@@ -104,29 +95,6 @@ class UserForgetPassword(EmailMixin, UpdateAction):
             },
         )
         return PW_FORGET_EMAIL_TEMPLATE.format_map(body_format)
-
-    def sendmail(
-        self, mail_client: Any, email: str, email_subject: str, email_body: str
-    ) -> bool:
-        try:
-            self.send_email(
-                mail_client,
-                EmailSettings.default_from_email,
-                email,
-                email_subject,
-                email_body,
-                html=False,
-            )
-        except SMTPRecipientsRefused as e:
-            self.logger.error(f"SMTPRecipientsRefused: {str(e)}")
-            return False
-        except SMTPServerDisconnected as e:
-            self.logger.error(f"SMTPServerDisconnected: {str(e)}")
-            return False
-        except SMTPDataError as e:
-            self.logger.error(f"SMTPDataError: {str(e)}")
-            return False
-        return True
 
     def check_permissions(self, instance: Dict[str, Any]) -> None:
         pass
--- a/openslides_backend/action/actions/user/forget_password.py
+++ b/openslides_backend/action/actions/user/forget_password.py
@@ -20,7 +20,7 @@ from ...util.default_schema import DefaultSchema
 from ...util.register import register_action
 from ...util.typing import ActionData
 
-PW_FORGET_EMAIL_TEMPLATE = """You are receiving this email, because you have requested a new password for your OpenSlides-account.
+PW_FORGET_EMAIL_TEMPLATE = """You are receiving this email because you have requested a new password for your OpenSlides-account.
 
 Please open the following link and choose a new password:
 /login/forget-password-confirm?user_id={user_id}&token={token}
--- a/openslides_backend/action/actions/user/forget_password.py
+++ b/openslides_backend/action/actions/user/forget_password.py
@@ -9,6 +9,7 @@ from smtplib import (
 from ssl import SSLCertVerificationError
 from time import time
 from typing import Any, Dict
+from urllib.parse import quote
 
 from ....models.models import User
 from ....shared.exceptions import ActionException
@@ -91,8 +92,7 @@ class UserForgetPassword(EmailMixin, UpdateAction):
                 )
 
     def get_token(self, user_id: int, email: str) -> str:
-        # get token without 'bearer '
-        return self.auth.create_authorization_token(user_id, email)[7:]
+        return quote(self.auth.create_authorization_token(user_id, email))
 
     def get_email_body(self, user_id: int, token: str, username: str) -> str:
 
--- a/openslides_backend/action/actions/user/forget_password.py
+++ b/openslides_backend/action/actions/user/forget_password.py
@@ -95,7 +95,6 @@ class UserForgetPassword(EmailMixin, UpdateAction):
         return quote(self.auth.create_authorization_token(user_id, email))
 
     def get_email_body(self, user_id: int, token: str, username: str) -> str:
-
         body_format = format_dict(
             None,
             {
--- a/openslides_backend/action/actions/user/forget_password.py
+++ b/openslides_backend/action/actions/user/forget_password.py
@@ -1,0 +1,136 @@
+from collections import defaultdict
+from smtplib import (
+    SMTPAuthenticationError,
+    SMTPDataError,
+    SMTPRecipientsRefused,
+    SMTPSenderRefused,
+    SMTPServerDisconnected,
+)
+from ssl import SSLCertVerificationError
+from time import time
+from typing import Any, Dict
+
+from ....models.models import User
+from ....shared.exceptions import ActionException
+from ....shared.filters import FilterOperator
+from ...generics.update import UpdateAction
+from ...mixins.send_email_mixin import EmailMixin, EmailSettings
+from ...util.default_schema import DefaultSchema
+from ...util.register import register_action
+from ...util.typing import ActionData
+
+PW_FORGET_EMAIL_TEMPLATE = """You receive this email, because you have requested a new password for your OpenSlides-account.
+
+Please open the following link and choose a new password:
+/login/forget-password-confirm?user_id={user_id}&token={token}
+
+For completeness your username: {username}"""
+PW_FORGET_EMAIL_SUBJECT = "Reset your OpenSlides password"
+
+
+class format_dict(defaultdict):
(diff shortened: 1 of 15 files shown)


reiterl:
I rebased it and added a fix to the test.

openslides_backend/action/actions/user/forget_password.py
```diff
@@ -0,0 +1,136 @@
...
+from ....models.models import User
+from ....shared.exceptions import ActionException
+from ....shared.filters import FilterOperator
+from ...generics.update import UpdateAction
+from ...mixins.send_email_mixin import EmailMixin, EmailSettings
+from ...util.default_schema import DefaultSchema
+from ...util.register import register_action
+from ...util.typing import ActionData
+
+PW_FORGET_EMAIL_TEMPLATE = """You receive this email, because you have requested a new password for your OpenSlides-account.
```
jsangmeister, line 22:
```suggestion
PW_FORGET_EMAIL_TEMPLATE = """You are receiving this email because you have requested a new password for your OpenSlides account.
```

openslides_backend/action/actions/user/forget_password.py
```diff
@@ -0,0 +1,136 @@
...
+
+Please open the following link and choose a new password:
+/login/forget-password-confirm?user_id={user_id}&token={token}
```
jsangmeister, line 25: Mmh, without a host this link will obviously not work. Maybe we need to do the work and make the host available in the backend? @GabrielInTheWorld @gsiv

openslides_backend/action/actions/user/forget_password.py
```diff
@@ -0,0 +1,136 @@
...
+    skip_archived_meeting_check = True
+
+    def get_updated_instances(self, action_data: ActionData) -> ActionData:
+        for instance in action_data:
+            email = instance.pop("email")
+
+            # check emailif valid email adress
+            if not self.check_email(EmailSettings.default_from_email):
+                raise ActionException(
+                    f"email '{EmailSettings.default_from_email}' is not a valid sender email address."
```
jsangmeister, line 56: I don't think the user should be told if there is a server configuration issue. A more general error message like "The server was configured improperly. Please contact your administrator." or something like that would be more appropriate I think.

openslides_backend/action/actions/user/forget_password.py
```diff
@@ -0,0 +1,136 @@
...
+        title="user forget password schema",
+        additional_required_fields={"email": {"type": "string"}},
+    )
+    skip_archived_meeting_check = True
+
+    def get_updated_instances(self, action_data: ActionData) -> ActionData:
+        for instance in action_data:
+            email = instance.pop("email")
+
+            # check emailif valid email adress
```
jsangmeister, line 53:
```suggestion
            # check if email adress is valid
```

openslides_backend/action/actions/user/forget_password.py
```diff
@@ -0,0 +1,136 @@
...
+                raise ActionException(f"'{email}' is not a valid email adress.")
+
+            # search for users with email
+            filter_ = FilterOperator("email", "=", email)
+            results = self.datastore.filter(
+                self.model.collection, filter_, ["id", "username"]
+            )
+
+            # try to send the mail.
+            try:
```
jsangmeister, line 68: With the current code, the email will be sent to all users until an error happens. After the error, the method exits. I think we should unify this so all users are tried, regardless if an error happens or not. Afterwards, if any error happened, it should be returned in the action results instead of throwing an exception.

openslides_backend/action/actions/user/forget_password.py
```diff
@@ -0,0 +1,136 @@
...
+                            ),
+                        ):
+                            yield {"id": user["id"], "last_email_send": round(time())}
+            except SMTPAuthenticationError as e:
+                raise ActionException(f"SMTPAuthenticationError: {str(e)}")
+            except SMTPSenderRefused as e:
+                raise ActionException(f"SMTPSenderRefused: {str(e)}")
+            except SSLCertVerificationError as e:
+                raise ActionException(f"SSLCertVerificationError: {str(e)}")
+            except Exception as e:
```
jsangmeister, line 88: Can this whole exception handling be extracted into a function in the email mixin? I don't think it's good to handle all exception in every function calling the mixin.

openslides_backend/action/actions/user/forget_password.py
```diff
@@ -0,0 +1,136 @@
...
+                raise ActionException(
+                    f"Unspecified mail connection exception on sending invitation email to server {EmailSettings.host}, port {EmailSettings.port}: {str(e)}"
+                )
+
+    def get_token(self, user_id: int, email: str) -> str:
+        # get token without 'bearer '
+        return self.auth.create_authorization_token(user_id, email)[7:]
+
+    def get_email_body(self, user_id: int, token: str, username: str) -> str:
+
```
jsangmeister, line 98: remove empty line

openslides_backend/action/actions/user/forget_password.py
```diff
@@ -0,0 +1,136 @@
...
+            {
+                "user_id": user_id,
+                "token": token,
+                "username": username,
+            },
+        )
+        return PW_FORGET_EMAIL_TEMPLATE.format_map(body_format)
+
+    def get_email_subjext(self) -> str:
+        return PW_FORGET_EMAIL_SUBJECT
```
jsangmeister, line 110: unnecessary function, you can just use the constant directly

openslides_backend/action/actions/user/forget_password.py
```diff
@@ -0,0 +1,136 @@
...
+            )
+        except SMTPRecipientsRefused as e:
+            self.logger.error(f"SMTPRecipientsRefused: {str(e)}")
+            return False
+        except SMTPServerDisconnected as e:
+            self.logger.error(f"SMTPServerDisconnected: {str(e)}")
+            return False
+        except SMTPDataError as e:
+            self.logger.error(f"SMTPDataError: {str(e)}")
+            return False
```
jsangmeister, line 132: same as above for these errors

openslides_backend/action/actions/user/forget_password.py
```diff
@@ -0,0 +1,136 @@
...
+            except SSLCertVerificationError as e:
+                raise ActionException(f"SSLCertVerificationError: {str(e)}")
+            except Exception as e:
+                raise ActionException(
+                    f"Unspecified mail connection exception on sending invitation email to server {EmailSettings.host}, port {EmailSettings.port}: {str(e)}"
+                )
+
+    def get_token(self, user_id: int, email: str) -> str:
+        # get token without 'bearer '
+        return self.auth.create_authorization_token(user_id, email)[7:]
```
jsangmeister, line 95: Just encode and decode the token properly and the space should not be a problem (should be converted to a `%20`)

openslides_backend/action/actions/user/forget_password_confirm.py
```diff
@@ -0,0 +1,41 @@
...
+        user_id = instance.pop("user_id")
+        new_password = instance.pop("new_password")
+        token = instance.pop("authorization_token")
+        self.check_token(user_id, token)
+        instance["id"] = user_id
+        instance["password"] = self.auth.hash(new_password)
+        return instance
+
+    def check_token(self, user_id: int, token: str) -> None:
+        if not self.auth.verify_authorization_token(user_id, "bearer " + token):
```
jsangmeister, line 37: same here, with proper encoding you don't need to add the prefix manually

jsangmeister (changes requested):
did not check the tests yet

openslides_backend/action/actions/user/forget_password.py
GabrielInTheWorld: Yes, I agree. At some time, we have to do this work anyway - therefore, we can do it now.

openslides_backend/action/actions/user/forget_password.py
GabrielInTheWorld: I agree, here.

openslides_backend/action/actions/user/forget_password_confirm.py
reiterl: Use quote/unquote here.

openslides_backend/action/actions/user/forget_password.py
reiterl: Use quote/unquote here.

openslides_backend/action/actions/user/forget_password.py
reiterl: Removed

openslides_backend/action/actions/user/forget_password.py
reiterl: Removed.

openslides_backend/action/actions/user/forget_password.py
reiterl: Renamed.

openslides_backend/action/actions/user/forget_password.py
reiterl: Done.

openslides_backend/action/actions/user/forget_password.py
reiterl: Renamed.

openslides_backend/action/actions/user/forget_password.py
reiterl: If there is a general problem with sending mails, it should stop. 
The exceptions SMTPAuthenticationError, SMTPSenderRefused, SSLCertVerificationError are general exception.
The sending email mechanism is broken. So I stop sending mails.
The SMTPRecipientsRefused, SMTPServerDisconnected, SMTPDataError exception don't stop sending emails.

openslides_backend/action/actions/user/forget_password.py
reiterl: Simplify the exception code here. Don't leak the exception to the user. In the general case (here). log the error and show a generic error message.

openslides_backend/action/actions/user/forget_password.py
reiterl: Add a send_mail_safe static method to the EmailMixin, which catches the exceptions.

openslides_backend/action/actions/user/forget_password.py
jsangmeister: Ok, we'll do that in a new issue.

openslides_backend/action/actions/user/forget_password.py
jsangmeister: Ah, ok, I see.

tests/system/action/user/test_forget_password.py
```diff
@@ -0,0 +1,73 @@
...
+
+from openslides_backend.action.mixins.send_email_mixin import (
+    ConnectionSecurity,
+    EmailSettings,
+)
+from tests.system.action.base import BaseActionTestCase
+from tests.system.action.mail_base import AIOHandler, AiosmtpdServerManager
+
+
+class UserForgetPassword(BaseActionTestCase):
```
jsangmeister, line 11: Please add a test with a non-existent email address. The request should succeed just like a test with an existing email, but no email should actually be sent (so an attacker cannot filter which email addresses exist and which do not).

tests/system/action/user/test_forget_password_confirm.py
```diff
@@ -0,0 +1,30 @@
+from urllib.parse import quote
+
+from tests.system.action.base import BaseActionTestCase
+
+
+class UserForgetPasswordConfirm(BaseActionTestCase):
```
jsangmeister, line 6: Add a test with an invalid token and assert that the request fails.

jsangmeister (changes requested):
Rest looks good!

tests/system/action/user/test_forget_password.py
```diff
@@ -0,0 +1,73 @@
...
+    def setUp(self) -> None:
+        super().setUp()
+        EmailSettings.connection_security = ConnectionSecurity.NONE
+        EmailSettings.host = "127.0.0.1"
+        EmailSettings.port = 25
+        EmailSettings.timeout = 5
+        EmailSettings.user = ""
+        EmailSettings.password = ""
+        EmailSettings.accept_self_signed_certificate = False
+        EmailSettings.default_from_email = "noreply@example.com"
```
jsangmeister, line 21: Also, please extract this into a helper function since it is used in other tests as well.

tests/system/action/user/test_forget_password.py
reiterl: Add a test.

tests/system/action/user/test_forget_password_confirm.py
reiterl: I added a test and improved the code. I added a needed try/except block.

tests/system/action/user/test_forget_password.py
reiterl: Add a helper function. It is also used in test_send_invitation_mail, use there the helper too.

jsangmeister (approved):

jsangmeister (changes requested):
whoops, wrong status

tests/system/action/user/test_forget_password.py
```diff
@@ -58,6 +52,12 @@ def test_forget_password_two_users_with_email(self) -> None:
         assert handler.emails[1]["to"][0] == "test@ntvtn.de"
         assert "For completeness your username: test2" in handler.emails[1]["data"]
 
+    def test_forget_password_no_user_found(self) -> None:
+        handler = AIOHandler()
+        with AiosmtpdServerManager(handler):
+            response = self.request("user.forget_password", {"email": "info@ntvtn.de"})
+        self.assert_status_code(response, 200)
```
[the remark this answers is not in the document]
reiterl: Added an assert.

jsangmeister (approved):

PR #1013, opened 2021-11-06
Extended datastore adapter

jsangmeister:


jsangmeister:
This got quite big. Some notable changes:
- `additional_relation_models` was renamed to `changed_models` to better reflect its purpose.
- `fetch_model` was merged with the default `get`. All methods of the extended datastore adapter now respect the changed models by default.
- `InstanceAdditionalBehaviour` was removed and replaced by a simple boolean flag, `use_changed_models`. The two modes `OnlyAdditional` and `DbBeforeAdditional` were never used and I can't image a case where they would be.
- Simplified the `skip_archived_meeting_check` flag since I ran into errors with it. Now, no `getattr` calls are necessary anymore, but the instance variable can just be directly accessed.
- Implemented custom enum for easier filter creation, since python does not follow its own principles.
- Removed the `type` parameter from min/max methods since it is always `int`.
- Removed the optionality of `mapped_fields` from all extended datastore adapter methods since I didn't want to implement the extra handling for all these cases and it was barely used anyway. For pure db access (with `use_changed_models=False`) it can still be passed `mapped_fields=[]` to access all fields instead of only specific ones.
  - This required rewriting the `export_helper.py` to work with `get_many` requests, which is more performant anway.
- Simplified `sort_linear` usage while removing the `additional_db_instances` from it, which are now covered by the changed_moels as well.
- Removed temporary solutions until this feature was done, e.g. the weight map in `motion_submitter.create`.
- Added `meeting_id` to some filters already (rest can follow in #1203).
- Added the `meta_new` flag so the datastore is not acessed or models which were generated in this action, since they would only throw a `MODEL_DOES_NOT_EXIST` anyway. Funnily, this sped up tests on my PC from about 7min to about 5min.
- Reduced amount of requests in generic delete action.
- Some tests which tested locked fields errors were either reduced in their broken locks or even succeed now. Please pay special attention to them to check that this is okay! (e.g. by searching for "broken" in the diff view).

openslides_backend/services/datastore/extended_adapter.py
```diff
@@ -0,0 +1,476 @@
...
+    -   all filter-based requests may take two calls to the datastore to succeed. The first call is
+        always necessary, since the changed_models are never complete. If, however, a model in the
+        changed_models matches the filter which it did not in the database AND some fields are
+        missing in the changed_models which are needed through the mapped_fields, a second request
+        is needed to fetch the missing fields. This can be circumvented by always storing (more or
+        less) "full" models in the changed_data, meaning all relevant fields which are requested in
+        future calls are present. This is the case for most applications in the backend.
+    -   filters are only evaluated separately on the changed_models and the datastore. If, for
+        example, a model in the datastore does not fit the filter, but through a change in the
+        changed_models would fit it, BUT does not fit the filter from the changed_models alone, it
```
r-peschke, line 45: For me it seems to be more consistent, if we would apply 

- all fields from datastore content, that are NOT in corresponding  changed_models respecting None-values and
- the missing fields including the filter fields for all changed_models, that are not in datastore content

 to the changed_models. Then we could filter the changed_models and the results are reflecting the updated data.
The data to read seems nearly identical.

Example with filtered ids as set:
datastore content: {1, 2, 3}
changed_models: {2, 3, 4, 5}
filtered_changed_models: {2, 4}
FInal result set with ids: (datastore_content - changed_models) | filtered_changed_models = {1, 2, 4}
This is the expected result for a filter on current data, i.e. with use_changed_models = True.

r-peschke (commented):
Still missing some files, especially the test with `broken locks`.

r-peschke:
Concerning the broken fields:
There is a difference in locking the fields found in addtional_relation_models/changed_models
The old way locked all the fields you got from additional_relation_models.
The changed_models are implemented without position. If you get a field directly from a changed_model, it is not locked. If you have to read the field as missing field, it will be locked. This can only happen, if the payload has more than 1 action in the payload. The locked_fields are deleted, the changed_models kept.

If we take good care of the changed_models fields including their relation-models this may be a way for creating 2 instances in 1 request. With complete tests we could try. Otherwise we would need to lock the fields from changed_models, too.

openslides_backend/services/datastore/extended_adapter.py
jsangmeister: This will not work, if I understand you correctly, since you do not have all data and do not know what you need to fetch. Imagine the following:

datastore: `{"model/1": {"a": 1}}`
changed_models: `{"model/1": {"b": 2}}`
filter request: `a=1 AND b=2`

The filter request to the datastore will return an empty result, as well as the one to the changed_models, but `model/1` would have to be returned for a correct result. But as written further down in the comment, this should not be a problem for us since we only filter on static/readonly fields like `meeting_id` (at least I never encountered any other filter yet).

jsangmeister:
> Concerning the broken fields: There is a difference in locking the fields found in addtional_relation_models/changed_models The old way locked all the fields you got from additional_relation_models. The changed_models are implemented without position. If you get a field directly from a changed_model, it is not locked. If you have to read the field as missing field, it will be locked. This can only happen, if the payload has more than 1 action in the payload. The locked_fields are deleted, the changed_models kept.
> 
> If we take good care of the changed_models fields including their relation-models this may be a way for creating 2 instances in 1 request. With complete tests we could try. Otherwise we would need to lock the fields from changed_models, too.

The way it was previously implemented was not correct, I think. As you say, the changed_models do not have a position and cannot, since they represent the changes done in this request. If, however, a field is written to the changed_models it is either based directly on the payload (in which case nothing needs to be locked, since the payload takes priority anyway) or it is based on some calculation done beforehand for which the locking already happened.

That's why I wanted you to look at the tests: I think they are correct now and multiple (simple) instances can be created in the same request. Simple meaning here that no filter requests are needed, because get/get_many can be done fully on the changed_models without a request to the datastore. More complex requests are still failing because of broken locks even though they technically would not need to because the datastore is accessed in every filter request and the fields are locked accordingly, but this is fine for us.

jsangmeister:
Should be done now. I write all events created into the `changed_models`. While rebasing, I simplified some weight handling code which was duplicated a thousand times.

openslides_backend/services/datastore/extended_adapter.py
```diff
@@ -0,0 +1,476 @@
...
+        return isinstance(self.changed_models.get(fqid), DeletedModel)
+
+    def is_new(self, fqid: FullQualifiedId) -> bool:
+        return self.changed_models.get(fqid, {}).get("meta_new") is True
+
+    def reset(self) -> None:
+        super().reset()
+        self.changed_models.clear()
+
+    def _filter_additional_models(
```
r-peschke, line 326: Minor: Change name to _filter_changed_models

r-peschke (changes requested):
2 minor changes, but seems very good

openslides_backend/action/actions/meeting/export_helper.py
```diff
@@ -3,67 +3,75 @@
...
-        "assignment_candidate",
-        "mediafile",
-        "projector",
-        "projection",
-        "projector_message",
-        "projector_countdown",
-        "chat_group",
-        "chat_message",
+def get_relation_fields() -> List[RelationListField]:
+    return [
```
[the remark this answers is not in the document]
jsangmeister: Much better this way I think and even caught two errors in the manual list, so good suggestion ;)

r-peschke (approved):

PR #1017, opened 2021-11-11
Fix routes to include the service as a prefix

jsangmeister:
#921 introduced this bug because I thought the proxy would change the routes, but it didn't. Now everything should work again


tsiegleauq (approved):
tested

PR #1028, opened 2021-11-17
Fix internal routes & add tests

jsangmeister:


peb-adr:
I'm still getting `curl: (52) Empty reply from server` on the first try with `"cmd": "migrate"`

    /srv # curl backend:9002/internal/migrations -H "Content-Type: application/json" -d '{"cmd": "stats"}'
    {"success": true, "output": "- Registered migrations for migration index 8\n- Datastore has 894 positions with 4875 events\n- The positions have a minimal migration index 4 and a maximal\n  migration index 4\n  -> Migration/Finalization is needed\n- There are 0 migration positions: 0 fully migrated,\n  0 partial migrated\n  -> Migration and finalization needed.\n- 894 positions have to be migrated (including partially migrated ones)\n"}
    /srv # curl backend:9002/internal/migrations -H "Content-Type: application/json" -d '{"cmd": "migrate"}'
    ### ... several seconds of silence, about as long as migrations should take ... ###
    curl: (52) Empty reply from server

as soon as curl returns with `empty reply` the backend logs

    [2021-11-17 17:50:00 +0000] [33] [CRITICAL] WORKER TIMEOUT (pid:35)
    [2021-11-17 17:50:00 +0000] [35] [INFO] Worker exiting (pid: 35)
    [2021-11-17 17:50:00 +0000] [37] [INFO] Booting worker with pid: 37

If I now execute the same curl it works

    /srv # curl backend:9002/internal/migrations -H "Content-Type: application/json" -d '{"cmd": "migrate"}'
    ### ... several seconds of silence ... ###
    {"success": true, "output": "Running migrations.\nPosition 524 from MI 4 to MI 8 ...\nPosition 525 from MI 4 to MI 8 ...\nPosition 526 from MI 4 to MI 8 ...

also The response does not seem to contain all migrated positions, is there a limit to the payload? (This particular DB has 894 positions)
In "output" a significant chunk from the start was cut off.
`finalize` afterwards works no problem.

jsangmeister:
Thank you for your observations. It seems that this is a timeout issue: If you run `stats` after you get the empty reply, it shows that only a part of the positions were migrated and the rest not, indicating that the migration process was cancelled.

Since the migration process already takes more than 30 seconds (the default timeout, see https://docs.gunicorn.org/en/latest/settings.html#timeout, and also the amount of time until the first empty reply occurs) with just under 1000 positions and we assume that we could have millions of positions, the request might take several hours to succeed. Gunicon offers the option of setting the timout to 0 = infinite, but I don't know how well HTTP requests which are open for several hours behave. Also, I don't know if this might negatively impact the overall performance of the backend, since if for some reason a route is poorly programmed and the backend worker gets stuck in a loop, this worker will never be killed off.

The quick fix is of course settings the timeout to 0 or a high enough value, but I think we need to talk alternative solutions.

jsangmeister:
Works for now

peb-adr:
Trying to access a media file on this branch returns this error message:

    {"message":"Media-Server: The server responded with an unexpected code 404: b'{\"success\": false, \"message\": \"The requested URL was not found on the server. If you entered the URL manually please check your spelling and try again.\"}'"}

This is also the case on the current master, where I narrowed it down to b8a77c7. One commit before it's working.

Using an - a bit wonky - debugging setup I think I found the issue is that at 
https://github.com/jsangmeister/openslides-backend/blob/af9cd4a60f835d8d3d5b06f5a20bdb6cfc1b70cd/openslides_backend/http/views/base_view.py#L93
 a path is falsely matched against `/internal` and thus returning 404.

jsangmeister:
seems like the media service calls the wrong URL: https://github.com/OpenSlides/openslides-media-service/blob/7f200eae1427ee47f19158bf49146ba785f41cd7/src/auth.py#L61 (a "/handle_request") is missing

jsangmeister:
merging this now. follow-up in https://github.com/OpenSlides/openslides-backend/issues/1053. If any other issues occur, let me know.

PR #1029, opened 2021-11-18
Internal action routes

### LINKED ISSUE #987: Opening common actions for internal use
How can I use common actions like `user.create` internally with the manage service.

I want to solve https://github.com/OpenSlides/openslides-manage-service/issues/88 and I think it might be nice not to push the new user directly into the datastore but trigger the `user.create` action like a superadmin would do. But I do not have the superadmin credentials (because the superadmin might have changed his initial password) and I also think it is not reasonable to make a login request first.

So can you please provide me a way to trigger the action service that way? If reasonable we can use a shared docker secret.

jsangmeister:
closes #987 

All actions are now accessible via `/internal/handle_request`. I tested the actions `user.{create,update,delete,set_password}` which work. Not all actions will work though: The user id is set to -1*, so any action which needs to access the request user will fail. If the manage service needs access to such an action, we'll need to revisit this, but for now it works.

\* This is somewhat arbitrary, but we need a value which can be written into the DS for the generated write requests. A value >= 1 is not suitable since these represent real users, while 0 might lead to confusion since it generally stands for the anonymous user. So I chose -1. Another possibility would be to make the user id not required in the SQL schema and then just leave it empty for such requests. I don't really care which of these possibilities we choose, do you have any thoughts?

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,6 +143,8 @@ General schema for internal routes: `/internal/<route>`
 * `/system/action/handle_request`: Main route of the service, is used to execute actions.
 * `/system/action/handle_separately`: Same function as `handle_request`, but the request is not executed atomically,
   meaning each action is executed and the result sent to the datastore separately.
+* `/internal/handle_request`: Same as the first route, but only for internal usage: All permission checks are skipped
+  and created write requests will have id -1.
 * `/internal/health`: Returns a list of all possible actions with their respective JSON schema.
 * `/internal/migrations`: Provides remote access to the migration tool.
 
--- a/openslides_backend/action/action.py
+++ b/openslides_backend/action/action.py
@@ -125,6 +125,7 @@ class Action(BaseAction, metaclass=SchemaProvider):
         """
         Entrypoint to perform the action.
         """
+        # breakpoint()
         self.user_id = user_id
         self.index = 0
         for instance in action_data:
--- a/openslides_backend/action/action.py
+++ b/openslides_backend/action/action.py
@@ -125,7 +125,6 @@ class Action(BaseAction, metaclass=SchemaProvider):
         """
         Entrypoint to perform the action.
         """
-        # breakpoint()
         self.user_id = user_id
         self.index = 0
         for instance in action_data:
--- a/openslides_backend/action/action_handler.py
+++ b/openslides_backend/action/action_handler.py
@@ -88,13 +88,18 @@ class ActionHandler(BaseHandler):
             yield name, info
 
     def handle_request(
-        self, payload: Payload, user_id: int, atomic: bool = True
+        self,
+        payload: Payload,
+        user_id: int,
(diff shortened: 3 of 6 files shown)


openslides_backend/action/action.py
```diff
@@ -125,6 +125,7 @@ def perform(
         """
         Entrypoint to perform the action.
         """
+        # breakpoint()
```
normanjaeckel, line 128: Please remove this line.

README.md
```diff
@@ -124,11 +124,29 @@ The action component listens to port 9002. The presenter component listens to po
...
+General schema for public routes: `/system/<service>/<route>`
+General schema for internal routes: `/internal/<route>`
+
+### Action Service
+
+* `/system/action/handle_request`: Main route of the service, is used to execute actions.
+* `/system/action/handle_separately`: Same function as `handle_request`, but the request is not executed atomically,
+  meaning each action is executed and the result sent to the datastore separately.
+* `/internal/health`: Returns a list of all possible actions with their respective JSON schema.
+* `/internal/migrations`: Provides remote access to the migration tool.
```
normanjaeckel, line 147: Do you want to document `/internal/handle_request` here?

normanjaeckel (approved):
Ok. Thanks. Do I have to do any authentication using a shared secret?

openslides_backend/main.py
```diff
@@ -50,7 +50,7 @@ def load_config(self) -> None:
         options = {
             "bind": DEFAULT_ADDRESSES[self.view_name],
             "worker_tmp_dir": "/dev/shm",  # See https://pythonspeed.com/articles/gunicorn-in-docker/
-            "timeout": int(os.environ.get("OPENSLIDES_BACKEND_WORKER_TIMEOUT", "30")),
+            "timeout": int(os.environ.get("OPENSLIDES_BACKEND_WORKER_TIMEOUT", "0")),
```
normanjaeckel, line 53: Please check this default value again. What timeout do we want in production? See https://docs.gunicorn.org/en/stable/settings.html#timeout

jsangmeister: see https://github.com/OpenSlides/openslides-backend/pull/1028#issuecomment-972678259 and the email from @peb-adr: this is a temporary fix.

README.md
jsangmeister: yes, thanks for the hint

jsangmeister:
> Ok. Thanks. Do I have to do any authentication using a shared secret?

Currently not, but this is needed to ensure security, as discussed in https://github.com/OpenSlides/OpenSlides/issues/6257. So please go ahead and send a secret in a suitable header of your choice and provide a way for the datastore to access this in the productive setup. Then, I can implement the validation of this.

jsangmeister:
also tested `organization.initial_import`. Merging this now since no further comments came up

PR #1040, opened 2021-11-22
Add chat_group.clear and update permissions of chat_message.create.

reiterl:
Add some tests,

### DIFF
--- a/openslides_backend/action/actions/chat_group/clear.py
+++ b/openslides_backend/action/actions/chat_group/clear.py
@@ -1,7 +1,6 @@
 from ....models.models import ChatGroup
 from ....permissions.permissions import Permissions
 from ....shared.patterns import FullQualifiedId
-from ....shared.schema import required_id_schema
 from ...generics.update import UpdateAction
 from ...util.default_schema import DefaultSchema
 from ...util.register import register_action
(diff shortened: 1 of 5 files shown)


openslides_backend/action/actions/chat_group/clear.py
```diff
@@ -0,0 +1,37 @@
...
+@register_action("chat_group.clear")
+class ChatGroupClear(ChatEnabledMixin, UpdateAction):
+    """
+    Action to clear a chat group.
+    """
+
+    model = ChatGroup()
+    schema = DefaultSchema(ChatGroup()).get_default_schema(
+        title="Schema for clear a chat group.",
+        additional_required_fields={"chat_group_id": required_id_schema},
```
jsangmeister, line 22: I would think the `chat_group_` prefix is implicitly given by the action name, so only `id` as the parameter name should suffice. (@GabrielInTheWorld)

jsangmeister: (then you can just use the normal update schema here)

openslides_backend/action/actions/chat_group/clear.py
```diff
@@ -0,0 +1,37 @@
...
+    )
+    permission = Permissions.Chat.CAN_MANAGE
+    permission_id = "chat_group_id"
```
jsangmeister, line 25: ... and remove this line

openslides_backend/action/actions/chat_group/clear.py
```diff
@@ -0,0 +1,37 @@
...
+    def get_updated_instances(self, action_data: ActionData) -> ActionData:
+        for instance in action_data:
+            chat_group = self.datastore.get(
+                FullQualifiedId(self.model.collection, instance["chat_group_id"]),
+                ["chat_message_ids"],
+            )
+            delete_action_data = []
+            for id_ in chat_group.get("chat_message_ids", []):
+                delete_action_data.append({"id": id_})
+            self.execute_other_action(ChatMessageDelete, delete_action_data)
```
jsangmeister, line 36: execute after loop to potentially remove the amount of calls

jsangmeister (changes requested):

openslides_backend/action/actions/chat_group/clear.py
reiterl: Done.

openslides_backend/action/actions/chat_group/clear.py
reiterl: Done.

openslides_backend/action/actions/chat_group/clear.py
reiterl: Done.

jsangmeister (approved):

PR #1041, opened 2021-11-22
Add GHActions to create issues when files change

tsiegleauq:
Automatically create issues when certain files change.

Namely:
global/meta/models.yml
global/meta/permission.yml
global/data/initial-data.json


ostcar (approved):
Looks good. Did not test it.

tsiegleauq:
As just introduced, i'll merge this now so we can work on it

PR #1043, opened 2021-11-23
Issue995 realization of presenter search_users_by_name_email

r-peschke:


openslides_backend/presenter/search_users_by_name_email.py
```diff
@@ -0,0 +1,143 @@
...
+from .presenter import register_presenter
+
+search_users_by_name_email_schema = fastjsonschema.compile(
+    {
+        "$schema": schema_version,
+        "type": "object",
+        "title": "search_users_by_name_email",
+        "description": "get lists of id, first-, last-name and email for tuples of exact (username, emails)-tuples.",
+        "properties": {
+            "permission_scope": {
```
jsangmeister, line 30: Will this ever be called from a commitee or organization scope? Isn't the meeting scope sufficient since the feature is meant for meeting admins?

openslides_backend/presenter/search_users_by_name_email.py
```diff
@@ -0,0 +1,143 @@
...
+                ],
+            },
+            "search": {
+                "type": "array",
+                "minItems": 1,
+                "items": {
+                    "type": "array",
+                    "minItems": 2,
+                    "maxItems": 2,
+                    "prefixItems": [
```
jsangmeister, line 46: Why do you use this strange tuple notation? I would just use an array of objects like
```
[{
  "username": "xy",
  "email": "x@y",
}]
```
I had to look up what `prefixItems` even does :D

openslides_backend/presenter/search_users_by_name_email.py
```diff
@@ -0,0 +1,143 @@
...
+                    FilterOperator("email", "=", search[1]),
+                )
+            elif search[0]:
+                filter_ = FilterOperator("username", "=", search[0])
+            elif search[1]:
+                filter_ = FilterOperator("email", "=", search[1])
+            else:
+                continue
+            instances = self.datastore.filter(
+                Collection("user"), filter_, ["id", "first_name", "last_name", "email"]
```
jsangmeister, line 83: looking at this, this will probably lead to performance issues when there are millions of users, since there is no index to use...

openslides_backend/presenter/search_users_by_name_email.py
jsangmeister: On another note, we do not need to seperate between email and username. The use case is that the admin enters the email OR username into ONE input field, which is then sent to the server. The client does not have the possibility to differentiate between username and email and it is not needed.

Also, on the same note, an array is unnecessary since only one input will ever be sent at a time. But if you want to support multiple inputs I guess that's fine, too (just see the other comment below).

openslides_backend/presenter/search_users_by_name_email.py
```diff
@@ -0,0 +1,143 @@
...
+            )
+
+            result[f"{search[0]}/{search[1]}"] = [
```
jsangmeister, line 86: Mmh, using this index means the object is potentially indexed with something like `"Prof. Peter Meier/pmeier@xy.de"` which I do not like. I would think just an array of results suffices.

If you want to support multiple inputs (an array of inputs, see above), you run into problem though: How do you differentiate which result belongs to which search input? This is another reason why I think restricting the payload to a single search string is better.

openslides_backend/presenter/search_users_by_name_email.py
```diff
@@ -0,0 +1,143 @@
...
+                (
+                    instance["id"],
+                    instance.get("first_name", ""),
+                    instance.get("last_name", ""),
+                    instance.get("email", ""),
+                )
```
jsangmeister, line 92: For the result, I woud prefer object notation as well just becase it's more readable. Else the client developer has to look up everytime if email was the second or the last field etc. Also you can then just use the result instances directly without generating the tuple first ;)

openslides_backend/presenter/search_users_by_name_email.py
jsangmeister: Then, you could just drop the whole `permission_scope` section and just provide a `meeting_id` and check below whether the user has permissions for this meeting, the meeting's committee oder the whole organization.

jsangmeister (changes requested):

openslides_backend/presenter/search_users_by_name_email.py
```diff
@@ -0,0 +1,127 @@
...
+        "type": "object",
+        "title": "search_users_by_name_email",
+        "description": "get lists of id, first-, last-name and email for tuples of exact (username, emails)-tuples.",
+        "properties": {
+            "permission_type": {
+                "type": "integer",
+                "enum": [1, 2, 3],
+            },  # 1=meeting, 2=committee, 3=organization
+            "permission_id": {
+                "type": "integer",
```
jsangmeister, line 35: You can add `min: 1` here

openslides_backend/action/actions/user/user_scope_permission_check_mixin.py
```diff
@@ -18,9 +18,9 @@
 
 
 class UserScope(Enum):
```
jsangmeister, line 20: If you change this to `UserScope(int, Enum)` you should be able to use it below with the `.value` suffix

openslides_backend/presenter/search_users_by_name_email.py
```diff
@@ -0,0 +1,127 @@
...
+        else:
+            meeting = self.datastore.get(
+                FullQualifiedId(Collection("meeting"), permission_id),
+                ["committee_id"],
+            )
+            if has_committee_management_level(
+                self.datastore,
+                self.user_id,
+                CommitteeManagementLevel.CAN_MANAGE,
+                meeting.get("committee_id", 0),
```
jsangmeister, line 124: I would prefer to use `[]` here since the `committee_id` is required and should always be present. If it is not, it is suitable to throw a 500, so the user and eventually we notice it and can debug and hopefully fix it. By providing `0` to `has_committee_management_level`, no error will be thrown because it will try to access the field `committee_$0_management_level` on the user which will always be `None`. This will probably go unnoticed or may even lead to hard-to-debug bugs.

tests/system/presenter/test_search_users_by_name_email.py
```diff
@@ -0,0 +1,277 @@
...
+                    "last_name": "last3",
+                    "email": "userX@test.de",
+                },
+                {
+                    "id": 4,
+                    "first_name": "first4",
+                    "last_name": "last4",
+                    "email": "userX@test.de",
+                },
+            ],
```
jsangmeister, line 109: Mmh, intuitively I'd thought that `username` and `email` are combined with an `AND`, so this should return nothing. But if the use case requires it, that's fine with me.

But we wanted to do CSV import validation with this as well and wanted to validate ALL of `first_name`, `last_name` and  `email`, right? Then we should do thi uniformly.

openslides_backend/presenter/__init__.py
```diff
@@ -6,6 +6,7 @@
     get_user_related_models,
     get_users,
     number_of_users,
+    search_users_by_name_email,
```
jsangmeister, line 9: Thinking about this now, I think the shorter name `search_users` is more suitable, since it describes good enough what the presenter does. What do you think?

jsangmeister (changes requested):

openslides_backend/presenter/__init__.py
r-peschke: I think this is no `search` in a usual kind, because you need to fill the both fields in an exact manner. It's more a `get`, but not with `id`. Also we have a presenter named `get_users`.

jsangmeister: It is a search by name or email ;) but agree to disagree. At least I would like to rename it to `search_users_by_name_or_email` to make it sound more fluent.

r-peschke: done

tests/system/presenter/test_search_users_by_name_email.py
r-peschke: nothing to do:D

openslides_backend/presenter/search_users_by_name_email.py
r-peschke: This is made because of mypy-check. `has_committee_management_level` expects an `int`-type. [] is not allowed, because it should be `int`. Changed it to `cast`.

openslides_backend/action/actions/user/user_scope_permission_check_mixin.py
r-peschke: You mean, I could use it without `.value` suffix?! Done

jsangmeister: yes, sorry :D

openslides_backend/presenter/search_users_by_name_email.py
r-peschke: done (minimum)

openslides_backend/presenter/search_users_by_name_email.py
jsangmeister: if you do `cast(int, meeting["committee_id"])` this should either work if committee id is correctly filled or throw an attribute error if it is empty. As I said, I prefer the 500 here since we want the use to notice if the database is corrupt. Or am I missing some mypy caveat here?

jsangmeister (approved):

openslides_backend/presenter/search_users_by_name_email.py
r-peschke: I will check it and throw an exception. On casting, I tried it, the error message is very abstruse (committee/None is not a valid FQID).

jsangmeister: But if you use the index operator you will never get to this point because you will get an attribute error before you even cast :D that's what I wrote above. But anyway, this PR does not have to wait because of such a detail

r-peschke: Yes, you are right with the index operator. But now we have decided message:D

PR #1046, opened 2021-11-23
Issue1005 cannot remove users from meeting

r-peschke:
Check was made for user scope, which should not be used for group C field user-"group"

### DIFF
--- a/openslides_backend/action/actions/user/create_update_permissions_mixin.py
+++ b/openslides_backend/action/actions/user/create_update_permissions_mixin.py
@@ -268,59 +268,6 @@ class CreateUpdatePermissionsMixin(UserScopePermissionCheckMixin):
             touch_meeting_ids: Set[int] = set(
                 map(int, instance.get("group_$_ids", dict()).keys())
             )
-            remove_meeting_ids: Set[int] = set(
-                map(
-                    lambda item: int(item[0]),
-                    filter(
-                        lambda item: not item[1],
-                        instance.get("group_$_ids", dict()).items(),
-                    ),
-                )
-            )
-            old_meeting_ids: Set[int] = set()
-            if instance_user_id := instance.get("id"):
-                old_meeting_ids = set(
-                    map(
-                        int,
-                        self.datastore.get(
-                            FullQualifiedId(Collection("user"), instance_user_id),
-                            ["group_$_ids"],
-                        ).get("group_$_ids", []),
-                    )
-                )
-
-            # Check scope on removing/adding meetings
-            if remove_meeting_ids or touch_meeting_ids - old_meeting_ids:
-                all_meeting_ids = touch_meeting_ids | old_meeting_ids
-                if len(all_meeting_ids) > 1:
-                    committee_ids = list(
-                        set(
-                            map(
-                                lambda x: x.get("committee_id"),
-                                self.datastore.get_many(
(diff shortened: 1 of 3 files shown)


jsangmeister (commented):
Are you sure everything can just be removed here? The case of removing a user from a group should work of course, but what about the cases like
- adding a user to a group of another meeting,
- as meeting admin: removing a user from a gruop of another meeting where you don't have manage permissions
- etc.

These should fail, do they still do this? (Sorry, I don't have an overview of the thousands of tests for user handling... :D)

openslides_backend/action/actions/user/create_update_permissions_mixin.py
```diff
@@ -268,59 +268,6 @@ def check_group_C(
             touch_meeting_ids: Set[int] = set(
                 map(int, instance.get("group_$_ids", dict()).keys())
```
r-peschke, line 269: This is the basis idea: All meetings touched by group are mentioned here and the user has to have the rights (OML, CML for the meeting or meeting) to fulfill. I expect all cases of adding or removing groups to the user are covered with this expression.

openslides_backend/action/actions/user/create_update_permissions_mixin.py
```diff
@@ -268,59 +268,6 @@ def check_group_C(
...
-                            "You need OrganizationManagementLevel.can_manage_users, because you try to add or remove meetings in Organization-scope!"
-                        )
-                    if committee_ids[0] not in permstore.user_committees:
-                        raise PermissionDenied(
-                            f"You need CommitteeManagementLevel.can_manage permission for committee {committee_ids[0]}, because you try to add or remove meetings in Committee-scope!"
-                        )
-                    return
-
             # Check permission for each change operation/meeting
             if diff := touch_meeting_ids - permstore.user_committees_meetings:
```
r-peschke, line 272: This subtract the meetings that are covered by committee-right

openslides_backend/action/actions/user/create_update_permissions_mixin.py
```diff
@@ -268,59 +268,6 @@ def check_group_C(
...
                 if diff := diff - permstore.user_meetings:
```
r-peschke, line 273: and this the meetings covered by user-manage rights in meeting

r-peschke:
> Are you sure everything can just be removed here? The case of removing a user from a group should work of course, but what about the cases like
> 
>     * adding a user to a group of another meeting,
> 
>     * as meeting admin: removing a user from a gruop of another meeting where you don't have manage permissions
> 
>     * etc.
> 
> 
> These should fail, do they still do this? (Sorry, I don't have an overview of the thousands of tests for user handling... :D)

The basis idea in checking is, that

> Are you sure everything can just be removed here? The case of removing a user from a group should work of course, but what about the cases like
> 
>     * adding a user to a group of another meeting,
> 
>     * as meeting admin: removing a user from a gruop of another meeting where you don't have manage permissions
> 
>     * etc.
> 
> 
> These should fail, do they still do this? (Sorry, I don't have an overview of the thousands of tests for user handling... :D)

See my remarks at the 3 deciding positions of code

jsangmeister (approved):
Ok sure, if you say that it's covered I trust you :D was just confused because code was only removed.

PR #1062, opened 2021-11-25
Organization enable chat

reiterl:
Add field to model, generate models.py,
add field to orga.update 
add check to EnableChatMixin and update tests.

### DIFF
--- a/tests/system/action/chat_group/test_create.py
+++ b/tests/system/action/chat_group/test_create.py
@@ -31,6 +31,19 @@ class ChatGroupCreate(BaseActionTestCase):
         self.assert_status_code(response, 400)
         assert "Chat is not enabled." in response.json["message"]
 
+    def test_create_chat_not_enabled_in_organization(self) -> None:
+        self.set_models(
+            {
+                "organization/1": {"enable_chat": False},
+                "meeting/1": {"is_active_in_organization_id": 1, "enable_chat": True},
+            }
+        )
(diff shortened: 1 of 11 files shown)


tests/system/action/chat_group/test_create.py
```diff
@@ -5,7 +5,10 @@
 class ChatGroupCreate(BaseActionTestCase):
```
jsangmeister, line 5: Add a test where `meeting/enable_chat` is true, but `organization/enable_chat` is false and assert it fails

tests/system/action/chat_group/test_sort.py
```diff
@@ -6,6 +6,7 @@ class ChatGroupSortActionTest(BaseActionTestCase):
     def setUp(self) -> None:
```
jsangmeister, line 6: same here

jsangmeister (changes requested):

tests/system/action/chat_group/test_create.py
reiterl: Done.

tests/system/action/chat_group/test_sort.py
reiterl: Done.

reiterl:
I also did a rebase to solve a conflict.

reiterl:
Added the two missing tests. I think this can be merged.

Issue #1071, opened 2021-11-26
committee.delete doesn't remove the cml-rights of users


r-peschke:
The CommitteeManagementLevel of user is a templateCharField, but no relation field. The reason, as far as I know, is the fact, that we want to have the ability of qualifying this relation, lets say to have not only CAN_MANAGE, but also CAN_DO_xy with the relation. Therefore technically the deletion of the committee has no impact on user.committee__management_level. We could implement this in committee.delete code with the disadvantage, that on recovery the user-CML could not be recovered.

Now this leads to problem of @emanuelschuetze that after deleting the default-committee you cant add a new one. This is because the admin still is the manager of the deleted committee. 

Workaround or more: With this Issue the calculated field handler will be modified, that it will recognize this fault on a per user base and will not modify such a left-over committee_management_level field.

reiterl:
As far as I understand this workaround:
In user/committee__management_level the trash of a deleted committee will stay.
e.g. user/committee_$_management_level = ["1"]
       user/committee_$1_management_level ="can_manage"
DELETE committee/1
in user stays:
       user/committee_$_management_level = ["1"]
       user/committee_$1_management_level ="can_manage"
Right?

I dislike this workaround. This can lead to multiple problems, because then user/committee__management_level will be inconsistent.

jsangmeister:
I would propose another approach: Updating only the template field and not the structured field. Example:

User 1 before deletion of committee 2:
```
"user/1": {
  "committee_$_management_level": ["2"],
  "committee_$2_management_level": "can_manage"
}
```
User 1 after committee deletion:
```
"user/1": {
  "committee_$_management_level": [],
  "committee_$2_management_level": "can_manage"
}
```
Since `committee/1` is deleted anyway, the extra information saved inside `user/1` does not hurt and will never be requested by any action that works on non-deleted models. When you restore the committee, you could potentially restore the CML, too.

To implement this, one would currently have to iterate all (!) users when a committee is deleted and check if the committee is present in the template field. This is too much work. Therefore, I would propose to extend our template-relation handling: Currently, each template field either has a `replacement_collection` or has a tag. This implicitely already separates the two types of fields. We could make two separate classes out of it (I'm talking `models.py` here, not `models.yml`) and for the former case, introduce an _additional relation_ in the replacement collection from where one could access all template fields which include it. This would mean the sarting scenario above would look the following:
```
"committee/2": {
[...]

PR #1072, opened 2021-11-26
Issue1071 Prevent changes for deleted committees on user CML-rights

r-peschke:


reiterl:
I won't approve this workaround, as mentioned in #1071

tests/system/action/committee/test_create.py
```diff
@@ -311,3 +311,38 @@ def test_permission(self) -> None:
...
+            },
+        )
+        self.assert_status_code(response, 200)
+        self.assert_model_deleted("committee/1", {"user_ids": [1]})
+        self.assert_model_exists("committee/2", {"name": "committee2", "user_ids": [1]})
+        user = self.assert_model_exists(
+            "user/1",
+            {
+                "committee_$1_management_level": "can_manage",
+                "committee_$1_management_level": "can_manage",
```
reiterl, line 345: See below.

reiterl (changes requested):

tests/system/action/committee/test_create.py
r-peschke: and here also

tests/system/action/committee/test_update.py
```diff
@@ -741,3 +741,35 @@ def test_remove_cml_manager_from_user21(self) -> None:
...
+            },
+        )
+        self.assert_status_code(response, 200)
+        self.assert_model_deleted("committee/1", {"user_ids": [1]})
+        self.assert_model_exists("committee/2", {"name": "committee2", "user_ids": [1]})
+        user = self.assert_model_exists(
+            "user/1",
+            {
+                "committee_$1_management_level": "can_manage",
+                "committee_$1_management_level": "can_manage",
```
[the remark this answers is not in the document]
r-peschke: Jupp, should be committee/2

openslides_backend/action/relations/user_committee_calculate_handler.py
```diff
@@ -36,6 +37,10 @@ def process_field(
             exception=False,
         )
         db_committee_ids = set(db_user.get("committee_ids", []) or [])
+        db_committee_cml = set(
+            map(int, db_user.get("committee_$_management_level", []) or [])
```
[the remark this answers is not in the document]
r-peschke: Because the result `None` does not fit in map or set-function

migrations/migrations/0012_committee_user_relation.py
```diff
@@ -0,0 +1,304 @@
...
+            self.committee_user_ids, list(committee_ids), [user_id], add=True
+        )
+        if not cml_committee_ids:
+            return None
+
+        # change cml-fields and create event for committee
+        for committee_id in cml_committee_ids:
+            event.data.pop(f"committee_${committee_id}_management_level")
+        event.data["committee_$_management_level"] = [cml_permission]
+        event.data["committee_$can_manage_management_level"] = cml_committee_ids
```
jsangmeister, line 92:
```suggestion
        event.data[f"committee_${cml_permission}_management_level"] = cml_committee_ids
```

migrations/migrations/0012_committee_user_relation.py
```diff
@@ -0,0 +1,304 @@
...
+            set(new_cml_committee_ids) - set(old_cml_committee_ids)
+        )
+        remove_cml_committee_ids = list(
+            set(old_cml_committee_ids) - set(new_cml_committee_ids)
+        )
+        # change cml-fields and create event for committee
+        for committee_id in new_cml_committee_ids:
+            event.data.pop(f"committee_${committee_id}_management_level")
+        event.data["committee_$_management_level"] = [cml_permission]
+        event.data["committee_$can_manage_management_level"] = new_cml_committee_ids
```
jsangmeister, line 238: see above

jsangmeister (approved):
Mostly reviewed apart from the tests, looks good. Just very minor remarks :)

migrations/migrations/0012_committee_user_relation.py
r-peschke: done

migrations/migrations/0012_committee_user_relation.py
r-peschke: done...

global/meta/models.yml
```diff
@@ -644,6 +650,17 @@ committee:
     restriction_mode: A
     read_only: true
     description: "Calculated field."
+
+  user_$_management_level:
+    type: template
+    restriction_mode: B
+    replacement_enum:
+      - can_manage
```
ostcar, line 658: This field was strange before but it now even stranger.

The reason of template fields is, that there is a variable number of replacement values that create field names that are only known at run-time.

In this case, this is always only one field `user_$can_manage_management_level`. Since there is no other possible field, there is not even a need to look into the template field.

Would it be possible for you to replace this field by a normal field like

```yml
manager_ids:
  type: relation-list
  to: user/committee_manager_ids
description: List of users ids, that have the committee management permission
```

This would make the code much easier to read

r-peschke: This field is a new field and we need it to get all related users in case of deleting a committee. It's absence was the reason that committee.delete left the database in a corrupt state, where it was impossible to create a new committee.
The (now!) related field  user.commitee_$_management_level was incorparated with the idea, that there will be more than one permission, a user could have for a specific committee and the opposite relation has to reflect this.

ostcar: I still don't understand why a template field is needed with only one possible replacement. Why can't you use a normal relation field?

PR #1076, opened 2021-12-02
Update migration route & restructure requirements

### LINKED ISSUE #1053: Support long-running requests like `migrate` while keeping timeout
follow-up to https://github.com/OpenSlides/openslides-backend/pull/1028. The gunicorn timeout is temporarily set to 0 to support these types of request but this solution is not adequat. We need to find a way to support this while keeping the default timeout of 30 seconds, which should be more than enough for all other purposes. Ideas are to execute the migration in a threaded/threaded worker/async worker.

jsangmeister:
Implement threading for migration route to support long-running migrations. Restructure some of the setup to deduplicate code. Add dev setup which uses the local datastore to support developing backend and datastore side-by-side.

closes #1053


peb-adr:
Yo, looks quite good already.
Though I found a problem.
When there are pending migrations and I issue an invalid command it seems it sets the progress status to 1, e.g.:
```bash
$ curl localhost:9002/internal/migrations -H "Content-Type: application/json" -d '{"cmd": "asd"}'
<html>
  <head>
    <title>Internal Server Error</title>
  </head>
  <body>
    <h1><p>Internal Server Error</p></h1>

  </body>
</html>

$ curl localhost:9002/internal/migrations -H "Content-Type: application/json" -d '{"cmd": "progress"}'
{"success": true, "status": 1, "output": ""}
```
this means I'm now in a state where no other commands then `progress` are permitted , but no migration is runnning. So the only way to resolve is to restart the container.

Also the `stats` command seems to be unreasonable slow.
```bash
$ curl localhost:9002/internal/migrations -H "Content-Type: application/json" -d '{"cmd": "stats"}'
```
Could take up to 10 seconds.

while
```bash
$ python3 migrations/migrate.py stats
```
Is always <1 seconds.

Makefile
```diff
@@ -94,6 +91,23 @@ run-tests:
 	dev/run-tests.sh
 
 
+# Build and run development container with local datastore in use
+
+start-dev-local-ds:
```
r-peschke, line 96: Really good!

openslides_backend/migrations/migration_handler.py
```diff
@@ -0,0 +1,118 @@
...
+
+    def execute_migrate_command(self, command: str, verbose: bool) -> None:
+        MigrationHandler.migration_running = True
+        time.sleep(10)
+        handler = MigrationWrapper(verbose, self.write_line)
+        try:
+            handler.execute_command(command)
+        except MigrationException as e:
+            raise View400Exception(str(e))
+        MigrationHandler.migration_running = False
```
r-peschke, line 107: Shouldn't you put this in a finally clause to disable migration_running-state?

r-peschke (changes requested):

openslides_backend/main.py
```diff
@@ -46,13 +46,13 @@ def __init__(self, view_name: str, *args: Any, **kwargs: Any) -> None:
         super().__init__(*args, **kwargs)
 
     def load_config(self) -> None:
-        loglevel = "debug" if is_dev_mode() else "info"
+        dev_mode = is_dev_mode()
         options = {
             "bind": DEFAULT_ADDRESSES[self.view_name],
             "worker_tmp_dir": "/dev/shm",  # See https://pythonspeed.com/articles/gunicorn-in-docker/
```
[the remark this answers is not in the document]
jsangmeister: I did not look into this, @normanjaeckel wrote this i think. We can talk about this, but I this would be a separate issue.

openslides_backend/migrations/migration_handler.py
jsangmeister: Yes, good catch. That should also solve @peb-adr's issue with the invalid state after an invalid command. Will implement it.

jsangmeister:
> Also the `stats` command seems to be unreasonable slow.

Well, I guess the `time.sleep(10)` I still had for test purposes had something to do with it :D Should be solved now, as well as the other issues. Please re-review.

peb-adr:
@jsangmeister 
Ok, cool
issues are resolved couldn't find any new ;)

peb-adr (approved):

PR #1077, opened 2021-12-02
Make health routes public

jsangmeister:


gsiv (approved):
I expected the health status to report `unhealthy` when, e.g., services that it depends on, such as Redis or the datastore service, are unavailable.  Instead, it happily responds `running` now, no matter what.  I assume, my expectations were simply wrong.

In general, the health route made public works and the patch can be merged, as far as I'm concerned.  Making the health feature return more useful information is going to be a separate discussion/task.

normanjaeckel:
Do you really want to provide such internal information to everybody out in the world? For docker's HEALTHCHECK this is not required, I think.

jsangmeister:
> I expected the health status to report `unhealthy` when, e.g., services that it depends on, such as Redis or the datastore service, are unavailable. Instead, it happily responds `running` now, no matter what. I assume, my expectations were simply wrong.

Checking this would be a lot more work and I don't know how feasible/reliable the information would be. I would suggest to independently check the status of redis/datastore/etc. to evaluate if the instance is healthy.

PS: At startup, it is checked that all dependencies are up and running, otherwise the backend doesn't start. But I assume you talk about the case that one of the services shuts down while the backend is running.

@normanjaeckel: I don't see an issue providing these informations publicly. Everyone can access the instance (the login mask) publicly anyway, and generally speaking, if the instance is unhealthy, the login mask will throw some error when trying to log in, so this information is already kind of "public".

gsiv:
> I would suggest to independently check the status of redis/datastore/etc. to evaluate if the instance is healthy.

You are probably right, especially for the near future.  However, please note that those checks would be complementary at best.  They are not an alternative to a proper health check because they can't tell us whether the backend can actually *connect* to the generally available Redis, for example.

> PS: At startup, it is checked that all dependencies are up and running, otherwise the backend doesn't start.

That's a good point, thanks.

> For docker's HEALTHCHECK this is not required, I think.

You're right.  The idea is to be able to query the information efficiently and in an orchestrator-agnostic way.

PR #1089, opened 2021-12-09
Improve the performance of vote.create by getting the ids for create in bulk.

reiterl:


openslides_backend/action/action.py
```diff
@@ -265,6 +266,13 @@ def get_updated_instances(self, action_data: ActionData) -> ActionData:
         """
         yield from action_data
 
+    def pre_get_ids(self, action_data: ActionData) -> ActionData:
```
jsangmeister, line 269: I would like a more generic name, like `prepare_action_data`/`prefetch`/`setup`. The method may be used for anything before the actual execution starts and when we implement the cache, this method might as well be used for that.

jsangmeister (changes requested):

openslides_backend/action/action.py
reiterl: Renamed into `prepare_action_data`.

jsangmeister (approved):

PR #1104, opened 2021-12-15
Update dependencies

jsangmeister:
Removed unneeded dependencies & added some typing supports.

I changed the dependency import for the local datastore to use the requirements file from github instead of duplicating the needed dependencies. This leads to two places where the datastore commit needs to be fixed, so I created a single script to enter this commit which must be executed before installing the dependencies. I think this is cleaner than fixing the commit in two places, but let me know what you think.


r-peschke (approved):
Yes, this way it is cleaner:-)

PR #1106, opened 2021-12-16
Update datastore

jsangmeister:


jsangmeister:
Dev setup does not work without these changes since the master of the datastore is always used in the compose files.

normanjaeckel (approved):

GabrielInTheWorld (approved):
I haven't tested it, but it looks good!

PR #1107, opened 2021-12-16
Adjust models.yml

jsangmeister:
Also set some other fields to `text` which might have longer content.

### DIFF
--- a/global/meta/models.yml
+++ b/global/meta/models.yml
@@ -959,7 +959,7 @@ meeting:
     required: true
     restriction_mode: B
   motions_preamble:
-    type: string
+    type: text
     default: "The assembly may decide:"
     restriction_mode: B
   motions_default_line_numbering:
@@ -1083,7 +1083,7 @@ meeting:
     default: Motions
     restriction_mode: B
   motions_export_preamble:
-    type: string
+    type: text
     restriction_mode: B
   motions_export_submitter_recommendation:
     type: boolean
@@ -1211,7 +1211,7 @@ meeting:
     default: Elections
     restriction_mode: B
   assignments_export_preamble:
-    type: string
+    type: text
     restriction_mode: B
 
   # Assignment polls
@@ -2572,7 +2572,7 @@ poll:
     type: number
     restriction_mode: A
   description:
-    type: string
+    type: text
     restriction_mode: A
   title:
     type: string
@@ -2825,7 +2825,7 @@ assignment:
     default: search
     restriction_mode: B
   default_poll_description:
-    type: string
+    type: text
     restriction_mode: B
   number_poll_candidates:
     type: boolean
--- a/global/meta/models.yml
+++ b/global/meta/models.yml
@@ -1,6 +1,6 @@
 ---
 # Types:
-#     - Nativ datatypes: string, number, boolean, JSON
+#     - Nativ datatypes: string, text, number, boolean, JSON
 #     - HTMLStrict: A string with HTML content.
 #     - HTMLPermissive: A string with HTML content (with video tags).
 #     - float: Numbers that are expected to be non-integer. Formatted as in rfc7159.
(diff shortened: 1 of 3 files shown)


r-peschke (approved):

ostcar:
Please wait before merging this. I would like to have a look. I will do it this weekend

ostcar:
I think the following fields should also be text:
* meeting/welcome_title
* meeting/description
* meeting/motion_preamble
* meeting/motions_export_preamble ???
* meeting/users_pdf_welcometitle
* meeting/assignments_export_preamble
* agenda_item/comment
* speaker/note
* motion_change_recommendation/other_description
* poll/description
* vote/value
* assignment/default_poll_description
* projector_countdown/description

global/meta/models.yml
```diff
@@ -1,6 +1,6 @@
 ---
 # Types:
-#     - Nativ datatypes: string, number, boolean, JSON
+#     - Nativ datatypes: string, text, number, boolean, JSON
```
ostcar, line 3: Could you add a short explanation what the different between string and text is?

ostcar (changes requested):

ostcar:
_From #1101_
> @ostcar do you really use the comment of the `models.yml` to generate code? I find that highly suspicious, but I can add the comment of course.
No. I hardcoded all possible types: https://github.com/OpenSlides/openslides-autoupdate-service/blob/92bd8f7cf8da951350c0558f9e2af8c888b74951/pkg/datastore/gen_request/main.go#L257-L291

I have to add the `text` type to that list. But I think that it is sensible first to change the specification before changing the code.

> > I think it is very important, that someone checks all of this fields an makes a decision, if they need more then 256 chars.
> 
> I don't think it's such a big problem: If someone tries to set more than 256 characters in a string field, he will just get an error. If enough people do that for a specific field, we can go ahead and just change that to `text` as well.
We don't test OpenSlides with real life data. So if someone finds a field that needs more then 256 characters it could be a big problem for him. It can not be fixed fast since it needs an update in the models.yml.

I think it would be better, that unlimited would be the default string-length. I don't understand why this limitation is necessary.

global/meta/models.yml
r-peschke: Only in backend: The string is limited to 256 char, text is unlimited

ostcar:
vote/value can not have a restrction. For assignment polls, this value can get big when there are a lot of options. See https://github.com/OpenSlides/OpenSlides/wiki/poll.vote

In most cases, a vote should be less then 100 bytes. But in special cases, the vote should not fail, because the field only accepts some values.

For the (future) vote decrypt service, I calculate with 1kb. But even this could not be enough in some cases.

jsangmeister:
> vote/value can not have a restrction. For assignment polls, this value can get big when there are a lot of options. See https://github.com/OpenSlides/OpenSlides/wiki/poll.vote
> 
> In most cases, a vote should be less then 100 bytes. But in special cases, the vote should not fail, because the field only accepts some values.
> 
> For the (future) vote decrypt service, I calculate with 1kb. But even this could not be enough in some cases.

The `poll.vote` action payload differs from the actual `vote/value` field. As @r-peschke said, `vote/value` only contains the actual value of the single vote and refers to a specific vote from exactly _one_ user to exactly _one_ option in _one_ poll. The payload from `poll.vote` is translated into multiple `vote.create` calls.

ostcar:
Ah. I did not realise this.

Can you tell and explain what the biggest value of the field could theoretically be?

jsangmeister:
> Can you tell and explain what the biggest value of the field could theoretically be?

As far as I know, the value is always restricted to `Y`, `N`, `A`. If we add more poll methods, this could theoretically be enhanced in the future, but I doubt that we'll ever need more than one char.

ostcar:
OK.

PR #1121, opened 2022-01-07
add opentelemetry

ostcar:


openslides_backend/main.py
```diff
@@ -65,6 +77,19 @@ def load(self) -> WSGIApplication:
...
         logging_module: LoggingModule = logging  # type: ignore
 
+        span_exporter = OTLPSpanExporter(
+            # optional
+            endpoint="collector:4317",
+            # credentials=ChannelCredentials(credentials),
+            # headers=(("metadata", "metadata")),
+        )
+        tracer_provider = TracerProvider(
+            resource=Resource.create({SERVICE_NAME: "bakend"})
```
emanuelschuetze, line 87: typo: "backend"

ostcar:
I don't plan to work on this branch. I only created the code for the demonstration. I was asked to create this drave-PR to "publish" my code.

The code I wrote does not have any standards. It is probably worse then the code finn wrote some time ago, that I have liked in #1153.

I think the opentelementry code should be added by someone, who has more inside in the backend-code. The code I wrote is not enough. On all important places, new spans have to be created. Important log-messages have to be written into the spans. The success-status of the span has to be logged into the span.

jsangmeister:
@peb-adr @ostcar if none of you need the content of this branch, I would like to delete it to keep the upstream clean.

peb-adr:
peb-adr:opentelemetry (#1201) contains the commit from this branch and built upon that. So nothing will be lost when deleting this branch.
> I would like to delete it to keep the upstream clean.

-> Ok from me

PR #1125, opened 2022-01-07
Issue1123 new user permission can_manage_presence

r-peschke:


openslides_backend/action/actions/user/set_present.py
```diff
@@ -71,6 +71,11 @@ def check_permissions(self, instance: Dict[str, Any]) -> None:
             self.user_id,
             Permissions.User.CAN_MANAGE,
             instance["meeting_id"],
+        ) or has_perm(
```
jsangmeister, line 74: You can remove the permission check above since the permissions are nested

jsangmeister (changes requested):

tsiegleauq (approved):
looks fine for me

openslides_backend/action/actions/user/set_present.py
r-peschke: done

PR #1136, opened 2022-01-10
Explicitly pull remote docker repos from main

tsiegleauq:
Pulls docker remotes from main instead from default (which is always
master in dockers context)
Fixes make run-tests

### DIFF
--- a/.github/docker-compose/docker-compose.prod.yml
+++ b/.github/docker-compose/docker-compose.prod.yml
@@ -18,7 +18,7 @@ services:
             - postgres_password
     reader:
         build:
-            context: "https://github.com/OpenSlides/openslides-datastore-service.git"
+            context: "https://github.com/OpenSlides/openslides-datastore-service.git#main"
             args:
                 MODULE: "reader"
                 PORT: "9010"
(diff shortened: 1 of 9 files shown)


reiterl (approved):
Looks good, if the tests pass, it is okay.

jsangmeister (approved):
I still find it astonishing that git/github uses and recommends `main` as the default branch's name, but doesn't automatically switch it here. (Or is it docker-compose which does not change the default name here?)

GabrielInTheWorld:
> I still find it astonishing that git/github uses and recommends `main` as the default branch's name, but doesn't automatically switch it here. (Or is it docker-compose which does not change the default name here?)

As I found here [docker-compose#git-repositories](https://docs.docker.com/engine/reference/commandline/build/#git-repositories), it seems that the branch "master" is used as the default branch in the docker-compose context.

tsiegleauq:
Seems like there is something deeper.

```
  fatal: remote error:
    The unauthenticated git protocol on port 9418 is no longer supported.
```

tsiegleauq:
https://github.blog/2021-09-01-improving-git-protocol-security-github/#no-more-unauthenticated-git

January 11, 2022
> Final brownout.
This is the full brownout period where we’ll temporarily stop accepting the deprecated key and signature types, ciphers, and MACs, and the unencrypted Git protocol. This will help clients discover any lingering use of older keys or old URLs.

r-peschke (approved):

PR #1146, opened 2022-01-14
Add a patch for a problem with mediafile/attachment_ids and clone.

reiterl:
I wrote a test, which shows the problem.
I wrote a fix for the problem in `meeting/clone` and tried to improve the checker code.

### DIFF
--- a/openslides_backend/action/actions/meeting/clone.py
+++ b/openslides_backend/action/actions/meeting/clone.py
@@ -67,6 +67,13 @@ class MeetingClone(MeetingImport):
                 value = instance.pop(field)
                 self.get_meeting_from_json(meeting_json)[field] = value
 
+        for mediafile_id in instance["meeting"]["mediafile"]:
+            if (
+                instance["meeting"]["mediafile"][mediafile_id].get("attachment_ids")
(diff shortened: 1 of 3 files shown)


openslides_backend/action/actions/meeting/clone.py
```diff
@@ -67,6 +67,13 @@ def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
                 value = instance.pop(field)
                 self.get_meeting_from_json(meeting_json)[field] = value
 
+        for mediafile_id in instance["meeting"]["mediafile"]:
```
r-peschke, line 70: Could you use get("mediafile", [])?

r-peschke (changes requested):
A smaller one;-)

openslides_backend/action/actions/meeting/clone.py
reiterl: Yes,  I could. But why? The meeting comes from the export_meeting function and must have a mediafile, or?
Okay, changed.

r-peschke (approved):

PR #1148, opened 2022-01-14
Issue1144 delete projection with historical projector in meeting.delete

r-peschke:
Allows projection.delete on meeting.delete-action, even without current_ or preview_projector_id. This coontinues to be forbidden for a single projection.delete

### DIFF
--- a/openslides_backend/action/actions/projector/delete.py
+++ b/openslides_backend/action/actions/projector/delete.py
@@ -26,8 +26,10 @@ class ProjectorDelete(DeleteAction):
             ["used_as_reference_projector_meeting_id", "meeting_id"],
         )
         if (
-            meeting_id := projector.get("used_as_reference_projector_meeting_id")
-        ) and not self.is_meeting_deleted(meeting_id):
+            (meeting_id := projector.get("used_as_reference_projector_meeting_id"))
+            and not self.is_meeting_deleted(meeting_id)
+            and self.parent_action != "meeting.delete"
+        ):
             raise ActionException(
                 "A used as reference projector is not allowed to delete."
             )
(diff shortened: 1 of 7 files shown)


emanuelschuetze (changes requested):
error: "It is not permitted to remove a candidate from a finished assignment!"

emanuelschuetze (approved):

openslides_backend/action/actions/projector/delete.py
```diff
@@ -26,8 +26,10 @@ def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
             ["used_as_reference_projector_meeting_id", "meeting_id"],
         )
         if (
-            meeting_id := projector.get("used_as_reference_projector_meeting_id")
-        ) and not self.is_meeting_deleted(meeting_id):
+            (meeting_id := projector.get("used_as_reference_projector_meeting_id"))
+            and not self.is_meeting_deleted(meeting_id)
+            and self.parent_action != "meeting.delete"
```
jsangmeister, line 31: Did you actually run into an error here? This is interesting because I explicitly added the line above to archieve the same goal, if the meeting is in the process of being deleted, the action is permitted. I would favor a uniform approach instead of adding more and more parameters and if statements which make the code more convoluted.

r-peschke: I run into the error on projection.delete. I searched for more conditional preventing of delete and found this one and more.
I agree, one approach is enough and should change the code. I would prefer my approach, because its more intuitive to see and may help with the parent._action in more situations. But you may decide:-)

jsangmeister: I would prefer an approach which makes the code less convoluted and does not need explicit exceptions for certain actions. If we run into more problem like these, you might as well have to add a list for action where you don't need to throw the exception. On the other hand, using the `is_deleted` approach, you only have to check if the "parent object" was deleted to check if this deletion is feasible.

PR #1149, opened 2022-01-14
Add sequential_number to some models. Add tests and migrations.

reiterl:
The migration code is with lot of code duplication. I tried to simplify it, but then it doesn't work.


reiterl:
@jsangmeister  Perhaps you can look at this and suggest a way to dedublicate the migration code.

migrations/migrations/0010_add_sequential_number_to_assignment.py
```diff
@@ -0,0 +1,46 @@
+from collections import defaultdict
```
jsangmeister, line 1: Please move all migrations into one. This means that you cannot use the base `AddFieldMigration`, but instead have to write the code yourself, but has the advantage that you can do all steps as one and can deduplicate the code.

migrations/migrations/0010_add_sequential_number_to_assignment.py
```diff
@@ -0,0 +1,46 @@
...
+        if self.sequential_numbers_map is None:
+            self.init_sequential_numbers_map()
+        self.sequential_numbers_map[event.data["meeting_id"]] += 1
+        return self.sequential_numbers_map[event.data["meeting_id"]]
+
+    def init_sequential_numbers_map(self) -> None:
+        ids = self.new_accessor.get_all_ids_for_collection(self.collection)
+        self.sequential_numbers_map = defaultdict(int)
+        for id_ in ids:
+            fqid = self.collection + "/" + str(id_)
```
jsangmeister, line 30:
```suggestion
            fqid = self.collection + KEYSEPARATOR + str(id_)
```

migrations/tests/test_0010_add_sequential_number_to_assignment.py
```diff
@@ -0,0 +1,92 @@
+def test_migration(write, finalize, assert_model):
+    write({"type": "create", "fqid": "meeting/1", "fields": {"id": 1}})
+    write(
+        {"type": "create", "fqid": "assignment/1", "fields": {"id": 1, "meeting_id": 1}}
+    )
+    write(
+        {"type": "create", "fqid": "assignment/2", "fields": {"id": 2, "meeting_id": 1}}
+    )
+
+    finalize("0010_add_sequential_number_to_assignment")
```
jsangmeister, line 10: Since the migration is more complex, we should test that it works with the two-stage migration. You should be able to create a `migrate` fixture analogous to the `finalize` one (see https://github.com/OpenSlides/openslides-backend/blob/9fe6aad0c7f61e2b9b9ef7d7a2c682b281d2f6ba/migrations/tests/conftest.py#L49). Use it here, write another event after that und then do a `finalize`. Assert that the sequential numbers are correct.

Also add another meeting to assert that the migration works correctly for multiple meetings.

openslides_backend/action/actions/assignment/create_update_delete.py
```diff
@@ -16,12 +19,20 @@
...
 ):
     dependencies = [AgendaItemCreate, ListOfSpeakersCreate]
 
+    def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
+        instance = super().update_instance(instance)
+        instance["sequential_number"] = self.get_sequential_number(
+            instance["meeting_id"]
+        )
+        return instance
+
```
jsangmeister, line 35: If you overwrite the `update_instance` method in the mixin, you can save these lines in each action.

jsangmeister (changes requested):

migrations/migrations/0010_add_sequential_number_to_assignment.py
reiterl: Done.

migrations/migrations/0010_add_sequential_number_to_assignment.py
reiterl: Done.

openslides_backend/action/actions/assignment/create_update_delete.py
reiterl: Done.

migrations/tests/test_0010_add_sequential_number_to_assignment.py
reiterl: Done.

migrations/tests/conftest.py
```diff
@@ -29,6 +29,12 @@ def clear_datastore(setup) -> None:
     writer: Writer = injector.get(Writer)
     writer.truncate_db()
 
+    def _clear_datastore() -> None:
+        writer: Writer = injector.get(Writer)
+        writer.truncate_db()
```
jsangmeister, line 34:
```suggestion
    def _clear_datastore() -> None:
        writer: Writer = injector.get(Writer)
        writer.truncate_db()
    _clear_datastore()
```

openslides_backend/action/actions/motion_workflow/create.py
```diff
@@ -57,6 +58,9 @@ class MotionWorkflowCreateSimpleWorkflowAction(CreateAction):
     )
 
     def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
+        instance["sequential_number"] = self.get_sequential_number(
+            instance["meeting_id"]
+        )
```
jsangmeister, line 63:
```suggestion
        instance = super().update_instance(instance)
```

openslides_backend/action/actions/motion_workflow/create.py
```diff
@@ -126,6 +130,9 @@ class MotionWorkflowCreateComplexWorkflowAction(CreateAction):
...
```
jsangmeister, line 135:
```suggestion
        instance = super().update_instance(instance)
```

jsangmeister (changes requested):
some minor remarks left, rest looks good!

migrations/tests/conftest.py
reiterl: Done.

openslides_backend/action/actions/motion_workflow/create.py
reiterl: Done

openslides_backend/action/actions/motion_workflow/create.py
reiterl: Done.

jsangmeister (approved):

PR #1151, opened 2022-01-14
Issue1150 Errors on meeting.clone

r-peschke:


emanuelschuetze:
retested: clone a meeting with a single motion:
`motion/49: Invalid fields agenda_type (value: internal) motion/49/recommendation_extension_reference_ids:`

emanuelschuetze:
two errors after cloning:
- `motion/48: Invalid fields workflow_id (value: 5) `
- `motion/41/amendment_paragraph_$: Missing amendment_paragraph_$2 since it is given as a replacement`

openslides_backend/action/actions/agenda_item/agenda_creation.py
```diff
@@ -56,18 +56,23 @@ def check_dependant_action_execution_agenda_item(
         )
         agenda_item_creation = meeting.get("agenda_item_creation")
         agenda_create = instance.pop("agenda_create", None)
+        result_value: Optional[bool] = None
```
jsangmeister, line 59: Shouldn't you be able to use 
```suggestion
        result_value: bool
```
since it is always assigned a value in the if-else? Then you can also drop the `cast` below

openslides_backend/action/actions/agenda_item/agenda_creation.py
```diff
@@ -56,18 +56,23 @@ def check_dependant_action_execution_agenda_item(
...
         if agenda_item_creation == "always":
             return True
```
jsangmeister, line 61: I would favor a uniform approach and also use `result_value` here

openslides_backend/models/fields.py
```diff
@@ -319,8 +318,10 @@ def get_schema(self) -> Schema:
         return self.extend_schema(super().get_schema(), **fqid_list_schema)
 
     def validate(self, value: Any, payload: Dict[str, Any] = {}) -> Any:
-        assert isinstance(value, list)
-        return [string_to_fqid(fqid) for fqid in value]
+        if value or self.required:
```
jsangmeister, line 321: I would suggest
```suggestion
        if value is not None or self.required:
```
to remove the implication that `0` or something similar could be a valid value and also remove the possibility that it becomes one by accident

openslides_backend/models/fields.py
```diff
@@ -465,5 +466,6 @@ def validate(self, value: Any, payload: Dict[str, Any] = {}) -> Any:
             return {key: sup.validate(struc) for key, struc in value.items()}
         elif type(value) == list:
             return value
-        else:
-            raise NotImplementedError
+        elif value is None:
+            return None
+        raise NotImplementedError
```
jsangmeister, line 471: Not form your changes, but noticing this just now: all the `raise NotImplementedError` lines are missing the brackets: `raise NotImplementedError()`

tests/system/action/meeting/test_clone.py
```diff
@@ -530,3 +531,71 @@ def test_permissions_missing_payload_committee_permission(self) -> None:
             "Missing CommitteeManagementLevel: can_manage for committee 2",
             response.json["message"],
         )
+
+    def test_clone_with_created_topic_and_agenda_type(self) -> None:
+        self.set_models(self.test_models)
+        response = self.request(
```
jsangmeister, line 537: If you just want to execute an action to use the result to test another action (here, you are not interested in the result or functionality of `topic.create` but onlywant to create the topic to use it later), you can use `execute_action`. This should be much faster since it skips a lot of checks like permission etc. Also the test code becomes (hopefully) clearer because all `request` calls are actually for the tested action.

tests/system/action/meeting/test_clone.py
```diff
@@ -530,3 +531,71 @@ def test_permissions_missing_payload_committee_permission(self) -> None:
...
+                "meeting_id": 1,
+                "title": "test",
+                "agenda_type": AgendaItem.INTERNAL_ITEM,
+                "agenda_duration": 60,
+            },
+        )
+        self.assert_status_code(response, 200)
+        topic_fqid = f'topic/{response.json["results"][0][0]["id"]}'
+        topic = self.assert_model_exists(
+            topic_fqid, {"agenda_type": None, "agenda_duration": None}
```
jsangmeister, line 549: Better test explicitly that the fields are not present. This creates the impression that the fields are present in the model and set to `None`.

jsangmeister (changes requested):

emanuelschuetze (approved):

openslides_backend/action/actions/agenda_item/agenda_creation.py
r-peschke: done

openslides_backend/action/actions/agenda_item/agenda_creation.py
r-peschke: done, see above

openslides_backend/models/fields.py
r-peschke: done

tests/system/action/meeting/test_clone.py
r-peschke: done

tests/system/action/meeting/test_clone.py
r-peschke: done

tests/system/action/meeting/test_clone.py
```diff
@@ -530,3 +531,72 @@ def test_permissions_missing_payload_committee_permission(self) -> None:
...
+                "content_object_id": topic_fqid,
+            },
+        )
+
+        response = self.request("meeting.clone", {"meeting_id": 1})
+        self.assert_status_code(response, 200)
+
+    def test_clone_with_created_motion_and_agenda_type(self) -> None:
+        self.set_models(self.test_models)
+        response = self.request(
```
r-peschke, line 567: Here the request is needed, because it includes other_actions calls with data.user_id

PR #1155, opened 2022-01-17
Add get_user_scope presenter and test.

reiterl:


reiterl:
I have discussed this issue in the daily with @r-peschke . We came to the conclusion, that we want the user_scope_mixin in the shared folder.

openslides_backend/presenter/get_user_scope.py
```diff
@@ -0,0 +1,55 @@
...
+                "item": {"type": "integer"},
+            },
+        },
+        "required": ["user_ids"],
+        "additionalProperties": False,
+    }
+)
+
+
+class UserScope(int, Enum):
```
r-peschke, line 29: You should import this from the shared folder

r-peschke (changes requested):

openslides_backend/presenter/get_user_scope.py
reiterl: Done.

jsangmeister (approved):

PR #1164, opened 2022-01-19
Issue1150 Use execute_action_internally

r-peschke:
Missing part from review of #1151
Maybe there is more to do on meeting.clone problem:

motion/48: Invalid fields workflow_id (value: 5)  After simple amendment motion on new crated meeting

### DIFF
--- a/openslides_backend/action/action.py
+++ b/openslides_backend/action/action.py
@@ -102,7 +102,6 @@ class Action(BaseAction, metaclass=SchemaProvider):
         relation_manager: RelationManager,
         logging: LoggingModule,
         skip_archived_meeting_check: bool = False,
-        parent_action: str = "",
     ) -> None:
         self.services = services
         self.auth = services.authentication()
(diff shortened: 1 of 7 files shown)


jsangmeister (changes requested):
You can do the same in the test below ( and maybe in other tests there too?)

jsangmeister:
> motion/48: Invalid fields workflow_id (value: 5) After simple amendment motion on new crated meeting

Can you evaluate when and why this problem occurs? It looks like the action data field `workflow_id` is not correctly deleted from the instance, but on a quick glance it seems like it is always correctly `pop`ed out.

r-peschke:
> > motion/48: Invalid fields workflow_id (value: 5) After simple amendment motion on new crated meeting
> 
> Can you evaluate when and why this problem occurs? It looks like the action data field `workflow_id` is not correctly deleted from the instance, but on a quick glance it seems like it is always correctly `pop`ed out.

It's a bug report by @emanuelschuetze : Create a meeting, write a motion and an motion amendment. The error should occur on meeting.clone, but doesn't. I asked him to help.

jsangmeister (approved):

PR #1169, opened 2022-01-20
Set e-voting true at initial-data

### LINKED ISSUE #1147: initial-data: Enable electronic voting by default
In `initial-data.json`, `enable_electronic_voting` should be set to `true` by default.  @emanuelschuetze, please confirm.

JLkp:
Fixes #1147


reiterl (approved):
looks okay for me. As emanuel has confirmed it.

Issue #1161, opened 2022-01-19
poll content_object


ostcar:
I have some questions for this already merged PR: https://github.com/OpenSlides/openslides-backend/pull/1154

Currently, the field [poll/content_object_id](https://github.com/OpenSlides/openslides-backend/blob/ba27ed3435d4246c339273df53857cdd033b6ca4/global/meta/models.yml#L2675-L2685) is not required.

Do I understand correctly, that there are no stand alone polls anymore but that every poll that does not belong to a motion or assignment belongs to a topic now? In this case, this field `required` has to be set to `true`.

Please also change the restriction wiki page: https://github.com/OpenSlides/OpenSlides/wiki/poll-restrictions

If the field is required, then you only have to change `Other/no content object` to `topic`.

This issue is urged, because this has to be fixed, before I make the required changes to the autoupdate-restrictor.

jsangmeister:
No, content_object_id is still not required, only `topic` was added as an additional thing to be voted about, nothing else was changed.

jsangmeister:
Additional info: The standalone polls are still not implemented in the client, but only in the backend, since it has not been a priority. However, it should be an easy task to extend the client accordingly, so we should IMO not remove this feature from the backend without need.

ostcar:
I talked about this feature with @emanuelschuetze on our "Entwickertreffen" and if I am not mistaken, he does not need a stand alone poll, if there are topic-polls.

If a topic-poll is something different then a standalone poll, could you check that the restriction wiki is correct? Should everyone in a meeting see a topic-poll or should a user need to see linked topic or at least the agenda?

jsangmeister:
> I talked about this feature with @emanuelschuetze on our "Entwickertreffen" and if I am not mistaken, he does not need a stand alone poll, if there are topic-polls.

I do not know about this, but as I said, it's easiest if we just leave it as-is with the option to add it easily later if needed. The initial thought behind this was that standalone polls could be used for short opinion gatherings like "What do we want to eat?"; things that are not on the agenda or even in the OS system. But I can imagine it is more a "community" feature for smaller assemblies instead of one for big conferences where everything has way more structure and order ;)

> If a topic-poll is something different then a standalone poll, could you check that the restriction wiki is correct? Should everyone in a meeting see a topic-poll or should a user need to see linked topic or at least the agenda?

[...]

PR #1170, opened 2022-01-20
Set poll/content_object_id to required

ostcar:
See discussion in #1161

This change is a blocker for the autoupdate-service. There are a lot of models.yml changes but I can not add them to the autoupdate-service, until the poll is fixed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -83,7 +83,7 @@ stop-dev:
 	docker-compose -f dev/docker-compose.dev.yml down --volumes
 
 run-dev-attach:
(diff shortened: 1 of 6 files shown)


ostcar:
I just changed the models.yml. I don't know the python code or why the tests fails.

If this should be mergent, someone else from the backend team has to fix this.

jsangmeister:
see https://github.com/OpenSlides/openslides-backend/issues/1161

jsangmeister:
Fixed the tests. Removed possibility to set a topic as `option/content_object_id` (we never use topics for options, I don't know why this was in there). Fixed a dev setup bug which broke the autocomplete in the shell.

r-peschke (approved):

PR #1174, opened 2022-01-24
Active archived meetings in clone meeting. Add some tests.

reiterl:


tests/system/action/meeting/test_clone.py
```diff
@@ -399,16 +399,29 @@ def test_limit_of_meetings_error(self) -> None:
...
+    def test_limit_of_meetings_error_archived_meeting(self) -> None:
         self.test_models["organization/1"]["limit_of_meetings"] = 1
         self.test_models["organization/1"]["active_meeting_ids"] = [3]
         self.test_models["meeting/1"]["is_active_in_organization_id"] = None
         self.set_models(self.test_models)
 
+        response = self.request("meeting.clone", {"meeting_id": 1})
+        self.assert_status_code(response, 400)
+        self.assertIn(
+            "You cannot clone an active meeting, because you reached your limit of 1 active meetings.",
```
r-peschke, line 411: 1. Shouldn't we change the text to "You cannot clone a meeting...", because in some cases the user tries to clone an inactive meeting? What do you think?
2. We should talk with Emanuel, because this solution gives the user the ability to reuse an archived meeting without beeing superadmin.

reiterl: Renamed the error message and update the tests.

reiterl: @emanuelschuetze  Is it okay to activate a cloned archived meeting?

r-peschke (approved):

PR #1188, opened 2022-02-01
Issue1181 Fix migration errors in update-event

r-peschke:


migrations/migrations/0012_committee_user_relation.py
```diff
@@ -234,65 +228,69 @@ def handle_user_restore_event(
...
+                self.new_accessor.get_model_ignore_deleted(f"meeting/{meeting_id}")[
+                    0
+                ].get("committee_id")
+                for meeting_id in meeting_ids
+            )
+        else:
+            new_committee_ids = old_committee_ids.copy()
+
+        old_cml_committee_ids = cast(List[int], user.get(user_cml_permission_field, []))
+        if user_cml_permission_field in event.data:
```
reiterl, line 258: Why should be `user_cml_permission_field` in event.data? This would only happen, if you migrate a part twice.

reiterl (commented):
Looked thru, I have found one possible problem.

jsangmeister (approved):
looks fine besides @reiterl's remark

migrations/migrations/0012_committee_user_relation.py
r-peschke: It appears, because in migrate_event the event was changed from original format (committee_$1_management_level) to new format (this one, committee_$can_manage_management_level). More can't be done in migrate_event, because the necessary meeting possibly still not exists. The (modified) event and, if it exists, the user in this state, are gathered in a list self.event_user.
In get_additional_events all events are stored in "database", the necessary meeting is created. Now get_additional_events works on the list self.event_users and calls the handle_user_update_event (and others) with the same event (in new format, see above) again.

PR #1194, opened 2022-02-03
Update the permission check of the committee of meeting.clone.

reiterl:
Remove the permission and remove the permission test of this permission.

### DIFF
--- a/openslides_backend/action/actions/meeting/clone.py
+++ b/openslides_backend/action/actions/meeting/clone.py
@@ -291,13 +291,20 @@ class MeetingClone(MeetingImport):
         )
 
     def check_permissions(self, instance: Dict[str, Any]) -> None:
-        if payload_committee_id := instance.get("committee_id"):
-            if not has_committee_management_level(
-                self.datastore,
-                self.user_id,
-                CommitteeManagementLevel.CAN_MANAGE,
-                payload_committee_id,
-            ):
-                raise PermissionDenied(
-                    f"Missing {CommitteeManagementLevel.CAN_MANAGE.get_verbose_type()}: {CommitteeManagementLevel.CAN_MANAGE} for committee {payload_committee_id}"
-                )
+        if instance.get("committee_id"):
+            committee_id = instance["committee_id"]
+        else:
+            meeting = self.datastore.fetch_model(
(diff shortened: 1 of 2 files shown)


openslides_backend/action/actions/meeting/clone.py
```diff
@@ -291,23 +291,7 @@ def build_write_request_helper(
...
-            CommitteeManagementLevel.CAN_MANAGE,
-            committee_id,
-        ):
-            raise PermissionDenied(
-                f"Missing {CommitteeManagementLevel.CAN_MANAGE.get_verbose_type()}: {CommitteeManagementLevel.CAN_MANAGE} for committee {committee_id}"
-            )
-        if (
-            payload_committee_id := instance.get("committee_id")
-        ) and payload_committee_id != committee_id:
+        if payload_committee_id := instance.get("committee_id"):
```
jsangmeister, line 294: `committee_id` is not required in the payload. As stated in the linked issue, if it is not given, the target meeting is equivalent to the source meeting and the permission check has to be done as before.

tests/system/action/meeting/test_clone.py
```diff
@@ -519,28 +519,6 @@ def test_permissions_oml_can_manage(self) -> None:
             "meeting/2", {"is_active_in_organization_id": 1, "committee_id": 2}
         )
 
-    def test_permissions_missing_meeting_committee_permission(self) -> None:
```
jsangmeister, line 522: Also leave this test in and check instead that it suceeds to "prove" that the changes work

jsangmeister (changes requested):

openslides_backend/action/actions/meeting/clone.py
reiterl: Ah, I understand.

reiterl: If a committee_id is in the payload, I use this for the check, else I would use the committee_id from the meeting.

tests/system/action/meeting/test_clone.py
reiterl: Add a test, which shows, that the permission check is in place.

r-peschke (approved):

PR #1201, opened 2022-02-09
Opentelemetry

peb-adr:
This is the kinda more serious and up-to-date version of #1121 .

In addition to ostcars PR otel can be turned on/off via env var.
For this I put otel-related functionality into a seperate module, which handles initiation and span creation.
Feel free to extend. E.g. I can imagine the `make_span` method could use some more parameters at some point.

Also I adjusted the compose-files and added suitable make targets to be able to view traces for the dev-stack.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -122,10 +122,10 @@ run-dev-otel: | start-dev-otel run-dev-attach
 # Build and run development container with local datastore and opentelemetry enabled and jaeger tracing (web UI binds to port 16686)
 
 start-dev-local-ds-otel:
-	USER_ID=$$(id -u $${USER}) GROUP_ID=$$(id -g $${USER}) docker-compose -f dev/docker-compose.dev.yml -f dev/dc.otel.yml -f dev/dc.otel.yml up --build --detach
+	USER_ID=$$(id -u $${USER}) GROUP_ID=$$(id -g $${USER}) docker-compose -f dev/docker-compose.dev.yml -f dev/dc.local-ds.yml -f dev/dc.otel.yml up --build --detach
 
 stop-dev-local-ds-otel:
-	docker-compose -f dev/docker-compose.dev.yml -f dev/dc.otel.yml -f dev/dc.otel.yml down --volumes
+	docker-compose -f dev/docker-compose.dev.yml -f dev/dc.local-ds.yml -f dev/dc.otel.yml down --volumes
 
 run-dev-local-ds-otel: | start-dev-local-ds-otel run-dev-attach
 
--- a/Makefile
+++ b/Makefile
@@ -108,6 +108,28 @@ run-dev-attach-local-ds:
 run-dev-local-ds run-bash-local-ds: | start-dev-local-ds run-dev-attach-local-ds
 
 
+# Build and run development container with opentelemetry enabled and jaeger tracing (web UI binds to port 16686)
+
+start-dev-otel:
+	USER_ID=$$(id -u $${USER}) GROUP_ID=$$(id -g $${USER}) docker-compose -f dev/docker-compose.dev.yml -f dev/dc.otel.yml up --build --detach
+
+stop-dev-otel:
+	docker-compose -f dev/docker-compose.dev.yml -f dev/dc.otel.yml down --volumes
+
+run-dev-otel: | start-dev-otel run-dev-attach
+
+
+# Build and run development container with local datastore and opentelemetry enabled and jaeger tracing (web UI binds to port 16686)
+
+start-dev-local-ds-otel:
+	USER_ID=$$(id -u $${USER}) GROUP_ID=$$(id -g $${USER}) docker-compose -f dev/docker-compose.dev.yml -f dev/dc.otel.yml -f dev/dc.otel.yml up --build --detach
+
+stop-dev-local-ds-otel:
+	docker-compose -f dev/docker-compose.dev.yml -f dev/dc.otel.yml -f dev/dc.otel.yml down --volumes
+
+run-dev-local-ds-otel: | start-dev-local-ds-otel run-dev-attach
+
+
 # Build standalone development container (not usable inside the docker container)
 
 build-dev:
--- a/dev/dc.otel.yml
+++ b/dev/dc.otel.yml
@@ -1,0 +1,22 @@
+version: "3"
+services:
+
+  collector:
+    image: otel/opentelemetry-collector:0.41.0
+    command: ["--config=/etc/otel-collector-config.yml"]
+    volumes:
+      - ./otel-collector-config.yml:/etc/otel-collector-config.yml
+    networks:
+      - datastore
+      - auth
+      - otel
+
+  jaeger:
+    image: jaegertracing/all-in-one
+    ports:
+      - "16686:16686"
+    networks:
+      - otel
+
+networks:
+  otel:
--- a/dev/dc.otel.yml
+++ b/dev/dc.otel.yml
@@ -1,5 +1,20 @@
 version: "3"
 services:
+  backend:
+    environment:
+      - OPENTELEMETRY_ENABLED=1
+  reader:
+    environment:
+      - OPENTELEMETRY_ENABLED=1
+  writer:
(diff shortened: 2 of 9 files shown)


Makefile
```diff
@@ -108,6 +108,28 @@ run-dev-attach-local-ds:
...
+stop-dev-otel:
+	docker-compose -f dev/docker-compose.dev.yml -f dev/dc.otel.yml down --volumes
+
+run-dev-otel: | start-dev-otel run-dev-attach
+
+
+# Build and run development container with local datastore and opentelemetry enabled and jaeger tracing (web UI binds to port 16686)
+
+start-dev-local-ds-otel:
+	USER_ID=$$(id -u $${USER}) GROUP_ID=$$(id -g $${USER}) docker-compose -f dev/docker-compose.dev.yml -f dev/dc.otel.yml -f dev/dc.otel.yml up --build --detach
```
jsangmeister, line 125:
```suggestion
	USER_ID=$$(id -u $${USER}) GROUP_ID=$$(id -g $${USER}) docker-compose -f dev/docker-compose.dev.yml -f dev/dc.otel.yml up --build --detach
```

Makefile
```diff
@@ -108,6 +108,28 @@ run-dev-attach-local-ds:
...
+
+stop-dev-local-ds-otel:
+	docker-compose -f dev/docker-compose.dev.yml -f dev/dc.otel.yml -f dev/dc.otel.yml down --volumes
```
jsangmeister, line 128:
```suggestion
	docker-compose -f dev/docker-compose.dev.yml -f dev/dc.otel.yml down --volumes
```

dev/docker-compose.dev.yml
```diff
@@ -19,6 +19,7 @@ services:
             - ../global:/app/global
             - ../scripts:/app/scripts
         environment:
+            - OPENTELEMETRY_ENABLED=0
```
jsangmeister, line 22: Isn't the default value `false` anyway? Then I would just remove it from the dc file so that it doesn't become even more bloated.

dev/otel-collector-config.yml
```diff
@@ -0,0 +1,21 @@
...
+
+processors:
+  batch:
+
+service:
+  pipelines:
+    traces:
+      receivers: [otlp]
+      processors: [batch]
+      exporters: [ jaeger]
```
jsangmeister, line 21:
```suggestion
      exporters: [jaeger]
```

openslides_backend/shared/otel.py
```diff
@@ -0,0 +1,56 @@
+import os
```
jsangmeister, line 1: Since this file doubles so much with the datastore one, you could just import the functionality from the datastore here, since it is embedded as a python package. You could still keep the file so it is easily extendable for the backend if needed (and for backend specific functionality, like the requests instructor).

requirements/partial/requirements_production.txt
```diff
@@ -9,3 +9,7 @@ bleach==4.1.0
 PyPDF2==1.26.0
 lxml==4.7.1
 git+https://github.com/OpenSlides/openslides-auth-service.git@9e252f33ccde41137a0a7174e8958d8fa3beb2f1#egg=authlib&subdirectory=auth/libraries/pip-auth # authlib
+opentelemetry-exporter-otlp
+opentelemetry-api
+opentelemetry-sdk
+opentelemetry-instrumentation-requests
```
jsangmeister, line 15: This also needs to be added to the dev requirements

openslides_backend/shared/otel.py
```diff
@@ -0,0 +1,56 @@
...
+
+"""
+Initializes the opentelemetry components and connection to the otel collector.
+"""
+def init(service_name):
+    if not is_truthy(os.environ.get("OPENTELEMETRY_ENABLED", "false")):
+        return
+
+    span_exporter = OTLPSpanExporter(
+        endpoint=f"http://collector:4317",
```
jsangmeister, line 21: Obsolete if you import this form the datastore, but this should probably also be configurable via env var

dev/dc.otel.yml
```diff
@@ -0,0 +1,22 @@
+version: "3"
```
jsangmeister, line 1: You also need to overwrite `OPENTELEMETRY_ENABLED` for backend, reader and writer here so that it works in the dev setup, or do you do this somewhere else?

jsangmeister (changes requested):

openslides_backend/shared/otel.py
ostcar: There already is an environment variable. See https://opentelemetry-python.readthedocs.io/en/latest/exporter/otlp/otlp.html

Since OpenTelementry is build for micro-services, they made a standard for environment variables. With this variables, you do not only configure your own services, but also other servies like caddy or redis (if they add support for OpenTelementry in the future).

I would recommend, that we use this environment variables instead of inventing our own.

dev/dc.otel.yml
ostcar: I would propose not to use `OPENTELEMETRY_ENABLED` at all.  To disable OpenTelementry, you can set the sammpler to off with the default varialbes. If the sampler is off, then all opentelementry functions are noops

requirements/partial/requirements_production.txt
jsangmeister: You are correct! my bad

openslides_backend/shared/otel.py
jsangmeister: I agree

dev/dc.otel.yml
jsangmeister: also good for me, but I would still argue to enable it in the docker compose file, because there is not really a "manual" way to activate it for us developers in the dev setup.

peb-adr: @ostcar @jsangmeister 
while I agree that we should use the default `OTEL_*` env vars for configuring e.g. the the endpoint of the exporter if needed. I think having our own var for turning it off alltogether can be more assuring for us, that there will no logging of any potentially sensitive data in production.
The introduction of the [env var specification](https://opentelemetry.io/docs/reference/specification/sdk-environment-variables/) reads

> The goal of this specification is to unify the environment variable names between different OpenTelemetry SDK implementations. SDKs MAY choose to allow configuration via the environment variables in this specification, but are not required to. If they do, they SHOULD use the names listed in this document.

Which doesn't make me comfortable in relying on every SDK implementation we might use in the future to yield the same behavior for the the same env var values passed. It would require careful testing everytime.

reiterl:
@peb-adr here is some work to do.

openslides_backend/shared/otel.py
jsangmeister: Can you change this to use the OTEL env vars?

openslides_backend/shared/otel.py
```diff
@@ -0,0 +1,62 @@
...
+from typing import Any, Dict
+
+from .env import is_truthy
+
+
+def init(service_name: str) -> None:
+    """
+    Initializes the opentelemetry components and connection to the otel collector.
+    """
+    if not is_truthy(os.environ.get("OPENTELEMETRY_ENABLED", "false")):
```
jsangmeister, line 19: Define `OPENTELEMETRY_ENABLED` as its own variable in `env.py` (or maybe diectly use a explicit helper function, similar to `is_dev_mode`

jsangmeister (changes requested):

reiterl:
Strange mypy errors. I don't know why mypy doesn't find the attributes of the opentelemetry module. It should be installed.

jsangmeister:
@reiterl take a look at https://githubhot.com/repo/open-telemetry/opentelemetry-python/issues/1957 and check if this already solves the problem. Also the other problems (isort/flake) and remarks from the review can be resolved independently

openslides_backend/shared/otel.py
reiterl: Use two helper function in env.py to get otel url and is otel enabled.

PR #1207, opened 2022-02-11
Add motion_state.sort action and add tests.

reiterl:


tests/system/action/motion_state/test_sort.py
```diff
@@ -0,0 +1,44 @@
+from openslides_backend.permissions.permissions import Permissions
```
jsangmeister, line 1: Add 2 more tests:
- an id that is present in the meeting is missing in the payload
- the payload has an extra id which does not exist in the meeting

jsangmeister (changes requested):

tests/system/action/motion_state/test_sort.py
reiterl: Added two additional test: one with extra id and one with missing id.

jsangmeister (approved):

PR #1208, opened 2022-02-11
Update the unique name check of chat_groups.

reiterl:
It checks that the names unique in a meeting.

### DIFF
--- a/openslides_backend/action/actions/chat_group/mixins.py
+++ b/openslides_backend/action/actions/chat_group/mixins.py
@@ -1,21 +1,32 @@
 from typing import Any, Dict
 
+from ....models.models import Model
+from ....services.datastore.interface import DatastoreService
 from ....shared.exceptions import ActionException
-from ....shared.filters import FilterOperator
+from ....shared.filters import And, FilterOperator
 from ....shared.patterns import Collection, FullQualifiedId
 from ...action import Action
 
 ONE_ORGANIZATION_ID = 1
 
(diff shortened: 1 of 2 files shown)


openslides_backend/action/actions/chat_group/mixins.py
```diff
@@ -1,21 +1,32 @@
...
+    model: Model
+
+    def fetch_meeting_id(self, instance: Dict[str, Any]) -> int:
         meeting_id = instance.get("meeting_id")
         if not meeting_id:
             chat_group = self.datastore.get(
                 FullQualifiedId(self.model.collection, instance["id"]), ["meeting_id"]
             )
             meeting_id = chat_group["meeting_id"]
+        return meeting_id
```
jsangmeister, line 24: This should be the same code as in https://github.com/OpenSlides/openslides-backend/blob/16f989d520aba0d5241031a34eb9f4a187e6f631/openslides_backend/action/action.py#L235 so you should be able to remove this class completely and just use the method `get_meeting_id` from the action class

jsangmeister (changes requested):

openslides_backend/action/actions/chat_group/mixins.py
reiterl: Remove `fetch_meeting_id` and use `get_meeting_id.`

jsangmeister (approved):

PR #1211, opened 2022-02-14
Update user.forget_password to include the orga.url

reiterl:
Update the model for cleanup. Add two migrations. Update tests and actions.


migrations/migrations/0016_remove_users_pdf_url.py
```diff
@@ -0,0 +1,13 @@
+from datastore.migrations import RemoveFieldMigration
```
jsangmeister, line 1: merge into one migration for less overhead

openslides_backend/action/actions/user/send_invitation_email.py
```diff
@@ -191,10 +191,8 @@ def __missing__(self, key: str) -> str:
             "password": user.get("default_password", ""),
             **subject_format,
         }
-        if meeting_id:
-            body_dict["url"] = mail_data.get("users_pdf_url", "")
-        else:
-            body_dict["url"] = mail_data.get("url", "")
+
+        body_dict["url"] = mail_data.get("url", "")
```
jsangmeister, line 195: move into definition of the `body_dict` above

jsangmeister (changes requested):

jsangmeister:
@GabrielInTheWorld (and @emanuelschuetze, I guess): These changes have to be reflected in the client before a new version of the services is pushed to the main repo, since the client sitll uses the old fields.

migrations/migrations/0016_remove_users_pdf_url.py
reiterl: Why that? The two migration make their work and are configured default migrations.

jsangmeister: Two reasons:
1) Each migration produces additional overhead. Going once through thousands of positions and applying some changes is more performant than doing the same thing twice.
2) Same or similar changes should be gathered in one migration to for easier overview.
The latter is discussable since "remove field x" might also be seen as a simple atomic change, but the more important reason IMO is 1), anyway.

openslides_backend/action/actions/user/send_invitation_email.py
reiterl: Done.

migrations/migrations/0016_remove_users_pdf_url.py
reiterl: Okay, write a merged migration and updated index.

jsangmeister (approved):

PR #1221, opened 2022-02-17
Fill entitled_users with all entitled users

### LINKED ISSUE #1220: wrong list of entitled users
For a poll the calculated list of entitled users is wrong.

Currently there are listed all _present_ users of meeting. That's not correct.
Expected: all users of defined voting group independent of present state (e.g. all delegates).

Probably [this for loop in poll/mixins.py#L165](https://github.com/OpenSlides/openslides-backend/blob/main/openslides_backend/action/actions/poll/mixins.py#L165) has to be fixed...

jsangmeister:
fixes #1220


emanuelschuetze (changes requested):
now, _all_ users are listed as entitled users

emanuelschuetze (approved):

PR #1229, opened 2022-02-21
Add an export presenter.

reiterl:
Needs a meeting_id and uses export_meeting helper to export the meeting
of the meeting_id.

### DIFF
--- a/openslides_backend/presenter/__init__.py
+++ b/openslides_backend/presenter/__init__.py
@@ -1,6 +1,7 @@
 from . import (  # noqa
     check_mediafile_id,
     check_resource_id,
+    export,
     get_active_users_amount,
     get_forwarding_meetings,
     get_history_information,
--- a/openslides_backend/presenter/__init__.py
+++ b/openslides_backend/presenter/__init__.py
@@ -1,7 +1,7 @@
 from . import (  # noqa
     check_mediafile_id,
     check_resource_id,
-    export,
+    export_meeting,
     get_active_users_amount,
     get_forwarding_meetings,
     get_history_information,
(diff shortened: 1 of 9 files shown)


openslides_backend/presenter/__init__.py
```diff
@@ -1,6 +1,7 @@
 from . import (  # noqa
     check_mediafile_id,
     check_resource_id,
+    export,
```
jsangmeister, line 4: Maybe `export_meeting` is more concise?

jsangmeister (changes requested):
Change name and add migration index to the exported file. You can find the method for this here: https://github.com/OpenSlides/openslides-backend/blob/f0fe824b8199011543732114240da11334828f3e/migrations/__init__.py#L32 You have to split this method up and extract the datastore migration index.

openslides_backend/presenter/__init__.py
reiterl: Okay, I renamed it into `export_meeting`.

reiterl:
Add _migration_index to export meeting presenter. Extract the way to get the mig index from the assert_migration_index()

jsangmeister (changes requested):
Did not realize that both migration indices must be the same so you can of course also use the backend migration index. But while checking this, I noticed that an `assert_migration_index` is still missing for the presenter view at this line: https://github.com/OpenSlides/openslides-backend/blob/811cde32605ab62799b0c37f557a8540dd6ed03c/openslides_backend/http/views/presenter_view.py#L20 (analogous to https://github.com/OpenSlides/openslides-backend/blob/2c1493c2c10143d9ed0d9a6a5fc362ff5169eced/openslides_backend/http/views/action_view.py#L26)

Please add this to make sure that the migration indices are alway equal.

jsangmeister (approved):

PR #1231, opened 2022-02-21
Issue1215 Adjust  is_public calculation for mediafiles, calculate always

r-peschke:
Can be reviewed, migrations are ready. Still missing recreation with calculated field handlers

### DIFF
--- a/migrations/migrations/0019_calc_mediafile_is_public.py
+++ b/migrations/migrations/0019_calc_mediafile_is_public.py
@@ -1,5 +1,5 @@
 from collections import defaultdict
-from typing import Any, Dict, List, Optional, Set, cast
+from typing import Any, Dict, List, Optional, Set, Tuple, cast
 
 from datastore.migrations import (
     BaseEvent,
@@ -9,7 +9,6 @@ from datastore.migrations import (
     UpdateEvent,
 )
 from datastore.shared.util import collection_and_id_from_fqid
-from helper.helper import calculate_inherited_groups_helper
 
 
 class Migration(BaseMigration):
@@ -133,3 +132,27 @@ class Migration(BaseMigration):
         for child in self.mediafiles[id_].get("childs", []) or []:
             if changed or child in self.mediafiles.keys():
                 self.check_recursive(child)
+
+def calculate_inherited_groups_helper(
+    access_group_ids: Optional[List[int]],
+    parent_is_public: Optional[bool],
+    parent_inherited_access_group_ids: Optional[List[int]],
+) -> Tuple[bool, List[int]]:
+    """
+    TODO: If migration and backend are combined one day,
+    import this function from openslides_backend/models/helper.py
+    """
+    inherited_access_group_ids: List[int]
+    is_public = False
+    if parent_inherited_access_group_ids and access_group_ids:
+        inherited_access_group_ids = [
+            id_ for id_ in access_group_ids if id_ in parent_inherited_access_group_ids
+        ]
+    elif access_group_ids:
+        inherited_access_group_ids = access_group_ids
+    elif parent_inherited_access_group_ids:
+        inherited_access_group_ids = parent_inherited_access_group_ids
+    else:
+        is_public = parent_is_public is not False
+        inherited_access_group_ids = []
+    return is_public, inherited_access_group_ids
--- a/migrations/migrations/0019_calc_mediafile_is_public.py
+++ b/migrations/migrations/0019_calc_mediafile_is_public.py
@@ -129,9 +129,9 @@ class Migration(BaseMigration):
             )
             self.events.append(update_event)
 
-        for child in self.mediafiles[id_].get("childs", []) or []:
-            if changed or child in self.mediafiles.keys():
-                self.check_recursive(child)
+        for child_id in self.mediafiles[id_].get("child_ids", []) or []:
+            if changed or child_id in self.mediafiles.keys():
+                self.check_recursive(child_id)
 
 
 def calculate_inherited_groups_helper(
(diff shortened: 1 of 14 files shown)


tests/system/action/mediafile/test_upload.py
```diff
@@ -199,3 +166,157 @@ def test_create_added_mimetype_ttf(self) -> None:
...
+        self.media.upload_mediafile.assert_called_with(file_content, 11, "text/plain")
+        self.assert_model_exists(
+            "group/1",
+            {
+                "mediafile_access_group_ids": [11],
+                "mediafile_inherited_access_group_ids": [11],
+            },
+        )
+
+    def test_createX_media_directory_group(self) -> None:
```
jsangmeister, line 222:
```suggestion
    def test_create_media_directory_group(self) -> None:
```

jsangmeister (approved):
looks good besides missing migrations. I'm not sure how a migration for this would be implemented the best - theoretically, you would need to recalculate the field (call the `calculated_fields_helper`) for each media file, top-down in the meeting's tree structure. This is hard to do in a migration. Maybe we rather provide a small script which does exactly this and can be called manually for instances where this problem occurs? It would then look like this:
- fetch all mediafiles from this instance with their `parent_id`
- rebuild the tree structure for each meeting
- go through each tree bottom-down, call `mediafile.update` for each mediafile with only the id as payload

This is basically also what would need to be done in a migration, but seems way harder to implement.

jsangmeister:
> looks good besides missing migrations. I'm not sure how a migration for this would be implemented the best - theoretically, you would need to recalculate the field (call the `calculated_fields_helper`) for each media file, top-down in the meeting's tree structure. This is hard to do in a migration. Maybe we rather provide a small script which does exactly this and can be called manually for instances where this problem occurs? It would then look like this:
> 
> * fetch all mediafiles from this instance with their `parent_id`
> * rebuild the tree structure for each meeting
> * go through each tree bottom-down, call `mediafile.update` for each mediafile with only the id as payload
> 
> This is basically also what would need to be done in a migration, but seems way harder to implement.

Talked with @r-peschke and came to the conclusion that a separate script is not the correct solution, but a "one-shot" update is  better and less work then a "real" migration where all events are updated. Will look into the datastore code and check how easily this is possible.

jsangmeister:
As discussed: A "real" migration is the way to go, "one-shot" migrations are not easily implementable

migrations/migrations/0019_calc_mediafile_is_public.py
```diff
@@ -0,0 +1,202 @@
...
+                self.mediafiles[id_] = self.new_accessor.get_model_ignore_deleted(
+                    event.fqid
+                )[0]
+            if remove_dict := event.remove:
+                for key, value in remove_dict.items():
+                    if not value:
+                        continue
+                    assert isinstance(
+                        old_value := self.mediafiles[id_].get(key), list
+                    ), f"'{event.fqid}' should have values for '{key}', because there is a ListUpdate.remove!"
```
jsangmeister, line 68: This is wrong, see https://github.com/OpenSlides/OpenSlides/wiki/Datastore-Service#interface: If a `remove` happens on a non-existent field, it is simply ignored.

migrations/migrations/0019_calc_mediafile_is_public.py
```diff
@@ -0,0 +1,202 @@
...
+
+
+def calculate_inherited_groups_helper(
+    access_group_ids: Optional[List[int]],
+    parent_is_public: Optional[bool],
+    parent_inherited_access_group_ids: Optional[List[int]],
+) -> Tuple[bool, List[int]]:
+    """
+    TODO: If migration and backend are combined one day,
+    import this function from openslides_backend/models/helper.py
```
jsangmeister, line 187: Thought again about this and I think it's actually good to keep the migration separated from the actual code. So I would just remove this TODO.

openslides_backend/action/actions/group/delete.py
```diff
@@ -14,3 +22,159 @@ class GroupDeleteAction(DeleteAction):
...
     permission = Permissions.User.CAN_MANAGE
+
+    def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
+        instance = super().update_instance(instance)
+        group = self.datastore.get(
+            FullQualifiedId(Collection("group"), instance["id"]), []
+        )
+        self.mediafile_ids: List[int] = list(
+            (set(group.get("mediafile_access_group_ids", set())) or set())
+            | (set(group.get("mediafile_inherited_access_group_ids", set())) or set())
```
jsangmeister, line 33: If I counted the brackets correctly, you write here `set(...) or set()`, so you should be able to remove the second `set()`

jsangmeister (changes requested):
Did not check every test, but the gist of it looks good. Please go ahead and test this on actual meeting data:
1) Create a meeting and upload a mediafile (or multiple, as you wish) with the current backend main
2) Verify that cloning this meeting fails, since `is_public` is set wrong
3) Shutdown the backend and restart it with this PR
4) Migrate by executing `python migrations/migrate.py finalize`
5) Assert that you can clone the meeting now

Process described from the top of my head, may contain errors.

migrations/migrations/0019_calc_mediafile_is_public.py
r-peschke: done

openslides_backend/action/actions/group/delete.py
r-peschke: done

peb-adr (approved):
Ok, I tested this successfully now.
Tried provocating other errors by e.g. cloning the clone, removing and adding files in the clones and the original meeting with different users.
All seems stable to me now.

PR #1238, opened 2022-02-23
Use is_partial in checker and update tests.

reiterl:
What is meant with `empty`?
- Fields with null/None for sure
- Lists with []: I think, too, but ?
- Strings with "": I think, this is not empty!?


jsangmeister:
> What is meant with `empty`?
> 
> * Fields with null/None for sure
yes
> * Lists with []: I think, too, but ?
yes
> * Strings with "": I think, this is not empty!?
yes, if not required

You should be able to remove the "non-partial" mode altogether from the checker, I think (or is it then used anywhere? I think all data is partial then). (I did not look at the code yet)

reiterl:
I have remove the is_partial option. Is_partial is the new standard. Updated the checker.

openslides_backend/models/checker.py
```diff
@@ -169,14 +168,9 @@ def __init__(
             in data, because they exist in same database.
         all: All collections are valid and has to be in the data
 
-        is_partial=True disables the check, that *all* collections have to be
-        explicitly given, so a non existing (=empty) collection will not raise
-        an error. Additionally, missing fields (=None) are ok, if they are not
-        required nor have a default (so required fields or fields with defaults
-        must be present).
+        Not all collections must be given and missing fields are ignore.
```
jsangmeister, line 171:
```suggestion
        Not all collections must be given and missing fields are ignored, but required fields and fields with a default value must be present.
```

global/data/example-data.json
```diff
@@ -57,7 +56,6 @@
             "default_number": "",
             "default_structure_level": "",
```
jsangmeister, line 57: remove empty strings as well

openslides_backend/models/checker.py
```diff
@@ -362,11 +353,7 @@ def check_normal_fields(self, model: Dict[str, Any], collection: str) -> bool:
             for field in self.models[collection]().get_fields()
             if field.required or field.default is not None
         )
-        necessary_fields = (
-            required_or_default_collection_fields
-            if self.is_partial
-            else all_collection_fields
-        )
+        necessary_fields = required_or_default_collection_fields
```
jsangmeister, line 356: unnecessary variable

jsangmeister (changes requested):

openslides_backend/models/checker.py
reiterl: Updated.

openslides_backend/models/checker.py
reiterl: Removed.

global/data/example-data.json
reiterl: Empty strings removed.

global/data/example-data.json
```diff
@@ -4,9 +4,7 @@
         "1": {
```
jsangmeister, line 4: Some more fields that can be removed from both JSON files, if set to an empty/null value:
- all boolean variables which are set to false, if not required. (Maybe they are all required? Then it is fine)
- `meeting`: `custom_translations`, `start_time`, `end_time`
- `mediafile/pdf_information`
- `projection/options`
- empty collections in `example_data.json`: `motion_statute_paragraph`, `chat_message`
- all empty collections in `initial_data.json`

Also make sure to remove all unnecessary fields from the import tests.

jsangmeister (changes requested):

reiterl:
Cleanup of the initial data, remove most collections.

global/data/example-data.json
reiterl: Done.

global/data/initial-data.json
```diff
@@ -4,29 +4,16 @@
...
-            ],
-            "active_meeting_ids": [
-                1
-            ],
-            "archived_meeting_ids": [],
-            "template_meeting_ids": [],
-            "resource_ids": [],
-            "organization_tag_ids": [],
+            "committee_ids": [],
+            "active_meeting_ids": [],
```
jsangmeister, line 13: You should also be able to remove these

global/data/initial-data.json
```diff
@@ -36,50 +23,15 @@
...
-            "option_$_ids": [],
-            "vote_$_ids": [],
-            "vote_delegated_vote_$_ids": [],
-            "assignment_candidate_$_ids": [],
-            "projection_$_ids": [],
-            "vote_delegated_$_to_id": [],
-            "vote_delegations_$_from_ids": [],
-            "chat_message_$_ids": [],
-            "meeting_ids": []
+	    "meeting_ids": [],
```
jsangmeister, line 31: same for this

tests/system/action/meeting/test_import.py
```diff
@@ -351,10 +302,6 @@ def get_user_data(self, obj_id: int, data: Dict[str, Any] = {}) -> Dict[str, Any
             "id": obj_id,
             "password": "",
```
jsangmeister, line 303: Also remove this and all other empty strings. Same goes for all the other helper methods, there are still some fields left.

jsangmeister (changes requested):
Still some places left. Also remove the `needed_collections` part from `test_import.py` since they are no longer needed.

global/data/initial-data.json
reiterl: user/1 meeting_ids is needed.

global/data/initial-data.json
reiterl: Removed.

tests/system/action/meeting/test_import.py
reiterl: Some of the empty strings are needed, like password.

global/data/initial-data.json
reiterl: user/meeting_ids is a `NumberArrayField`. These fields are special fields with a default `default` of []. So the checker wants the `user/meeting_ids` field.

reiterl: Updated this.

tests/system/action/meeting/test_import.py
jsangmeister: But not all of them, something like `title` or `pronoun` should not be required. Also, for e.g. password, we should also note in the models file that it is required.

jsangmeister (changes requested):
See open comment. Also the `needed_collection` are still to be removed.

reiterl:
I have removed the needed_collections part from the test_import.

reiterl:
I removed some more empty fields from the meeting/test_import

tests/system/action/meeting/test_import.py
reiterl: Removed not needed (user) fields like `title`, `pronoun` and `password`.

reiterl: Removed not needed fields like `title`,`pronoun` and `password`.

jsangmeister (approved):

PR #1249, opened 2022-03-01
Some migration route fixes

### LINKED ISSUE #1227: Document migration HTTP routes
The migration command "progress" is not mentioned in the migrate.py but this command can be called via HTTP route. It would be nice if someone can cleanup this either by extending the migrate.py script and/or moving some logic from the HTTP route code to another layer.

### LINKED ISSUE #1248: Unify migrations output
Here https://github.com/OpenSlides/openslides-backend/blob/main/openslides_backend/migration_handler/migration_handler.py#L57 the migrations output is "message". In all other cases it is "output". Please use the same property in all cases.

Question for me: Is it correct that the content of "status" is always an integer (0, 1 or 2) and the content of "output" is always a string?

jsangmeister:
closes #1248 
closes #1227 

Also add some tests


jsangmeister:
Also added error handling to the migration routes which was previously missing. Please check if this suits your needs.

peb-adr (approved):
Looks good to me

docs/migration_route.md
```diff
@@ -0,0 +1,41 @@
...
+
+{
+    "success": true,
+    "status"?: MigrationProgressState,
+    "output"?: str,
+    "exception"?: str,
+}
+```
+The output differs slightly, depending on the command:
+- `migrate`/`finalize`/`progress`: `status` can only be 0 if `progress` is called while no migration is running, otherwise it will always be greater than 0. If `status` is 1, `output` contains the output until now of either the command issued in this request or, in the case of a `progress` command, of the threaded migration. If the migration is finished (`status=2`), the full output is returned and stored until a new migration is started, meaning repeated `progress` requests after this point will always return the same result with the full output.
```
normanjaeckel, line 29: Is it true that if migrations are required and I send the migrate or finalize command (not progress) the handler returns immediately and status is always 1? It can never be 2 in this case, right? What if no migrations are required? Is status then 0 or 2? If it is 2 I think this is missleading because there aren't any migrations that were recently finished.

I propose the following changes:

```js
enum MigrationProgressState {
    NO_MIGRATIONS_REQUIRED = "no_migrations_required"
    MIGRATIONS_REQUIRED = "migrations_required"
    MIGRATIONS_RUNNING = "migrations_running"
    PREPARED_MIGRATIONS_PENDING = "prepared_migrations_pending"
}
```

Of course the flags can be less verbose but this is what they should tell me.

Migrate, finalize progress and stats should always return a status. Only for reset and clear-collectionfied-tables the status can be omitted.

I also strongly recommend using strings and not integers. This is more verbose for the user of this API (that's me in the manage tool). If you insist in integers, I can live with it but then please do not use 0. The 0 should be the same as a not given property. This would make many things a lot easier for me in Golang code.

normanjaeckel (changes requested):

docs/migration_route.md
normanjaeckel: As discussed: Status also for reset.

jsangmeister:
Some rework together with https://github.com/OpenSlides/openslides-datastore-service/pull/182/. Please review again.

jsangmeister:
done

jsangmeister:
@normanjaeckel ping

docs/migration_route.md
```diff
@@ -0,0 +1,58 @@
...
+enum MigrationState {
+    MIGRATION_RUNNING = "migration_running"
+    MIGRATION_REQUIRED = "migration_required"
+    FINALIZATION_REQUIRED = "finalization_required"
+    NO_MIGRATION_REQUIRED = "no_migration_required"
+}
+
+{
+    "success": true,
+    "status"?: MigrationProgressState,
```
normanjaeckel, line 24: Typo: `MigrationState` instead of `MigrationProgressState`

normanjaeckel (approved):
I checked only the docs. Looks good to me. If I run in any issues in the future I will tell you. So this can be merged (after fixing the typo).

Issue #904, opened 2021-08-16
Reduce the number of get(-many) requests


FinnStutzenstein:
Before making the "big solution" with a cache layer one can improve single, seemingly heavy actions. This is an excerpt of tests from @reiterl and @jsangmeister:

```
=================================================
Anzahl get Aufrufe, Anzahl get_many Aufrufe, Test
=================================================
147 2 meeting.test_delete.MeetingDeleteActionTest.test_delete_full_meeting
16 0 meeting.test_create.MeetingCreateActionTest.test_create_simple
12 1 
motion_category.test_sort.MotionCategorySortActionTest.test_sort_complex_correct
12 1 agenda_item.test_sort.AgendaItemSortActionTest.test_sort_complex_correct
10 1 
list_of_speakers.test_delete_all_speakers.ListOfSpeakersDeleteAllSpeakersActionTester.test_delete_all_correct
9 2 meeting.test_import.MeetingImport.test_double_import
9 1 meeting.test_import.MeetingImport.test_replace_ids_and_write_to_datastore
7 1 projector.test_project.ProjectorProject.test_project
6 6 user.test_update.UserUpdateActionTest.test_update_change_group
6 1 user.test_update.UserUpdateActionTest.test_update_template_fields
=================================================
```
- Especially sorting (`motion_category.test_sort.MotionCategorySortActionTest.test_sort_complex_correct`, ...) and `list_of_speakers.test_delete_all_speakers.ListOfSpeakersDeleteAllSpeakersActionTester.test_delete_all_correct` seems to make lineary many get requests; This needs to be checked and improved
- `projector.test_project.ProjectorProject.test_project`: Seems to be heavy, since it is a very common operation during an assembly. It should be fast.

Generally: You should redo the measurements with a real setup. Maybe, when testing with the client, other pain points arise.

jsangmeister:
Just ran a few tests and you seem to be correct:
- `agenda_item.sort` with 15 items: Depending on the tree structure, about 80-90 GET calls
- `list_of_speakers.delete_all_speakers` with about 25 speakers: 104/16
- `projector.project`: about 17 GETs on my instance, but I don't exactly know on what it depends
We could start by locally optimizing a first action and see by how much we can improve this and then, if we want to, still implement the generic cache layer solution.

jsangmeister:
[...]

PR #1251, opened 2022-03-02
Multiple votes accumulated on one person

MJJojo97:
See [https://github.com/OpenSlides/OpenSlides/pull/6376](url)


global/meta/models.yml
```diff
@@ -2683,6 +2683,10 @@ poll:
     type: number
     default: 1
     restriction_mode: A
+  max_votes_per_option:
+    type: number
+    default: 1
```
jsangmeister, line 2688: Add `minimum: 1` and `required: true`

jsangmeister (changes requested):
The changes itself look good, but some tests for it are missing. You can add them by doing the following (I don't know how familiar you are with pytest and the development setup itself, so I'm trying to be rather more precise):
1) Replace the source in https://github.com/OpenSlides/openslides-backend/blob/a5ea1164f60ce939b4efb0fb51905ce2f5eccbab/dev/docker-compose.dev.yml#L93 with your own fork and branch so you can use the changed vote service code
2) Start the backend dev setup with `make run-dev`
3) Write your tests in the appropriate files (`test_{create|update|vote}.py`). The other tests should provide enough orientation on how to do this; if not, always feel free to ask for help. You should test at least the following things (you can also alter existing tests and just add the field, if appropriate):
  - you can set `max_votes_per_option` in `poll.create` and `poll.update`
  - you cannot set values <0 for the field
  - you cannot update this field if the poll is not in the `Created` state
  - if someone votes multiple times for the same option, the result is correct (note that the tests in `poll_vote.py` automatically start and stop the poll) - without and with vote weights
4) Execute your tests with `pytest` - see their docs for more details, but `pytest -k <test-name>` is most of the time the command of choice to not execute all tests, which is the default
5) Clean up your code by just executing `make` (inside the container!) and fix all errors which flake/mypy report
Also, you'll have to rebase again since the example data changed so the CI can ru successfully. Thank you for you contribution so far and feel free to ask if you run into problems!

openslides_backend/action/actions/poll/mixins.py
```diff
@@ -95,18 +95,25 @@ def on_stop(self, instance: Dict[str, Any]) -> None:
...
                         vote_value = poll["pollmethod"]
-                        vote_weight *= value
-
-                    option_results[option_id][vote_value] += vote_weight
+                        # print("Summing up")
+                        # print(vote_value)
+                        # print(value)
+                        # print(vote_weight)
+                        # print(ballot["weight"])
+                        # vote_weight *= value
```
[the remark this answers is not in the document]
MJJojo97: Changed in new commit.

global/meta/models.yml
MJJojo97: Changed in new commit.

MJJojo97:
Thank you for the explanations. They were very helpful!

I made all the changes. However, I currently just added the test cases for "<0" as comments because currently you actually can set negative values. Since this holds for values like max_votes_amount, too, I was not sure how I should fix this problem. Could you give me a hint on how I can fix that?

jsangmeister (approved):
waiting for client. vote service remote has to be changed before merging

dev/docker-compose.dev.yml
```diff
@@ -90,7 +90,7 @@ services:
             - cache
     vote:
         build:
-            context: "https://github.com/OpenSlides/openslides-vote-service.git#main"
+            context: "https://github.com/MJJojo97/openslides-vote-service.git#main"
```
ostcar, line 93: Is it possible to undo this line and still merge this PR?

tests/system/action/poll/test_create.py
```diff
@@ -828,3 +832,26 @@ def test_create_permissions_topic(self) -> None:
...
+    #             "global_abstain": True,
+    #             "onehundred_percent_base": "Y",
+    #             "min_votes_amount": 5,
+    #             "max_votes_amount": 10,
+    #             "max_votes_per_option": -1,
+    #             "backend": "long",
+    #         },
+    #     )
+    #     self.assert_status_code(response, 400)
+    #     self.assert_model_not_exists("poll/1")
```
ostcar, line 857: Can you either remove this test or uncomment it?

ostcar (changes requested):

ostcar:
Sry. Wrong click :(

MJJojo97:
I opened a new PR for the models.yml changes.

dev/docker-compose.dev.yml
MJJojo97: I removed this change. Can you try whether it is possible to merge it, now? Or what shall I do for testing that?

ostcar:
> waiting for client. vote service remote has to be changed before merging
The vote service is updated. We don't have to wait for the client, since the old client should be compatible with this change.

emanuelschuetze:
@ostcar Client is ready for merge. Can we merge this backend PR?

ostcar:
I don't want to give a review in the backend since I don't have enough knowledge.

If the backend developers approve it, then it can be merged.

emanuelschuetze (approved):
Tested with current client PR #904

PR #1253, opened 2022-03-04
Reader cache experiment

ostcar:
This is my experiment with a datastore reader cache. It only caches get and get_many requests.

I tested this with the user create action.

It reduces the amount of reader request by a factor of 3. There is still one cache miss and one filter-request per user.

I would expect, this to make the call faster. But it is nearly the same (sometimes a bit faster or a bit slower).

It seems, that my converting code is so slow, that there is no big speed gain.

Do you think this code can be optimized or is it a dead end?

### DIFF
--- a/openslides_backend/services/datastore/extended_adapter.py
+++ b/openslides_backend/services/datastore/extended_adapter.py
@@ -471,6 +471,6 @@ class ExtendedDatastoreAdapter(DatastoreAdapter):
             for fqid, fields in missing_fields_per_fqid.items()
         ]
         results = super().get_many(
-            get_many_requests, None, DeletedModelsBehaviour.ALL_MODELS, lock_result
+            get_many_requests, None, DeletedModelsBehaviour.NO_DELETED, lock_result
         )
         return results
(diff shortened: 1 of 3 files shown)


openslides_backend/services/datastore/extended_adapter.py
```diff
@@ -471,6 +471,6 @@ def _fetch_missing_fields_from_datastore(
             for fqid, fields in missing_fields_per_fqid.items()
         ]
         results = super().get_many(
-            get_many_requests, None, DeletedModelsBehaviour.ALL_MODELS, lock_result
+            get_many_requests, None, DeletedModelsBehaviour.NO_DELETED, lock_result
```
ostcar, line 474: @jsangmeister is this a bug in the current code? I think it is not necessary here to ask for all models

jsangmeister: Yes, this was probably a leftover from a previous state where I wanted to handle deleted models as well.

jsangmeister:
closing in favor of https://github.com/OpenSlides/openslides-backend/pull/1289

PR #1254, opened 2022-03-05
Added OPENSLIDES_LOGLEVEL and OPENSLIDES_BACKEND_NUM_WORKERS env vari…

normanjaeckel:
…ables.


openslides_backend/main.py
```diff
@@ -45,13 +47,28 @@ def __init__(self, view_name: str, *args: Any, **kwargs: Any) -> None:
 
         super().__init__(*args, **kwargs)
 
+    def loglevel(self) -> str:
+        if is_dev_mode():
+            return "DEBUG"
+        lvl = os.environ.get("OPENSLIDES_LOGLEVEL", "INFO").upper()
+        if lvl not in ("DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"):
+            raise ValueError("Invalid OPENSLIDES_LOGLEVEL: {lvl}")
+        return lvl
```
jsangmeister, line 56: I think this belongs in `openslides_backend/shared/env.py`

openslides_backend/main.py
```diff
@@ -45,13 +47,28 @@ def __init__(self, view_name: str, *args: Any, **kwargs: Any) -> None:
 
         super().__init__(*args, **kwargs)
 
+    def loglevel(self) -> str:
+        if is_dev_mode():
+            return "DEBUG"
```
jsangmeister, line 52: I think if given, the explicit log level should override the default one from the dev mode (and I think @ostcar also wanted this to be able to remove the log output while testing in dev mode). So, the handling should be the following:
- `OPENSLIDES_LOGLEVEL` set? use the given level
- else:
  - dev mode? -> `DEBUG`
  - else -> default log level `INFO`

Also for unity purposes, we should then put all default values into variables (like you already started doing) and use variables for the names of env vars (like in `env.py`)

jsangmeister (changes requested):

openslides_backend/main.py
normanjaeckel: Refactord

openslides_backend/main.py
normanjaeckel: Refactored

normanjaeckel:
I refactored some things. Please review again. I had to delete two tests. They were wrong and it is not easy to insert them again because of a missconfiguration of the test system (the tests use the environment of the host system and do not offer any possibility to override them inside a single test because everything is done inside a low level setUp() method.

openslides_backend/shared/env.py
```diff
@@ -1,28 +1,93 @@
...
 
-def get_internal_auth_password() -> str:
-    if is_dev_mode():
-        return DEV_PASSWORD
-    filename = os.environ.get(INTERNAL_AUTH_PASSWORD_FILE)
-    if filename:
-        with open(filename) as file_:
-            return file_.read()
-    else:
-        raise ServerError("No internal auth password specified.")
```
jsangmeister, line 28: I would keep this method in the new `Environment` class instead of moving it to the view.

openslides_backend/shared/env.py
```diff
@@ -1,28 +1,93 @@
...
+from typing import Any
 
-from .exceptions import ServerError
-
-INTERNAL_AUTH_PASSWORD_FILE = "INTERNAL_AUTH_PASSWORD_FILE"
-OPENSLIDES_DEVELOPMENT = "OPENSLIDES_DEVELOPMENT"
 DEV_PASSWORD = "openslides"
 
 
+class Loglevel(Enum):
```
jsangmeister, line 7:
```suggestion
class Loglevel(str, Enum):
```
will simplify the handling below, since you can then remove the `.name` when accessing members and can create members with `LogLevel(value)` etc.

openslides_backend/main.py
```diff
@@ -119,13 +120,22 @@ def sigterm_handler(signalnum: int, current_stack_frame: Any) -> None:
...
+    component = env.OPENSLIDES_BACKEND_COMPONENT
     if component == "action":
-        start_action_server()
+        start_action_server(env)
     elif component == "presenter":
-        start_presenter_server()
+        start_presenter_server(env)
     elif component == "all":
-        start_them_all()
+        if is_truthy(env.OPENSLIDES_DEVELOPMENT):
```
jsangmeister, line 130: Please also move the old `is_dev_mode` method into the `Environment` for easier access

tests/system/action/test_internal_actions.py
```diff
@@ -88,22 +74,3 @@ def test_internal_no_password_in_request(self) -> None:
...
-    @patch("openslides_backend.shared.env.is_dev_mode")
-    def test_internal_try_access_backend_internal_action(
-        self, is_dev_mode: Mock
-    ) -> None:
-        is_dev_mode.return_value = False
-        response = self.internal_request(
-            "option.create", {"meeting_id": 1, "text": "test"}
-        )
-        self.assert_status_code(response, 400)
-        self.assert_model_not_exists("option/1")
```
jsangmeister, line 109: If you keep the `is_dev_mode` method, you can also keep these tests. Also, how are they wrong?

jsangmeister (changes requested):

openslides_backend/main.py
normanjaeckel: Done.

openslides_backend/shared/env.py
normanjaeckel: I think this is part of the action view and does not belong to generic environment variable logic. And it is only used in this view. If we reuse it elsewhere we can move it to the shared module.

openslides_backend/shared/env.py
jsangmeister: Just tested it and `auto()` seems to always use the index as the value, so in this case `FOO.BAR.value == '1'`. So you would have to either assign the values manually (or use `FOO["BAR"]` instead, but this kinda eliminates the advantage of doing this change in the first place). I think the "double work" by manually assigning values is weighed out by the easier and more intuitive usage.

openslides_backend/shared/env.py
jsangmeister: I disagree, but we can leave it this way.

openslides_backend/shared/env.py
normanjaeckel: If it is ok for all, I would like to keep the code as is for the moment. The "right" way I think would be usinge a customized Enum class with customized `__getattr__` method. But this is not important at the moment because using `.name` i a few lines does not harm.

normanjaeckel:
I refactored some things more. It was more work than I expected. I hope it is ok for now. Minor things can be done in an extra issue. So if everything important is ok, please merge. Thanks. ;-)

tests/system/action/test_internal_actions.py
normanjaeckel: I reinserted the test but changed some things. The env vars are set only during `setUp` so it is not possible to manipulate them inside a single test. Now everything should work.

tests/system/action/test_internal_actions.py
```diff
@@ -89,21 +83,53 @@ def test_internal_no_password_in_request(self) -> None:
...
+    ) -> None:
+        is_dev_mode.return_value = False
+        response = self.internal_request(
+            "user.create", {"username": "my username"}, "wrong pw"
+        )
+        self.assert_status_code(response, 401)
+        self.assert_model_not_exists("user/2")
+
+    @patch("openslides_backend.shared.env.Environment.is_dev_mode")
     def test_internal_try_access_backend_internal_action(
```
jsangmeister, line 127: This test is wrong now: Backend-internal actions should only be accessed from inside the backend (and, for testing purposes, in dev mode from the outside). But in production mode, they can never be accessed from the outside which this test previously reflected. I don't know why this suddenly succeeds since the check in `action_hanlder.py` is still present, but please fix this before merging.

openslides_backend/shared/interfaces/env.py
```diff
@@ -0,0 +1,17 @@
+from typing import Protocol
+
+
+class Env(Protocol):  # pragma: no cover
+    """
+    Interface for the object containing all environment variables and some
+    helper methods.
+    """
+
+    def __getattr__(self, attr: str) -> str:
```
jsangmeister, line 10: minor remark: I don't think this method should be part of the protocol

jsangmeister (changes requested):

tests/system/action/test_internal_actions.py
normanjaeckel: Sorry, I do not know how to test that the route can not be accessed from outside. I even do not know how the backend should know what is "inside" and "outside". I think we have the proxy config for this. So I think the test is correct. It shows that one can reach the internal action using the correct password.

openslides_backend/shared/interfaces/env.py
normanjaeckel: Mypy complains if I remove this ...

jsangmeister: Mmh ok, then nevermind

tests/system/action/test_internal_actions.py
jsangmeister: Sorry, I somehow missed that you also changed the called action from `option.create` to `user.create`. So, yes, this test is correct, but we still also need the old test with `option.create`.

There are 3 types of backend actions (see also https://github.com/OpenSlides/openslides-backend/blob/2c1493c2c10143d9ed0d9a6a5fc362ff5169eced/openslides_backend/action/util/action_type.py): "normal" or "public" actions which are accessible from the "outside" (i.e. the client), "stack internal" actions which are only accessible from inside the stack (i.e. the manage service) and "backend internal" which should not be accessed from anyone else but the backend and are only used for internal operations. This test tests that a "backend internal" action cannot be accessed from the outside, even from inside the stack with the correct password. "Outside meaning here via HTTP.

So, please add back the previous test, which asserted that `option.create` cannot be called from outside of the backend.

normanjaeckel: Done.

Issue #1130, opened 2022-01-08
Add sequential numbers for important meeting-specific collections


jsangmeister:
Analogous to `motion/sequential_number`, we need such a field for all collections with a detail view in the client, namely (from a short grep of the client):
- assignment
- motion
- motion_block
- motion_category
- motion_workflow
- poll
- projector
- topic
- list_of_speakers
- motion_statute_paragraph
- motion_comment_section

The code form motion can be extracted and reused in a mixin, so it should not be much work. We need a migration which adds a unique sequential number to each motion, counting up for each meeting individually, so it looks like it's been there all along.

GabrielInTheWorld:
I found some additional collections with detail view:
- list_of_speakers
- motion_statute_paragraph
- motion_comment_section

(users have also a detail view within a meeting, but I think, we might leave this as it is)

jsangmeister:
Thanks, added it to the issue description. For users, this is not even possible, since they do not belong to a single meeting. We would have to generate a sequential number for each meeting the user is part of, which would get _really_ messy. But fortunately, as you say, I don't think it's needed.

reiterl:
These are new fields. What restriction mode should they have?

reiterl:
There exists a mixin in `motion/sequential_number_mixins.py`. I will use it.

GabrielInTheWorld:
> These are new fields. What restriction mode should they have?

If this is still necessary to know: Since the sequential_number of any model should be used to navigate instead of its id, a sequential_number should have the same restriction mode as the id. This should be in the most cases restriction mode `A`.

jsangmeister:
> There exists a mixin in `motion/sequential_number_mixins.py`. I will use it. This code uses the `datastore.max` method. This is a problem, because in an action with multiple payload, this leads to not unique sequential_numbers. I think @jsangmeister is working on a fix.

Yes, you can just ignore this for now

jsangmeister:
> > These are new fields. What restriction mode should they have?
> 
> If this is still necessary to know: Since the sequential_number of any model should be used to navigate instead of its id, a sequential_number should have the same restriction mode as the id. This should be in the most cases restriction mode `A`.

I think the same. This also means that the restriction mode of `motion/sequential_number` has to be adjusted (currently C).

reiterl:
How should the migration work? The simple approach with a class/object attribute in the migration is too short, because if you migrate and finalize it will deliver wrong numbers.

[...]

PR #1256, opened 2022-03-08
#1130 Second migration

r-peschke:
Repeats the migration 0010_add_sequential_numbers, because there are not unique sequence_numbers under special circumstances.


peb-adr (approved):
Ok, I tested this successfully now.
Tried provocating other errors by e.g. cloning the clone, removing and adding files in the clones and the original meeting with different users.
All seems stable to me now.

EDIT: whoops wrong issue ...

Issue #1203, opened 2022-02-10
Add meeting id to all filters where it is possible


jsangmeister:
Everytime a filter-based request (filter, min/max, ...) is made, a lock based on the filter is created. It is cirtical for the datastore that this filter contains a meeting id (if possible) so that the database query can be excuted performant. So, alls calls like this should be checked and a filter for the meeting id added, if possible. The exception from this are of course organization-wide objects like users etc.

reiterl:
Done with #1257 .

PR #1257, opened 2022-03-08
Add meeting_id filter to some filters 

reiterl:
I am not sure if I understand the issue https://github.com/OpenSlides/openslides-backend/issues/1203 .So I did some update
of filters as example. Before I continue here @jsangmeister  need to have a look.

### DIFF
--- a/openslides_backend/action/actions/motion/update.py
+++ b/openslides_backend/action/actions/motion/update.py
@@ -5,8 +5,8 @@ from typing import Any, Dict
 from ....models.models import Motion
 from ....permissions.permission_helper import has_perm
 from ....permissions.permissions import Permissions
+from ....services.datastore.commands import GetManyRequest
 from ....shared.exceptions import ActionException, PermissionDenied
-from ....shared.filters import FilterOperator
 from ....shared.patterns import (
     KEYSEPARATOR,
     POSITIVE_NUMBER_REGEX,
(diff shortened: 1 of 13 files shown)


reiterl:
Checked min/max datastore calls. They all use meeting_id.

jsangmeister (approved):
Looks good! Did not check whether any more places exist, but we will see if more come up. Please make sure to also add the meeting id to all future filter-based requests where possible.

reiterl:
I have gone thru filter, min, max, count and exists.
At one exists I was not sure, how to change it: `action/actions/motion/update.py` Perhaps s.o. other looks at it.

jsangmeister:
> I have gone thru filter, min, max, count and exists. At one exists I was not sure, how to change it: `action/actions/motion/update.py` Perhaps s.o. other looks at it.

I assume you mean this:
https://github.com/OpenSlides/openslides-backend/blob/d4818faffbdec4a7e1b0d8f9eb7693ec1209faca/openslides_backend/action/actions/motion/update.py#L115

That should not be an `exist` query, but rather a simple `get_many` on all possible ids to determine if they exist or not. You can go ahead and change this in this PR.

reiterl:
Update the `motion.update` action: replace `exist` call with get_many.

openslides_backend/action/actions/motion/update.py
```diff
@@ -108,17 +108,18 @@ def set_recommendation_extension_reference_ids(
...
-            exists = self.datastore.exists(
-                collection=Collection(collection),
-                filter=FilterOperator("id", "=", int(id_)),
-            )
-            if exists:
+            motion_ids.append(int(id_))
+        gm_request = GetManyRequest(Collection("motion"), motion_ids, ["id"])
+        gm_result = self.datastore.get_many([gm_request]).get(Collection("motion"), {})
+        for motion_id in motion_ids:
+            if motion_id in gm_result:
```
jsangmeister, line 120: Iterate over the `gm_result` directly to save time

jsangmeister (changes requested):
Minor improvement

openslides_backend/action/actions/motion/update.py
reiterl: Done.

jsangmeister (approved):

PR #1263, opened 2022-03-11
Refactor calc inherited_access_groups and is_public.

reiterl:


jsangmeister (changes requested):
Not exactly what I had in mind. Generally speaking, I'm not a fan of giving the models datastore access (although this is just a helper function, but still). My goal was this: The `calculate_inherited_groups_helper` gets called every time, indepenent from whether a `parent_id` is present or not. If no parent exists, `None` is given for the parent values. I created some relevant test cases (see below) for which I think the current helper is already working (so the helper itself does not need to be changed, just the places calling it).

Keeping the helper independent from the datastore also means that unit tests are easier. The following unit test cases should be implemented and asserted to be correct (input on the left side, output on the right):
```
access_groups		parent_is_public	parent_inherited_groups	|	is_public	inherited_access_groups
None == []		None			None == []		|	True		[]
None == []		True			None == []		|	True		[]
None == []		False			None == []		|	False		[]
None == []		False			[1]			|	False		[1]
[1]			None			None == []		|	False		[1]
[1]			True			None == []		|	False		[1]
[1]			False			None == []		|	False		[]
[1]			False			[2]			|	False		[]
[1, 2]			False			[2, 3]			|	False		[2]

**Comment by reiterl** [comment]
Ahh, okay, I have moved the with_parent_id helper function to another place and added the unittests in test_model_helper.
I found a small bug in the calc function and fixed it.

File: openslides_backend/action/actions/mediafile/calculate_mixins.py [hunk 1]
```diff
@@ -57,3 +58,22 @@ def handle_children(
...
+            FullQualifiedId(Collection("mediafile"), parent_id),
+            ["is_public", "inherited_access_group_ids"],
+        )
+        return calculate_inherited_groups_helper(
+            access_group_ids,
+            parent.get("is_public"),
+            parent.get("inherited_access_group_ids"),
+        )
+    else:
+        return (not bool(access_group_ids), access_group_ids)
```
jsangmeister, line 79:
```suggestion
    if parent_id:
        parent = datastore.get(
            FullQualifiedId(Collection("mediafile"), parent_id),
            ["is_public", "inherited_access_group_ids"],
        )
    else:
        parent = {}
    return calculate_inherited_groups_helper(
        access_group_ids,
        parent.get("is_public"),
        parent.get("inherited_access_group_ids"),
    )
```
This was the initial hang up on why this change was needed, so the information about how to calculate these fields is gathered in one place ;)

jsangmeister (changes requested):

tests/unit/test_model_helper.py
```diff
@@ -0,0 +1,39 @@
+from typing import List, Optional, Tuple
+from unittest import TestCase
+
+from openslides_backend.models.helper import calculate_inherited_groups_helper
+
+
+class ModelHelper(TestCase):
```
jsangmeister, line 7: Also rename this class and file to accurately reflect what is being tested - the `calculate_inherited_groups_helper`

tests/unit/test_model_helper.py
```diff
@@ -0,0 +1,39 @@
...
+            ([], True, [], True, []),
+            (None, False, None, False, []),
+            (None, False, [1], False, [1]),
+            ([1], None, None, False, [1]),
+            ([1], True, [], False, [1]),
+            ([1], False, [], False, []),
+            ([1], False, [2], False, []),
+            ([1, 2], False, [2, 3], False, [2]),
+        ]
+        for entry in data:
```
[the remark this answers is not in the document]
reiterl: Renamed entry.

tests/unit/test_model_helper.py
```diff
@@ -0,0 +1,39 @@
...
+
+from openslides_backend.models.helper import calculate_inherited_groups_helper
+
+
+class ModelHelper(TestCase):
+    """
+    Test methods of model/helper functions.
+    """
+
+    def test_calculate_inherited_groups_helper(self) -> None:
```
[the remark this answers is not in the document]
reiterl: Add some comments.

openslides_backend/action/actions/mediafile/calculate_mixins.py
reiterl: Done.

tests/unit/test_model_helper.py
reiterl: Renamed.

jsangmeister (approved):

Issue #1258, opened 2022-03-08
I still cannot create committees after deleting one


GabrielInTheWorld:
First, I created a committee and delete it.
Then, I try to create a new one and get this error message:
```ts
{
    "success": false,
    "message": "Datastore service sends HTTP 400. Model 'committee/3' does not exist."
}
```

Here are some logs:
```
docker-backend-1           | DEBUG:openslides_backend.shared.handlers.base_handler:Perform action committee.create.
docker-backend-1           | DEBUG:openslides_backend.services.datastore.adapter:Start GET request to datastore with the following data: GetRequest(fqid='user/1', mapped_fields=['organization_management_level', 'meta_position'], position=None, get_deleted_models=<DeletedModelsBehaviour.NO_DELETED: 1>)
docker-backend-1           | DEBUG:openslides_backend.services.datastore.adapter:Start RESERVE_IDS request to datastore with the following data: Collection: committee, Amount: 1
docker-backend-1           | DEBUG:urllib3.connectionpool:Starting new HTTP connection (1): datastore-writer:9011
docker-datastore-writer-1  | 2022-03-08T14:39:01.568: [/app/datastore/writer/core/writer_service.py] [INFO] 1 ids reserved
docker-datastore-writer-1  | 172.20.0.12 - - [08/Mar/2022 14:39:01] "POST /internal/datastore/writer/reserve_ids HTTP/1.1" 200 -
docker-backend-1           | DEBUG:urllib3.connectionpool:http://datastore-writer:9011 "POST /internal/datastore/writer/reserve_ids HTTP/1.1" 200 12
docker-backend-1           | DEBUG:openslides_backend.services.datastore.adapter:Get response with status code 200: {'ids': [17]}
docker-backend-1           | DEBUG:openslides_backend.services.datastore.adapter:Start GET request to datastore with the following data: GetRequest(fqid='organization/1', mapped_fields=['committee_ids', 'meta_position'], position=None, get_deleted_models=<DeletedModelsBehaviour.NO_DELETED: 1>)
docker-backend-1           | DEBUG:openslides_backend.services.datastore.adapter:Start GET request to datastore with the following data: GetRequest(fqid='user/1', mapped_fields=['committee_$can_manage_management_level', 'meta_position'], position=None, get_deleted_models=<DeletedModelsBehaviour.NO_DELETED: 1>)
[...]

PR #1264, opened 2022-03-11
#1258  sequential_numbers and mediafile.is_public required, add seq-n…

r-peschke:
…umber to workflow.import

Sets sequential_numbers of all 11 collection required
fixes motion_workflow.import and sets a sequential_number

### DIFF
--- a/openslides_backend/action/actions/organization/initial_import.py
+++ b/openslides_backend/action/actions/organization/initial_import.py
@@ -134,8 +134,10 @@ class OrganizationInitialImport(SingularActionMixin, Action):
         }
         if self.backend_migration_index > self.data_migration_index:
             result["message"] = "Data imported, but must be migrated!"
+            result["migration_needed"] = True
         else:
             result[
                 "message"
             ] = f"Data imported, Migration Index set to {self.backend_migration_index}"
+            result["migration_needed"]: False
         return result
--- a/openslides_backend/action/actions/organization/initial_import.py
+++ b/openslides_backend/action/actions/organization/initial_import.py
@@ -121,7 +121,8 @@ class OrganizationInitialImport(SingularActionMixin, Action):
         Add Migration Index to the one and only write request
         """
         write_request = super().process_write_requests()
-        write_request.migration_index = self.data_migration_index
+        if write_request:
+            write_request.migration_index = self.data_migration_index
         return write_request
 
     def create_action_result_element(
--- a/openslides_backend/action/actions/organization/initial_import.py
+++ b/openslides_backend/action/actions/organization/initial_import.py
@@ -2,6 +2,8 @@ from typing import Any, Dict, Iterable, Optional, Tuple
 
 from datastore.shared.util import DeletedModelsBehaviour
 
+from migrations import get_backend_migration_index
+
 from ....models.checker import Checker, CheckException
 from ....models.models import Organization
 from ....shared.exceptions import ActionException
@@ -56,6 +58,8 @@ class OrganizationInitialImport(SingularActionMixin, Action):
             data = get_initial_data_file(INITIAL_DATA_FILE)
             instance["data"] = data
 
+        self.check_migration_index(instance)
+
         # check datavalidation
         checker = Checker(data=data, mode="all")
         try:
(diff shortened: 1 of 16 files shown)


r-peschke:
Couldn't we set mediafile.is_public = True to all mediafiles, except the meeting one's?

openslides_backend/action/actions/organization/initial_import.py
```diff
@@ -75,6 +79,27 @@ def check_empty_datastore(self) -> None:
...
 
+    def check_migration_index(self, instance: Dict[str, Any]) -> None:
+        data_migration_index = instance["data"].get("_migration_index")
+        if data_migration_index is None:
+            raise ActionException(
+                "Data must have a valid migration index in `_migration_index`."
+            )
+        if data_migration_index < 1:
+            raise ActionException(
+                f"Data must have a valid migration index >= 1, bus has {data_migration_index}."
```
jsangmeister, line 90:
```suggestion
                f"Data must have a valid migration index >= 1, but has {data_migration_index}."
```

openslides_backend/action/actions/organization/initial_import.py
```diff
@@ -75,6 +79,27 @@ def check_empty_datastore(self) -> None:
...
+            )
+
+        backend_migration_index = get_backend_migration_index()
+        if backend_migration_index < data_migration_index:
```
jsangmeister, line 94:
```suggestion
        if backend_migration_index == data_migration_index:
```
The backend can't function with older data

openslides_backend/action/actions/organization/initial_import.py
```diff
@@ -75,6 +79,27 @@ def check_empty_datastore(self) -> None:
...
+            raise ActionException(
+                f"Migration indices do not match: Data has {data_migration_index} and the backend has {backend_migration_index}"
+            )
+        self.data_migration_index = data_migration_index
+        self.migration_needed = False
+        if backend_migration_index > data_migration_index:
+            self.migration_needed = True
```
jsangmeister, line 101: Does this flag do anything? Unless you want to migrate on the fly during the initial import (which I'm against) I would argue that only the current MI can be imported. In practice, this will always be the case anyway - we always keep our initial/example data up-to-date.

jsangmeister: Another advantage: You can remove the whole MI handling from the action code again and just set it below in the `create_write_requests`. It is invalid anyway for any action except `initial_import` to set a migration index, so we should not give all actions possibility to do so. See below for details

openslides_backend/action/actions/organization/initial_import.py
```diff
@@ -87,6 +112,7 @@ def create_write_requests(self, instance: Dict[str, Any]) -> Iterable[WriteReque
                         fqid,
                         "initial import",
                         entry,
+                        migration_index=self.data_migration_index,
```
jsangmeister, line 115: Instead of giving it as an argument to the function, just create the write request first and then set the MI manually after, and then append it to the list of write requests.

jsangmeister (changes requested):

jsangmeister:
> Couldn't we set mediafile.is_public = True to all mediafiles, except the meeting one's?

Sure, would be okay for me.

openslides_backend/action/actions/organization/initial_import.py
r-peschke: > The backend can't function with older data
I think you mean !=, not ==, to raise the exception.

openslides_backend/action/actions/organization/initial_import.py
r-peschke: okay, will do

openslides_backend/action/actions/organization/initial_import.py
r-peschke: I think we should migrate "on the fly"! Even better, the need to migrate should be checked by the starting procedure. 
Reason: It is not guaranteed that the initial import has the correct MI. Imagine a customized initial-data file and the latest version will  be or is deployed. Nobody wants to do the migration by hand, but it is easy for us implement such a logic.
Let's talk about it tomorrow.
That's the reason for the self.migration_needed flag.

> > Couldn't we set mediafile.is_public = True to all mediafiles, except the meeting one's?
> 
> Sure, would be okay for me.



> > Couldn't we set mediafile.is_public = True to all mediafiles, except the meeting one's?
> 
> Sure, would be okay for me.

Including setting the field required?

openslides_backend/action/actions/organization/initial_import.py
jsangmeister: Right, my bad :D

r-peschke: After our conversion: I will not change this. If the backend_migration_index is greater than the MI from import-data, there will be a message in the result. In this case the backend will not accept any action, until the needed migrations are done.

jsangmeister (changes requested):
Looks good! Minor change

jsangmeister (approved):

PR #1266, opened 2022-03-14
Update models.yml for multiple votes per option

MJJojo97:
See https://github.com/OpenSlides/openslides-backend/pull/1251


ostcar (approved):

global/meta/models.yml
```diff
@@ -2654,11 +2654,17 @@ poll:
...
     restriction_mode: A
   max_votes_amount:
     type: number
     default: 1
+    minimum: 1
     restriction_mode: A
+  max_votes_per_option:
+    type: number
+    default: 1
+    minimum: 1
```
ostcar, line 2667: `restriction_mode: A` is missing

ostcar (changes requested):

global/meta/models.yml
MJJojo97: Oh yes. I am very sorry about that!

ostcar:
Ohh, I see there is a check, that the example_data is valid. I did not realise that. This is strange, because your field is not required. So why do you have to add it to the example data?

Anyway, could you add it?

MJJojo97:
Sorry for the inconvenience! I didn't have the full overview on what make-tasks I have to run. Otherwise, I would have seen that problem before.

I hope that everything is working now.

jsangmeister:
> Ohh, I see there is a check, that the example_data is valid. I did not realise that. This is strange, because your field is not required. So why do you have to add it to the example data?

Because it has a a default value - and yes, I know how strange that sounds :D

> Sorry for the inconvenience! I didn't have the full overview on what make-tasks I have to run. Otherwise, I would have seen that problem before.

No problem, it's a lot one has to take care of when changing the models definition. You can be very happy that no migration is required here :D

jsangmeister (approved):

PR #1269, opened 2022-03-17
Add a merging of update events after the relation handling

reiterl:
Merges all Update Events with `fields` into one. 
It doesn't merge the update events with `list_fields`.

### DIFF
--- a/openslides_backend/action/action.py
+++ b/openslides_backend/action/action.py
@@ -414,13 +414,11 @@ class Action(BaseAction, metaclass=SchemaProvider):
 
         result: List[Event] = []
         for fqid in events_by_fqid:
-            result.extend(self.merge_update_events_helper(events_by_fqid[fqid]))
+            result.extend(self.merge_update_events_for_fqid(events_by_fqid[fqid]))
 
         return result
 
-    def merge_update_events_helper(self, events: List[Event]) -> List[Event]:
-        if len(events) < 2:
-            return events
+    def merge_update_events_for_fqid(self, events: List[Event]) -> List[Event]:
         result: List[Event] = []
         trailing_index: Optional[int] = None
         count = 0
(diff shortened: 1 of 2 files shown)


openslides_backend/action/action.py
```diff
@@ -397,10 +397,48 @@ def process_write_requests(
...
+
+        result: List[Event] = []
+        for fqid in events_by_fqid:
+            result.extend(self.merge_update_events_helper(events_by_fqid[fqid]))
+
+        return result
+
+    def merge_update_events_helper(self, events: List[Event]) -> List[Event]:
+        if len(events) < 2:
+            return events
```
jsangmeister, line 423: I think the algorithm also works for one element so you can remove this check, right?

openslides_backend/action/action.py
```diff
@@ -397,10 +397,48 @@ def process_write_requests(
...
+        for event in update_events:
+            events_by_fqid[event["fqid"]].append(event)
+
+        result: List[Event] = []
+        for fqid in events_by_fqid:
+            result.extend(self.merge_update_events_helper(events_by_fqid[fqid]))
+
+        return result
+
+    def merge_update_events_helper(self, events: List[Event]) -> List[Event]:
```
jsangmeister, line 421:
```suggestion
    def merge_update_events_for_fqid(self, events: List[Event]) -> List[Event]:
```
helper is always very generic

jsangmeister (changes requested):
Looks good! Tested it with the performance test from https://github.com/OpenSlides/openslides-backend/pull/1247 and the time used for 100 votes was reduced from about 1.9s to 1.1s and the amount of events from ~900 to ~200.

openslides_backend/action/action.py
reiterl: Removed, the code doesn't need it.

openslides_backend/action/action.py
reiterl: Done. Renamed.

jsangmeister (approved):
In the long run, we can look to streamline the relation handling, so this is no longer necessary, but works good for now.

PR #1274, opened 2022-03-22
Update to python 3.10 and update dependencies

jsangmeister:
Also cleaned up some code which was improved by the newer python version.


jsangmeister:
Test in conjunction with https://github.com/OpenSlides/openslides-datastore-service/pull/185. Need to update datastore remote before merging.

jsangmeister:
@r-peschke it seems like some of the default email security setting were changed which leads to the failing tests. Maybe https://bugs.python.org/issue45487 can help here. I don't have enough knowledge of email security or our implementation of it, can you take a look and hopefully fix the tests?

r-peschke:
Can't fix the test for ssl/tls. The error occurs in test of aiosmtpd-package, too and has a pull-request, see https://github.com/aio-libs/aiosmtpd/pull/294. It is not approved, because use of insecure protocol TLS v1.1., see https://lgtm.com/projects/g/aio-libs/aiosmtpd/rev/pr-f99a420015bd7f698b29ebe86761f86b1ddb64cf. 
Maybe we'll wait for a fix of aiosmtpd!

r-peschke:
Mail tests are okay, but there is still a test from initial_import failing. Will have a look at it tomorrow.

openslides_backend/shared/env.py
```diff
@@ -1,28 +1,99 @@
...
-def is_dev_mode() -> bool:
-    dev = os.environ.get(OPENSLIDES_DEVELOPMENT, "off")
-    return is_truthy(dev)
+class Environment:
+    """
+    Contains all environment variables. See the vars attribute for all defaults.
+    """
+
+    vars = {
+        "ACTION_PORT": "9002",
```
r-peschke, line 27: I miss the email settings

jsangmeister:
I think you butchered the history here...

jsangmeister:
I rebased the commits to another branch and now it looks good. Can you check that I did not miss any changes or yours @r-peschke? https://github.com/OpenSlides/openslides-backend/compare/main...jsangmeister:update-python-and-deps-temp?expand=1

r-peschke:
> I rebased the commits to another branch and now it looks good. Can you check that I did not miss any changes or yours @r-peschke? https://github.com/OpenSlides/openslides-backend/compare/main...jsangmeister:update-python-and-deps-temp?expand=1

Seems okay for me

What is about the centralized env-variables and the email settings? And does the presenter really needs the writer?

openslides_backend/shared/env.py
jsangmeister: I will have a look

.github/docker-compose/docker-compose.prod.yml
```diff
@@ -1,12 +1,28 @@
...
+            - reader
+        secrets:
+            - postgres_password
+    backendPresenter:
+        build: ../..
+        image: openslides-backend
+        ports:
             - "9003:9003"
         environment:
+            - OPENSLIDES_BACKEND_COMPONENT=presenter
```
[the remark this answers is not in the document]
jsangmeister: No, this was probably missed in https://github.com/OpenSlides/openslides-backend/pull/1254. I will remove it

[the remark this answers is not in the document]
jsangmeister: It is not as easy to remove it as I thought, since it is tethered to the entrypoint which is always executed in the backend. So I would propose to leave it as-is.

openslides_backend/shared/env.py
jsangmeister: These are only the defaults; email settings previously did not have default settings, either. So this should be fine.

PR #1275, opened 2022-03-22
Fix relation handling bug in generic delete action

jsangmeister:
Tricky bug in the generic delete: When deleting a model with a template relation field, the reverse relation was sometimes not updated correctly. Root cause for this was that too few fields were fetched from the datastore when calculating the reverse relations so they were assumed to be empty. The reason this error only occured when the user was a submitter of some motion is that through a second mistake all fields were fetched from the datastore when actually none should have been fetched (if `[]` is given as `mapped_fields` to the `get` method, it is interpreted as "give me the whole model). So there needed to be two template relation fields in the model for this error to occur, one which was correctly fetched and one which was missed because of that.

@emanuelschuetze I would not want to write any migration for this, since it could potentially affect all delete actions, not only user. By deleting the invalid models (e.g. the poll with wrong `voted_ids`) and recreating them, everything should be working again as long as they don't go back to this exact point in time with the history mode. Alternatively, they can of course reset their instance, depending on how many data was already filled in.

### DIFF
--- a/openslides_backend/action/actions/motion/create.py
+++ b/openslides_backend/action/actions/motion/create.py
@@ -20,7 +20,6 @@ class MotionCreate(MotionCreateBase):
 
     schema = DefaultSchema(Motion()).get_create_schema(
         optional_properties=[
-            "title",
             "number",
             "state_extension",
             "sort_parent_id",
--- a/openslides_backend/action/generics/delete.py
+++ b/openslides_backend/action/generics/delete.py
@@ -20,17 +20,13 @@ class DeleteAction(Action):
         """
         Takes care of on_delete handling.
         """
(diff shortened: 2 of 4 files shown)


openslides_backend/action/actions/motion/create.py
```diff
@@ -20,7 +20,6 @@ class MotionCreate(MotionCreateBase):
 
     schema = DefaultSchema(Motion()).get_create_schema(
         optional_properties=[
-            "title",
```
jsangmeister, line 23: noticed while testing that `title`was doubled in the optional and required properties, so removed it here.

emanuelschuetze (approved):

PR #1278, opened 2022-03-23
Update vote action poll.start, poll.stop and vote/adapter 

reiterl:
* move vote.start into update_instance
* catch exception at vote.clear and log an error
* Change the endpoint from create to start at start.

**Notice**:
You need the pr https://github.com/OpenSlides/openslides-vote-service/pull/113 for this. Both prs should be merged the same time-


reiterl:
The tests will fail, because 113 is not merged.

ostcar:
Tell me, when I should merge the PR in the Vote services

jsangmeister (changes requested):
> When the datastore-write requests fails, then you SHOULD call "clear" in the vote-service.

You should do this also. You can either implement a `get_on_failure` method in the action, analogous to the `get_on_success` method, or change the latter to somethind like `get_on_finish` and provide the success status as a flag. I'm not sure what the better approach is here. Either way, analogous to the clearing in the success routine of `poll.stop`, the poll should be cleared in the failure routine of `poll.start`.

openslides_backend/action/action_handler.py
```diff
@@ -245,4 +249,6 @@ def perform_action(
             # -1: error which cannot be directly associated with a single action data
             if action.index > -1:
                 exception.action_data_error_index = action.index
+            if on_failure := action.get_on_failure(action_data):
+                self.on_failure.append(on_failure)
```
jsangmeister, line 253: You can't gether the `on_failure` calls here because `perform_action` is possibly called multiple times per action (see `execute_write_requests` above: If the datastore returns a `DatastoreLockedException`, the request is retried). So suppose we want to execute `poll.start`. The first try fails and adds the `on_failure` routine to the list, then the second try succeeds. Then, at the end, the `on_failure` routine is still executed and the poll is reset even though it shouldn't have been, which leads to an inconsistent state.

The easiest solution is probably to execute the `on_failure` routine here directly. So you don't need the `on_failure` list at all.

jsangmeister (changes requested):

openslides_backend/action/action_handler.py
reiterl: Done.

jsangmeister (approved):

jsangmeister:
@ostcar this should be ready now, you can merge your PR

ostcar:
> @ostcar this should be ready now, you can merge your PR

done

PR #1280, opened 2022-03-25
Motion_change_recommendation.create: Fix the collide check.

reiterl:
Two closed intervals are disjunct, iff they don't "touch" each other.
Updated tests and code.

### DIFF
--- a/tests/system/action/motion_change_recommendation/test_create.py
+++ b/tests/system/action/motion_change_recommendation/test_create.py
@@ -195,7 +195,11 @@ class MotionChangeRecommendationLineValidationTest(BaseActionTestCase):
     def test_create_change_recommendation_min_overlap_1(self) -> None:
         self.create_change_recommendation(1, 5)
         response = self.cr_request(5, 10)
-        self.assert_status_code(response, 200)
+        self.assert_status_code(response, 400)
+        self.assertIn(
+            "The recommendation collides with an existing one",
+            response.json["message"],
+        )
 
     def test_create_change_recommendation_min_overlap_2(self) -> None:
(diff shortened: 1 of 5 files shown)


emanuelschuetze (changes requested):
There are still some collisions in my tests which needs to discussed...

tests/system/action/motion_change_recommendation/test_create.py
```diff
@@ -210,3 +214,12 @@ def test_create_change_recommendation_other_motion_not_colliding(self) -> None:
...
         self.assert_status_code(response, 200)
+
+    def test_create_change_recommendation_collide_check(self) -> None:
+        self.create_change_recommendation(2, 5)
+        response = self.cr_request(1, 2)
+        self.assert_status_code(response, 400)
+        self.assertIn(
+            "The recommendation collides with an existing one",
+            response.json["message"],
+        )
```
jsangmeister, line 225: This test can be removed again as it is the same as min_overlap_2

jsangmeister (changes requested):
some cleanup, otherwise looks good

tests/system/action/motion_change_recommendation/test_create.py
reiterl: I removed this surplus test.

jsangmeister (approved):
We should wait for the merge until is the client is updated as well so the migration works correctly and we do not have mixed data.

GabrielInTheWorld:
I think, you can merge this PR and I will check, that the client's changes (OpenSlides/openslides-client#988) will fit.

PR #1282, opened 2022-03-27
Introduce setting allowing to enable or disable max_votes_per_option

MJJojo97:


r-peschke (approved):
You forgot to add the field to the action description in openslides/openslides wiki! It's done

PR #1288, opened 2022-04-05
Fix the changed_model filtering for empty mapped_fields

### LINKED ISSUE #1285: Error in datastore.adapter getting data from changed_models
The method _get_many_from_changed_models in extended_adapter.py is not correct if fqid is in self.changed_models and mapped_fields is an empty list, which stands for `read all fields`. In this case an empty dict will be returned as read-data, the missing fields are also empty.
On the other side, if the fqid is not in the changed_models, the empty mapped_fields were returned and in consequence the datastore.get will read fields. That is what one may expect, but in generic delete-action we shouldn't read it this way, because the mapped fields are used for specific fields and if there aren't some, we shouldn't read at all. 

@jsangmeister lets talk about it

jsangmeister:
fixes #1285

### DIFF
--- a/tests/unit/extended_datastore_adapter/test_get.py
+++ b/tests/unit/extended_datastore_adapter/test_get.py
@@ -20,7 +20,10 @@ class TestGetExtendedDatastoreAdapter(BaseTestExtendedDatastoreAdapter):
     ) -> Dict[str, Any]:
         if fqid.id in self.mock_datastore_content.get(fqid.collection, {}):
             model = self.mock_datastore_content[fqid.collection][fqid.id]
-            return {field: model[field] for field in mapped_fields}
+            if mapped_fields:
+                return {field: model[field] for field in mapped_fields}
+            else:
+                return model
         else:
             raise DatastoreException("mock_db: model does not exist")
 
@@ -101,3 +104,14 @@ class TestGetExtendedDatastoreAdapter(BaseTestExtendedDatastoreAdapter):
             )
         self.db_method_mock.assert_called()
         self.add_get_many_mock.assert_not_called()
+
+    def test_get_empty_mapped_fields_and_changed_models(self) -> None:
+        self.set_additional_models(
+            {
+                "test/1": {"id": 1, "changed": 3},
+            }
+        )
+        result = self.adapter.get(FullQualifiedId(self.collection, 1), [])
+        assert result == {"id": 1, "f": 1, "unused": 2, "changed": 3}
+        self.db_method_mock.assert_called()
+        self.add_get_many_mock.assert_called()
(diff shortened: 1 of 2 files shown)


tests/unit/extended_datastore_adapter/test_get.py
```diff
@@ -101,3 +104,14 @@ def test_get_only_db_exception(self) -> None:
             )
         self.db_method_mock.assert_called()
         self.add_get_many_mock.assert_not_called()
+
+    def test_get_empty_mapped_fields_and_changed_models(self) -> None:
```
r-peschke, line 108: IMO this test only tests the mock, but not the changed _get_many_from_changed_models-method. From code-review I think the change is okay, but ...;-)

r-peschke (changes requested):

tests/unit/extended_datastore_adapter/test_get.py
jsangmeister: No, I wrote the test before I changed the code and it failed. The changed method is not mocked here, only the datastore - the fake mok datastore used here had to be changed because it did not know about the "empty mapped-fields" thing ;)

r-peschke: You are right, the changed method is called via mock

r-peschke: But there are 2 more things in the get-method itself.
1. Performance issue: Should have a flag that the whole model is read into change_models "cache" to prevent reading the whole model each time, when there are no mapped fields
2. The get method has only 1 fqid and can only work with 1 fqid, but the method pretends to be usable with more fqids. Shouldn't we implement it without fqid-dict or made it usable for more than 1 fqid?

jsangmeister: > 1. Performance issue: Should have a flag that the whole model is read into change_models "cache" to prevent reading the whole model each time, when there are no mapped fields

I don't think so, this is a task for the real cache - see #1289.

> 2. The get method has only 1 fqid and can only work with 1 fqid, but the method pretends to be usable with more fqids. Shouldn't we implement it without fqid-dict or made it usable for more than 1 fqid?

I don't understand what you mean. What get method? Both the one in `extended_datastore_adapter.py` as well as the `_get_mock` in this file only take one fqid.

r-peschke: > > 1. Performance issue: Should have a flag that the whole model is read into change_models "cache" to prevent reading the whole model each time, when there are no mapped fields
> 
> I don't think so, this is a task for the real cache - see #1289.
> 
> > 2. The get method has only 1 fqid and can only work with 1 fqid, but the method pretends to be usable with more fqids. Shouldn't we implement it without fqid-dict or made it usable for more than 1 fqid?
> 
> I don't understand what you mean. What get method? Both the one in `extended_datastore_adapter.py` as well as the `_get_mock` in this file only take one fqid.

The adapter get-method: With the missing_fields_per_fqid-dict it seems to be made for more than 1 fqid, but it in't. We could use a simple fields-list instead a dict with list.

PR #1289, opened 2022-04-06
Improved reader cache

jsangmeister:
First test sees improvement of about 20%, which is slightly less then I expected. Probable cause for this is that as @ostcar noticed, although cache hits are faster than always accessing the database, the access to the cache should also be reduced by streamlining the code (e.g. checking permissions only once for all action data and not for every instance).

Only a raw draft for now, will continue work on this.

### DIFF
--- a/openslides_backend/services/datastore/cache_adapter.py
+++ b/openslides_backend/services/datastore/cache_adapter.py
@@ -10,9 +10,7 @@ from ...shared.patterns import Collection, FullQualifiedId
 from ...shared.typing import ModelMap
 from .adapter import DatastoreAdapter
 from .commands import GetManyRequest
-from .interface import Engine, LockResult, PartialModel
-
-MappedFieldsPerFqid = Dict[FullQualifiedId, List[str]]
+from .interface import Engine, LockResult, MappedFieldsPerFqid, PartialModel
 
 
 class CacheDatastoreAdapter(DatastoreAdapter):
--- a/openslides_backend/services/datastore/cache_adapter.py
+++ b/openslides_backend/services/datastore/cache_adapter.py
@@ -40,7 +40,7 @@ class CacheDatastoreAdapter(DatastoreAdapter):
             )
 
         mapped_fields_per_fqid = {fqid: mapped_fields}
-        # fetch results from changed models
+        # fetch results from cached models
         results, missing_fields_per_fqid = self._get_many_from_cached_models(
             mapped_fields_per_fqid
         )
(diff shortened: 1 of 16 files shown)


jsangmeister:
Only thing left to do is to replace the `prefetch` method in `poll.stop` with a generic approach to prefetch the needed data (which needs to be extended to `vote.create` and the other called actions as well). Otherwise this should be done. I'm not exactly happy with the code of the cache adapter since a huge part of it is the same as the extended datastore adapter, but I did not find a good way to unify these, since they differ in important parts which are nested deep in the respective methods.

openslides_backend/services/datastore/cache_adapter.py
```diff
@@ -0,0 +1,153 @@
...
+            return super().get(
+                fqid,
+                mapped_fields,
+                position,
+                get_deleted_models,
+                lock_result,
+            )
+
+        mapped_fields_per_fqid = {fqid: mapped_fields}
+        # fetch results from changed models
```
r-peschke, line 43: Seems the wrong comment

r-peschke (commented):
To much for me to detect:
- why do you use the rest-methods (with parameter sometimes needed This just hides what it does in the different classes
- what is the relation between cache and models_changed, ho wto get from database, how to get from changed_models
- if you use cProfile, show or document with a small example, for instance how to get the requests response, how tho fill parameters
- ...

jsangmeister:
> To much for me to detect:
> 
> * why do you use the rest-methods (with parameter sometimes needed This just hides what it does in the different classes
> * what is the relation between cache and models_changed, ho wto get from database, how to get from changed_models
> * if you use cProfile, show or document with a small example, for instance how to get the requests response, how tho fill parameters
> * ...

I don't think I understand what you mean. What rest-methods? Do you mean `super`? The general structure is `ExtendedAdapter` -> `CacheAdapter` -> `BaseAdapter`, so the cache adapter works as a cache layer between the extended adapter and the actual datastore. I have to use the `super`-methods so the correct methods of all classes are called. The cache and `changed_models` do not have any relation: `changed_models` works just as before. If a field is not found there, the request is delegated to the next layer (the cache layer), where it is checked if it was already cached. This does, however, not influence the behaviour of the extended adapter or any using classes at all. In case of a cache miss, the base adapter is called to fetch the data from the actual datastore. The only thing connecting the `ExtendedDatastoreAdapter` and the `CacheAdapter` (besides inheritance) is that they have very similar code since both are looking data up in an in-memory dict, but since both are doing slightly different things, they could not be united code-wise.

For a profiling example, check out https://github.com/OpenSlides/openslides-backend/wiki/Profiling. The test currently creates the file `test_stop_performance.prof` which you can inspect as you wish.

r-peschke (approved):
LGTM with reasonable amount of time;-)

jsangmeister:
Generified the prefetch and added more utils for performance/datastore access testing.

r-peschke (approved):

Issue #1074, opened 2021-12-01
Migrate data in `meeting.import`


jsangmeister:
When importing data from an old OS3 instance, the data must first be migrated before the import can be executed. The alternative would be to update OS3 with every new migration index and every change to the models file, since the export will then differ from the required format. My proposed solution is the following:

Change the migration code so that it also works in-memory or extract/restructure the necessary parts. The logic is already there, so this should be doable. In the import action, before doing any importing/validation/etc., create a dummy structure which represents exactly one position for the migration process and insert all models with all their fields as `CREATE` events. The exported data must get a migration index which will be fixed once this system is implemented. Migrate the dummy position to the newest migration index. Extract the migrated data from the dummy position (should be very easy, since it is just `CREATE` events which hold all data) and then proceed normally with validation and importing.

jsangmeister:
Since I did not hear any objections, this should be ready to implement.

PR #1293, opened 2022-04-08
#1074 in-memory migrations for meeting import

r-peschke:
Dependency: https://github.com/OpenSlides/openslides-datastore-service/pull/188 okay

State: Everything concerning the migration seems okay, including the migration of an OS3-exported database with 150MB JSON-file.
But after migrations there is always the consistence checker, complaining about empty fields with default values and other problems. First one is solved within the checker and a repair-parameter.


openslides_backend/action/actions/meeting/import_.py
```diff
@@ -55,6 +68,13 @@ def perform(
         self.index = 0
         action_data = self.get_updated_instances(action_data)
         instance = next(iter(action_data))
+
+        if self.name == "meeting.import":
```
jsangmeister, line 72: I'd rather move this into one method which is called every time and can be overriden to do nothing in the subclass

openslides_backend/action/actions/meeting/import_.py
```diff
@@ -131,11 +169,11 @@ def check_usernames_and_generate_new_ones(self, json_data: Dict[str, Any]) -> No
         used_usernames = set()
```
jsangmeister, line 169: Can you unify this method with the one in `user.create` so the behaviour does not diverge again? If you want, we can also do this in a seperate issue to not blow up this one even more

openslides_backend/action/actions/meeting/import_.py
```diff
@@ -430,3 +519,91 @@ def check_permissions(self, instance: Dict[str, Any]) -> None:
...
+        self, instance: Dict[str, Any]
+    ) -> List[CreateEvent]:
+        json_data = instance["meeting"]
+        import_create_events = []
+        for collection in json_data:
+            for entry in json_data[collection].values():
+                # Necessary to remove None-values in event
+                for k, v in list(entry.items()):
+                    if v is None:
+                        entry.pop(k)
```
jsangmeister, line 533: Why? Does the datastore throw an error in that case? Iterating over all entries takes a lot of time so we should only do that if really necessary

openslides_backend/action/actions/meeting/import_.py
```diff
@@ -430,3 +519,91 @@ def check_permissions(self, instance: Dict[str, Any]) -> None:
...
+
+    def create_instance_from_migrated_events(
+        self, instance: Dict[str, Any], migrated_events: List[BaseEvent]
+    ) -> Dict[str, Any]:
+        data: Dict[str, Dict] = defaultdict(dict)
+        for event in migrated_events:
+            collection, id_ = collection_and_id_from_fqid(event.fqid)
+            if event.type == CreateEvent.type:
+                data[collection].update({str(id_): event.data})
+            elif event.fqid.split("/")[0] in ("organization", "committee", "user"):
```
jsangmeister, line 602: Use `collection_from_fqid` here (or at least the `KEYSEPARATOR` constant)

jsangmeister (changes requested):
Looks good! did not check the tests thoroughly. I would like to have this tested with real data to assert that it works as intended. Can you export a real OS3 instance with the latest update where the MI was added and try to import it?

openslides_backend/action/actions/meeting/import_.py
r-peschke: done

openslides_backend/action/actions/meeting/import_.py
r-peschke: The BaseEvent in datastore.migartions calls check and the _ModelEvent implements a check, that a field may not have a None value and raises a BadEventException.

openslides_backend/action/actions/meeting/import_.py
r-peschke: > Use `collection_from_fqid` here (or at least the `KEYSEPARATOR` constant)

done

openslides_backend/action/actions/meeting/import_.py
r-peschke: done

PR #1296, opened 2022-04-12
Add the new forwarding including committee user.

reiterl:


tests/system/action/motion/test_create_forwarded.py
```diff
@@ -63,6 +64,7 @@ def test_correct_origin_id_set(self) -> None:
                 "all_origin_ids": [12],
             },
         )
+        assert model.get("forwarded")
```
r-peschke, line 67: Would you please use the optional field reason to check it overwrites the motions reasons

r-peschke (changes requested):
Just 2 small things so far.-)

tests/system/action/motion/test_create_forwarded.py
reiterl: Added a test for 'reason'.

r-peschke (approved):
Except the dummy user

jsangmeister (approved):
Looks good to me, but @r-peschke should probably double-check it, since he wrote the concept

tests/system/action/motion/test_create_forwarded.py
```diff
@@ -50,19 +54,94 @@ def test_correct_origin_id_set(self) -> None:
                 "meeting_id": 2,
                 "origin_id": 12,
                 "text": "test",
+                "reason": "reason_jLvcgAMx",
             },
         )
         self.assert_status_code(response, 200)
+        model = self.assert_model_exists(
```
r-peschke, line 61: Please test the correct submitter in target-motion

tests/system/action/motion/test_create_forwarded.py
```diff
@@ -28,6 +30,7 @@ def setUp(self) -> None:
             "motion_state/34": {
                 "name": "name_state34",
                 "meeting_id": 2,
+                "allow_motion_forwarding": True,
```
r-peschke, line 33: The test data isn't correct. motion_state/34 is a state of the target-meeting, but is also used in the source motion/12 of meeting/1 as current state.
All this data is valid only in their own meeting and cannot be used in the other meeting, too.
The target-motion should use the default workflow of the new meeting with the first state of it.
Please check, if there are more fields form source-motion used in target-motion, i.e. amendments, attachments. The may not be copied to the target motion.
Use tests to show the behavior in this cases.

r-peschke (changes requested):

tests/system/action/motion/test_create_forwarded.py
reiterl: Add code to test for submitter, motion_submitter and link in user/2.

tests/system/action/motion/test_create_forwarded.py
reiterl: Added a state 30 of the source meeting.
Check the state_id of the created motion.
The create_forwarded action is a **create action**, so no coping of amendmens, etc. No need to test that.

r-peschke (approved):

PR #1299, opened 2022-04-14
Allow to set 'name' as updated value in meeting.clone

reiterl:
And update the name suffix code to work with setting of name.
Update the tests.


jsangmeister (approved):

GabrielInTheWorld (approved):
Works!

PR #1300, opened 2022-04-14
Extends the configuration options for PDF export

m-schieder:
* Add the options Line height, Page margin left/top/right/bottom.


jsangmeister (changes requested):
I would suggest to set appropriate min/max values (if needed) for all the settings. If you want the defaults to be applied to all existing meetings, you will have to write a migration to add these fields, which will be similar to e.g. https://github.com/OpenSlides/openslides-backend/blob/82a2c3eeb50852605d127a9e2b5bf49ea57aeb64/migrations/migrations/0004_add_motion_state_weight.py

We should, however, do all fields in one migration which is currently not possible with the `AddFieldMigration`. This has to be changed in the datastore to accomodate for more than one field - if you want, I can implement this for you or you try it out, your choice.

jsangmeister:
Ah, and you also have to add the new fields to the `meeting.import` tests, since they expect a complete meeting with all its fields.

jsangmeister:
> We should, however, do all fields in one migration which is currently not possible with the `AddFieldMigration`. This has to be changed in the datastore to accomodate for more than one field - if you want, I can implement this for you or you try it out, your choice.

Went ahead and already implemented it - see https://github.com/OpenSlides/openslides-datastore-service/pull/189.

jsangmeister:
Implemented the migration & set min values for all settings - please check if these are suitable or should ne changed. Afterwards, this can be merged from my point of view.

jsangmeister (approved):

global/data/example-data.json
```diff
@@ -1,5 +1,5 @@
 {
-    "_migration_index": 21,
+    "_migration_index": 26,
```
jsangmeister, line 2: @r-peschke it seems like the migration index of the example data is checked nowhere

m-schieder:
We have noticed that it makes more sense to set the default values for the page margins to 0 and the users set the margins from this value because otherwise the values are not so well understandable for them. The client then adds the current default values to the input for the PDF export. So it must also be possible to specify negative values to get the page spacing smaller than the default value.

emanuelschuetze (changes requested):
Please use the new defaults.

jsangmeister:
> So it must also be possible to specify negative values to get the page spacing smaller than the default value.

Then you can just go ahead and remove the `minimum` lines again where they are not needed. If the default should really be 0 now (I'm a little confused as your comment and the commit contradict each other), you can also go ahead and remove them from the migration again as 0 is the implicit default for numbers if they are not present.

global/data/example-data.json
```diff
@@ -372,6 +372,11 @@
             "export_csv_separator": ";",
             "export_pdf_pagenumber_alignment": "center",
             "export_pdf_fontsize": 10,
+            "export_pdf_line_height": 1.25,
+            "export_pdf_page_margin_left": 75,
+            "export_pdf_page_margin_top": 90,
+            "export_pdf_page_margin_right": 75,
+            "export_pdf_page_margin_bottom": 50,
```
emanuelschuetze, line 379: Here are still the old defaults.

emanuelschuetze (approved):

GabrielInTheWorld:
@m-schieder: Please rebase and merge.

PR #1301, opened 2022-04-20
Add motion.start_line_number.

reiterl:
Add it to the model. Add a way to access it in motion.update. Update
tests.

### DIFF
--- a/global/meta/models.yml
+++ b/global/meta/models.yml
@@ -2025,6 +2025,7 @@ motion:
   start_line_number:
     type: number
     restriction_mode: A
+    minimum: 1
 
   lead_motion_id:
     type: relation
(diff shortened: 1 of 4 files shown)


global/meta/models.yml
```diff
@@ -2022,6 +2022,9 @@ motion:
     type: timestamp
     read_only: true
     restriction_mode: C
+  start_line_number:
+    type: number
+    restriction_mode: A
```
jsangmeister, line 2027: add `minimum: 1`

reiterl: Add the minimum to the start_line_number

jsangmeister (approved):

PR #1308, opened 2022-04-21
Update start_line_number default.

reiterl:
Update json data, add migration, update test.

### DIFF
--- a/global/data/example-data.json
+++ b/global/data/example-data.json
@@ -1,5 +1,5 @@
 {
-    "_migration_index": 26,
+    "_migration_index": 27,
     "organization": {
         "1": {
             "id": 1,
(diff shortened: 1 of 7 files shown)


r-peschke (commented):
What is about setting the start_line_number? If the client wants to do this,this should be in motion.create/update payload!?

reiterl:
The action to set the start_line_number is the `motion.update` action. This is already in the code.

r-peschke (approved):

jsangmeister:
migration would not have been necessary since the client uses `1` as the default anyway

r-peschke:
> migration would not have been necessary since the client uses `1` as the default anyway

Yes and no: I think the client did, the pdf-export not: IMO it's more secure to fix this with migration in the database;-)

PR #1318, opened 2022-04-29
Issue1052 bulk performance poll

r-peschke:
Some performance improvement on low level, measurement poll.stop action

### DIFF
--- a/openslides_backend/action/actions/meeting/import_.py
+++ b/openslides_backend/action/actions/meeting/import_.py
@@ -1,12 +1,14 @@
 import time
 from collections import defaultdict
-from typing import Any, Dict, Iterable, List, Optional, Tuple
+from typing import Any, Dict, Iterable, List, Optional, Tuple, cast
 
 from datastore.migrations import BaseEvent, CreateEvent
 from datastore.shared.util import collection_and_id_from_fqid, collection_from_fqid
 
 from migrations import get_backend_migration_index
 from migrations.migrate import MigrationWrapper
+
+from openslides_backend.models.models import User
 
 from ....models.base import model_registry
 from ....models.checker import Checker, CheckException
@@ -25,6 +27,7 @@ from ....permissions.management_levels import (
     OrganizationManagementLevel,
 )
 from ....permissions.permission_helper import has_committee_management_level
+from ....services.datastore.interface import GetManyRequest
 from ....shared.exceptions import ActionException, MissingPermission
 from ....shared.interfaces.event import EventType
 from ....shared.interfaces.write_request import WriteRequest
@@ -65,6 +68,10 @@ class MeetingImport(SingularActionMixin, LimitOfUserMixin, UsernameMixin):
         """
         self.user_id = user_id
         self.index = 0
+
+        # prefetch as much data as possible
+        self.prefetch(action_data)
+
         action_data = self.get_updated_instances(action_data)
         instance = next(iter(action_data))
         instance = self.preprocess_data(instance)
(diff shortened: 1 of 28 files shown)


openslides_backend/action/actions/poll/stop.py
```diff
@@ -65,18 +66,7 @@ def prefetch(self, action_data: ActionData) -> None:
...
-                    Collection("user"),
-                    [self.user_id],
-                    [
-                        "organization_management_level",
-                        *[f"group_${meeting_id}_ids" for meeting_id in meeting_ids],
-                    ],
-                )
-            )
-        self.datastore.get_many(requests)
+        self.datastore.get_many(requests, use_changed_models=False)
```
r-peschke, line 69: I don't know why, but removing this `get` from prefetch, reduces the time for the write in the datastore-service by roundabout  33%.

profile method with locked fields from request user in prefetch/removed from prefetch, poll with 100 users:
/app/openslides_backend/services/datastore/http_engine.py:37(retrieve) tottime: 0.000/0.000    cumtime:0.455/0.304

The difference are the 2 locked fields  user.group_$meeting_ids and user.organization_management_level. 
Could you have a look into the datastore, why the impact of these 2 fields is so big? The percentage remains constant, even with 2000 votes intest.

openslides_backend/shared/patterns.py
```diff
@@ -32,9 +32,10 @@ def __repr__(self) -> str:
         return f"Collection({repr(str(self))})"
 
     def __eq__(self, other: object) -> bool:
-        if not isinstance(other, Collection):
-            return NotImplemented
-        return self.collection == other.collection
+        try:
+            return self.collection == other.collection  # type: ignore
```
jsangmeister, line 36: rather use `cast()` instead of `# type: ignore` - this should make no difference runtime-wise

tests/system/action/meeting/test_delete.py
```diff
@@ -144,9 +144,11 @@ def test_delete_full_meeting(self) -> None:
         for i in range(3):
             user = self.get_model(f"user/{i+1}")
             for field in User().get_fields():
-                if isinstance(
-                    field, BaseTemplateField
-                ) and field.replacement_collection == Collection("meeting"):
+                if (
+                    isinstance(field, BaseTemplateField)
+                    and field.replacement_collection
+                    and field.replacement_collection == Collection("meeting")
```
jsangmeister, line 150: what purpose serves this?

jsangmeister (changes requested):

tests/system/action/meeting/test_delete.py
r-peschke: Could be there is no replacement_collection, but a replacement_enum. With change of eq-operator to try/except construction in this case there will be an exception. Before there has been a silent `False`.

openslides_backend/shared/patterns.py
r-peschke: Done

openslides_backend/shared/patterns.py
```diff
@@ -59,9 +60,13 @@ def __repr__(self) -> str:
         return f"FullQualifiedId({repr(str(self))})"
 
     def __eq__(self, other: object) -> bool:
-        if not isinstance(other, FullQualifiedId):
-            return NotImplemented
-        return self.collection == other.collection and self.id == other.id
+        try:
+            return (
+                self.collection.collection == other.collection.collection  # type: ignore
+                and self.id == other.id  # type: ignore
```
jsangmeister, line 66: Also replace these "type: ignore"s and those below

tests/system/action/meeting/test_delete.py
jsangmeister: but here is no try/except :D you just added `and field.replacement_collection` which I would not think to make any difference (except maybe that the `Collection("meeting")` is not executed?)

jsangmeister (changes requested):

openslides_backend/action/actions/meeting/import_.py
```diff
@@ -570,10 +613,12 @@ def migrate_data(self, instance: Dict[str, Any]) -> Dict[str, Any]:
                     "resource_ids",
                     "organization_tag_ids",
                 ],
+                lock_result=False,
```
jsangmeister, line 616: This must be locked! Everytime relations are fetched, they have to be locked to prevent parallel alterations and therefore lost updates on it. Things that must not be locked are only static fields like `meeting_id` that do not ever change and fields for simple condition checks where the rationale is "if this is changed during this request, we'll just pretend that this request was issued slightly earlier".

openslides_backend/action/actions/meeting/import_.py
```diff
@@ -570,10 +613,12 @@ def migrate_data(self, instance: Dict[str, Any]) -> Dict[str, Any]:
...
             )
             committee = self.datastore.get(
                 FullQualifiedId(Collection("committee"), instance["committee_id"]),
                 ["meeting_ids"],
+                lock_result=False,
```
jsangmeister, line 621: same here

openslides_backend/permissions/permission_helper.py
```diff
@@ -118,6 +118,8 @@ def has_committee_management_level(
         user = datastore.get(
             FullQualifiedId(Collection("user"), user_id),
             ["organization_management_level", *cml_fields],
+            lock_result=False,
```
jsangmeister, line 121: same here, since the CML is a relation field

openslides_backend/shared/patterns.py
```diff
@@ -85,13 +91,14 @@ def __repr__(self) -> str:
...
-        return (
-            self.collection == other.collection
-            and self.id == other.id
-            and self.field == other.field
-        )
+        try:
+            return (
+                self.collection == other.collection  # type: ignore
+                and self.id == other.id  # type: ignore
+                and self.field == other.field  # type: ignore
```
jsangmeister, line 98: change this too please

openslides_backend/shared/patterns.py
```diff
@@ -118,9 +125,13 @@ def __repr__(self) -> str:
         return f"CollectionField({repr(str(self))})"
 
     def __eq__(self, other: object) -> bool:
-        if not isinstance(other, CollectionField):
-            return NotImplemented
-        return self.collection == other.collection and self.field == other.field
+        try:
+            return (
+                self.collection.collection == other.collection.collection  # type: ignore
+                and self.field == other.field  # type: ignore
```
jsangmeister, line 131: this too

openslides_backend/shared/patterns.py
```diff
@@ -32,9 +32,10 @@ def __repr__(self) -> str:
         return f"Collection({repr(str(self))})"
 
     def __eq__(self, other: object) -> bool:
-        if not isinstance(other, Collection):
-            return NotImplemented
-        return self.collection == other.collection
+        try:
+            return self.collection == cast(Collection, other).collection
+        except Exception as e:
+            raise NotImplementedError(e)
```
jsangmeister, line 38: Maybe we should just return `False` here? There should be no reason to not be able to compare on of these classes with some random object, they are just not equal then.

tests/system/action/meeting/test_delete.py
jsangmeister: I still don't understand what you mean, but it's not important

tests/system/action/meeting/test_import.py
```diff
@@ -1367,3 +1372,16 @@ def test_all_migrations(self) -> None:
...
         self.assert_model_exists("motion_workflow/1", {"sequential_number": 1})
         self.assert_model_exists("projector/2", {"sequential_number": 1})
+
+    @performance
+    def test_big_file(self) -> None:
+        data = {}
+        data["meeting"] = get_initial_data_file(
+            "global/data/big_data145MB_with_MI3.json"
+        )
+        # data["meeting"]["_migration_index"] = current_migration_index
```
jsangmeister, line 1382: remove comments here

tests/system/util.py
```diff
@@ -142,6 +143,27 @@ def mock_datastore_method(self, method: str) -> None:
...
+            elif orig_method.__name__ == "get":
+                self.calls_detail.append(
+                    (
+                        orig_method.__name__,
+                        {args[0]: args[1]},
+                        kwargs.get("lock_result", "undefined"),
+                    )
+                )
+            else:
+                self.calls_detail.append((orig_method.__name__, "NotImplemented"))
```
jsangmeister, line 166: For this I had the verbose flag. But if you like this more structured approach, please also print this data below instead of my rudimentary approach (just `print(self.call_details[-1]` should suffice)

jsangmeister (changes requested):

openslides_backend/shared/patterns.py
r-peschke: done

openslides_backend/shared/patterns.py
r-peschke: done

openslides_backend/shared/patterns.py
r-peschke: IMO we should insist on comparable objects. If we will change the type from FqId to string it will be useful to see the type.

jsangmeister: ok

PR #1319, opened 2022-05-02
Use -2 as default value for votes* fields in analog polls

jsangmeister:


luisa-beerboom:
This may just be Python-illiteracy, but as far as I understand it you are currently starting by unconditionally setting the field to -2 if it isn't set.

luisa-beerboom (approved):
Looks good to me

PR #1321, opened 2022-05-09
Update export_meeting

reiterl:
Add users to export and remove organization tags.

### DIFF
--- a/openslides_backend/presenter/export_meeting.py
+++ b/openslides_backend/presenter/export_meeting.py
@@ -50,12 +50,17 @@ class Export(BasePresenter):
             msg += f" Missing permission: {OrganizationManagementLevel.SUPERADMIN}"
             raise PermissionDenied(msg)
         export_data = export_meeting(self.datastore, self.data["meeting_id"])
-        self.exclude_organization_tags(export_data)
+        self.exclude_organization_tags_and_default_meeting_for_committee(export_data)
         self.add_users(export_data, self.data["meeting_id"])
         return export_data
 
-    def exclude_organization_tags(self, export_data: Dict[str, Any]) -> None:
+    def exclude_organization_tags_and_default_meeting_for_committee(
+        self, export_data: Dict[str, Any]
+    ) -> None:
         self.get_meeting_from_json(export_data)["organization_tag_ids"] = None
+        self.get_meeting_from_json(export_data)[
+            "default_meeting_for_committee_id"
+        ] = None
 
     def add_users(self, export_data: Dict[str, Any], meeting_id: int) -> None:
         user_ids = self.get_meeting_from_json(export_data)["user_ids"]
@@ -84,6 +89,7 @@ class Export(BasePresenter):
                 "last_email_send",
                 "is_demo_user",
                 "organization_management_level",
+                "is_present_in_meeting_ids",
             ],
         )
         users: Any = self.datastore.get_many([gmr])[Collection("user")]
@@ -120,7 +126,11 @@ class Export(BasePresenter):
                             str(meeting_id)
                         ]
             users[user_key]["meeting_ids"] = [meeting_id]
-            users[user_key]["is_present_in_meeting_ids"] = [meeting_id]
+
+            if meeting_id in (users[user_key].get("is_present_in_meeting_ids") or []):
+                users[user_key]["is_present_in_meeting_ids"] = [meeting_id]
+            else:
+                users[user_key]["is_present_in_meeting_ids"] = None
 
         export_data["user"] = users
 
--- a/openslides_backend/presenter/export_meeting.py
+++ b/openslides_backend/presenter/export_meeting.py
@@ -112,9 +112,9 @@ class Export(BasePresenter):
             + [field_pair[0] for field_pair in fields],
         )
         users: Any = self.datastore.get_many([gmr])[Collection("user")]
-        # change keys to strings and remove meta_* keys
+        # remove meta_* keys
         users = {
-            str(key): {
+            key: {
                 key_inner: value_inner
                 for key_inner, value_inner in value.items()
                 if not key_inner.startswith("meta_")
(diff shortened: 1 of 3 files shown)


openslides_backend/presenter/export_meeting.py
```diff
@@ -41,4 +50,90 @@ def get_result(self) -> Any:
...
+                        TemplateCharField,
+                        TemplateHTMLStrictField,
+                        TemplateDecimalField,
+                        TemplateRelationListField,
+                    ),
+                ):
+                    field_name = field.get_structured_field_name(meeting_id)
+                    user = self.datastore.get(
+                        FullQualifiedId(Collection("user"), int(user_key)), [field_name]
+                    )
```
jsangmeister, line 122: combine with the `get_many` before

openslides_backend/presenter/export_meeting.py
```diff
@@ -41,4 +50,90 @@ def get_result(self) -> Any:
...
+        users = {
+            key: {
+                key_inner: value_inner
+                for key_inner, value_inner in value.items()
+                if not key_inner.startswith("meta_")
+            }
+            for key, value in users.items()
+        }
+
+        for user_key in users:
```
jsangmeister, line 108: combine into one loop over the result

openslides_backend/presenter/export_meeting.py
```diff
@@ -41,4 +50,90 @@ def get_result(self) -> Any:
...
+                "default_vote_weight",
+                "last_email_send",
+                "is_demo_user",
+                "organization_management_level",
+                "is_present_in_meeting_ids",
+            ],
+        )
+        users: Any = self.datastore.get_many([gmr])[Collection("user")]
+        # change keys to strings
+        users = {str(key): value for key, value in users.items()}
```
jsangmeister, line 97: Is this even necessary? The result is converted to JSON somewhere at the end and JSON only uses string keys anyway.

jsangmeister (changes requested):

r-peschke (approved):
LGTM, except Joshua's comment.
I will remove the add_empty_fields-method in my import-meeting performance issue, but you don't need to worry whilst you are using the export_meeting method.

openslides_backend/presenter/export_meeting.py
reiterl: Done.

openslides_backend/presenter/export_meeting.py
reiterl: Done.

openslides_backend/presenter/export_meeting.py
reiterl: okay, I removed this code. The tests are okay.

openslides_backend/presenter/export_meeting.py
```diff
@@ -41,4 +50,91 @@ def get_result(self) -> Any:
...
+                "default_structure_level",
+                "default_vote_weight",
+                "last_email_send",
+                "is_demo_user",
+                "organization_management_level",
+                "is_present_in_meeting_ids",
+            ]
+            + [field_pair[0] for field_pair in fields],
+        )
+        users: Any = self.datastore.get_many([gmr])[Collection("user")]
```
r-peschke, line 114: - Do you need the `Any`? Because of mypy-errors? 
- Use optional parameters  `lock_results` and `use_changed_models` for performance reasons on get_many.

openslides_backend/presenter/export_meeting.py
```diff
@@ -41,4 +50,91 @@ def get_result(self) -> Any:
...
+                key_inner: value_inner
+                for key_inner, value_inner in value.items()
+                if not key_inner.startswith("meta_")
+            }
+            for key, value in users.items()
+        }
+
+        for user_key in users:
+            for field_name, field_template_name in fields:
+                if users[user_key].get(field_name):
```
r-peschke, line 127: use `(field_name := users[user_key].get(field_name)): Syntax and in next line `field_name`

openslides_backend/presenter/export_meeting.py
```diff
@@ -41,4 +50,91 @@ def get_result(self) -> Any:
...
+                "last_email_send",
+                "is_demo_user",
+                "organization_management_level",
+                "is_present_in_meeting_ids",
+            ]
+            + [field_pair[0] for field_pair in fields],
+        )
+        users: Any = self.datastore.get_many([gmr])[Collection("user")]
+        # remove meta_* keys
+        users = {
```
r-peschke, line 116: Can you combine this user loop with the following one. I think this is, what Joshua reclaimed.
And in this loop, you don't need to reconstruct the whole user.

To combine them, you could do something like:
for key, values in user.items():
     user[key] = new values without metas in your comprehensive expression
     do the template fields stuff

r-peschke (changes requested):

openslides_backend/presenter/export_meeting.py
reiterl: I removed the `Any`. it is not needed.
Added lock_results and use_changed_models.

openslides_backend/presenter/export_meeting.py
reiterl: I don't like the walroff notation. But I changed it a bit.

openslides_backend/presenter/export_meeting.py
reiterl: Yes, I could, but it is not what Joshua reclaimed. And it would obscure the code. I won't do it.

reiterl: Remove meta_* keys is not needed anymore, because of `lock_result = False` . I removed the code.

r-peschke (approved):

openslides_backend/presenter/export_meeting.py
```diff
@@ -41,4 +50,83 @@ def get_result(self) -> Any:
...
+                "can_change_own_password",
+                "gender",
+                "email",
+                "default_number",
+                "default_structure_level",
+                "default_vote_weight",
+                "last_email_send",
+                "is_demo_user",
+                "organization_management_level",
+                "is_present_in_meeting_ids",
```
jsangmeister, line 110: Generally, not locking relation fields should be forbidden as explained in https://github.com/OpenSlides/openslides-backend/pull/1318/files#r870980336. In this case, it's okay since it is just for the export and not for any updates, but you should keep this in mind for the future.

PR #1322, opened 2022-05-09
Change default values for updates to -2

jsangmeister:
see https://github.com/OpenSlides/openslides-client/issues/777#issuecomment-1118606683

### DIFF
--- a/openslides_backend/action/actions/option/update.py
+++ b/openslides_backend/action/actions/option/update.py
@@ -126,16 +126,11 @@ class OptionUpdateAction(UpdateAction):
         self, instance: Dict[str, Any], poll: Dict[str, Any]
     ) -> None:
         data = self._get_data(instance)
-
         if poll.get("type") == "analog":
-            if poll.get("pollmethod") == "N":
-                instance["no"] = data.get("no", "0.000000")
-            else:
-                instance["yes"] = data.get("yes", "0.000000")
-                if poll.get("pollmethod") in ("YN", "YNA"):
-                    instance["no"] = data.get("no", "0.000000")
-                if poll.get("pollmethod") == "YNA":
-                    instance["abstain"] = data.get("abstain", "0.000000")
+            pollmethod = poll["pollmethod"]
+            for key in ("yes", "no", "abstain"):
+                if key[0].upper() in pollmethod:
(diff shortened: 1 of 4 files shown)


tests/system/action/option/test_update.py
```diff
@@ -126,6 +126,42 @@ def test_update_Y(self) -> None:
...
+            {
+                "id": 57,
+                "Y": "1.000000",
+            },
+        )
+        self.assert_status_code(response, 200)
+        option = self.get_model("option/57")
+        assert option.get("yes") == "1.000000"
+        assert option.get("no") == "-2.000000"
+        assert option.get("abstain") == "0.000000"
```
r-peschke, line 163: Don't you think a -2 (=not documented) would be better? I understood that the client don't show the"not documented" values, but may show the 0.0000 value as "0 votes for this option".

openslides_backend/action/actions/option/update.py
```diff
@@ -126,16 +126,11 @@ def _handle_poll_option_data(
...
-                instance["no"] = data.get("no", "0.000000")
-            else:
-                instance["yes"] = data.get("yes", "0.000000")
-                if poll.get("pollmethod") in ("YN", "YNA"):
-                    instance["no"] = data.get("no", "0.000000")
-                if poll.get("pollmethod") == "YNA":
-                    instance["abstain"] = data.get("abstain", "0.000000")
+            pollmethod = poll["pollmethod"]
+            for key in ("yes", "no", "abstain"):
+                if key[0].upper() in pollmethod:
```
r-peschke, line 132: Maybe better to set all not given keys to -2 not_documented without `if`

jsangmeister: My preferred solution would be to just prevent keys which do not match the pollmethod - the client should not even display them and there should be no input for them anyway. Changing the values even thought the pollmethod does not match feels just as wrong as ignoring it.

r-peschke: IMO the keys not matching the pollmethod should be `None` (best solution), or -2, not documented, but 0.0000 is a valid, setted voting value. Maybe this is just your test setup?

jsangmeister:
Also forbid votes that don't match the pollmethod

Issue #1324, opened 2022-05-10
Error in analog poll, introduced with PR1322


r-peschke:
method OptionUpdateAction._handle_poll_option_data raises an error on each letter of YNA, which is not in pollmethod, no matter whether it is used or not.

PR #1325, opened 2022-05-10
#1324 fix poll option error

r-peschke:


openslides_backend/action/actions/option/update.py
```diff
@@ -121,13 +121,13 @@ def _get_option_and_poll(
     def _handle_poll_option_data(
         self, instance: Dict[str, Any], poll: Dict[str, Any]
     ) -> None:
-        data = self._get_data(instance)
         if poll.get("type") == "analog":
+            data = self._get_data(instance)
```
reiterl, line 125: change this by the other methods, too.

openslides_backend/action/actions/option/update.py
```diff
@@ -121,13 +121,13 @@ def _get_option_and_poll(
...
             pollmethod = poll["pollmethod"]
             for letter, key in option_keys_map.items():
                 if letter in pollmethod:
                     instance[key] = data.get(key, "-2.000000")
-                else:
+                elif data.get(key):
```
reiterl, line 130: Why?

tests/system/action/option/test_update.py
```diff
@@ -111,7 +125,7 @@ def test_update_invaid_keys(self) -> None:
             "option.update",
             {
                 "id": 57,
-                "A": "1.000000",
+                "A": "0.000000",
```
reiterl, line 128: Why is this test changed?

reiterl (commented):
I don't get it, why is this change needed.

openslides_backend/action/actions/option/update.py
r-peschke: Because with value, this poll.option is not allowed and should raise the following exception. Without beeing used in data (data.get(key) returns None) or beeing set to None the the exception shouldn't be raised.
Without these change, a pollmethod missing one of the letters "YNA" ALWAYS raise the exception, because for example the "A" is not in the pollmethod, but in the list of letters.

tests/system/action/option/test_update.py
r-peschke: Finally doesn't matter, but I wanted to demonstrate, that 0 votes should raise the exception

reiterl: Okay.

openslides_backend/action/actions/option/update.py
reiterl: If "Y" is not in pollmethod, providing "Y" in the payload should raise an **exception**. If "Y": None it is an client error,

reiterl (commented):
AFAIK, if a letter is not in the pollmethod, there should be an exception.

reiterl (approved):
After the discussion and looking in the wiki, LGTM.

PR #1328, opened 2022-05-12
Update motion.reset_state change created timestamp.

reiterl:


openslides_backend/action/actions/motion/reset_state.py
```diff
@@ -70,7 +71,6 @@ def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
             FullQualifiedId(Collection("motion_state"), instance["state_id"]),
             ["set_created_timestamp"],
         )
-        instance["created"] = None
-        if state.get("set_created_timestamp"):
+        if not motion.get("created") and state.get("set_created_timestamp"):
```
jsangmeister, line 74: Only fetch the state if needed, i.e. if `created` is not set

tests/system/action/motion/test_reset_state.py
```diff
@@ -124,7 +124,7 @@ def test_reset_state_correct_number_value(self) -> None:
         model = self.get_model("motion/22")
         assert model.get("state_id") == 76
         assert model.get("number_value") == 23
-        assert model.get("created") is None
+        assert model.get("created") == 1111111
```
jsangmeister, line 127: Add a test where the timestamp is set in the reset action

jsangmeister (changes requested):

openslides_backend/action/actions/motion/reset_state.py
reiterl: Done.

tests/system/action/motion/test_reset_state.py
jsangmeister: Ah, did not see it, you are right

jsangmeister (approved):

PR #1335, opened 2022-05-20
Implement ~= operator from datastore and filter emails in presenter case insensitive

### LINKED ISSUE #1334: The presenter `search_users_by_name_or_email` works case-sensitive
Please adjust the presenter so, that it works case-insensitive (so it doesn't matter, if the client sends `MaxMustermann` or `maxmustermann`, `mm@example.com` or `Mm@example.com`).

jsangmeister:
test with https://github.com/OpenSlides/openslides-datastore-service/pull/196

fixes #1334

I assumed that only the email should be matched case insensitive, since "MaxMustermann" and "maxmustermann" might be different users. @GabrielInTheWorld is this correct or should the username also be checked case insensitive?


r-peschke (changes requested):
Wait for @GabrielInTheWorld  response, in #1334 he wanted the username also o be case-insensitive.
Otherwise LGTM

GabrielInTheWorld:
> test with [OpenSlides/openslides-datastore-service#196](https://github.com/OpenSlides/openslides-datastore-service/pull/196)
> 
> fixes #1334
> 
> I assumed that only the email should be matched case insensitive, since "MaxMustermann" and "maxmustermann" might be different users. @GabrielInTheWorld is this correct or should the username also be checked case insensitive?

@jsangmeister You're right. I think, the username should also be checked case-insensitive. However, this can lead into some more problems, because already existing users can have usernames with a capitalized letter and the same username with a lower case version of the letter. @emanuelschuetze can you tell us, whether it is better to search through usernames case-insensitive or case-sensitive (which is the current state)?

jsangmeister:
Currently, everything is case sensitive. I just pushed the update which makes the username also case insensitive, but this is only 2 lines, I can change it back if @emanuelschuetze wants this.

emanuelschuetze:
> the username should also be checked case-insensitive

+1

r-peschke:
Please, think about **checked** (we may have duplicates in existing data and need to handle this problem, preventing the "more security" by case sensitivity) and **search** (we may find more than one with a username, but don't need to change user data). On login IMO we should stay case sensitive

jsangmeister:
> Please, think about **checked** (we may have duplicates in existing data and need to handle this problem, preventing the "more security" by case sensitivity) and **search** (we may find more than one with a username, but don't need to change user data). On login IMO we should stay case sensitive

Yes, of course. This PR only affects the `search_users_by_name_or_email` presenter, which is not used for logins.

PR #1345, opened 2022-05-30
fix changed_models filter creation for datastore issue 197

r-peschke:
A change made in datastore with [Issue197](https://github.com/OpenSlides/openslides-datastore-service/issues/197l) create a different sql for database filter, which leads to about 30 tests-errors in generation of memory filter for changed_models. This PR fixes these

### DIFF
--- a/openslides_backend/services/datastore/extended_adapter.py
+++ b/openslides_backend/services/datastore/extended_adapter.py
@@ -19,7 +19,10 @@ from .handle_datastore_errors import raise_datastore_error
 from .interface import Engine, LockResult, MappedFieldsPerFqid, PartialModel
 
 MODEL_FIELD_SQL = "data->>%s"
-COMPARISON_VALUE_SQL = "%s::text"
+MODEL_FIELD_NUMERIC_SQL = r"\(data->%s\)::numeric"
+MODEL_FIELD_NUMERIC_REPLACE = "(data->%s)::numeric"
+COMPARISON_VALUE_TEXT_SQL = "%s::text"
+COMPARISON_VALUE_SQL = "%s"
 
 
 class ExtendedDatastoreAdapter(CacheDatastoreAdapter):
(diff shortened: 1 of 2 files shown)


openslides_backend/services/datastore/extended_adapter.py
```diff
@@ -343,7 +346,7 @@ def _filter_changed_models(
         # transform query into valid python code
         filter_code = sql_query.lower().replace("null", "None").replace(" = ", " == ")
         # regex for all FilterOperators which were translated by the SqlQueryHelper
-        regex = rf"(?:{MODEL_FIELD_SQL}|lower\({MODEL_FIELD_SQL}\)) (<|<=|>=|>|==|!=|is|is not) ({COMPARISON_VALUE_SQL}|lower\({COMPARISON_VALUE_SQL}\)|None)"
+        regex = rf"(?:{MODEL_FIELD_SQL}|lower\({MODEL_FIELD_SQL}\)|{MODEL_FIELD_NUMERIC_SQL}|lower\({MODEL_FIELD_NUMERIC_SQL}\)) (<|<=|>=|>|==|!=|is|is not) ({COMPARISON_VALUE_SQL}|lower\({COMPARISON_VALUE_SQL}\)|{COMPARISON_VALUE_TEXT_SQL}|lower\({COMPARISON_VALUE_TEXT_SQL}\)|None)"
```
reiterl, line 349: Why is the `lower\({MODEL_FIELD_NUMERIC_SQL}\)` needed here? How can you lower a numeric value?

openslides_backend/services/datastore/extended_adapter.py
```diff
@@ -363,15 +366,20 @@ def _filter_changed_models(
                 formatted_args.append(f'model.get("{arguments[i]}")')
             i += 1
             # if comparison happens with a value, append it as well
-            if (
-                match[1] == COMPARISON_VALUE_SQL
-                or match[1] == f"lower({COMPARISON_VALUE_SQL})"
+            if match[1] in (
+                COMPARISON_VALUE_SQL,
+                f"lower({COMPARISON_VALUE_SQL})",
```
reiterl, line 371: same here.

reiterl (commented):
I'm not sure, that you should use the lower there.

openslides_backend/services/datastore/extended_adapter.py
r-peschke: The generated SQL for postgres may use the lower function inits code and our regex needs to find such an expression. Without it, the presenter for username_and_email will not work anymore

PR #1351, opened 2022-06-09
Issue1350 superadmin-user can't redraw superadmin-right or set himsel…

r-peschke:
…f inactive


peb-adr (approved):
Look good to me.
Also built and tested the production image.
Works as expected
  - can change (add/revoke) superadmin permission for other user
  - but not my own

PR #1354, opened 2022-06-14
Remove publish_immediately from write req. of poll and option

reiterl:
Publish immediately is not a poll fields. Rm it at the end of the
update_instance. Add a migration and update tests.

### DIFF
--- a/global/data/example-data.json
+++ b/global/data/example-data.json
@@ -1,5 +1,5 @@
 {
-    "_migration_index": 28,
+    "_migration_index": 29,
     "organization": {
         "1": {
             "id": 1,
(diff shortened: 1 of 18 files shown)


reiterl:
This is for the first part of the issue. The check in the write requests is missing.

r-peschke (changes requested):
SO far the code is okay for me. But the same happens in collection **option** with the field "publish_immediatelly". Please remove it also and write tests for it (for update you can use the option/test_update.py test_update_Y, where the field will be written to database.

reiterl:
Also fix the publish_immediately bug in option.update. option.create doesn't need a fix.

jsangmeister (approved):

r-peschke (approved):
LGTM

Issue #1137, opened 2022-01-11
Add option for threaded actions


jsangmeister:
Some actions like `meeting.import` (may) take longer than the timeout of 30s, so we need the option to execute the action in a thread to circumvent this timeout.

r-peschke:
Related to https://github.com/OpenSlides/openslides-datastore-service/pull/207 and https://github.com/OpenSlides/openslides-datastore-service/issues/206 as precondition

r-peschke:
@peb-adr @jsangmeister @rrenkert :
Proposal for some Gunicorn settings (https://docs.gunicorn.org/en/stable/settings.html):
workers: Gunicorn-Doc:  2-4 x $(NUM_CORES) , others recommend 2*$(NUM_CORES) + 1
worker_class: gthread  (see https://docs.gunicorn.org/en/stable/design.html)
threads: Gunicorn-Doc:  2-4 x $(NUM_CORES)  setting only available for gthread worker class
timeout: 30 sec is okay. Backend refreshes the timeout every 10 secs (fixed in code) by worker.tmp.notify()

Fixed in Backend:
wait until notify: 10s Fixed in backend/action_worker.py gunicorn_post_worker. Period to prolongate the alive cycle of gunicorns arbiter watcher
THREAD_WATCH_TIMEOUT: 1.0sec. Time for majority of short actions to end without action_worker-notification in database
Separation of action server and presenter server, both using the same settings: we should think about having different settings

Postgres:
Min and max DATASTORE_MIN/MAX_CONNECTIONS: minimum 6 set in pg_connection_handler.py belonging to datastore-package. Explanations: IMO should be number of Gunicorn-threads + 1 for special routes to reserve_ids and to write action_worker-Notification objects per Server-process. Last ones needed to write, even if the processing of the actions occupies all db-connections for a longer time.

Extra: We should be sure that the datastore writer standard route works like a singleton, otherwise the locked-fields mechanism doesn't work as designed. The writer route to get reserved_ids could be critical, too, but as far as I know has special isolation scope. The action_worker updates aren't critical IMO.

r-peschke:
Some notes on checking by test_action_worker.py -k test_action_worker_permanent_stress. No Gunicorn in test used, but the hook is implemened in test base.py!!!
1. 4 threads started in tests
2. gunicorn timeout 30sec
3. datastore_min/max_connections: 10
4. test 4 typen a 200 Sätze in 4 request threads
[...]

PR #1355, opened 2022-06-15
Issue1137 thread_watcher for actions with timeout

r-peschke:


luisa-beerboom:
Can you give a description (or point me to appropriate resources) as to what exactly happens from the client developer point of view when the timeout occurs (what will the 202 message look like, what will the watchers that I will subscribe to look like, where in the 202-response will I find the Fqid)?

r-peschke:
> Can you give a description somewhere (or point me to appropriate resources) as to what exactly happens from the client developer point of view when the timeout occurs (what will the 202 message look like, what will the watchers that I will subscribe to look like, etc)?

see https://github.com/OpenSlides/openslides-backend/issues/1137#issuecomment-1191543586

r-peschke:
State: Using worker_class GThread the timeout problem of Gunicorn processes seems solved. 
Importing a 150 MB meeting is still not possible, because of memory problems. I miss used the action_worker result field to write some memory dates into the database, the last ones cannot be written to the db, but are in log. May we should abort the watcher thread, if memory consumption is about 95%.

openslides_backend/action/action_worker.py
```diff
@@ -0,0 +1,288 @@
...
+    def __init__(
+        self,
+        user_id: int,
+        starttime: int,
+        logging: LoggingModule,
+        action_names: str,
+        datastore: DatastoreService,
+    ) -> None:
+        self.user_id = user_id
+        self.starttime = starttime
```
jsangmeister, line 94: Can't you just set the time here instead of using a parameter? I don't see how these milliseconds will change anything and saving parameters (especially if there are this many) is always better.

But If you insist on it, please use snake case (`start_time`).

openslides_backend/action/action_worker.py
```diff
@@ -0,0 +1,288 @@
...
+        self.logger.debug(f"action_worker: {message}, written:{self.written}")
+        return message
+
+    def continue_action_worker_write(self) -> None:
+        current_time = round(time())
+        with self.datastore.get_database_context():
+            ram = psutil.virtual_memory()
+            response = {
+                "success": False,
+                "message": f"ram total:{ram.total} available:{ram.available} percent:{ram.percent} used:{ram.used} free.{ram.free}",
```
jsangmeister, line 146: Do you really want to spam the DB with these informations? I would think they rather belong in the logs instead of the db, and maybe not on each iteration, but on each xth (x=100? 1000?).

openslides_backend/action/action_worker.py
```diff
@@ -0,0 +1,288 @@
...
+                        Event(
+                            type=EventType.Update,
+                            fqid=self.fqid,
+                            fields={
+                                "timestamp": current_time,
+                                "result": response,
+                            },
+                        )
+                    ],
+                    information={self.fqid: ["timestamp during running action_worker"]},
```
jsangmeister, line 162: see above

openslides_backend/action/action_worker.py
```diff
@@ -0,0 +1,288 @@
...
+                    user_id=self.user_id,
+                    locked_fields={},
+                )
+            )
+            self.logger.debug(
```
jsangmeister, line 167: Mmh, don't you think that this is too much spam? Printing the current time is also not really meaningful. But the message itself might help to determine whether the thread is still running or not, I'm not sure.

openslides_backend/action/action_worker.py
```diff
@@ -0,0 +1,288 @@
...
+                            type=EventType.Update,
+                            fqid=self.fqid,
+                            fields={
+                                "state": state,
+                                "timestamp": current_time,
+                                "result": response,
+                            },
+                        )
+                    ],
+                    information={self.fqid: ["finish action_worker"]},
```
jsangmeister, line 205: see above

openslides_backend/action/action_worker.py
```diff
@@ -0,0 +1,288 @@
...
+                                fields={
+                                    "id": self.new_id,
+                                    "name": self.action_names,
+                                    "state": "running",
+                                    "created": self.starttime,
+                                    "timestamp": current_time,
+                                },
+                            )
+                        ],
+                        information={self.fqid: ["create action_worker"]},
```
jsangmeister, line 125: please remove. We only want meaningful entries in the history, as introduced in https://github.com/OpenSlides/openslides-backend/pull/1356

openslides_backend/action/action_worker.py
```diff
@@ -0,0 +1,288 @@
...
+                            },
+                        )
+                    ],
+                    information={self.fqid: ["finish action_worker"]},
+                    user_id=self.user_id,
+                    locked_fields={},
+                )
+            )
+            self.logger.debug(
+                f"action_worker '{self.fqid} {self.action_names}': {current_time}"
```
jsangmeister, line 211: Again, the current time does not really make sense, but I would expect some wording of "final" or "finished".

openslides_backend/action/action_worker.py
```diff
@@ -0,0 +1,288 @@
...
+    curr_thread = threading.current_thread()
+    for thread in threading.enumerate():
+        if (
+            thread.name == "action_worker"
+            and cast(ActionWorker, thread).thread_executor_pid == curr_thread.native_id
+        ):
+            action_worker = cast(ActionWorker, thread)
+            break
+
+    if not action_worker:
```
jsangmeister, line 273:
```suggestion
    else:
```
finally a meaningful usage! :D

openslides_backend/action/action_worker.py
```diff
@@ -0,0 +1,288 @@
...
+    if lock.acquire(timeout=THREAD_WATCH_TIMEOUT):
+        if hasattr(action_worker_thread, "exception"):
+            logger.debug("Action request finished with exception within timeout.")
+            lock.release()
+            raise action_worker_thread.exception
+        if hasattr(action_worker_thread, "response"):
+            logger.debug("Action request finished successfully within timeout.")
+            lock.release()
+            return action_worker_thread.response
+        lock.release()
```
jsangmeister, line 62: you can just release the lock as the first thing above and remove the 3 separate calls

setup.cfg
```diff
@@ -51,6 +51,15 @@ ignore_missing_imports = true
...
 ignore_missing_imports = true
 
+[mypy-gunicorn.http.message]
+ignore_missing_imports = true
+
+[mypy-gunicorn.http.wsgi]
+ignore_missing_imports = true
+
+[mypy-gunicorn.workers.gthread]
+ignore_missing_imports = true
```
jsangmeister, line 61: Can you combine that with `[mypy-gunicorn.*]`?

tests/system/action/motion/test_create_sequential_number.py
```diff
@@ -144,8 +146,13 @@ def test_create_sequential_numbers_deleted_motion(self) -> None:
         model = self.get_model("motion/2")
         self.assertEqual(model.get("sequential_number"), 2)
 
+    @pytest.mark.skip(
+        reason="Not working with watch-thread, because the testlock is unknown in action_worker"
+    )
     def test_create_sequential_numbers_race_condition(self) -> None:
         """
+        !!!We could delete this test or implement a switch-off for the action_worker procedure at all!!!
```
jsangmeister, line 154: I would support the second option. Maybe with a header like `disable-action-worker`? I think this test is valuable and should be kept.

tests/system/action/topic/test_create.py
```diff
@@ -56,11 +56,40 @@ def test_create_multiple_requests(self) -> None:
                 },
             ],
         )
-        self.assert_status_code(response, 400)
-        self.assertIn(
-            "Datastore service sends HTTP 400. The following locks were broken: 'list_of_speakers/meeting_id', 'list_of_speakers/sequential_number', 'topic/meeting_id', 'topic/sequential_number'",
-            response.json["message"],
-        )
+        if response.status_code == 202:
```
jsangmeister, line 59: Mmh, this can potentially happen with every request, right? Can't you just set the `THREAD_WATCH_TIMEOUT` to something high enough (or even better: 0 to disable it, if it's not too much work) so we don't need this kind of special handling? (Or alternatively use the header mentioned above)

I really don't like that random tests could potentially fail now because of this timeout when this has nothing to do with what we're trying to test, also adding this special handling to every test is no solution either.

tests/system/base.py
```diff
@@ -73,6 +74,17 @@ def setUp(self) -> None:
         self.vote_service.clear_all()
         self.anon_client = self.create_client()
 
+    def tearDown(self) -> None:
```
jsangmeister, line 77: This can then be removed as well if we somehow disable this behaviour, I think

openslides_backend/action/action_handler.py
```diff
@@ -44,17 +45,21 @@
         "description": "An array of actions",
         "type": "array",
         "items": {
-            "type": "object",
-            "properties": {
-                "action": {
-                    "description": "Name of the action to be performed on the server",
-                    "type": "string",
-                    "minLength": 1,
+            "anyOf": [
```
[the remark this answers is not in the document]
r-peschke: yup and done

openslides_backend/action/action_worker.py
```diff
@@ -0,0 +1,288 @@
...
+    handler: ActionHandler,
+) -> ActionsResponse:
+    starttime = round(time())
+    curr_thread = threading.current_thread()
+    logger = handler.logging.getLogger(__name__)
+    lock = threading.Lock()
+    try:
+        action_names = ",".join(elem.get("action", "") for elem in payload)
+    except Exception:
+        action_names = "Cannot be determined"
```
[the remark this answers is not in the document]
r-peschke: Yes, but this meta information will be build before the action_worker-thread is even started and the schema will be checked

openslides_backend/action/action_worker.py
r-peschke: done and done

openslides_backend/action/action_worker.py
r-peschke: This is the "Still alive"-signal for the client or anybody else, who wants to wait for the end of the action. It will be written every 10seconds, see lock.acquire in method gunicorn_post_request.
I shortened the memory information, but for now, it may be usefull.

openslides_backend/action/action_worker.py
r-peschke: removed

openslides_backend/action/action_worker.py
r-peschke: All the action_worker writing will be made only by very rare actions, the big majority of them don't need a second and will not use action_worker-notifiers. This intermediate update-writing appears after 11sec for the first time. That are not common actions using the action_worker-notifyings.

openslides_backend/action/action_worker.py
r-peschke: removed

openslides_backend/action/action_worker.py
r-peschke: done

openslides_backend/action/action_worker.py
r-peschke: done

openslides_backend/action/action_worker.py
r-peschke: done, but even more: most of this logger-entry are really useless, because this is the "business-as-usual"-case, except the last one, which is an heavy error

setup.cfg
r-peschke: done

tests/system/base.py
r-peschke: Let's talk about it!

tests/system/action/topic/test_create.py
r-peschke: This test covers both possible cases and I think this is a really good test as an example, how backend-actions have to be handled. This is implemented in the client with https://github.com/OpenSlides/openslides-client/pull/1423.

tests/system/action/motion/test_create_sequential_number.py
r-peschke: This is not as easy as it seems to be in the initial part of the job, because each action is processed in the action_worker-thread.
There is a new issue issue #1473 to fix this, because for now we want the functionality of the thread watcher

openslides_backend/action/action_worker.py
jsangmeister: ok

tests/system/action/topic/test_create.py
jsangmeister: Having a specific test (or multiple) to test the action worker behaviour is good and needed, but as I see it, currently every test might fail randomly if it takes longer than 1s since the backend then returns a 202 instead of the code expected by the test (200/400/...). This is a problem and should be fixed before we merge this. My proposal (since implementing some kind of off-switch for the whole action worker is as you say more work to implement): Set the `THREAD_WATCH_TIMEOUT` to `-1` automatically in every test (by mocking it or overwriting it, your choice). This tells `lock.acquire()` to wait indefinitely (https://docs.python.org/3/library/threading.html#threading.Lock.acquire), which would basically be the same as disabling the action worker if I understand it correctly.

tests/system/base.py
jsangmeister: see above

r-peschke: This can't be removed, because the action_worker-thread is always used

tests/system/action/topic/test_create.py
r-peschke: done

openslides_backend/action/action_handler.py
```diff
@@ -54,7 +55,7 @@
                 "data": action_data_schema,
             },
             "required": ["action", "data"],
-            "additionalProperties": False,
+            "additionalProperties": [False],
```
jsangmeister, line 58: I think this was an accident?

openslides_backend/action/action_handler.py
```diff
@@ -65,7 +66,7 @@ class ActionHandler(BaseHandler):
     Action handler. It is the concrete implementation of Action interface.
     """
 
-    MAX_RETRY = 3
+    MAX_RETRY = 100
```
jsangmeister, line 69: This was probably for debugging I guess, please revert

openslides_backend/http/application.py
```diff
@@ -63,10 +63,21 @@ def dispatch_request(self, request: Request) -> Union[Response, HTTPException]:
                 raise
         except HTTPException as exception:
             return exception
+        if type(response_body) == dict:
+            status_code = response_body.get("status_code", 200)
+        elif request.path == r"/system/presenter/handle_request":
```
jsangmeister, line 68: Mmh, I don't like that the generic application should have specific knowledge of the underlying views. Can you find a more generic way to handle this here? What is even the problem?

tests/system/action/speaker/test_create.py
```diff
@@ -189,7 +189,7 @@ def test_create_add_2_speakers_in_2_actions(self) -> None:
         )
         self.assert_status_code(response, 400)
         self.assertIn(
-            "Datastore service sends HTTP 400. The following locks were broken: 'speaker/list_of_speakers_id', 'speaker/meeting_id', 'speaker/weight'",
+            "Datastore service sends HTTP 400. The following locks were broken: 'speaker/list_of_speakers_id', 'speaker/meeting_id', 'speaker/weight",
```
jsangmeister, line 192: Probably a mistake as well

tests/system/action/base.py
```diff
@@ -347,7 +366,7 @@ def base_permission_test(
             self.set_models(models)
         response = self.request(action, action_data)
         if permission:
-            self.assert_status_code(response, 200)
+            self.assertIn(response.status_code, (200, 202))
```
jsangmeister, line 369: You can probably revert this with the new timout of -1

jsangmeister (changes requested):

openslides_backend/action/action_handler.py
r-peschke: Done

openslides_backend/action/action_handler.py
r-peschke: Reverted

tests/system/action/speaker/test_create.py
r-peschke: done

tests/system/action/base.py
r-peschke: done

openslides_backend/http/application.py
r-peschke: Talked with Raimund: I don't know a more generic way. First case should be the standard: Response with status_code in a dict for actions, presenter or anything else, that may be.
2nd case: Presenter without status code: Finally a kind of error, but in practice working, consequence: Let it pass and assume a 200.
2rd case: An action or anything else without status code: throw an exception, this is not allowed, because potentially dangerous.
I don't have a better solution!

requirements/partial/requirements_production.txt
```diff
@@ -16,3 +16,4 @@ opentelemetry-instrumentation-requests
 opentelemetry-instrumentation-flask
 protobuf==3.20.1
 Babel==2.10.2
+psutil==5.9.1
```
peb-adr, line 19: won't be needed if ram usage will no longer be reported in `openslides_backend/action/action_worker.py`

peb-adr (changes requested):

openslides_backend/http/views/action_view.py
```diff
@@ -42,9 +43,9 @@ def action_route(self, request: Request) -> Tuple[ResponseBody, Optional[str]]:
         handler = ActionHandler(self.env, self.services, self.logging)
         Translator.set_translation_language(request.headers.get("Accept-Language"))
         is_atomic = not request.environ["RAW_URI"].endswith("handle_separately")
-        response = handler.handle_request(request.json, user_id, is_atomic)
-
-        self.logger.debug("Action request finished successfully.")
+        response = handle_action_in_worker_thread(
+            request.json, user_id, is_atomic, handler
+        )
```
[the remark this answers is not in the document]
r-peschke: done

openslides_backend/action/action_worker.py
```diff
@@ -0,0 +1,289 @@
...
+        self.logger.debug(f"action_worker: {message}, written:{self.written}")
+        return message
+
+    def continue_action_worker_write(self) -> None:
+        current_time = round(time())
+        with self.datastore.get_database_context():
+            ram = psutil.virtual_memory()
+            response = {
+                "success": False,
+                "message": f"Still running, percentage ram used {ram.percent} from total:{ram.total}.",
```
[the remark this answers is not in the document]
r-peschke: response removed => done

openslides_backend/action/action_worker.py
```diff
@@ -0,0 +1,289 @@
...
+                    )
+                )
+                self.datastore.get(
+                    self.fqid, [], lock_result=False, use_changed_models=False
+                )
+                message = f"Action lasts too long. {self.fqid} written to database. Get the result from database, when the job is done."
+                self.written = True
+            except DatastoreException as e:
+                message = f"Action lasts too long, exception on writing {self.fqid}: {e.message}. Get the result later from database."
+        self.logger.debug(f"action_worker: {message}, written:{self.written}")
```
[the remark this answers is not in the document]
r-peschke: use info-level log and removed explicitly `written`-key

openslides_backend/action/action_worker.py
```diff
@@ -0,0 +1,289 @@
...
+        current_time = round(time())
+        with self.datastore.get_database_context():
+            state = "end"
+            if hasattr(action_worker_thread, "exception"):
+                response = {
+                    "success": False,
+                    "message": str(action_worker_thread.exception),
+                    "action_error_index": 0,
+                    "action_data_error_index": 0,
+                }
```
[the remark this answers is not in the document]
r-peschke: done

openslides_backend/action/action_worker.py
```diff
@@ -0,0 +1,289 @@
...
+            elif hasattr(action_worker_thread, "response"):
+                response = action_worker_thread.response
+            else:
+                state = "aborted"
+                response = {
+                    "success": False,
+                    "message": "action_worker aborted without any specific message",
+                    "action_error_index": 0,
+                    "action_data_error_index": 0,
+                }
```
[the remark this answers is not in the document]
r-peschke: done

openslides_backend/action/action_worker.py
```diff
@@ -0,0 +1,289 @@
...
+                            },
+                        )
+                    ],
+                    user_id=self.user_id,
+                    locked_fields={},
+                )
+            )
+            self.logger.debug(
+                f"finish action_worker '{self.fqid} {self.action_names}': {current_time}"
+            )
```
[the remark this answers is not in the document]
r-peschke: done

openslides_backend/main.py
```diff
@@ -57,9 +58,14 @@ def load_config(self) -> None:
             "workers": int(self.env.OPENSLIDES_BACKEND_NUM_WORKERS),
             "worker_tmp_dir": "/dev/shm",  # See https://pythonspeed.com/articles/gunicorn-in-docker/
             "timeout": int(self.env.OPENSLIDES_BACKEND_WORKER_TIMEOUT),
+            "graceful_timeout": 30,  # time to finish after receiving restart signal
             "loglevel": self.env.get_loglevel().lower(),
             "reload": self.env.is_dev_mode(),
             "reload_engine": "auto",  # This is the default however.
+            "worker_class": "gthread",  # async gthread with unlimite prolongation possibility
+            "threads": 3,  # Threads per Worker(process)
```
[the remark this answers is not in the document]
r-peschke: done in env.py

requirements/partial/requirements_datastore_as_package.txt
```diff
@@ -1 +1 @@
-git+https://github.com/OpenSlides/openslides-datastore-service.git@${DATASTORE_COMMIT_HASH}
+git+https://github.com/openslides/openslides-datastore-service.git@main
```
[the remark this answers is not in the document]
r-peschke: done

requirements/partial/requirements_production.txt
r-peschke: removed

setup.cfg
```diff
@@ -69,6 +69,9 @@ ignore_missing_imports = true
 [mypy-authlib.*]
 ignore_missing_imports = true
 
+[mypy-psutil]
+ignore_missing_imports = true
+
```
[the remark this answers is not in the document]
r-peschke: removed

requirements/partial/requirements_development.txt
```diff
@@ -15,6 +15,6 @@ git+https://github.com/foutrelis/aiosmtpd.git@e3321f474c0d8f63a17186a6e727f490d9
 
 # typing
 types-PyYAML==6.0.5
-types-bleach==4.1.5
+types-bleach==4.1.4
```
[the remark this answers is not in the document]
r-peschke: Today github is happy also with 4.1.5, which is now back again

openslides_backend/http/application.py
```diff
@@ -63,10 +63,21 @@ def dispatch_request(self, request: Request) -> Union[Response, HTTPException]:
...
+            raise ViewException(f"Unknown type of response_body:{response_body}.")
+
         self.logger.debug(
-            f"All done. Application sends HTTP 200 with body {response_body}."
+            f"All done. Application sends HTTP {status_code} with body {response_body}."
+        )
+        response = Response(
+            json.dumps(response_body),
+            status=status_code,
+            content_type="application/json",
```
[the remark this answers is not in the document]
peb-adr: ok

requirements/partial/requirements_development.txt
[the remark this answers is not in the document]
peb-adr: oh well :P

peb-adr (approved):

PR #1356, opened 2022-06-15
More verbose history informations

jsangmeister:


jsangmeister:
Some open things:
- for OS3, updated to e.g. polls or motion comments also triggered an entry for the motion in the history. In OS4, this is no longer easily possible since the motion itself does not get updated. We could either ignore these history entries or find a way to insert it into the event log anyway. (This is also why 3 tests are still failing)
- The history information format changed from a mapping from FQID to string to a list of strings, analogous to OS3 (the first string is the actual message, the rest are arguments which are encoded as `{argx}` in the message for easier translations.
  - The client has to be adjusted
  - What do we do with old history entries in the dict format? I would rather not keep both functionalities in the client, so the easiest would be to drop all existing informations alltogether or don't display them in the client (a simple "is object" check should be sufficent)

r-peschke (commented):
Like you wrote there are still errors in test.
I looked for the Issue or a description what this PR is for, but didn't find any. Would you please write some?

jsangmeister:
> Like you wrote there are still errors in test. I looked for the Issue or a description what this PR is for, but didn't find any. Would you please write some?

This was discussed at the second to last planning meeting and introduced by @emanuelschuetze - the history informations are too generic and vague and should be replaced by less positions with any information at all and more specific informations, if any are present.

r-peschke (approved):

jsangmeister:
Btw, our migrations are currently not able to change the history informations, so we would either have to drop the history information manually for all instances (a simple SQL command should suffice, which could maybe batch-like deployed to all instances), enhance the migrations (which seems overkill for just this use case - I don't think we'll have a similar case again) or take the other route and ignore old entries in the client (also not preferable, since we keep deprecated code in our application).

r-peschke (approved):

jsangmeister:
no discussion needed, will be merged together with https://github.com/OpenSlides/openslides-client/pull/1429

emanuelschuetze (approved):
Tested

emanuelschuetze:
Talked with @peb-adr that we have to run `UPDATE positions SET information=NULL` in postgres

PR #1361, opened 2022-06-17
Remove old restriction modes that are not needed anymore

ostcar:


openslides_backend/models/models.py
```diff
@@ -2,8 +2,9 @@
 
 from openslides_backend.models import fields
 from openslides_backend.models.base import Model
+from openslides_backend.shared.patterns import Collection
```
jsangmeister, line 5: Where does this come from/is this needed?

jsangmeister: Ok, it seems to be a relict in `generate_models.py`. Can you remove it from the template there?

PR #1362, opened 2022-06-19
Poll stop no locked users

ostcar:


r-peschke (changes requested):
- deleting all user related locked-fields is not secure in terms of database inetegrity
- this way this patch is injected into poll.stop (wanted I assume) and poll.publish (don't know if you know/want this, too)

emanuelschuetze (approved):

Issue #1363, opened 2022-06-20
poll.stop lasts to long on filled database


r-peschke:
1. Minimize locked fields for users, because they are very expensive on a database with al lot of events
   user/is_present_in_meeting, organization_management_level, group_$_ids, vote_delegated_$_to_id, group_$x_ids, vote_delegated_$x_to_id: doesn't matter in poll.stop, because the voting is done and changing one of this fields doesn't affect the result and none of them will be changed by the resulting write_request -> #1364

  **Remaining locked_fields for user:**,
  user.vote_$_ids and user.vote_$n_ids: necessary, because other vote in same/other meeting could be lost in racing condition
  user.poll_voted_$_ids and poll_voted_$n_ids: necessary, because poll in same/other meeting could be lost in racing condition
  vote_delegated_vote_$_ids and vote_delegated_vote_$n_ids: necessary, because delegated vote sin same/other meeting could be lost in racing condition
 
2. Accelerate the datastore access for the locked fields by better SQL and setting indices -> https://github.com/OpenSlides/openslides-datastore-service/pull/199

jsangmeister:
Can this be closed? The specific problem is solved as far as I know and I don't think we should keep all performance-related issues open just because we can still improve the performance

PR #1364, opened 2022-06-20
#1363 Remove unneccessary locked_fields, some improvements

r-peschke:
Reduces the amount of locked_field, but increases the amount of database calls, because calls without locked_results are not cached. In this case, for each voting user we need 1 database call extra. 
Opened #1366 for this


jsangmeister (approved):
looks good minus comments

PR #1369, opened 2022-06-22
Fix errors on import for merged users

r-peschke:
Calculated fields had problems with meeting.import and meeting.clone, because the trigger field group_$_ids in the imported/cloned instance doesn't contain the values for the existing users. The import-code uses list-fields with add.
Tosolve the problem forcalculated fields, the PR removes the "remove" part of list_field event for the actions meeting.import an dmeeting.clone.


openslides_backend/action/relations/user_committee_calculate_handler.py
```diff
@@ -79,7 +79,11 @@ def process_field(
         )
         new_committees_ids.update(committee_ids)
         added_ids = new_committees_ids - db_committee_ids
-        removed_ids = db_committee_ids - new_committees_ids
+        removed_ids = (
+            db_committee_ids - new_committees_ids
+            if action not in ("meeting.import", "meeting.clone")
+            else []
```
reiterl, line 85: Yes, that works and seems to be correct, but I would add this kind of code in `meeting.import`.

r-peschke:
Closed in favor of #1370

PR #1379, opened 2022-06-27
Fix the oml key error and add a test.

reiterl:
The problem was to reproduce the failure with a test. I found one test.


jsangmeister (approved):
looks good

PR #1380, opened 2022-06-27
Fix deletion of motion in projector history

jsangmeister:
If a motion was projected and then moved in to the projection history, it could no longer be deleted because of the projection blocking it.


peb-adr (approved):
perfekt.

Built and tested on production clone. Fixed the problem.
Test looks reasonable as well.

r-peschke (approved):

PR #1384, opened 2022-06-29
use local datastore for writer including debug

r-peschke:
Also removes some really time-consuming right settings to 777, that nobody seems to need.

Debugging needs also [PR](https://github.com/OpenSlides/openslides-datastore-service/pull/202) from datastore with modified entry.

### DIFF
--- a/dev/Dockerfile.dev
+++ b/dev/Dockerfile.dev
@@ -22,8 +22,6 @@ COPY setup.cfg .
 COPY dev/entrypoint.sh ./
 
 RUN chmod 777 -R .
-RUN chmod 777 /root
-RUN chmod 777 -R /usr/local/lib/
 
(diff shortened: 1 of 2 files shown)


dev/Dockerfile.dev
```diff
@@ -22,8 +22,6 @@ COPY setup.cfg .
 COPY dev/entrypoint.sh ./
 
 RUN chmod 777 -R .
-RUN chmod 777 /root
-RUN chmod 777 -R /usr/local/lib/
```
jsangmeister, line 26: I'm sure there was a reason for this. But we can remove it for now and see when problems pop up.

r-peschke: @peb-adr We removed the 777-right settings from backends Dockerfile.**dev**

PR #1386, opened 2022-06-30
Update mail invitations: inner loop catches exception and put count a…

reiterl:
…nd index into the message.

### DIFF
--- a/openslides_backend/action/actions/user/send_invitation_email.py
+++ b/openslides_backend/action/actions/user/send_invitation_email.py
@@ -87,6 +87,8 @@ class UserSendInvitationMail(EmailMixin, UpdateAction):
                         result["message"] = e.message
                     except SMTPDataError as e:
                         result["message"] = f"SMTPDataError: {str(e)}"
+                    except SMTPSenderRefused as e:
+                        raise e
                     except Exception as e:
                         result["message"] = f"Exception: {str(e)}"
 
@@ -97,7 +99,7 @@ class UserSendInvitationMail(EmailMixin, UpdateAction):
                     else:
                         result["message"] = (
                             str(result["message"])
-                            + f" Count {len(list(action_data))}, Index {self.index}"
+                            + f" Mail {self.index+1} from {len(list(action_data))}"
                         )
 
                     self.results.append(result)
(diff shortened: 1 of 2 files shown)


tests/system/action/user/test_send_invitation_email.py
```diff
@@ -208,7 +208,7 @@ def test_SMTPSenderRefused_not_authenticated(self) -> None:
             )
         self.assert_status_code(response, 200)
         self.assertIn(
-            f"SMTPSenderRefused: (530, b'5.7.0 Authentication required', '{EmailSettings.default_from_email}')",
+            f"Exception: (530, b'5.7.0 Authentication required', '{EmailSettings.default_from_email}') Count 1, Index 0",
```
r-peschke, line 211: This should not occur. Please catch the SMTPSenderRefused in the inner part and throw it again to be caught in outer asserts. Reason: This occurs in sending email in **connected state**, but the sender is always the same and there is no need to assert each of the mails separately.
You'll stay with the problem of writing a test for this. You could use the AIOHandler in tests/system/action/mail_base.py and raise an Exception for a special address in mail. But that's on your fantasy;-)

openslides_backend/action/actions/user/send_invitation_email.py
```diff
@@ -87,11 +87,18 @@ def perform(
...
+                        result["message"] = f"Exception: {str(e)}"
 
                     if result["sent"]:
                         instance.pop("meeting_id", None)
                         write_request = self.create_write_requests(instance)
                         self.write_requests.extend(write_request)
+                    else:
+                        result["message"] = (
+                            str(result["message"])
+                            + f" Count {len(list(action_data))}, Index {self.index}"
```
r-peschke, line 100: Proposal: f"Mail {self.index+1} from {len(action_data)}" for the index-part of the message.

r-peschke (changes requested):

openslides_backend/action/actions/user/send_invitation_email.py
reiterl: Done.

tests/system/action/user/test_send_invitation_email.py
reiterl: I have reraised the SMTPSenderRefused Exception. And cleanup the code.

r-peschke (approved):

PR #1398, opened 2022-07-11
Issue1396 Remove forbidden fields to allow import of exported data

r-peschke:


reiterl (approved):
I find it a bit overkill to write a `remove_from_collection` function. But okay..
Else LGTM.

openslides_backend/action/actions/meeting/import_.py
```diff
@@ -142,22 +139,25 @@ def check_one_meeting(self, instance: Dict[str, Any]) -> None:
         if not len(meeting_json.get("meeting", {}).values()) == 1:
             raise ActionException("Need exactly one meeting in meeting collection.")
 
-    def check_not_allowed_fields(self, instance: Dict[str, Any]) -> None:
+    def remove_not_allowed_fields(self, instance: Dict[str, Any]) -> None:
         json_data = instance["meeting"]
+        regex_cml = re.compile(r"^committee_\$(\D)*_management_level$")
+
+        def remove_from_collection(
```
jsangmeister, line 146: I agree with @reiterl that this method is confusing since it is not used in any other place.

PR #1404, opened 2022-07-13
Issue1402 presenter search username email

r-peschke:


openslides_backend/presenter/search_users_by_name_or_email.py
```diff
@@ -63,31 +64,41 @@ class SearchUsersByNameEmail(BasePresenter):
...
-            elif email:
-                filter_ = FilterOperator("email", "~=", email)
-            else:
-                continue
-            instances = self.datastore.filter(
-                "user",
-                filter_,
-                ["id", "first_name", "last_name", "email"],
+            user_ids: Set[int] = userd.get(username.lower(), set()).union(
+                emaild.get(email.lower(), set())
```
jsangmeister, line 96:
```suggestion
            user_ids: Set[int] = userd[username.lower()] | emaild[email.lower()]
```
you can use the index operator directly since they are defaultdicts

jsangmeister (changes requested):

reiterl (approved):
I mainly reviewed the presenter code. LGTM. Yes, Joshua's comment could be useful.

openslides_backend/presenter/search_users_by_name_or_email.py
```diff
@@ -67,9 +67,9 @@ def get_result(self) -> Any:
         result: Dict[str, List[Dict[str, Union[str, int]]]] = defaultdict(list)
         filter_bulk_tuples: Set[Tuple[str, str]] = set()
         for search in self.data["search"]:
-            if username := search.get("username"):
+            if username := search.get("username", "").strip():
                 filter_bulk_tuples.add(("username", username))
-            if email := search.get("email"):
+            if email := search.get("email", "").strip():
```
jsangmeister, line 72: Although it would make sense, we currently do not strip the email field. So either remove this here or add this in the user mixin so the stripping is consistent.

jsangmeister (changes requested):

openslides_backend/presenter/search_users_by_name_or_email.py
r-peschke: done, stripped the email

PR #1407, opened 2022-07-18
Fix datastore commit

jsangmeister:
was mistakenly set to the wrong branch in https://github.com/OpenSlides/openslides-backend/pull/1273. Also updated the datastore.


jsangmeister:
also updated lxml according to the security alert

PR #1410, opened 2022-07-18
Add ex-users to export_meeting, add test, update clone.

reiterl:


openslides_backend/action/actions/meeting/export_helper.py
```diff
@@ -55,7 +58,35 @@ def export_meeting(datastore: DatastoreService, meeting_id: int) -> Dict[str, An
         else:
             export[str(collection)] = {}
 
-    add_users(meeting.get("user_ids", []), export, meeting_id, datastore)
+    user_ids = set()
+    for id_ in meeting.get("user_ids", []):
+        user_ids.add(id_)
+    for collection in export:
```
r-peschke, line 64: Instead of all this you could put the following lines after existing line 57, because all data is already read:
     
```
user_ids.update((result.get("user_id") for result in results[collection].values() if result.get("user_id")))
       for result in results[collection].values():
             user_ids.update(result.get("user_ids", ()))

```

openslides_backend/action/actions/meeting/clone.py
```diff
@@ -89,9 +89,10 @@ def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
 
         # needs an empty map for superclass code
         self.merge_user_map = {}
-        for id_ in self.get_meeting_from_json(meeting_json).get("user_ids", []):
+        imported_users_ids = [int(key) for key in meeting_json.get("user", {})]
```
r-peschke, line 92: Couldn't you set self.merge_user_map directly with a dict comprehension like
`self.merge_user_map = { id_ : id_ for key in ...}`
and for the len you could use `len(self.merge_user_map)`

tests/system/action/meeting/test_clone.py
```diff
@@ -167,6 +167,80 @@ def test_clone_with_users(self) -> None:
...
+                    "group_$_ids": ["1"],
+                    "group_$1_ids": [1],
+                    "meeting_ids": [1],
+                },
+                "user/2": {
+                    "username": "exuser1",
+                    "submitted_motion_$_ids": ["1"],
+                    "submitted_motion_$1_ids": [1],
+                },
+                "motion/1": {
```
r-peschke, line 185: Please add 2 more users, only with usernames, and add them via "admin_ids" and "user_ids" in the request-call. Check, if those were exported also.

For the users you added via submitter, motion etc. use different user_ids, especially not the 1, which is added via meeting.user_ids.

r-peschke (changes requested):

openslides_backend/action/actions/meeting/export_helper.py
```diff
@@ -55,7 +58,35 @@ def export_meeting(datastore: DatastoreService, meeting_id: int) -> Dict[str, An
         else:
             export[str(collection)] = {}
 
-    add_users(meeting.get("user_ids", []), export, meeting_id, datastore)
+    user_ids = set()
```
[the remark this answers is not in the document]
reiterl: Done.

openslides_backend/action/actions/meeting/export_helper.py
reiterl: Done.

openslides_backend/action/actions/meeting/clone.py
reiterl: Used, int of key, because the key is a string.

tests/system/action/meeting/test_clone.py
reiterl: admin_ids and user_ids users are *not* exported. Why should they be exported? I already have a test for for admin/user_ids users.

Okay, I use now 11.

reiterl: Added two users to the *ex users* test. (admin and user_ids)

r-peschke: > admin_ids and user_ids users are _not_ exported. Why should they be exported? I already have a test for for admin/user_ids users.
> 
> Okay, I use now 11.

Yes, you are right, the admin_ids/user_ids will not be exported

r-peschke (changes requested):
Please write a test for presenter meeting.export, that users not in meeting.user_ids, but submitter of a motion and, another user, with personal_note in meeting, will be exported.

reiterl:
Added a presenter test for the ex_user.

r-peschke (approved):

PR #1412, opened 2022-07-20
User with manager perms can change to all states.

reiterl:
Update `motion.set_state`.
Use `motion.can_manage_metadata`
Add and update tests.

### DIFF
--- a/tests/system/action/motion/test_follow_recommendation.py
+++ b/tests/system/action/motion/test_follow_recommendation.py
@@ -76,46 +76,6 @@ class MotionFollowRecommendationActionText(BaseActionTestCase):
         assert model.get("state_extension") == "test_test_test"
         assert model.get("last_modified", 0) >= check_time
 
-    def test_follow_recommendation_not_neighbour(self) -> None:
-        self.set_models(
-            {
-                "meeting/222": {
-                    "name": "name_SNLGsvIV",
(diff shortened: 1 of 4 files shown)


r-peschke:
Before approving the requirements should be cleared: which right allows to change the state without evaluation by next/prev-states. And this should be documented in https://github.com/OpenSlides/OpenSlides/wiki/motion.set_state.

tests/system/action/motion/test_follow_recommendation.py
```diff
@@ -76,46 +76,6 @@ def test_follow_recommendation_correct(self) -> None:
         assert model.get("state_extension") == "test_test_test"
         assert model.get("last_modified", 0) >= check_time
 
-    def test_follow_recommendation_not_neighbour(self) -> None:
```
jsangmeister, line 79: Keep this test and fix `motion.follow_recommendation`: There, the check for the next states should still exist.

jsangmeister (changes requested):

jsangmeister:
Updated the wiki.

tests/system/action/motion/test_follow_recommendation.py
reiterl: Readded the test and modify follow_recommendation and set_state, that the test passes.

jsangmeister (approved):

PR #1415, opened 2022-07-22
Add check_database presenter. Add tests.

reiterl:
Use `get_all()` to get all meeting_ids. Use export meeting to get the json and use the checker for the check of the meeting.
Use the checker with the parameters from clone.


openslides_backend/presenter/check_database.py
```diff
@@ -0,0 +1,78 @@
...
+            raise PermissionDenied(msg)
+
+        meeting_ids = self.get_all_meeting_ids()
+        errors: Dict[int, str] = {}
+        for meeting_id in meeting_ids:
+            export = export_meeting(self.datastore, meeting_id)
+            checker = Checker(
+                data=export,
+                mode="internal",
+                repair=True,
```
r-peschke, line 49: IMO should be False. The `all_*`-fields are not checked and the others doesn't matter, if they are on the errors list. I'm not quite sure, but internal should ignore relations to the other meetings.
But we want the missing default values as errors.

r-peschke (changes requested):

openslides_backend/presenter/check_database.py
reiterl: If repair is False, the checker will complain, that a lot of fields are missing, every field must be in the meeting, etc. This is not correct.

reiterl: Set the repair to False, and update the correct test. I added the default fields.

r-peschke (changes requested):
Could you please write one more test, which includes all the relations to collections out of the meeting to be checked: meeting.is_archived_in_organization_id, meeting.template_for_organization_id, meeting.mediafile_ids, meeting.logo__id, meeting.font_id, meeting.committee_id, meeting.default_meeting_for_committee_id, meeting.organization_tag_ids, meeting.present_user_ids, meeting.user_ids
motion_submitter.user_id,
vote.user_id, vote.delegated_user_id,
assignment_candidate.user_id

I hope that are all fields. For test to external relations use always a new instance, i.e for each reference to a user, use a new user. If possible test always with relation inside and outside the meeting.

reiterl:
Writing the test for the external relation costs some of my time and the test is bloated.

reiterl:
I tried to wrap up stuff in the tests.

r-peschke:
> Writing the test for the external relation costs some of my time and the test is bloated.
- I forget 2 relations to check: motion origin_id and motion.derived_motion_ids
- Remove the user_ids from group/1 and from meeting. Some of the new errors are strange
- The algorithm of  export_meeting in export_helper.py for adding users is not sufficient. It should find more users.

reiterl:
Okay, I removed `user_ids` from meeting and group/1. I found errors, that the relations are broken. Looks good to me.

reiterl:
Ah, not enough users are exported by the `export_meeting` function. I don't think, we should handle this here.
I opened: https://github.com/OpenSlides/openslides-backend/issues/1418 for this.

reiterl:
The `motion.origin_id` and  the `motion.derived_motion_ids`are also problematic. As far as I understand the check_database
code, the problem is, that the motions are in two different meetings and the check uses the export_meeting function to get just
one meeting and so one motion.

reiterl:
A possible solution for the `motion.origin_id` is to reintroduce the `repair` mode and remove this fields.

reiterl:
We go the solution to assert the error message in the `origin_id` test.

reiterl:
We use the repair True solution.

r-peschke (approved):

PR #1420, opened 2022-08-03
Hard coded some fields to get the user ids in export_meeting.

reiterl:


jsangmeister (changes requested):
Can you use a generic solution for all fields? If we leave it the current way, we will have to remember and always add new fields here which might refer to users. I imagine something like iterating over all fields of the collection (with `model_registry[collection].get_relation_fields()`) and then checking if the relation points to a user and then adding it if needed. You will probably need special handling for `meeting.user_ids` since it's calculated, but that's only one special case.

reiterl:
I updated the code to a more generic solution. It checks meeting.user_ids, RelationFields, RelationListFields and GenericRelationFields.

jsangmeister (approved):

jsangmeister:
Maybe wait for @r-peschke's opinion since this is not urgent.

openslides_backend/action/actions/meeting/export_helper.py
```diff
@@ -62,30 +66,47 @@ def export_meeting(datastore: DatastoreService, meeting_id: int) -> Dict[str, An
         else:
             export[str(collection)] = {}
 
+    user_ids = set()
```
r-peschke, line 69: user_ids are initialized in line 44 (in a better way)

r-peschke (changes requested):
Just one small change, after change you could merge:-)

openslides_backend/action/actions/meeting/export_helper.py
reiterl: Remove this surplus user_ids initialize lines. If the tests still work, I will merge it.

PR #1421, opened 2022-08-04
Update check_database to check the whole database with the checker

reiterl:


openslides_backend/presenter/check_database.py
```diff
@@ -39,40 +78,35 @@ def get_result(self) -> Any:
...
+        )
+        errors: str = ""
+        try:
+            checker.run_check()
+        except CheckException as ce:
+            errors = str(ce)
         if not errors:
             return {"ok": True, "errors": ""}
-        return {"ok": False, "errors": self.gen_error_message(errors)}
+        return {"ok": False, "errors": errors}
```
jsangmeister, line 93: you can combine this:
```
try:
    checker.run_check()
    return {"ok": True}
except CheckException as ce:
    return {"ok": False, "errors": str(ce)}
```

openslides_backend/presenter/check_database.py
```diff
@@ -39,40 +78,35 @@ def get_result(self) -> Any:
...
+
+    def get_all_database(self) -> Dict[str, Any]:
+        export: Dict[str, Any] = {"_migration_index": get_backend_migration_index()}
+        for collection in ALL_COLLECTIONS:
+            export[collection] = remove_meta_fields(
+                self.datastore.get_all(collection, [], lock_result=False)
+            )
+        return export
```
jsangmeister, line 101: You can use the `get_everything` method of the datastore to fetch all data. It is currently only available in dev mode, so we can either make the presenter also only available in dev mode or remove this restriction, but we can think about this later, maybe with @r-peschke. Also, the backend adapter of the DS does not know this route yet, since we did not need it yet. You will have to implement it there, but that should be fairly straight forward.

jsangmeister (changes requested):

openslides_backend/presenter/check_database.py
reiterl: Okay, done. I have also updated the presenter wiki.

openslides_backend/presenter/check_database.py
reiterl: Okay, perhaps improve this later.

jsangmeister (changes requested):
using `get_everything` is still open

r-peschke:
IMO reading the whole database into memory will be problematical, because we will run into memory problems, depending the organization.
I talked with Raimund and we want to keep the existing check of all meetings and create a second presenter checking the whole database. There we should use the get_everything-method. On clientside see https://github.com/OpenSlides/openslides-client/issues/1566 for having a start button.

r-peschke (changes requested):

reiterl:
Introduce two check_databases. Updated the second one to use get_everything().

openslides_backend/presenter/check_database.py
reiterl: I use get_everything from the datastore in check_datastore_all

openslides_backend/presenter/check_database.py
```diff
@@ -22,6 +23,44 @@
     }
 )
 
+ALL_COLLECTIONS = [
```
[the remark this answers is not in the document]
reiterl: Updated, use now get_everything().

r-peschke (approved):

reiterl:
@jsangmeister  I use now get_everything, pls don't forget to make the get_everything route usable in production setup.

openslides_backend/presenter/check_database_all.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+def remove_meta_fields(res: Dict[int, Any]) -> Dict[str, Any]:
+    dict_without_meta_fields = {}
+    for key in res:
+        new_entry = {}
+        for fieldname in res[key]:
+            if not is_reserved_field(fieldname):
+                new_entry[fieldname] = res[key][fieldname]
+        dict_without_meta_fields[str(key)] = new_entry
```
jsangmeister, line 71: You can use `datastore.shared.util.strip_reserved_fields` here

jsangmeister (changes requested):

openslides_backend/presenter/check_database_all.py
reiterl: Okay, use strip_reserved_fields.

openslides_backend/presenter/check_database_all.py
```diff
@@ -64,9 +64,6 @@ def get_everything(self) -> Dict[str, Any]:
 def remove_meta_fields(res: Dict[int, Any]) -> Dict[str, Any]:
```
jsangmeister, line 64: Please move inside the class

openslides_backend/presenter/check_database_all.py
```diff
@@ -64,9 +64,6 @@ def get_everything(self) -> Dict[str, Any]:
...
     dict_without_meta_fields = {}
     for key in res:
-        new_entry = {}
-        for fieldname in res[key]:
-            if not is_reserved_field(fieldname):
-                new_entry[fieldname] = res[key][fieldname]
-        dict_without_meta_fields[str(key)] = new_entry
+        strip_reserved_fields(res[key])
+        dict_without_meta_fields[str(key)] = res[key]
```
jsangmeister, line 68: Why do you copy into a new dict? Since this is a presenter, there should be no mutations on the dict and you can just change it in-place.

jsangmeister (changes requested):

openslides_backend/presenter/check_database_all.py
reiterl: Done.

openslides_backend/presenter/check_database_all.py
reiterl: I copy it, because of the int to str key transformation.

jsangmeister (approved):

PR #1430, opened 2022-08-16
Add l10n to the backend. Update forget_password.

reiterl:
Add test, add make targets.


openslides_backend/action/action.py
```diff
@@ -673,6 +675,19 @@ def get_on_failure(self, action_data: ActionData) -> Callable[[], None]:
...
+            return msg
+
+        if not hasattr(self, "language") or not self.language:
+            return null_translate
+        t = gettext.translation(
+            "messages", "/app/openslides_backend/locale", [self.language]
+        )
+        if not t:
+            return null_translate
+        return t.gettext
```
jsangmeister, line 689: Please extract this into a singleton/static class where the current language is always saved and the translate function can be accessed. I don't like polluting the action class any more

openslides_backend/action/action_handler.py
```diff
@@ -223,6 +223,10 @@ def perform_action(
         action = ActionClass(
             self.services, self.datastore, relation_manager, self.logging, self.env
         )
+        if hasattr(self, "language"):
+            action.language = self.language
+        else:
+            action.language = "en_US"
```
jsangmeister, line 229: this can be removed then

openslides_backend/http/views/action_view.py
```diff
@@ -39,6 +39,7 @@ def action_route(self, request: Request) -> Tuple[ResponseBody, Optional[str]]:
 
         # Handle request.
         handler = ActionHandler(self.env, self.services, self.logging)
+        handler.set_translation_language(request.headers.get("Accept-Language"))
```
jsangmeister, line 42: and here you can set the language directly on the translate service

openslides_backend/action/actions/user/forget_password.py
```diff
@@ -42,6 +34,16 @@ class UserForgetPassword(EmailMixin, UpdateAction):
...
+        _ = self.get_translate_function()
+        self.PW_FORGET_EMAIL_TEMPLATE = _(
+            """You are receiving this email because you have requested a new password for your OpenSlides-account.
+
+Please open the following link and choose a new password:
+{url}/login/forget-password-confirm?user_id={user_id}&token={token}
+
+For completeness your username: {username}"""
+        )
+        self.PW_FORGET_EMAIL_SUBJECT = _("Reset your OpenSlides password")
```
jsangmeister, line 46: I think you can also use `gettext` as just a marker function and do the actual translating later. Can you try this out, so you don't need to have hardcoded strings inside the class?

tests/system/action/base.py
```diff
@@ -59,6 +59,14 @@ def request_json(self, payload: Payload, anonymous: bool = False) -> Response:
         client = self.client if not anonymous else self.anon_client
         return client.post(ACTION_URL, json=payload)
 
+    def request_lang(
+        self, action: str, data: Dict[str, Any], lang: str = "en_US"
+    ) -> Response:
+        payload = [{"action": action, "data": [data]}]
+        headers = {"Accept_Language": lang}
+
+        return self.client.post(ACTION_URL, json=payload, headers=headers)
```
jsangmeister, line 68: I would rather integrate the `lang` parameter into the other request function to have it available everywhere and not have to differentiate between translate request and other ones.

openslides_backend/action/action.py
```diff
@@ -673,6 +675,19 @@ def get_on_failure(self, action_data: ActionData) -> Callable[[], None]:
...
         """
 
+    def get_translate_function(self) -> Callable[[str], str]:
+        def null_translate(msg: str) -> str:
+            return msg
+
+        if not hasattr(self, "language") or not self.language:
+            return null_translate
+        t = gettext.translation(
+            "messages", "/app/openslides_backend/locale", [self.language]
```
jsangmeister, line 685: Also, these should be constants

jsangmeister (changes requested):

openslides_backend/action/action.py
reiterl: We need a translation function per request. So I need some kind of factory, not a singleton.

jsangmeister: Yes, but a factory instance will have to be passed through, which is not what we want. Since we only handle one request at a time, a singleton should suffice.

reiterl: Okay, 2 requests should not be a problem, introduce a singleton Translator.

openslides_backend/action/action_handler.py
reiterl: Removed.

openslides_backend/http/views/action_view.py
reiterl: Use the translator Singleton here.

openslides_backend/action/action.py
reiterl: Add consts for magic strings.

openslides_backend/action/actions/user/forget_password.py
reiterl: The way I use gettext, it doesn't work with translation later.

tests/system/action/base.py
reiterl: Done.

openslides_backend/locale/translator.py
```diff
@@ -0,0 +1,30 @@
...
+        return t.gettext
+
+    def set_translation_language(self, lang: Optional[str]) -> None:
+        if lang is None:
+            self.language = DEFAULT_LANGUAGE
+        else:
+            self.language = lang
+
+
+Translator = TranslatorSingleton()
```
jsangmeister, line 30: I would like a function `translate` which accepts a string to be translated, fetches the translate function from the instance and returns the translated string. Then, you can import that via `... import translate as _` in other files and do not have to call the translator every time.

tests/system/action/base.py
```diff
@@ -48,24 +56,22 @@ def request_multi(
...
+        self, payload: Payload, anonymous: bool = False, lang: Optional[str] = None
     ) -> Response:
-        payload = [{"action": action, "data": [data]}]
-        headers = {"Accept_Language": lang}
-
-        return self.client.post(ACTION_URL, json=payload, headers=headers)
+        client = self.client if not anonymous else self.anon_client
+        headers = {}
+        if lang:
+            headers["Accept_Language"] = lang
```
jsangmeister, line 73:
```suggestion
            headers["Accept-Language"] = lang
```
I'm wondering why that even worked...

jsangmeister (changes requested):

tests/system/action/base.py
reiterl: Changed.

openslides_backend/locale/translator.py
reiterl: I don't know how to do that.

reiterl: _(): Seems to have a special meaning. I can do from ....locale.translator import Translator and self.PW... = Translator(..) but not import Translator as _ and _(...).

jsangmeister:
I pushed my suggestion on how to use the translate function as `_`, you can take a look and merge it if it looks good. The error you were getting was probably related to `_` being used as a local variable later, which is of course not possible when using this translation approach.

reiterl:
I looked at the changes of @jsangmeister . They look good. It is in a way, which can be used.

PR #1432, opened 2022-08-17
Small fix to the committee ml handling.

reiterl:
Add a check to make the management level handling more robust.

### DIFF
--- a/openslides_backend/presenter/get_user_related_models.py
+++ b/openslides_backend/presenter/get_user_related_models.py
@@ -14,7 +14,7 @@ from ..permissions.permission_helper import (
 )
 from ..permissions.permissions import Permissions
 from ..services.datastore.commands import GetManyRequest
-from ..shared.exceptions import MissingPermission
+from ..shared.exceptions import MissingPermission, PresenterException
 from ..shared.filters import And, FilterOperator
 from ..shared.patterns import fqid_from_collection_and_id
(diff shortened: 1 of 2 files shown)


openslides_backend/presenter/get_user_related_models.py
```diff
@@ -117,7 +117,12 @@ def get_committees_data(self, user_id: int) -> List[Dict[str, Any]]:
         }
         for level in user.get("committee_$_management_level", []):
             for committee_nr in user.get(f"committee_${level}_management_level", []):
-                committees[committee_nr]["cml"].append(level)
+                if committee_nr in committees:
+                    committees[committee_nr]["cml"].append(level)
+                else:
+                    raise PresenterException(
+                        f"Data error: committee/{committee_nr} is not user committee."
```
r-peschke, line 124: Make comment more meaningful: f"Data error: user has rights for committee {committee_nr}, but faultily is no member of committee." 
Don't want to write "database corrupt", but the fault should be clear. perhaps you have a better text

r-peschke (changes requested):

openslides_backend/presenter/get_user_related_models.py
reiterl: I don't have a better text. So I used your text.

r-peschke (approved):

PR #1434, opened 2022-08-19
Fix to KeyError of the mediafile.upload. Add test.

reiterl:


luisa-beerboom:
Don't know if this is a leftover, but I'm getting the following error:

`Database corrupt: The resource token has to be unique, but there are 2 tokens \"None\".`

jsangmeister (changes requested):
This is not the intended behaviour:
- If no token is given, the whole check in the `get_updated_instances` method can be skipped, because it must not be checked if the token is a duplicate.
- `None` is not a valid token, which should theoretically be caught by the schema, but if not, you will have to do this manually

With these changes, the errors from @luisa-beerboom should disappear.

reiterl:
Add a test for token is None and skip the token check code, if Token is not set.

jsangmeister (approved):
looks good, not tested. @luisa-beerboom can you check if this fixes your problem?

luisa-beerboom (approved):
This fixes my problem, feel free to merge

Issue #1441, opened 2022-08-24
Forwarding fails in the source committee, if there is already an account with the exact username of the source committee


alveolus:
### Description of the problem

- After a motion is forwarded, a message is displayed stating that the motion has been successfully forwarded to the target conference. But this motion does not appear in the target conference.
- A failed forwarding can be recognized by the fact that in a motion below the field "Creation date" the field "Motion forwarded to" is not visible.

### How to reproduce & Example

→ testintern → /86/motions/1

"Source Conference A" = Committee title = Event title
"Target Conference B" = Committee title = Event title

Motion from "Source Conference A" → "Target Conference B"

When forwarding a motion, an account is automatically created with the username of the committee from whose event the motion is to be forwarded. In my example, this is "Source Conference A". The motion will then (normally) end up at "Target Conference B". The submitter:in is then "Source Conference A".

However, if someone in the "Source Conference A" have already manually created an account with the exact name "Source conference A", so that this account is listed in the motion as the submitter, then the forwarding will fail. Because: By creating this account manually, there is already an account with exactly this name ("Source Conference A"). Therefore, no account can be created during the forwarding process. 

### Weiteres ToDo

- [ ] Revise notification message while/after forwarding. Only display "Motion successfully forwarded" if this is actually the case. Otherwise, display an error message with the reason for the failed forwarding.

r-peschke:
This doesn't work, because the backend looks for a registered user to replace the motion-applicant in the forwarded motion. If there is no user registered, it tries to create with **the name of the committee**. This fails, if the username is used, because it has to be unique. There is a bug in the client, that the error message is not shown, but instead the success message => see https://github.com/OpenSlides/openslides-client/issues/1530.

We should search for a solution and take account of:
1. the created user is a system user, set inactive and no physical user, but with the default group of the selected meeting. Forwarding to other meetings will automatically add their default group to the user.
2. Not all of the openslides customers will want the real motion-applicant automatically replaced by a committee_named user
3. To set the motion-applicant user manually is a lot of work and the name of the real petitioner at the original meeting gets lost

IMO from 1. results we shouldn't capture a user with an accidental same username as the committee without being registered as forwarding user of this committee. This will be a user with growing rights for all meetings forwardings will be made.
If the customer is not sure about the behavior of openslides and creates a special user for this use case, creating another user with a generated unique username and replace the wanted user is also a bad solution.

[...]

PR #1442, opened 2022-08-25
Issue1441 Ceate committe-user anyway

r-peschke:
Changed error message for giving the context of motion forwardings system user creation

### DIFF
--- a/openslides_backend/action/actions/motion/create_forwarded.py
+++ b/openslides_backend/action/actions/motion/create_forwarded.py
@@ -1,4 +1,5 @@
 import time
+from copy import deepcopy
 from typing import Any, Dict
 
 from ....models.models import Motion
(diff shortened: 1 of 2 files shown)


jsangmeister (changes requested):
This is unnecessary if we implement the solution suggested in https://github.com/OpenSlides/openslides-backend/issues/1441

openslides_backend/action/actions/motion/create_forwarded.py
```diff
@@ -65,18 +66,35 @@ def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
                 ]
                 self.execute_other_action(UserUpdate, user_update_payload)
         else:
+            username = committee.get("name", "Committee User")
             committee_user_create_payload = {
-                "username": committee.get("name", "Committee User"),
+                "username": username,
```
jsangmeister, line 71: You can just remove this line and the whole while-loop below, because without a username given, `user.create` will automatically generate one by appending a number.

jsangmeister (changes requested):

jsangmeister (approved):

PR #1455, opened 2022-09-07
Assert that assignment poll history works

jsangmeister:
There are already a few tests which test this functionality (that a history entry is created for the content object, in this case the assignment, when the poll is changed). This PR just verified that and adds another assertion.

Client side is still missing, I will create an issue for this.

### DIFF
--- a/openslides_backend/services/datastore/adapter.py
+++ b/openslides_backend/services/datastore/adapter.py
@@ -338,7 +338,7 @@ class DatastoreAdapter(BaseDatastoreService):
         self.logger.debug(
             f"Start HISTORY_INFORMATION request to datastore with the following data: {request}"
         )
-        return self.reader.history_information(request)
+        return dict(self.reader.history_information(request))
 
     def lock_collection_fields_from_filter(
         self,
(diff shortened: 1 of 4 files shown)


r-peschke:
The method BaseActionTestCase.assert_history_information itself is wrong in line 386. There must be a line like `assert informations[-1]["information"] is None`. 
But with this change the test motion/test_create_forwarded:test_all_origin_ids_complex fails and I think it must fail ,because both motions tested are changed with the same position (db-Transaction) and therefore with the same `information`.

r-peschke (changes requested):
The method BaseActionTestCase.assert_history_information itself is wrong in line 386. There must be a line like assert informations[-1]["information"] is None.
But with this change the test motion/test_create_forwarded:test_all_origin_ids_complex fails and I think it must fail ,because both motions tested are changed with the same position (db-Transaction) and therefore with the same information.

jsangmeister:
you are correct, I changed the test accordingly, since I don't see an easy way to implement this feature without completely changing the history structure

r-peschke (approved):

PR #1456, opened 2022-09-07
Add organization.user_ids and user.organization_id relation

reiterl:
Update the tests, and add test parts.
Add a migration.

### DIFF
--- a/global/meta/models.yml
+++ b/global/meta/models.yml
@@ -147,6 +147,10 @@ organization:
     to: mediafile/owner_id
     on_delete: CASCADE
     restriction_mode: A
+  user_ids:
+    type: relation-list
+    restriction_mode: A
+    to: user/organization_id
   users_email_sender:
     type: string
     default: OpenSlides
@@ -433,6 +437,11 @@ user:
     description: Calculated. All ids from group_$_ids as integers.
     read_only: true
     restriction_mode: E
+  organization_id:
+    type: relation
+    to: organization/user_ids
+    required: True
+    restriction_mode: F
 
 organization_tag:
   id:
--- a/global/meta/models.yml
+++ b/global/meta/models.yml
@@ -149,7 +149,7 @@ organization:
     restriction_mode: A
   user_ids:
     type: relation-list
-    restriction_mode: A
+    restriction_mode: C
     to: user/organization_id
   users_email_sender:
     type: string
--- a/openslides_backend/action/actions/meeting/import_.py
+++ b/openslides_backend/action/actions/meeting/import_.py
@@ -638,13 +638,11 @@ class MeetingImport(SingularActionMixin, LimitOfUserMixin, UsernameMixin):
                 new_user_ids.append(user_entry["id"])
 
         if new_user_ids:
-            write_requests.append(
-                self.build_write_request(
+            events.append(
+                self.build_event(
                     EventType.Update,
                     ONE_ORGANIZATION_FQID,
-                    f"import meeting {meeting_id}",
-                    None,
-                    {
+                    list_fields={
                         "add": {
                             "user_ids": new_user_ids,
                         },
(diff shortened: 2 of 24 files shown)


openslides_backend/action/actions/meeting/import_.py
```diff
@@ -549,9 +549,7 @@ def create_events(
                                 list_fields["add"][field] = value
                             else:
                                 fields[field] = value
-                        elif isinstance(
-                            model_field, (RelationListField, RelationField)
-                        ):
+                        elif isinstance(model_field, RelationListField):
```
r-peschke, line 552: What's about RelationField?

openslides_backend/action/actions/meeting/import_.py
```diff
@@ -631,6 +629,26 @@ def append_extra_events(
                 )
             )
 
+        # add new users to the organization.user_ids
+        new_user_ids = []
+        for user_entry in json_data.get("user", {}).values():
```
r-peschke, line 634: Are you sure, that the user_ids from json_data are the new ones or still the old ones from export?

r-peschke (commented):
Some points to check and we have to resolve with @ostcar the restriction mode F problem

openslides_backend/action/actions/meeting/import_.py
reiterl: Does not work and is not needed here and lead to errors.

openslides_backend/action/actions/meeting/import_.py
reiterl: I except the json_data["user"] to be merged ones and new ones. And then use only the new ones.

global/meta/models.yml
```diff
@@ -433,6 +437,11 @@ user:
     description: Calculated. All ids from group_$_ids as integers.
     read_only: true
     restriction_mode: E
+  organization_id:
+    type: relation
+    to: organization/user_ids
+    required: True
+    restriction_mode: F
```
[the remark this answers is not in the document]
ostcar: In this case F is correct. This means that you need the OML can_manage_users.

global/meta/models.yml
```diff
@@ -147,6 +147,10 @@ organization:
     to: mediafile/owner_id
     on_delete: CASCADE
     restriction_mode: A
+  user_ids:
+    type: relation-list
+    restriction_mode: A
```
ostcar, line 152: This is the place, where for performance reasons I would like you to use another restriction_mode.

openslides_backend/action/actions/meeting/import_.py
reiterl: Perhaps here is an error. I will look at it again.

openslides_backend/action/actions/meeting/import_.py
reiterl: The handling of the RelationField should still be in place.

openslides_backend/action/actions/meeting/import_.py
reiterl: Ah, there was a bug, I changed the code.

tests/system/action/meeting/test_import.py
```diff
@@ -1742,7 +1765,7 @@ def test_all_migrations(self) -> None:
         with CountDatastoreCalls(verbose=True) as counter:
             response = self.request("meeting.import", data)
         self.assert_status_code(response, 200)
-        assert counter.calls == 6
+        assert counter.calls == 7
```
r-peschke, line 1768: This is the test to test all migrations. Please remove the organization_id in instance data (`del data["meeting"]["user"]["1"]["organization_id"]` and insert a test that it was set by migration (`self.assert_model_exists("organization/1", {"user_ids": [1, 2], "active_meeting_ids": [1, 2]})`)

r-peschke (changes requested):
Still want to check whether the search works with the autoupdate restriction modes and https://github.com/OpenSlides/openslides-backend/issues/1459

ostcar (changes requested):
See my comment from 5 days ago. The restriction_mode of the field organization/user_ids should be changed

global/meta/models.yml
reiterl: I chatted with @r-peschke about this topic. I will change the restriction_mode of organization/user_ids to C.

tests/system/action/meeting/test_import.py
reiterl: Done.

global/meta/models.yml
reiterl: Done.

r-peschke (approved):
Please write a test for the change in user/create.py

PR #1458, opened 2022-09-09
Issue1429 meeting.import duplicate users

r-peschke:


emanuelschuetze (approved):

tests/system/action/user/test_create.py
```diff
@@ -12,20 +12,20 @@ def permission_setup(self) -> None:
...
             {
-                "username": "test_Xcdfgee",
+                "username": " test Xcdfgee ",
                 "organization_management_level": OrganizationManagementLevel.CAN_MANAGE_USERS,
             },
         )
         self.assert_status_code(response, 200)
         model = self.get_model("user/2")
-        assert model.get("username") == "test_Xcdfgee"
+        assert model.get("username") == "test Xcdfgee"
```
jsangmeister, line 28: This is wrong, the first name and last name should be joined WITHOUT a space between them to avoid spaces in usernames all together. Please revert the change.

r-peschke: No, this correct. If we build the username from first- and last-name, we remove all spaces, but if the username is explicitly given in create and import, we want to preserve the spaces in the username, only stripping leading and trailing spaces.

jsangmeister: You are correct, my bad! Did not notice that the username was given explicitely.

PR #1460, opened 2022-09-10
models.yml for vote-decrypt

ostcar:
This is only an example. Especially the names could be different.

### DIFF
--- a/global/meta/models.yml
+++ b/global/meta/models.yml
@@ -2712,7 +2712,7 @@ poll:
       - analog
       - named
       - pseudoanonymous
-      - crypt
+      - cryptographic
     restriction_mode: A
   backend:
     type: string
@@ -2807,14 +2807,14 @@ poll:
     read_only: true
     required: true
     restriction_mode: A
-  crypt_key:
+  cryptographic_key:
     type: string
-    description: base64 public key to encrypt votes.
+    description: base64 public key to cryptographic votes.
     read_only: true
     restriction_mode: A
-  crypt_key_signature:
+  cryptographic_key_signature:
     type: string
-    description: base64 signature of crypt_key.
+    description: base64 signature of cryptographic_key.
     read_only: true
     restriction_mode: A
   votes_raw:
--- a/global/meta/models.yml
+++ b/global/meta/models.yml
@@ -2712,6 +2712,7 @@ poll:
       - analog
       - named
       - pseudoanonymous
+      - crypt
     restriction_mode: A
   backend:
     type: string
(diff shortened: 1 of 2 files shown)


global/meta/models.yml
```diff
@@ -2700,6 +2700,7 @@ poll:
       - analog
       - named
       - pseudoanonymous
+      - crypt
```
jsangmeister, line 2703: I think I would favor "anonymous". If you disagree, I would at least write the full word ("cryptographic") to stay in lane with the other names.

jsangmeister (changes requested):

global/meta/models.yml
ostcar: I have no preference. 

anonymous sounds like the poll is for anonymous users.

The English term for "geheime Wahl" is "secret ballot". So another option would be "secret".

jsangmeister: That is right, also we already have a similar "problem" with "pseudoanonymous". But "cryptographic" but be better suited under these circumstances.

ostcar: done

global/meta/models.yml
```diff
@@ -2806,6 +2807,26 @@ poll:
     read_only: true
     required: true
     restriction_mode: A
+  cryptographic_key:
+    type: string
```
r-peschke, line 2811: I suppose this and also following names are not visible for the customer, then I would prefer shorter names: `crypt_key`

global/meta/models.yml
```diff
@@ -2806,6 +2807,26 @@ poll:
...
+    description: base64 public key to cryptographic votes.
+    read_only: true
+    restriction_mode: A
+  cryptographic_key_signature:
```
r-peschke, line 2815: `crypt_signature`, analogue to votes_raw and votes_signature

r-peschke (changes requested):
My proposal to use shorter names, if you both agree

global/meta/models.yml
ostcar: The keys are not visible. My first proposal was also "crypt_key". I changed this to have it consistent with the value of `poll/type`. @jsangmeister asked to change the value of `poll/type`. See the outdated discussion above.

global/meta/models.yml
ostcar: I don't see the analogy to `votes_raw` and `votes_signature`. What are your proposals for the names of these keys?

ostcar:
> My proposal to use shorter names, if you both agree

I have no preference. @r-peschke and @jsangmeister please discuss this among each other and change the PR afterwards.

global/meta/models.yml
r-peschke: votes_raw and votes_signature, but crypt_key and crypt_key_signature. In one case you strip the "usage" (raw), in the other case you leave the "usage" (key).

r-peschke (approved):

r-peschke:
@jsangmeister Fell free to change the names, if it is important for you

global/meta/models.yml
jsangmeister: I have no strong opinion on the name of these fields, so we can leave it as it currently is

PR #1464, opened 2022-09-13
Add migration to remove permission 'user.can_see_extra_data'.

reiterl:


r-peschke (approved):
user.can_see or user.can_manage is present in all groups we create (example_data and meeting.create), so we don't have to replace user.can_see_extra_data by user.can_see

openslides_backend/migrations/migrations/0029_remove_permission_user_can_see_extra_data.py
```diff
@@ -0,0 +1,44 @@
...
+        collection = collection_from_fqid(event.fqid)
+        if collection != "group":
+            return None
+
+        if isinstance(event, (CreateEvent, UpdateEvent)):
+            if self.group_permission in event.data.get("permissions", []):
+                event.data["permissions"].remove(self.group_permission)
+                return [event]
+        elif isinstance(event, ListUpdateEvent):
+            mark = False
```
jsangmeister, line 35: you can remove this `mark` handling and just always return `[event]`, since this is just what's used internally in case `None` is returned.

jsangmeister (changes requested):

jsangmeister (approved):

Issue #1461, opened 2022-09-10
Interface of vote-service has changed


ostcar:
To support cryptographic votes in the future, the interface between the backend as the vote-service has to change.

Here is the PR in the vote-service: https://github.com/OpenSlides/openslides-vote-service/pull/141

For this to work, there need to be some changes in the models.yml. Here is an suggestion: https://github.com/OpenSlides/openslides-backend/pull/1460

The required changes in the backend are:

On start request the returned key and key_signature have to be saved to the datastore.

On stop the returned votes and the votes_signature have to be saved to the datastore.

All votes have to be validated. Invalid votes have to be ignored and the number of invalid votes have to be written in the field `votesinvalid`.

r-peschke:
State 14.09.22: Changes will be reverted in vote-service (back to commit 31fbd957) and moved to a PR
We could continue this issue with the PR.

PR #1465, opened 2022-09-14
Use vote-service from local, not from github#main anymore

r-peschke:
Needed to work on #1461 with vote-services PR


r-peschke:
Not needed anymore

PR #1472, opened 2022-09-15
Create relation between all_origin_ids and all_derived_motion_ids for motions

jsangmeister:
Without the relation or any other handling, deleted motions were left in the field which led to errors when trying to forward a motion which had deleted motions in its `all_origin_ids` field. With this change, motions will be deleted from these fields as well. This leads to the need of a new field `origin_meeting_id` to save the information that the motion was at all forwarded.

Summary of the changes:
- Introduce relation between `all_origin_ids` and `all_derived_motion_ids` so deletion of items gets handled automatically.
- Introduce additional field `motion/origin_meeting_id` which contains the meeting id of the origin motion, so if the origin motion gets deleted, we can still display the meeting information
- Add migration which fixes existing broken relations and retroactively adds the new `origin_meeting_id`


emanuelschuetze (approved):
Tested in local setup

r-peschke (approved):

PR #1477, opened 2022-09-19
Add import tests with present users

jsangmeister:
@rrenkert I tried to indentify/reproduce the "is_present_in_meeting_ids" error, but was not able to. I could, however, localize the error source: `meeting.import`. Since the meeting in question was imported a few days before the user merge feature was merged (in https://github.com/OpenSlides/openslides-backend/commit/ce6dfdb0eb12717b10c0a074ff10dae1a481a0b9), I can only assume that the error no longer exists. Broken instances can be fixed with a script written for that purpose.

@r-peschke These are just additional tests I created while trying to reproduce the problem, I thought it was better to keep them.

### DIFF
--- a/tests/system/action/meeting/test_import.py
+++ b/tests/system/action/meeting/test_import.py
@@ -1,6 +1,6 @@
 import base64
 import time
-from typing import Any, Dict, List, cast
+from typing import Any, Dict, List, Optional, cast
 
 from openslides_backend.migrations import get_backend_migration_index
 from openslides_backend.models.models import Meeting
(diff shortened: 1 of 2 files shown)


tests/system/action/meeting/test_import.py
```diff
@@ -40,7 +40,7 @@ def setUp(self) -> None:
             }
         )
 
-    def create_request_data(self, datapart: Dict[str, Any]) -> Dict[str, Any]:
+    def create_request_data(self, datapart: Dict[str, Any] = {}) -> Dict[str, Any]:
```
reiterl, line 43: Don't use a empty list or empty dict as a default, it is a can of worms.
If you add sth to that dict, it will continue to be in the dict, the next time you call it.
So it is *bad* style.

reiterl (commented):
Small comment.

tests/system/action/meeting/test_import.py
jsangmeister: done

r-peschke (approved):

PR #1480, opened 2022-09-22
Fix broken migrations

jsangmeister:


peb-adr:
@jsangmeister What you did for migration 29 seems sensible. I also tested it for the DB where the problem occured and migration went through.
However I reverted the (small) changes for migration 32, since @r-peschke is working on our actual fix for this.

Can you guys approve this PR, so we can merge and build an image for migrating the production DB.

@rrenkert FYI as well

r-peschke (changes requested):
Could you please remove the 0032_test-file. I will move it to my PR for the 0032-migration

r-peschke (approved):

Issue #1474, opened 2022-09-16
Protect against deleting one self


ostcar:
As a superadmin I accidentally deleted the myself.

I think we had a protection against this in OS 3. Should we add one in OS4?

If it should be allowed to delete one self, you should at least get logged out, because the current behavior  is strange

r-peschke:
@emanuelschuetze @ostcar @rrenkert @reiterl 
We should discuss first, what we want.
IMO we should prevent a superadmin,to delete himself, eventually only if he is the only superadmin.

PR #1481, opened 2022-09-22
Prevent to delete yourself.

reiterl:
Update user.delete action. Add a test.


r-peschke (commented):
See my comment in #1474 
If we change something, we have to change wiki, too.

jsangmeister (approved):
I'm fine with this, since I don't see a point in deleting oneself. But we can discuss this first if @r-peschke wants to

r-peschke (approved):

reiterl:
I added "Prevent to delete oneself." to the action in the wiki.

PR #1483, opened 2022-09-23
Add meeting/users_enable_vote_delegations

reiterl:
Add a migration.


openslides_backend/action/actions/meeting/update.py
```diff
@@ -114,6 +114,7 @@
     "motion_poll_default_backend",
     "users_enable_presence_view",
     "users_enable_vote_weight",
+    "users_enable_vote_delegations",
```
r-peschke, line 117: Please add the new field to the wiki on action meeting.update

r-peschke (changes requested):
Please add the new field to the wiki description for meeting.update, see above

Check this setting in the backend: No forwarding allowed in setting and voting, if the flag is set to false. 
Document the impact of the setting in the various actions.
Otherwise it should be discussed/decided, how to handle this setting.

emanuelschuetze (changes requested):
Why `users_enable_vote_delegations` is true by default? 
Please set it to false (to prevent migration and to use the same default as `users_enable_vote_weight`). Vote delegation is an optional feature which should enabled by admins who need this. All others should use the simple default.

reiterl:
Okay changed the default of users_enable_vote_delegations to false and remove the migration code (and the test.
Update the data,

openslides_backend/action/actions/meeting/update.py
reiterl: Add the field to the action wiki

jsangmeister (approved):

r-peschke (approved):

PR #1484, opened 2022-09-23
migration 0032 with changing updateEvents

r-peschke:
fixed

### DIFF
--- a/openslides_backend/migrations/migrations/0032_fix_motion_origin_relations.py
+++ b/openslides_backend/migrations/migrations/0032_fix_motion_origin_relations.py
@@ -1,4 +1,5 @@
-from typing import Any, Dict, List, Optional
+from collections import defaultdict
+from typing import Any, Dict, List, Optional, Set
 
 from datastore.migrations.core.base_migration import BaseMigration
 from datastore.migrations.core.events import (
@@ -6,6 +7,7 @@ from datastore.migrations.core.events import (
     CreateEvent,
     DeleteEvent,
     ListUpdateEvent,
+    UpdateEvent,
 )
 from datastore.shared.util import (
     collection_and_id_from_fqid,
@@ -15,6 +17,9 @@ from datastore.shared.util import (
 
 class Migration(BaseMigration):
     target_migration_index = 33
+
+    def position_init(self) -> None:
+        self.created_motion_ids: Dict[int, Set[int]] = defaultdict(set)
 
     def migrate_event(
         self,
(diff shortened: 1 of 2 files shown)


r-peschke:
@rrenkert @peb-adr  FYI

openslides_backend/migrations/migrations/0032_fix_motion_origin_relations.py
```diff
@@ -63,4 +70,26 @@ def migrate_event(
...
                     )
             return new_events
+        elif isinstance(event, UpdateEvent):
+            """Changes the update event
+            Precondition from backend: The backend sorts the events by type: create/update/delete
+            Therefore we don't need the self.get_additional_events-method and
+            can modify the original event.
+            To fix the current migration deleted motions will be removed from the update event
+            """
+            if "all_derived_motion_ids" in event.data and self.created_motion_ids.get(
```
jsangmeister, line 80: You have to do this for `all_origin_ids` as well

openslides_backend/migrations/migrations/0032_fix_motion_origin_relations.py
```diff
@@ -63,4 +70,26 @@ def migrate_event(
                         )
                     )
             return new_events
+        elif isinstance(event, UpdateEvent):
+            """Changes the update event
+            Precondition from backend: The backend sorts the events by type: create/update/delete
+            Therefore we don't need the self.get_additional_events-method and
+            can modify the original event.
```
jsangmeister, line 77: This might not hold true in case some other migration ran before, which modified the events and appended another one at the end. I'm not sure if we have such a migration, you might want to double check that.

openslides_backend/migrations/migrations/0032_fix_motion_origin_relations.py
```diff
@@ -63,4 +70,26 @@ def migrate_event(
...
+            To fix the current migration deleted motions will be removed from the update event
+            """
+            if "all_derived_motion_ids" in event.data and self.created_motion_ids.get(
+                id
+            ):
+                new_derived_motion_ids: List[int] = []
+                for motion_id in event.data["all_derived_motion_ids"]:
+                    if motion_id in self.created_motion_ids[id]:
```
jsangmeister, line 85: Why do you need the `created_motion_ids`? Can't you just drop it altogether and just iterate over the fields and check if the motion was deleted?

tests/system/migrations/test_0032_fix_motion_origin_relations.py
```diff
@@ -129,3 +129,105 @@ def test_all_origin_ids(write, finalize, assert_model):
...
+    setup_data(write)
+    write(
+        {
+            "type": "delete",
+            "fqid": "motion/2",
+        },
+        {
+            "type": "update",
+            "fqid": "motion/1",
+            "fields": {"derived_motion_ids": [3]},
```
jsangmeister, line 144: This is wrong, `motion/3` is not created yet

tests/system/migrations/test_0032_fix_motion_origin_relations.py
```diff
@@ -129,3 +129,105 @@ def test_all_origin_ids(write, finalize, assert_model):
...
+        },
+        {
+            "type": "update",
+            "fqid": "meeting/1",
+            "fields": {"motion_ids": [1, 3]},
```
jsangmeister, line 149: same here

jsangmeister (changes requested):

tests/system/migrations/test_0032_fix_motion_origin_relations.py
r-peschke: changed

tests/system/migrations/test_0032_fix_motion_origin_relations.py
r-peschke: changed

openslides_backend/migrations/migrations/0032_fix_motion_origin_relations.py
r-peschke: this way i could be sure that the correct newly created motion is allowed in update. The created motion is still not applied to keyframe_models.
And it is cheaper, because don't need to ask the database

openslides_backend/migrations/migrations/0032_fix_motion_origin_relations.py
r-peschke: There are no motion-creates appended

PR #1489, opened 2022-09-28
Adjust dev setup to match main repo

jsangmeister:


jsangmeister:
Needed for https://github.com/OpenSlides/OpenSlides/pull/6495

dev/docker-compose.dev.yml
```diff
@@ -42,12 +43,13 @@ services:
             - "9010:9010"
         environment:
             - OPENSLIDES_DEVELOPMENT=1
+            - DATASTORE_DATABASE_NAME=openslides-test
         depends_on:
-            - postgresql
+            - postgres
```
r-peschke, line 48: should be postgres-test

dev/docker-compose.dev.yml
```diff
@@ -58,12 +60,13 @@ services:
             - "9011:9011"
         environment:
             - OPENSLIDES_DEVELOPMENT=1
+            - DATASTORE_DATABASE_NAME=openslides-test
         depends_on:
-            - postgresql
+            - postgres
```
r-peschke, line 65: s.o.

dev/docker-compose.dev.yml
```diff
@@ -118,15 +121,15 @@ services:
...
+    postgres-test:
         image: postgres:11
         environment:
             - POSTGRES_USER=openslides
             - POSTGRES_PASSWORD=openslides
-            - POSTGRES_DB=openslides
+            - POSTGRES_DB=openslides-test
         networks:
-            - postgresql
+            - postgres
```
r-peschke, line 132: maybe better to rename also to postgres-test!?

r-peschke (changes requested):

scripts/export_datastore_variables.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-export DATASTORE_DATABASE_HOST=${DATASTORE_DATABASE_HOST:-postgresql}
+export DATASTORE_DATABASE_HOST=${DATASTORE_DATABASE_HOST:-postgres}
```
r-peschke, line 3: This file is used entrypoint.sh for dev and not dev

jsangmeister: This should not be a problem, right? Because in dev setup, I explicitely given the database host, so this default value is not used.

dev/docker-compose.dev.yml
jsangmeister: Mmh, I don't like it. The network is still the same function-wise, only the container changed

r-peschke: Change the names the way you prefer:-). This was my first attempt to fix this one, but then there have been more problems.

scripts/export_datastore_variables.sh
r-peschke: I'm not sure about this!

r-peschke:
The cause of the test failures may be the option --wait-for-client, used in the dev-dockerfile for backend and datastore. This is a mydebug-option, that waits for the attached debug client, which is the only way to debug the startup phase of a service.

jsangmeister:
Pushing this to 4.1 to not introduce big changes to the setup before the release, but will continue the work

jsangmeister:
basically just changes the points mentioned in https://github.com/OpenSlides/OpenSlides/pull/6495#issuecomment-1359122090 - the previous changes of renaming the postgres container/db to `*-test` proved unnecessary since you cannot start the main setup and the backend setup in parallel anyway. For running backend tests etc. in the main setup, see https://github.com/OpenSlides/OpenSlides/pull/6495#issuecomment-1359432008.

rrenkert (approved):

r-peschke (approved):

PR #1495, opened 2022-09-30
Issue1461 vote interface changed

r-peschke:
Cryptographic parts works, correct cryptographic votes are stored in database on poll.stop
- vote errors reported by vote service in poll.stop still not handled => done
- failing (old) vote-tests 0=> fixed
-
- Waiting for vote service to fix some invalid votes checks


openslides_backend/action/actions/poll/mixins.py
```diff
@@ -74,22 +76,28 @@ def on_stop(self, instance: Dict[str, Any]) -> None:
...
         action_data = []
         votesvalid = Decimal("0.000000")
         option_results: Dict[int, Dict[str, Decimal]] = defaultdict(
             lambda: defaultdict(lambda: Decimal("0.000000"))
         )  # maps options to their respective YNA sums
-        for ballot in results["votes"]:
+        for ballot in votes:
             user_token = get_user_token()
-            vote_weight = Decimal(ballot["weight"])
+            vote_weight = Decimal(ballot.get("weight")) or Decimal("1.000000")
```
ostcar, line 92: This does not work if the attribute "weight" is not in ballot. This is the case for cryptographic votes. The following code works:

```python
vote_weight = Decimal("1.000000")
if ballot.get("weight",None) is not None:
    vote_weight = Decimal(ballot.get("weight"))
```

openslides_backend/action/actions/poll/mixins.py
```diff
@@ -110,7 +118,7 @@ def on_stop(self, instance: Dict[str, Any]) -> None:
                             **vote_template,
                         }
                     )
-            elif isinstance(ballot["value"], str):
+            elif isinstance(ballot("value"), str):
```
ostcar, line 121: Here is a typo. The correct line is:

```python
elif isinstance(ballot.get("value"), str):
```

ostcar (changes requested):
Looks good. I had to change the code on two places, But then it worked

ostcar:
@r-peschke Not so important, but it would be nice, if you could clear the new fields on poll.reset

openslides_backend/action/actions/poll/mixins.py
r-peschke: done

openslides_backend/action/actions/poll/mixins.py
r-peschke: done

r-peschke:
> @r-peschke Not so important, but it would be nice, if you could clear the new fields on poll.reset

Done

dev/docker-compose.dev.yml
```diff
@@ -91,7 +91,7 @@ services:
             - cache
     vote:
         build:
-            context: "https://github.com/OpenSlides/openslides-vote-service.git#main"
+            context: "https://github.com/ostcar/openslides-vote-service.git#crypto-preperation"
```
r-peschke, line 94: This has to be reverted before merging!

ostcar:
It is important, that when the poll.status is set to "started" in the datastore that the poll key is written at the same datastore-event. I have not checked, if you already do it.

This means, the backend has to call the vote-service before saving the data to the datastore. (Again, maybe you already do it this way).

r-peschke:
> It is important, that when the poll.status is set to "started" in the datastore that the poll key is written at the same datastore-event. I have not checked, if you already do it.
> 
> This means, the backend has to call the vote-service before saving the data to the datastore. (Again, maybe you already do it this way).
Yes, it is done this way, see start.py, line 58

tests/system/util.py
```diff
@@ -33,21 +47,67 @@ def convert_to_test_response(response: RequestsResponse) -> Response:
...
+        ).derive(shared_key)
+        nonce = os.urandom(nonceSize)
+        cipher = Cipher(algorithms.AES(derived_key), modes.GCM(nonce))
+        encryptor = cipher.encryptor()
+        value_string = str(data.get("value")).replace("'", '"')
+        encrypt_string = bytes(
+            f'{{"votes":{value_string},"token":"12345678"}}', encoding="utf-8"
+        )
+        encrypted = encryptor.update(encrypt_string)
+        encryptor.finalize()
```
ostcar, line 104: You have to add the return value of finalize to the encrypted data. In my tests, it always was a empty byte-string. But I don't know if this is always the case. If you don't want to add it, then you should add an assert statement, that it is empty.

The documentation of the function says, that it has to be added. See the example here: https://cryptography.io/en/latest/hazmat/primitives/symmetric-encryption/#cryptography.hazmat.primitives.ciphers.Cipher

https://cryptography.io/en/latest/hazmat/primitives/symmetric-encryption/#cryptography.hazmat.primitives.ciphers.CipherContext
> Once that is done call finalize() to finish the operation and obtain the remainder of the data.

PR #1497, opened 2022-10-04
Allow to delete a user, if they are a candidate.

reiterl:
This blocked the deleting of a user. So remove it.

### DIFF
--- a/openslides_backend/action/actions/assignment_candidate/delete.py
+++ b/openslides_backend/action/actions/assignment_candidate/delete.py
@@ -1,4 +1,8 @@
+from typing import Any, Dict
+
 from ....models.models import AssignmentCandidate
+from ....shared.exceptions import ActionException
+from ....shared.patterns import fqid_from_collection_and_id
 from ...generics.delete import DeleteAction
 from ...util.default_schema import DefaultSchema
 from ...util.register import register_action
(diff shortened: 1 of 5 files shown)


openslides_backend/action/actions/assignment_candidate/delete.py
```diff
@@ -17,23 +13,3 @@ class AssignmentCandidateDelete(PermissionMixin, DeleteAction):
...
-            ),
-            mapped_fields=["phase", "meeting_id"],
-        )
-        if assignment.get("phase") == "finished" and not self.is_meeting_deleted(
-            assignment.get("meeting_id", 0)
-        ):
-            raise ActionException(
-                "It is not permitted to remove a candidate from a finished assignment!"
-            )
-        return instance
```
jsangmeister, line 39: Please add this back in, it should still not be permitted to delete the assignment candidate, just the user

jsangmeister (changes requested):

openslides_backend/action/actions/assignment_candidate/delete.py
reiterl: Re-added the check and the test.

r-peschke (approved):

PR #1502, opened 2022-10-11
Translate created objects at meeting.create

reiterl:
Add organization/default_language and meeting/language.
Update organization.initial_import and tests.
Update meeting.create and add some tests.


reiterl:
The action wiki must be updated before merging this.

global/data/example-data.json
```diff
@@ -1,11 +1,12 @@
...
-    "_migration_index": 34,
+    "_migration_index": 35,
     "organization": {
         "1": {
             "id": 1,
             "name": "Test Organization",
             "legal_notice": "<a href=\"http://www.openslides.org\">OpenSlides</a> is a free web based presentation and assembly system for visualizing and controlling agenda, motions and elections of an assembly.",
-            "login_text": "Guten Morgen!",
+            "login_text": "Good Morning!",
+            "default_language": "de",
```
jsangmeister, line 9:
```suggestion
            "default_language": "en",
```

openslides_backend/action/actions/meeting/create.py
```diff
@@ -47,6 +51,14 @@ class MeetingCreate(CreateActionWithDependencies, MeetingPermissionMixin):
     skip_archived_meeting_check = True
 
     def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
+        # update Translator
+        if instance.get("language") is None:
+            organization = self.datastore.get(
+                ONE_ORGANIZATION_FQID, ["default_language"], lock_result=False
+            )
+            instance["language"] = organization.get("default_language")
```
jsangmeister, line 59: you can remove this and make it required in the payload

openslides_backend/action/actions/motion_workflow/create.py
```diff
@@ -63,29 +64,29 @@ def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
...
                 "allow_create_poll": True,
                 "allow_support": True,
                 "workflow_id": instance["id"],
                 "first_state_of_workflow_id": instance["id"],
                 "set_created_timestamp": True,
             },
             {
-                "name": "accepted",
+                "name": _("accepted"),
                 "recommendation_label": "Acceptance",
```
jsangmeister, line 76: also translate all recommendation labels here and below

global/meta/models.yml
```diff
@@ -111,6 +111,12 @@ organization:
     restriction_mode: B
     default: 0
     minimum: 0
+  default_language:
```
jsangmeister, line 114: make it required to enforce users to give it in the initial import

openslides_backend/action/actions/organization/initial_import.py
```diff
@@ -84,6 +87,24 @@ def check_empty_datastore(self) -> None:
         ):
             raise ActionException("Datastore is not empty.")
 
+    def translate_organization_and_theme(self, data: Dict[str, Any]) -> None:
+        if "organization" in data and "1" in data["organization"]:
```
jsangmeister, line 91: call this method after the checker so you can be sure that the organization and the id exist and you can skip this check

openslides_backend/action/actions/organization/initial_import.py
```diff
@@ -84,6 +87,24 @@ def check_empty_datastore(self) -> None:
...
+            organization = data["organization"]["1"]
+            Translator.set_translation_language(organization.get("default_language"))
```
jsangmeister, line 93:
```suggestion
            Translator.set_translation_language(organization["default_language"])
```
possible, if you set it as required

jsangmeister (changes requested):

global/data/example-data.json
reiterl: Changed to en.

openslides_backend/action/actions/meeting/create.py
reiterl: I have made "language" a required payload of meeting create and updated the tests.

global/meta/models.yml
reiterl: I have made orga default_language required.

openslides_backend/action/actions/organization/initial_import.py
reiterl: I have moved the translate method behind the checker and removed this check.

openslides_backend/action/actions/organization/initial_import.py
reiterl: Done.

tests/system/action/organization/test_initial_import.py
```diff
@@ -32,15 +33,28 @@ def test_initial_import_with_initial_data_file(self) -> None:
...
+                        "legal_notice",
+                        "users_email_subject",
+                        "users_email_body",
+                    ):
+                        if entry.get(field):
+                            entry[field] = Translator.translate(entry[field])
+                if collection == "theme":
+                    if entry.get("name"):
+                        entry["name"] = Translator.translate(entry["name"])
+                self.assert_model_exists(f"{collection}/{id_}", entry)
```
jsangmeister, line 58: add a test where you don't provide a `default_language`

jsangmeister (changes requested):
one test missing, rest looks good

tests/system/action/organization/test_initial_import.py
reiterl: Add a test with missing default_language.

jsangmeister (approved):
manage tool and client have to be adjusted at the same time

rrenkert:
> manage tool and client have to be adjusted at the same time

Can we merge this one or are the mentioned repos dependencies?
If the other repos need changes before we can merge, please create issues.

global/meta/models.yml
```diff
@@ -111,6 +111,13 @@ organization:
...
+    type: string
+    read_only: true
+    description: Filled at organization initial import.
+    default: en
```
jsangmeister, line 118: Remove the default here, it is not used in the action anyway

jsangmeister (changes requested):
We need to extend the migration to translate the existing group names etc., or the users will suddenly see the english names even in their german client. Or did we have another solution which I forgot?

@rrenkert I will do that.

global/meta/models.yml
jsangmeister: Also, we should restrict its values to the existing languages in the client, which are `en`, `de`, `it`, `es`, `ru` and `cs`

jsangmeister:
The manage tool does not need an update, I think, because the new field is in the `initial-import` payload and therefore we don't need a flag in the manage tool. Or @normanjaeckel are there any other changes needed?

normanjaeckel:
I looks to me that we don't need any changes in the manage tool here.

jsangmeister:
Talked with @rrenkert and we have two options, since we don't want to implement a migration to automatically translate all objects to german:
- Implement a "dynamic" migration which reads the target language from an environment variable and automatically translates all group names etc. to the target language via hardcoded translation in the migration. We could either force the user to supply such a variable and throw an error if not provided or implicitly set it to `en` if not given, which would result in no changes at all. The former would introduce the need for more admin interaction and might confuse some self-hosters, while the latter might result in users migrating by accident without setting a language (either by forgetting it or because they did not know that this needed to be done) and having an english interface, although they previously used another language.
- Implement a script which automatically updates all names etc. to the desired language, given as a parameter, and link it visibly in the README and CHANGELOG. This has the advantage that it can be used independently of the migrations and user can use it at anytime they find it as long as they have not translated their instance manually yet. The disadvantage is that it requires even more user interaction and some users might not find the script at all. Also, the history would not be changed with this script, so if any user uses the history mode, he will see the english names independent of his current or previous language.

jsangmeister:
Talked with @emanuelschuetze and @rrenkert and we decided to go with the second approach. The script should be placed in the `cli` folder. It should work by calling the reader and writer directly for fetching and writing the data. You can orientate yourself on 
[delete-projections.txt](https://github.com/OpenSlides/openslides-backend/files/10783831/delete-projections.txt) for the general structure of reading and writing.

Also, I left some minor change requests above which should also be resolved. If you have questions, feel free to ask. @reiterl

global/meta/models.yml
reiterl: Why should it be updated? What should happen then? I think that users may expect to update certain default stuff too. We need a concept for this.

reiterl: I removed read_only and description from the default_language field, because it should be allowed to be updated.

reiterl: Done.

reiterl:
Which stuff should be translated? From en->another language? or from allowed language to allowed language? Can you explain the code of delete projections?

jsangmeister:
> Which stuff should be translated?

- group names
- workflow names
- workflow state names & recommendation labels
- countdown names
- projector names
- values of meeting & organization settings, if they have content in english (check the `models.yml` for defaults)

For the last point I noticed that translations are also missing in `meeting.create`.

You can just pipe all these through the translator, since they are currently all translated in the client the same way.

> From en->another language? or from allowed language to allowed language?

Only from `en` to all other languages. Assert beforehand that `organization/default_language` is `en`, otherwise stop since trying to translate from any other language does not make sense.

> Where is the translation map?

You should hopefully be able to just import the translator in your code and use that, since it has no dependencies.

global/meta/models.yml
```diff
@@ -732,6 +743,11 @@ meeting:
   imported_at:
     type: timestamp
     restriction_mode: B
+  language:
```
jsangmeister, line 746: Also add the enum here

global/meta/models.yml
```diff
@@ -732,6 +743,11 @@ meeting:
...
+    type: string
+    restriction_mode: A
+    description: Filled at meeting creation.
```
jsangmeister, line 749: you can remove the description, almost all fields are filled at meeting creation

jsangmeister (changes requested):

global/meta/models.yml
reiterl: Removed. Done

reiterl:
Ehm, which fields of meeting and organization should be translated extra?

reiterl:
Add a translate.py script. Pls review.

jsangmeister:
> Ehm, which fields of meeting and organization should be translated extra?

- `organization`
  - `name`
  - `login_text`
  - `description`
- `meeting`
  - `name`
  - `description`
  - `welcome_title`
  - `welcome_text`
  - `motion_preamble`
  - `motions_export_title`
  - `assignments_export_title`
  - `users_pdf_welcometitle`
  - `users_pdf_welcometext`
  - `users_email_sender`
  - `users_email_subject`
  - `users_email_body
- `motion`
  - `recommendation_label`

I hope I didn't miss any.

cli/translate.py
```diff
@@ -0,0 +1,88 @@
...
+def read_collection(collection: str, fields: List[str]) -> Any:
+    reader: Reader = injector.get(Reader)
+    with reader.get_database_context():
+        response = reader.filter(
+            FilterRequest(
+                collection,
+                FilterOperator("meta_deleted", "=", False),
+                ["id", *fields],
+            )
+        )
```
jsangmeister, line 40: Don't use a filter, especially not with so many results. Fetch the relations in order (`organization`->`meeting`->meeting-related model`) and apply the translations in each step. Only use `get` and `get_many` with `deleted_models_behaviour=NO_DELETED`.

cli/translate.py
```diff
@@ -0,0 +1,88 @@
...
+    print("     Translates from en to <language>.")
+    print("     language could be de, en, it, es, ru, cs")
+
+
+def main() -> None:
+    if len(sys.argv) != 2:
+        print_help()
+        sys.exit(1)
+
+    language = sys.argv[1]
```
jsangmeister, line 55: assert that the language is valid

cli/translate.py
```diff
@@ -0,0 +1,88 @@
...
+from datastore.reader.core import FilterRequest, Reader
+from datastore.reader.services import register_services as register_reader_services
+from datastore.shared.di import injector
+from datastore.shared.util import FilterOperator, fqid_from_collection_and_id
+from datastore.writer.core import RequestUpdateEvent, Writer, WriteRequest
+from datastore.writer.services import register_services as register_writer_services
+
+from openslides_backend.locale.translator import Translator
+
+collection_to_fields_map = {
```
jsangmeister, line 13: Add the fields mentioned in https://github.com/OpenSlides/openslides-backend/pull/1502#issuecomment-1441767944

cli/translate.py
jsangmeister: If possible, use the enum already present in the models description for validation

cli/translate.py
```diff
@@ -0,0 +1,88 @@
...
+            if changed_fields:
+                events.append(
+                    RequestUpdateEvent(
+                        fqid_from_collection_and_id(collection, id_),
+                        changed_fields,
+                    )
+                )
+
+    # write events into the datastore
+    if events:
```
jsangmeister, line 81: Also update `organization.default_language` and `meeting.language` for all models. This should be done regardless of whether any events were created otherwise or not.

cli/translate.py
```diff
@@ -0,0 +1,88 @@
...
+        for id_, model in read_collection(collection, fields):
+            changed_fields: Dict[str, Any] = {}
+            for field in fields:
+                old_value = model[field]
+                possible_new_value = Translator.translate(old_value)
+                if possible_new_value != old_value and possible_new_value is not None:
+                    changed_fields[field] = possible_new_value
+            if changed_fields:
+                events.append(
+                    RequestUpdateEvent(
```
jsangmeister, line 74: Bundle the update events for the same objects

jsangmeister (changes requested):

cli/translate.py
reiterl: Done. Added the fields to the `collection_to_fields_map`.

cli/translate.py
reiterl: I use an ordered dictionary to set the relation order, and reorder the relations in collection_to_fields_map.

cli/translate.py
reiterl: Yes, that is done here. The changed_fields variable includes all fields of one fqid.

cli/translate.py
reiterl: I use get_all, that is what I want.

cli/translate.py
reiterl: I update these, too. Done.

README.md
```diff
@@ -7,6 +7,10 @@ Backend service for OpenSlides which
 
 Use docker to build and start the application with your favorite tools. See [Dockerfile](Dockerfile) for steps to setup the production build.
 
+## Translate
+
+To translate group/names etc. from English to another language look at the
+cli/translate.py script.
```
jsangmeister, line 13:
```suggestion
## Translations

We recently updated the source language in the database from english to the user's language. To migrate your data from english to your preferred language, use the provided [translation script](cli/translate.py).
```

cli/translate.py
jsangmeister: ok, but then you can remove the `OrderedDict` again

cli/translate.py
```diff
@@ -0,0 +1,126 @@
...
+        print_help()
+        sys.exit(2)
+
+
+def check_organization_language() -> None:
+    reader: Reader = injector.get(Reader)
+    with reader.get_database_context():
+        response = reader.get(GetRequest(ONE_ORGANIZATION_FQID, ["default_language"]))
+    if response["default_language"] != "en":
+        print("Need organization to have default language equals 'en'.")
```
jsangmeister, line 71:
```suggestion
        print("Cannot translate from source languages other than `en`.")
```

cli/translate.py
```diff
@@ -0,0 +1,126 @@
...
+        "motion": ["recommendation_label"],
+    }
+)
+possible_languages = Organization().default_language.constraints["enum"]
+
+
+def read_collection(collection: str, fields: List[str]) -> Any:
+    reader: Reader = injector.get(Reader)
+    with reader.get_database_context():
+        response = reader.get_all(
```
jsangmeister, line 50: use `DeletedModelsBehaviour.NO_DELETED` to not fetch deleted models.

jsangmeister (changes requested):
What's still missing is translating default values of some meeting settings on `meeting.create`

README.md
reiterl: Changed.

cli/translate.py
reiterl: Changed.

cli/translate.py
reiterl: Okay, I have added this option to the get_all, but I thought it is the default.

reiterl:
Rebased on main.

jsangmeister (approved):
wait for client with merge

bastianjoel (changes requested):
I think before this can be merged at least German translations should be added. With the client PR these fields will not be translated anymore.

reiterl:
Where do I get the German PO file from?

jsangmeister:
I added the german translations & merged the main branch. @rrenkert @m-schieder you should be able to test this now in conjunction with https://github.com/OpenSlides/openslides-client/pull/2072.

rrenkert:
> I added the german translations & merged the main branch. @rrenkert @m-schieder you should be able to test this now in conjunction with [OpenSlides/openslides-client#2072](https://github.com/OpenSlides/openslides-client/pull/2072).

Thanks! We will test this in the next days!

PR #1508, opened 2022-10-20
Add test and fix a problem with delete assignment candidate with empty user_id

reiterl:


reiterl:
I can reproduce the bug, and wrote a test case. Then I wrote a fix.

r-peschke (approved):

PR #1510, opened 2022-10-21
Add submitter_can_write and logic in motion_comment permissions system.

reiterl:


openslides_backend/action/actions/motion_comment/create_delete_update.py
```diff
@@ -41,6 +43,27 @@ def check_permissions(self, instance: Dict[str, Any]) -> None:
...
+        motion_id = instance.get("motion_id")
+        if not motion_id:
+            comment = self.datastore.get(
+                fqid_from_collection_and_id(self.model.collection, instance["id"]),
+                ["motion_id"],
+            )
+            motion_id = comment.get("motion_id")
+
+        if (
+            section.get("submitter_can_write")
```
r-peschke, line 55: Please check flag before reading the motion. The latter is not necessary, if the flag is false.

openslides_backend/action/actions/motion_comment/create_delete_update.py
```diff
@@ -41,6 +43,27 @@ def check_permissions(self, instance: Dict[str, Any]) -> None:
...
+                And(
+                    FilterOperator("user_id", "=", self.user_id),
+                    FilterOperator("motion_id", "=", motion_id),
+                ),
+            )
+        ):
+            return
+
         msg = f"You are not allowed to perform action {self.name}."
         msg += " You are not in the write group of the section or in admin group."
```
r-peschke, line 68: Use this message only, if the flag is False. Otherwise add something like `" You are not in the write group of the section or in admin group and no submitter."

r-peschke (changes requested):

openslides_backend/action/actions/motion_comment/create_delete_update.py
reiterl: Done.

r-peschke (approved):

PR #1516, opened 2022-11-02
Move personal data template fields into meeting_user

reiterl:
Update model and test.


openslides_backend/action/actions/user/toggle_presence_by_number.py
```diff
@@ -60,21 +60,23 @@ def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
...
-            return list(result.keys())[0]
+            return list(result.values())[0]["user_id"]
         elif len(result.keys()) > 1:
             raise ActionException("Found more than one user with the number.")
 
-        filter_ = And(
-            FilterOperator(f"number_${meeting_id}", "=", ""),
-            FilterOperator("default_number", "=", number),
-        )
+        filter_ = FilterOperator("number", "=", "")
```
r-peschke, line 70: Maybe it's easier to look for the default_number in the user-collection.
On 0 results, there is no user with this number, on 1 or more results read meeting_user with a filter with meeting_id and empty number. 
The intersection-set of both resulting user_ids has to have 1 element or is an error.

tests/system/action/meeting_user/test_create.py
```diff
@@ -0,0 +1,22 @@
+from tests.system.action.base import BaseActionTestCase
+
+
+class MeetingUserCreate(BaseActionTestCase):
+    def test_create(self) -> None:
```
r-peschke, line 5: Needs discussion: 
Write one more test to change the about_me-field without any right, but being member of the meeting as replacement for the field in user.update_self. You also have to change the permission check.

tests/system/action/meeting_user/test_update.py
```diff
@@ -0,0 +1,25 @@
...
+                },
+                "meeting_user/5": {"user_id": 1, "meeting_id": 10},
+            }
+        )
+        test_dict = {
+            "id": 5,
+            "comment": "test bla",
+            "number": "XII",
+            "structure_level": "A",
+            "about_me": "A very long description.",
```
r-peschke, line 20: Needs discussion:
Write one more test to change the about_me-field without any right, but being member of the meeting as replacement for the field in user.update_self. You also have to change the permission check.

r-peschke (changes requested):
We have to think on voting-service and to discuss the user.update_self concerning the `about_me`-field

openslides_backend/action/actions/user/create_update_permissions_mixin.py
```diff
@@ -169,7 +169,6 @@ class CreateUpdatePermissionsMixin(UserScopePermissionCheckMixin):
             "structure_level_$",
```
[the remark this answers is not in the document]
reiterl: Done.

openslides_backend/action/actions/user/toggle_presence_by_number.py
```diff
@@ -60,21 +60,23 @@ def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
         return instance
 
     def find_user_to_number(self, meeting_id: int, number: str) -> int:
-        filter_: Filter = FilterOperator(f"number_${meeting_id}", "=", number)
-        result = self.datastore.filter("user", filter_, ["id"])
+        filter_: Filter = FilterOperator("number", "=", number)
```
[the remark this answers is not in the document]
reiterl: Add the meeting_id filter with and to the filter.

openslides_backend/action/actions/user/toggle_presence_by_number.py
reiterl: Start with looking for the number in the user/default_numbers, and then filter for empty string in number and meeting_id.

tests/system/action/meeting_user/test_create.py
reiterl: I think we need a new action, like update_self, for this. This should be discussed. Removing the permission is not the solution.

global/meta/models.yml
```diff
@@ -452,6 +429,38 @@ user:
...
+    restriction_mode: A
+  number:
+    type: string
+    restriction_mode: A
+  structure_level:
+    type: string
+    restriction_mode: A
+  about_me:
+    type: HTMLStrict
+    restriction_mode: A
```
r-peschke, line 458: Set to description_mode B and insert a `description: "restriction_mode B is restriction_mode A or request user == user_id"`

r-peschke (changes requested):
Subject of changes is the field about_me, which may be changed by the user itself. Please cover this with a test in meeting_user.create/update

global/meta/models.yml
jsangmeister: I don't think the description of the restriction mode should be part of the `models.yml`. We already have these in the AU service and neither doubling the information nor becoming inconsistent with other models/restriction modes provides any value.

reiterl: Set restriction_mode(!) to B. As @jsangmeister said, I don't include a description.

r-peschke: This is a restriction mode we invented. If you only write "B" neither Oskar  and in some weeks nor we know what restriction mode B should be. If you have a better place to describe this for Oskar then document it there!.

openslides_backend/action/actions/meeting_user/create.py
```diff
@@ -14,6 +16,21 @@ class MeetingUserCreate(CreateAction):
...
+            "vote_weight",
+        ],
     )
     permission = Permissions.User.CAN_MANAGE
+
+    def check_permissions(self, instance: Dict[str, Any]) -> None:
+        if "about_me" in instance and not any(
+            field in instance
+            for field in ("comment", "number", "structure_level", "vote_weight")
+        ):
```
r-peschke, line 33: Better to ask if there is another field in instance. Your way you have to remember to change the code on adding one more field.

openslides_backend/action/actions/meeting_user/update.py
```diff
@@ -13,6 +16,26 @@ class MeetingUserUpdate(UpdateAction):
...
+            "about_me",
+            "vote_weight",
+        ],
     )
     permission = Permissions.User.CAN_MANAGE
+
+    def check_permissions(self, instance: Dict[str, Any]) -> None:
+        if "about_me" in instance and not any(
+            field in instance
+            for field in ("comment", "number", "structure_level", "vote_weight")
```
r-peschke, line 32: s.o.

global/meta/models.yml
```diff
@@ -471,6 +447,19 @@ meeting_user:
...
     restriction_mode: A
+  number:
+    type: string
+    restriction_mode: A
+  structure_level:
+    type: string
+    restriction_mode: A
+  about_me:
+    type: HTMLStrict
+    restriction_mode: B
```
r-peschke, line 458: Choose a good place to document B for Oskar or describe it here or invent a new tag in models.yml (may be the best option. if description is used anywhere)

r-peschke (changes requested):

openslides_backend/action/actions/meeting_user/update.py
reiterl: Changed the condition into` len(instance) ==1`.

global/meta/models.yml
reiterl: It is documented in the Action wiki, that is a good place to look at.

openslides_backend/action/actions/meeting_user/create.py
reiterl: There is another field in the instance: user_id and meeting_id. I added `len(instance) == 3` here.

global/meta/models.yml
reiterl: The action wiki is not a good place for the description. I talked to @rrenkert and @r-peschke  about it and added the description for now to the model.yml.

r-peschke: The action wiki is not a good place, because
- the action wiki permissions are write-permissions, the restriction modes are read-permissions
- there is no relation between the used abbreviations/group letters between backend actions and autoupdate restriction modes
- one field may appear in various actions. Would be difficult to find them all to have an idea, what a specific restriction mode is for.

r-peschke (approved):

global/meta/models.yml
jsangmeister: The place would be the documentation of all other restriction modes, the go files in the AU repo. To get the information to @ostcar and through that in the file, the usual way to go would be an issue in the according repo where the needed information is documented to be transferred into the file. We should remove this information from the models.yml again.

ostcar: Since I currently don't have a branch for this change, a Issue would be the best place.

PR #1519, opened 2022-11-04
Update motion.set_state, allow a submitter to withdraw.

reiterl:
Add tests, add two fields to motion_state, update motion_state.update.


openslides_backend/action/actions/motion/mixins.py
```diff
@@ -16,6 +16,9 @@ def is_allowed_and_submitter(self, submitter_ids: List[int], state_id: int) -> b
         )
         if not state.get("allow_submitter_edit"):
             return False
+        return self.is_submitter(submitter_ids, state_id)
+
```
r-peschke, line 20: really good "split":-)

r-peschke (approved):

PR #1521, opened 2022-11-08
Repair vote service

r-peschke:
- Set environment variables for direct access of vote service on backend database
- set required fields for poll-object where needed


reiterl (approved):
The inclusion of required data into the tests should be okay.
Adding the options in docker compose is okay, too.

PR #1524, opened 2022-11-10
WIP: First check with total memory consumption

r-peschke:


r-peschke:
May be it is not the best approach, solve the problem from the view of one process!

PR #1525, opened 2022-11-10
Move template field speaker id into meeting_user

reiterl:
Update model, actions and tests.


reiterl:
Against the wrong branch.

PR #1530, opened 2022-11-12
Send a dispatch message to autoupdate on changes

ostcar:
With this change to the github action, the backend should send a dispatch message to the autoupdate repo and trigger the following action, that should create a PR on the autoupdate service

https://github.com/OpenSlides/openslides-autoupdate-service/blob/main/.github/workflows/update-generated-files.yml

This should work, but I am not sure, since it is not possible to test it without making noise by changing the models.yml.

I could not find any documentation, if the [if-condition](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsif) supports something like a `or`. If it does, only one stop would be necessary, since both use the same event-type.

### DIFF
--- a/.github/workflows/create-issue-for-file-updates.yml
+++ b/.github/workflows/create-issue-for-file-updates.yml
@@ -43,16 +43,10 @@ jobs:
         with:
           filename: .github/workflows/announce-file-changes-template.md
       
-      - name: Send dispatch if models.yml was changed
-        if: contains(steps.changed-files.outputs.modified_files, 'global/meta/models.yml')
-        uses: peter-evans/repository-dispatch@v2
-        with:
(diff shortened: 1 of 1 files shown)


jsangmeister:
Looks really good! According to https://docs.github.com/en/actions/learn-github-actions/expressions#operators, `||` should be supported. I would like to change that and then run a few tests with dummy changes, which will create a few unnecessary issues/PRs, but we can close them directly again.

jsangmeister (approved):

jsangmeister:
merging this to test it out

ostcar:
Success https://github.com/OpenSlides/openslides-autoupdate-service/pull/584/files

PR #1537, opened 2022-11-17
Issue1536 Exception if there is no instance and no user_id given for get_user_scope

r-peschke:


r-peschke:
> LGTM. If self.request("get_user_scope", {"user_ids": [0]}), what does happen then?

Then you'll get a DatastoreException from the datastore.get, because  the raise_exception parameter is True per default.

PR #1540, opened 2022-11-18
Remove python minor version from GH actions workflow

jsangmeister:


jsangmeister:
Seems like setup-python only supports the current minor version of all major python versions so 3.10.3 could not be fetched, see https://github.com/actions/setup-python/issues/64#issuecomment-594837888. Replacing it with 3.10.x seems to fix the problem.

reiterl (approved):
Okay, it seems to clear the issue.

PR #1541, opened 2022-11-18
Add search_deleted_models presenter

jsangmeister:
We need a presenter for the client to have access to deleted models. Use case: A user gets deleted by accident and should be (manually, partially) recovered from the history. If the operating user does not remember the id (which will most likely be the case), he has no easy way of accessing the history of the deleted user. This presenter will enable the client to query for deleted models and present them to the operating user in a meaningful way.

### DIFF
--- a/openslides_backend/presenter/search_deleted_models.py
+++ b/openslides_backend/presenter/search_deleted_models.py
@@ -1,0 +1,89 @@
+from typing import Any, List
+
+import fastjsonschema
+from datastore.shared.util import DeletedModelsBehaviour
+
+from ..permissions.management_levels import OrganizationManagementLevel
+from ..permissions.permission_helper import has_organization_management_level
+from ..shared.exceptions import PermissionDenied
+from ..shared.filters import And, Filter, FilterOperator, Or
+from ..shared.schema import schema_version
+from .base import BasePresenter
+from .presenter import register_presenter
+
+schema = fastjsonschema.compile(
+    {
+        "$schema": schema_version,
+        "type": "object",
+        "properties": {
+            "collection": {"type": "string"},
+            "filter_string": {"type": "string"},
+            "meeting_id": {"type": "integer", "minimum": 1},
+        },
+        "required": ["collection", "filter_string"],
+        "additionalProperties": False,
+    }
+)
+
+
+search_fields = {
+    "assignment": ["title"],
+    "motion": ["number", "title"],
+    "user": [
+        "username",
+        "first_name",
+        "last_name",
+        "title",
+        "pronoun",
+        "structure_level",
+        "number",
+        "email",
+    ],
+}
+
+
+@register_presenter("search_deleted_models")
+class SearchDeletedModels(BasePresenter):
+    """
+    Searches all deleted models of the given collection for a given filter string.
+    """
+
+    schema = schema
+
+    def get_result(self) -> Any:
+        if not has_organization_management_level(
+            self.datastore, self.user_id, OrganizationManagementLevel.SUPERADMIN
+        ):
+            raise PermissionDenied("You are not a superadmin")
+
+        collection = self.data["collection"]
+        meeting_id = self.data["meeting_id"]
+        filters: List[Filter] = []
+        for search_field in search_fields[collection]:
+            filters.append(
+                FilterOperator(search_field, "%=", self.data["filter_string"])
+            )
+            filter: Filter
+        if len(filters) > 1:
+            filter = Or(*filters)
+        else:
+            filter = filters[0]
+
+        if collection == "user":
+            filter = And(
+                filter,
+                FilterOperator(f"group_${meeting_id}_ids", "!=", None),
+                FilterOperator(f"group_${meeting_id}_ids", "!=", []),
+            )
+        else:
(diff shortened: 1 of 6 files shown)


openslides_backend/services/datastore/extended_adapter.py
```diff
@@ -372,6 +372,8 @@ def _filter_changed_models(
                 formatted_args.append(
                     f'self._comparable(model.get("{arguments[i]}"), {val_str}) and model.get("{arguments[i]}")'
                 )
+            elif match[0] == "ilike":
+                raise NotImplementedError("%= is not implemented for changed models")
```
jsangmeister, line 376: out of scope for this PR. I don't know if we even need this feature, so I'll suppose we let it stay unimplemented for now.

openslides_backend/presenter/search_deleted_models.py
```diff
@@ -0,0 +1,89 @@
...
+    def get_result(self) -> Any:
+        if not has_organization_management_level(
+            self.datastore, self.user_id, OrganizationManagementLevel.SUPERADMIN
+        ):
+            raise PermissionDenied("You are not a superadmin")
+
+        collection = self.data["collection"]
+        meeting_id = self.data["meeting_id"]
+        filters: List[Filter] = []
+        for search_field in search_fields[collection]:
```
r-peschke, line 62: This raises an error with collections not in search field.
You should think about checking the collections in the schema

openslides_backend/presenter/search_deleted_models.py
```diff
@@ -0,0 +1,89 @@
...
+schema = fastjsonschema.compile(
+    {
+        "$schema": schema_version,
+        "type": "object",
+        "properties": {
+            "collection": {"type": "string"},
+            "filter_string": {"type": "string"},
+            "meeting_id": {"type": "integer", "minimum": 1},
+        },
+        "required": ["collection", "filter_string"],
```
r-peschke, line 23: The code made me assume you want the meeting_id required!?
Please, write the presenter documentation in the wiki.

tests/system/presenter/test_search_deleted_models.py
```diff
@@ -0,0 +1,140 @@
...
+            {
+                "collection": collection,
+                "filter_string": filter_string,
+                "meeting_id": meeting_id,
+            },
+        )
+        self.assertEqual(actual_status_code, status_code)
+        return data
+
+    def test_search_users_first_name(self) -> None:
```
r-peschke, line 90: Please ad tests for the error cases:
- searching without meeting_id ( required or not)
- searching with other collections/not existing connection

r-peschke (changes requested):

jsangmeister:
waiting for https://github.com/OpenSlides/openslides-autoupdate-service/issues/586

r-peschke:
@jsangmeister There are still some change requests. Would you please resolve them?

jsangmeister:
Still waiting for https://github.com/OpenSlides/openslides-autoupdate-service/issues/586 - before we have this feature in the AU, it does not make sense to implement it in the backend

tests/system/presenter/test_search_deleted_models.py
r-peschke: What is the real use case for this? Searching for mistakenly deleted users. This is what the presenter can do and don't need the autoupdate-service for this.
With the idea that we don't want the presenter anymore, we should only implement this here, if this an urgently needed feature. After 3 month I'm in doubt about this and think we should close this issue.

jsangmeister: > What is the real use case for this? 

Not only mistakenly deleted users, but any model which was deleted (by accident or not) somewhere down the line and one wants to access any history data about it. If I don't know the fqid of the model im looking for (which I most certainly do not, as a normal user), the only way to search for the information I want is by accessing a random time in the history of another model which still exists, which lays before the time the model I'm searching for was deleted. Then I'm able to write down the fqid and access the history data of it. This is a valid and imported use case which was discussed with @emanuelschuetze when this PR was created.

> This is what the presenter can do and don't need the autoupdate-service for this.

Currently, the client access the history information via the autoupdate route `history_information`. This lead to https://github.com/OpenSlides/openslides-autoupdate-service/issues/586 being created, since the route is broken for deleted models. Another way would be to use the backend presenter `get_history_information` (which was created by mistake, I think, and in parallel to the AU route), which would need an update of the client and probably changes in the backend presenter, but would be possible. It's important to differ between the two presenters `get_history_information` (already exists) and `search_deleted_models` (implemented in this PR), since they do completely different things.

> With the idea that we don't want the presenter anymore

Before just abandoning presenters, we need a replacement system for it, which I don't see yet.

> After 3 month I'm in doubt about this and think we should close this issue.

It was a feature request 3 months ago. If @emanuelschuetze now decides that we don't need it anymore, we can close it, otherwise I think since https://github.com/OpenSlides/openslides-autoupdate-service/issues/586 was closed without intention to change the AU the way to go would be to change the client to use the `get_history_information` presenter so this can be finalized and merged. This would also supposedly make @ostcar happy since the `history_information` route could be deleted from the AU ;)

jsangmeister:
Closing this for now at it does not seem to be needed any more.

PR #1542, opened 2022-11-22
Add correct state_extension handling

jsangmeister:
See updated section https://github.com/OpenSlides/OpenSlides/wiki/Motions#motion-recommendation

The `state_extension` field should be handled analogously to the `recommendation_extension` field since the same reference ability is present there. Therefore, a new relation is necessary for this.

Theoretically we would need a migration to fill the new relation and also consequently use `[<fqid>]` as the replacement pattern instead of `[<collection>:<id>]`, but we decided against it because of the urgency. Maybe we can add this later.


jsangmeister:
test in conjunction with https://github.com/OpenSlides/openslides-client/pull/1824

emanuelschuetze (approved):

r-peschke (commented):
I'm not able to review this. I don't understand, what is done here and the [motion-recommendation-doc](https://github.com/OpenSlides/OpenSlides/wiki/Motions#motion-recommendation) doesn't help me, too.

jsangmeister:
> I'm not able to review this. I don't understand, what is done here and the [motion-recommendation-doc](https://github.com/OpenSlides/OpenSlides/wiki/Motions#motion-recommendation) doesn't help me, too.

The field `recommendation_extension` may contain placeholders that are replaced by the client with the appropriate motion number. In OS3, the placeholder was `[motion:<id>]` and no further fields or anything were necessary, the replacement was done fully on the client's side. The field `state_extension` works analogously.

In OS4, some things changed:
- Since the client now does not have all models as opposed to OS3, we needed a way to fetch these models from the AU. Therefore, the relation `recommendation_extension_reference_ids` was introduced, which contains all models referenced in a motion's `recommendation extension`. But this was not done for the field `state_extension`, although this would have been neccessary. This is rectified in this PR.
- Since we now have fqids and the colon was always somewhat arbitrary, the format was changed to `[motion/<id>]` (or, more general, `[<fqid>]`. This, however, was only done partially: The backend was adjusted to respect these new changes (in export, import and `motion.update` to fill the new relation), but the client was not. This lead to the feature seemingly working, since the client replaced the placeholders correctly, but the relation was not filled by the backend and the client did not fetch the models correctly. This PR (and the client one) unifies the handling to use `[<fqid>]`, as intended.

As stated above, we will need a migration to fix the existing motions and fill both relations, but for time reasons, this fix is enough for now.

openslides_backend/action/actions/motion/follow_recommendation.py
```diff
@@ -47,12 +52,19 @@ def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
...
+        recommendation_extension_reference_ids = instance.pop(
+            "recommendation_extension_reference_ids", None
+        )
         if (
             recommendation_extension is not None
             and recommendation.get("show_state_extension_field")
             and recommendation.get("show_recommendation_extension_field")
         ):
             instance["state_extension"] = recommendation_extension
+            if recommendation_extension_reference_ids:
```
r-peschke, line 64: if this is None, the content of `state_extension_reference_ids` remains unchanged and differs from it's `state_extension`. Is this intended?

tests/system/action/motion/test_update.py
```diff
@@ -217,13 +217,14 @@ def test_update_correct_2(self) -> None:
...
-        assert model.get("state_extension") == "test_blablab_noon"
-        assert model.get("recommendation_extension") == "ext_sldennt [motion/112]"
+        assert model.get("state_extension") == "ext [motion/112] [motion/113]"
+        assert model.get("recommendation_extension") == "ext [motion/112] [motion/113]"
         assert model.get("category_id") == 4
         assert model.get("block_id") == 51
         assert model.get("supporter_ids") == []
         assert model.get("tag_ids") == []
         assert model.get("attachment_ids") == []
+        assert model.get("state_extension_reference_ids") == ["motion/112"]
```
r-peschke, line 227: What's about motion/113?

r-peschke (changes requested):
I'm missing tests for meeting.import and meeting.clone with the extension and referenced fields.

openslides_backend/action/actions/motion/follow_recommendation.py
jsangmeister: Good catch! Added a fix and a test for this case

tests/system/action/motion/test_update.py
jsangmeister: motion/113 does not exist and should therefore not be present in the relation. This test tests multiple things at once, maybe that's why it's a little confusing. Added a comment for this.

jsangmeister:
Added tests for `meeting.import` and `meeting.clone`

PR #1544, opened 2022-11-23
Extension field migration

jsangmeister:
Migration based on https://github.com/OpenSlides/openslides-backend/pull/1542/ to fix the broken fields and relations, as mentioned there.


jsangmeister:
@emanuelschuetze what's the plan here? I would like to have this merged soon, so the client codebase does not diverge too much.

jsangmeister:
@r-peschke review is still missing

r-peschke (approved):
**Before merging you have to see for correct datastore commit hash**

PR #1556, opened 2022-12-05
Bugfix: database error as string not handled

r-peschke:


openslides_backend/services/datastore/handle_datastore_errors.py
```diff
@@ -34,31 +34,39 @@ def raise_datastore_error(
...
+                raise DatastoreLockedException(error_message)
+            elif type_verbose == "MODEL_DOES_NOT_EXIST":
+                error_message = " ".join(
+                    (
+                        error_message,
+                        f"Model '{additional_error_message.get('fqid')}' does not exist.",
+                    )
+                )
+            else:
+                error_message = " ".join((error_message, str(additional_error_message)))
```
reiterl, line 69: Doesn't look right, why is this same error_message twice?

openslides_backend/services/datastore/handle_datastore_errors.py
```diff
@@ -34,31 +34,39 @@ def raise_datastore_error(
...
-                (
-                    error_message,
-                    f"The following locks were broken: {broken_locks}",
+        if isinstance(additional_error_message, dict):
+            type_verbose = additional_error_message.get("type_verbose", "")
+            if type_verbose == "MODEL_LOCKED":
+                broken_locks = (
+                    "'"
+                    + "', '".join(
+                        sorted(cast(str, additional_error_message.get("keys")))
```
reiterl, line 46: It doesn't look right here.

reiterl (commented):
Here are no tests. They are missing.

openslides_backend/services/datastore/handle_datastore_errors.py
reiterl: This cast to str doesn't make sense.

openslides_backend/services/datastore/handle_datastore_errors.py
r-peschke: IMO it is easier this way. 
The first one is to append the additional_error_message, which is a dict, but only conditionally.
The second one is to append it, if it is not a dict, but without any conditions. 
It is by chance that the statement is identical, the str-function takes all types as argument.

reiterl: Okay, accepted.

openslides_backend/services/datastore/handle_datastore_errors.py
reiterl: Okay, we discussed it and use the Iterable[str].

PR #1562, opened 2022-12-14
Replace template field amendment_paragraph with json field

reiterl:
Update model, actions and tests.

### DIFF
--- a/openslides_backend/action/actions/motion/create.py
+++ b/openslides_backend/action/actions/motion/create.py
@@ -6,8 +6,12 @@ from ....permissions.permission_helper import has_perm
 from ....permissions.permissions import Permissions
 from ....services.datastore.commands import GetManyRequest
 from ....shared.exceptions import ActionException, MissingPermission, PermissionDenied
-from ....shared.patterns import POSITIVE_NUMBER_REGEX, fqid_from_collection_and_id
-from ....shared.schema import id_list_schema, optional_id_schema
+from ....shared.patterns import fqid_from_collection_and_id
+from ....shared.schema import (
+    id_list_schema,
+    number_string_json_schema,
+    optional_id_schema,
+)
 from ...util.default_schema import DefaultSchema
 from ...util.register import register_action
 from ...util.typing import ActionData
@@ -42,10 +46,7 @@ class MotionCreate(AmendmentParagraphHelper, MotionCreateBase):
         additional_optional_fields={
             "workflow_id": optional_id_schema,
             "submitter_ids": id_list_schema,
-            "amendment_paragraph": {
-                "type": "object",
-                "propertyNames": {"pattern": POSITIVE_NUMBER_REGEX},
-            },
+            "amendment_paragraph": number_string_json_schema,
             **agenda_creation_properties,
         },
     )
--- a/openslides_backend/action/actions/motion/create.py
+++ b/openslides_backend/action/actions/motion/create.py
@@ -6,17 +6,18 @@ from ....permissions.permission_helper import has_perm
 from ....permissions.permissions import Permissions
 from ....services.datastore.commands import GetManyRequest
 from ....shared.exceptions import ActionException, MissingPermission, PermissionDenied
-from ....shared.patterns import POSITIVE_NUMBER_REGEX, fqid_from_collection_and_id
+from ....shared.patterns import fqid_from_collection_and_id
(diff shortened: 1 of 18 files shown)


r-peschke:
Please write tests with filled amendment_paragraph, each with valid html and invalid html
- meeting.import
- meeting.clone
- motion.create
- motion.update

openslides_backend/action/actions/motion/mixins.py
```diff
@@ -31,3 +33,15 @@ def is_submitter(self, submitter_ids: List[int], state_id: int) -> bool:
...
         )
+
+
+class AmendmentParagraphHelper:
+    def validate_amendment_paragraph(self, instance: Dict[str, Any]) -> None:
+        if not isinstance(instance["amendment_paragraph"], dict):
+            raise ActionException("Wrong amendment_paragraph except dict")
+        for key, html in instance["amendment_paragraph"].items():
+            if not str.isdigit(key):
+                raise ActionException(f"amendment_paragraph {key} not allowed.")
```
jsangmeister, line 44: I would rather put this directly in the schema - you cannot use the `optional_properties` then, but have to enter the schema manually into the `additional_optional_properties`, but it's cleaner and more uniform than checking the schema here in the code.

jsangmeister (changes requested):

openslides_backend/action/actions/motion/mixins.py
reiterl: I moved both check into the schema.

reiterl:
I have added the missing import and checker code and have added some tests for them.

openslides_backend/action/actions/motion/create.py
```diff
@@ -42,6 +42,10 @@ class MotionCreate(AmendmentParagraphHelper, MotionCreateBase):
         additional_optional_fields={
             "workflow_id": optional_id_schema,
             "submitter_ids": id_list_schema,
+            "amendment_paragraph": {
+                "type": "object",
+                "propertyNames": {"pattern": POSITIVE_NUMBER_REGEX},
```
jsangmeister, line 47: Also restrict the values here (currently, a value of e.g. `null` or `1` is also valid, when only strings shoul be valid). I would suggest to use `patternProperties` in combination with `additionalProperties: false`.

I would also extract this partial jsonschema definition and reuse it in `motion.update`.

openslides_backend/models/checker.py
```diff
@@ -562,6 +564,21 @@ def get_enum_from_collection_field(
...
+    def check_amendment_paragraph(self, model: Dict[str, Any], collection: str) -> None:
+        if collection != "motion":
+            return
+        if "amendment_paragraph" in model:
+            msg = f"{collection}/{model['id']}/amendment_paragraph error: "
+            if not isinstance(model["amendment_paragraph"], dict):
+                self.errors.append(msg + "Not a dict.")
+            for key, html in model["amendment_paragraph"].items():
+                if not str.isdigit(key):
+                    self.errors.append(msg + f"{key} is not a number.")
```
jsangmeister, line 576: also use the jsonschema here

openslides_backend/models/checker.py
```diff
@@ -562,6 +564,21 @@ def get_enum_from_collection_field(
         field_type = self.get_model(collection).get_field(field)
         return field_type.constraints.get("enum")
 
+    def check_amendment_paragraph(self, model: Dict[str, Any], collection: str) -> None:
```
jsangmeister, line 567: Can you rename this to something like `check_special_fields` and move the recommendation extension validation code from https://github.com/OpenSlides/openslides-backend/blob/d30f2a82f49176442b8ebc1c0ce15df8c87b414f/openslides_backend/models/checker.py#L684-L705 also in here to unify the handling of such fields with extra validation?

jsangmeister (changes requested):

openslides_backend/action/actions/motion/create.py
reiterl: Done.

openslides_backend/models/checker.py
reiterl: Done.

openslides_backend/models/checker.py
reiterl: Done.

jsangmeister (approved):

PR #1563, opened 2022-12-15
Move template field group_ids into meeting_user 

reiterl:
If all tests are okay, we should merge to let client, autoupdate and vote-service do their changes
Should be reviewed by @jsangmeister anyway.

There is a #1660 issue to reactivate delete and skipped test and reach again the 96% coverage


reiterl:
Look at the last two commits for more info.

jsangmeister:
Some general notes:
- the outside interface of the actions should stay the same! The client has to be able to still give all the `meeting_user` fields in the `user.create` payload, otherwise he would have to first execute the `user.create`, wait for the result with the id and then execute a `meeting_user.create`.
- the calculated fields can be left out for now and done in a different PR to not bloat this any more than necessary
- regarding the changes in the vote service: @ostcar can you create a branch in the vote repo with the appropriate changes? Before that, I don't think fixing the tests makes sense. One could introduce some kind of hack to make the vote tests work for the backend and the vote service, but I think that's cumbersome as we need the vote service changes anyway.
- For your questions regarding the actions/tests, you have to be more specific. I can't give such unspecific help for specific actions ;) I think you already got the general concept: Everytime `group_$_ids` was accessed, you need to fetch the appropriate `meeting_user` now instead and everytime it was written, you have to write to it.

reiterl:
@jsangmeister

- I don't think, the interface should stay the same, that would mean still use tf on client side.
- I think, that the calc fields handlers should be fixed in this pr. Else we need to comment out 100+ lines in the tests.

So you can help with changing the calcfields handlers, or work on meeting.import and meeting.clone.
There are some draft versions in the pr, but they don't handle the extra admin and default user you can add to the payload.

global/meta/models.yml
```diff
@@ -315,20 +315,6 @@ user:
...
-    type: template
-    replacement_collection: meeting
-    fields:
-      type: relation-list
-      to: group/user_ids
-    restriction_mode: A
-
   meeting_ids:
     type: number[]
     description: Calculated. All ids from group_$_ids as integers.
```
ostcar, line 320: The field `group_$_ids` does not exist anymore. The description has to be updated.

Is this field necessary at all? It should be quite easy to use user/user_meeting_ids/meeting_id

ostcar:
Should the field `user/is_present_in_meeting_ids` be replaced by `meeting_user/is_present`?

ostcar:
https://github.com/openslides/openslides-vote-service/tree/feature/vote-decrypt

I habe to fix some Tests, but the Code should work

ostcar:
Sry, wrong Link. Here is the correct link to the PR and the feature branch: https://github.com/OpenSlides/openslides-vote-service/pull/173

r-peschke:
> Should the field `user/is_present_in_meeting_ids` be replaced by `meeting_user/is_present`?

At the moment we will leave it, because the goal in this step is only to remove template fields

global/meta/models.yml
r-peschke: > The field `group_$_ids` does not exist anymore. The description has to be updated.
> 
> Is this field necessary at all? It should be quite easy to use user/user_meeting_ids/meeting_id

No, because there is still no mechanism to delete the user_meeting instance automatically, if the user isn't in some group of the meeting anymore. At the opposite we had to reactivate the meeting_user, if the user gains rights in this meeting again.

ostcar: But please change at least the description of the field.

openslides_backend/action/actions/meeting_user/helper.py
```diff
@@ -17,4 +18,8 @@ def create_or_get_meeting_user(self, meeting_id: int, user_id: int) -> int:
                 MeetingUserCreate,
                 [{"meeting_id": meeting_id, "user_id": user_id}],
             )
-            return action_results[0]["id"]  # type: ignore
+            id_ = action_results[0]["id"]  # type: ignore
+            self.datastore.changed_models.get(
+                fqid_from_collection_and_id("meeting_user", id_), {}
+            ).pop("meta_new", None)
```
jsangmeister, line 24: Why do you remove the `meta_new` here?

tests/system/action/base.py
```diff
@@ -280,40 +267,91 @@ def create_user_for_meeting(self, meeting_id: int) -> int:
...
+            [
+                int(k[1])
+                for key in self.created_fqids
+                if (k := key.split("/"))[0] == "meeting_user"
+            ]
+            or [0]
         )
+        meeting_users_new = {
+            meeting_id: {
+                "id": (last_meeting_user_id := last_meeting_user_id + 1),  # noqa: F841
```
jsangmeister, line 305: You can remove the assignment and the `noqa` if I see it correctly - the newly assigned value is never used

tests/system/action/conftest.py
```diff
@@ -0,0 +1,2 @@
+def pytest_runtest_logreport(report):  # type:ignore
+    report.nodeid = ">" + report.nodeid[-75:]
```
jsangmeister, line 2: What is this for?

jsangmeister (changes requested):
Only rough review, some things I noticed

global/meta/models.yml
```diff
@@ -314,19 +314,9 @@ user:
...
-  group_$_ids:
-    type: template
-    replacement_collection: meeting
-    fields:
-      type: relation-list
-      to: group/user_ids
+  poll_candidate_ids:
+    type: relation-list
+    to: poll_candidate/user_id
     restriction_mode: A
```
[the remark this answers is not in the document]
r-peschke: Done

openslides_backend/action/actions/meeting/import_.py
```diff
@@ -482,19 +473,64 @@ def replace_fn(match: re.Match[str]) -> str:
...
+            if meeting_user.get("user_id") == self.user_id:
+                new_meeting_user_id = int(meeting_user_id)
+                if new_meeting_user_id not in (group.get("meeting_user_ids", {}) or {}):
+                    data_json["meeting_user"][meeting_user_id]["group_ids"] = (
+                        data_json["meeting_user"][meeting_user_id].get("group_ids")
+                        or []
+                    ) + [admin_group_id]
+                break
+        if not new_meeting_user_id:
+            # und hier der request_user, der nicht im import drin ist
```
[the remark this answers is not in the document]
r-peschke: Removed

openslides_backend/action/actions/meeting/import_.py
```diff
@@ -482,19 +473,64 @@ def replace_fn(match: re.Match[str]) -> str:
...
+                fqid_user := fqid_from_collection_and_id("user", self.user_id),
+                ["id", "meeting_user_ids", "committee_management_ids", "committee_ids"],
+            )
+            request_user.pop("meta_position", None)
+            request_user["meeting_user_ids"] = (
+                request_user.get("meeting_user_ids") or []
+            ) + [new_meeting_user_id]
+            data_json["user"][str(self.user_id)] = request_user
+            self.replace_map["user"].update(
+                {0: self.user_id}
```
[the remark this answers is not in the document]
r-peschke: The mapping is from import-data to new number and event creating is triggered by the value/right-side of the map. The executing user is not in the import, but event creation is needed.

openslides_backend/action/actions/meeting/import_.py
```diff
@@ -589,31 +630,30 @@ def create_events(
...
+        #                 "meeting_user", self.new_meeting_user_id
+        #             ),
+        #             list_fields={
+        #                 "add": {
+        #                     "group_ids": [self.new_group_for_request_user],
+        #                 },
+        #                 "remove": {},
+        #             },
+        #         )
+        #     )
```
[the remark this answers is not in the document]
r-peschke: removed

openslides_backend/action/actions/meeting_user/helper.py
r-peschke: Otherwise I think, will be created again

openslides_backend/action/actions/motion/create_forwarded.py
```diff
@@ -84,43 +86,59 @@ def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
...
+                        MeetingUserUpdate,
+                        [
+                            {
+                                "id": meeting_user["id"],
+                                "group_ids": (meeting_user.get("group_ids") or [])
+                                + [target_meeting["default_group_id"]],
+                            }
+                        ],
+                    )
+
```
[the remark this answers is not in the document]
r-peschke: meeting_user.set_data is just a comfort layer, but here I know, if it is create or update

tests/system/action/base.py
r-peschke: done

tests/system/action/conftest.py
r-peschke: Removed
I use it for shifting the pytest output to see the errors better and not only the initial path

PR #1568, opened 2023-01-06
Set password to default password on import

### LINKED ISSUE #1567: Set password to default password on import
The newly generated passwords are useless, because no one knows them or has access to them, so we need to set them to the given default passwords.

jsangmeister:
fixes #1567

### DIFF
--- a/openslides_backend/action/actions/meeting/import_.py
+++ b/openslides_backend/action/actions/meeting/import_.py
@@ -324,7 +324,9 @@ class MeetingImport(SingularActionMixin, LimitOfUserMixin, UsernameMixin):
         # generate passwords
         for entry in json_data["user"].values():
             if entry["id"] not in self.merge_user_map:
-                entry["password"] = self.auth.hash(get_random_string(10))
+                if not entry.get("default_password"):
+                    entry["default_password"] = get_random_string(10)
+                entry["password"] = self.auth.hash(entry["default_password"])
 
         # set enable_anonymous
         meeting["enable_anonymous"] = False
--- a/openslides_backend/action/actions/meeting/import_.py
+++ b/openslides_backend/action/actions/meeting/import_.py
@@ -206,7 +206,13 @@ class MeetingImport(SingularActionMixin, LimitOfUserMixin, UsernameMixin):
                     "derived_motion_ids",
                     "all_origin_id",
                     "all_derived_motion_ids",
-                ]
+                ],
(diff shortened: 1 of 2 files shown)


r-peschke (commented):
This way is not secure: The importer of the meeting will receive all passwords in uncoded text. IMO we shouldn't do that!
In a final step we should generate a new default_password, set it as initial password and provide the password via secure channel.

jsangmeister:
> The importer of the meeting will receive all passwords in uncoded text

I don't think this is a problem: To import a meeting, you need to be a committee manager. To see a user's `default_password`, you need to have OML `can_manage_users` or `user.can_manage` in one of the user's meetings. While, strictly speaking, these are separate permissions, it is very likely that many request users will have both requirements fulfilled (and I think the committee manager cannot see users default passwords only because it hasn't come up as a use case, not because there is an inherent security problem with that). So in conclusion, either the importer has the right to see the passwords anyway after importing or it is at least not a problem.

That being said, I see the risk of importing clear-text passwords from a file which might have wandered through anyone's hands. So I'm fine with both solutions - it once again comes down to security vs. convenience ;)

openslides_backend/action/actions/meeting/import_.py
```diff
@@ -324,7 +324,9 @@ def update_meeting_and_users(self, instance: Dict[str, Any]) -> None:
         # generate passwords
         for entry in json_data["user"].values():
             if entry["id"] not in self.merge_user_map:
-                entry["password"] = self.auth.hash(get_random_string(10))
+                if not entry.get("default_password"):
```
r-peschke, line 327: After talking with @emanuelschuetze : Create always a new default_password and set it, i.e. never import the default_password.
We don't need to send the password automatically to the users, the admin will and can select users to send invitation mails.

r-peschke (changes requested):
Please add the following task to this issue:
- don't import default_password, password and last_email_send
-  include a test for these fields

emanuelschuetze (changes requested):
I imported a OS3-meeting-export into OS4. Open meeting. Go to participant detail view: the default password is the same as in OS3! But loging does NOT work with old password (OK). Reset of password helps to login successfully.
Expected: new default password should show correctly in detail view.

emanuelschuetze (approved):
After backend restart: Works now for me!

r-peschke (approved):

PR #1575, opened 2023-01-11
Ignore action_worker meeting-import/export and datastore checker.

emanuelschuetze:
Fixed warning (```Collections in file do not match with models.py. Invalid collections: action_worker.```) from (extensive) datastore checker (button in client on legal notice page) by adding missiong collection `action_worker`.


r-peschke (changes requested):

jsangmeister:
I don't think this should be run through the checker at all, but rather removed beforehand

r-peschke:
@joshua: Yes, it shouldn't be exported at all.

jsangmeister:
@r-peschke you tagged the wrong person :D

jsangmeister (approved):

r-peschke (approved):

PR #1579, opened 2023-01-13
Add migration to remove unnecessary events

jsangmeister:


jsangmeister:
Tested with @peb-adr on a big instance, no errors came up, events were reduced by about 25% (~2.1m to ~1.6m). If you approve @r-peschke I would implement the same logic into the actual datastore to ignore unnecessary events.

r-peschke (approved):

PR #1603, opened 2023-01-23
Fix a UnboundLocalError in checker.

reiterl:


openslides_backend/models/checker.py
```diff
@@ -631,7 +631,7 @@ def check_relation(
                     )
             elif self.mode == "external":
                 self.errors.append(
-                    f"{basemsg} points to {foreign_collection}/{foreign_id}, which is not allowed in an external import."
+                    f"{basemsg} points to {foreign_collection} and {foreign_ids}, which is not allowed in an external import."
```
jsangmeister, line 634:
```suggestion
                    f"{basemsg} points to {foreign_collection}/{foreign_field}, which is not allowed in an external import."
```

jsangmeister (changes requested):

openslides_backend/models/checker.py
reiterl: Done, changed the error msg.

jsangmeister (approved):

PR #1607, opened 2023-01-25
WIP:Issue1590 gunicorn memory 

r-peschke:
Status quo: There is no general way to release the memory of gunicorn processes. There is 1 unicorn-setting to solve this problem in a general way: [max-requests](https://docs.gunicorn.org/en/stable/settings.html#max-requests) for a worker-process to reuse and the related max_requests_jitter.

Here we will look for special cases with high memory consumption:
- the migration process
- the import/cloning data

Will remove the high label, because on current servers the problems can be resolved by the admin team


r-peschke:
Close this one, seems thre is no easy way to do and the new server configuration solved the current problems,

PR #1610, opened 2023-01-30
Add permissions array to `group.update`

jsangmeister:
Needed for easier implementation of https://github.com/OpenSlides/openslides-client/issues/1925: The permission updates are not sent one by one anymore, but instead as a whole, which is easier to do with this change. `group.set_permission` is not used anywhere else, so it is removed here.


openslides_backend/action/actions/group/update.py
```diff
@@ -12,5 +15,14 @@ class GroupUpdateAction(UpdateAction):
     """
 
     model = Group()
-    schema = DefaultSchema(Group()).get_update_schema(optional_properties=["name"])
+    schema = DefaultSchema(Group()).get_update_schema(
+        optional_properties=["name", "permissions"]
```
r-peschke, line 19: Proposal: Use 2 fields for changing permissions instead of the database field permissions:
add_permissions and remove_permissions with a possible list of permissions
This way all updates can be made inside database transaction including the read and the user or developer can't accidentally remove permissions from the group.

r-peschke (changes requested):
see notes above and
- describe the changes in the wiki for group.update

tests/system/action/group/test_update.py
```diff
@@ -3,44 +3,79 @@
...
         model = self.get_model("group/111")
         assert model.get("name") == "name_srtgb123"
 
-    def test_update_no_permissions(self) -> None:
+    def test_update_forbidden(self) -> None:
         self.base_permission_test(
-            {
-                "group/111": {"name": "name_srtgb123", "meeting_id": 1},
-            },
+            {},
```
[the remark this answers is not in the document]
jsangmeister: These are the basic permission tests that all actions have: Check that the action fails without sufficient permissions and that it succeeds with correct permissions. I just renamed them to not interfere with the new tests, which test the payload field `permissions`.

openslides_backend/action/actions/group/update.py
jsangmeister: I think you misunderstood the goal of these changes. Currently, the client sends a `group.set_permission` for every click in the permission matrix to save the changes directly. This behaviour should be changed so that all changes are only saved after clicking a save button (see the linked issue). Of course, that would be possible with the current action, but would require cumbersome tracking of the changes in the client and calculating the payloads, which is unnecessary to implement and to execute each time. With your proposals, this would still be the case, just with another payload format.

> all updates can be made inside database transaction including the read 

I'm not sure I understand this part or what this has to do with the issue

> the user or developer can't accidentally remove permissions from the group

The whole point of the change is to give the user full control of the permissions. It is unnecessary to put articificial burdens in the workflow. And, if that's a real concern for you, we would need to change all other update actions as well: E.g. the update of a user's groups is done the same way (and almost all other list updates as well), where the client sends the full new list and the backend just validates it and writes it to the datastore.

r-peschke: The [client issue](https://github.com/OpenSlides/openslides-client/issues/1925) explicitly demands tracking the changes and not send the whole matrix, that's done the best way with proposal 1.

> This way all updates can be made inside database transaction including the read and the user or developer can't accidentally remove permissions from the group.

Conflicts can be detected by transaction logic, if another user changed permissions of the group. Your way this will be silently overwritten.

r-peschke (changes requested):
Show the unresolved conversation. In the client issue you demand to track the changes. That's what I also want you to do.

jsangmeister:
That was before I came to the conclusion to also change the backend. The intention was to reduce the payload size and to not send a payload entry for every permission, but only for the changed ones - this approach does it far better and only needs one payload item per group. Do you really want to start this discussion all over again?

r-peschke (approved):
Please change the client issue, if you don't realized the change tracking

jsangmeister:
Waiting for https://github.com/OpenSlides/openslides-client/pull/1985

emanuelschuetze (approved):
tested in combination with https://github.com/OpenSlides/openslides-client/pull/1985

PR #1615, opened 2023-02-01
Add poll_candidate and poll_candidate_list.

reiterl:
Add models and actions and tests.
Update option.delete and poll delete test.

### DIFF
--- a/openslides_backend/action/actions/poll_candidate/create.py
+++ b/openslides_backend/action/actions/poll_candidate/create.py
@@ -1,14 +1,12 @@
 from ....models.models import PollCandidate
-from ...mixins.create_action_with_inferred_meeting import (
-    CreateActionWithInferredMeeting,
-)
+from ...generics.create import CreateAction
 from ...util.action_type import ActionType
 from ...util.default_schema import DefaultSchema
 from ...util.register import register_action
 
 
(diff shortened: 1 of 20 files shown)


jsangmeister (changes requested):
Looks really good!

openslides_backend/action/actions/poll_candidate/create.py
```diff
@@ -1,21 +1,23 @@
...
-)
+from ...generics.create import CreateAction
 from ...util.action_type import ActionType
 from ...util.default_schema import DefaultSchema
 from ...util.register import register_action
 
 
 @register_action("poll_candidate.create", action_type=ActionType.BACKEND_INTERNAL)
-class PollCandidateCreate(CreateActionWithInferredMeeting):
+class PollCandidateCreate(CreateAction):
```
jsangmeister, line 9: Why did you change the base classes for the create actions? I think it was correct before

r-peschke: Because it is not necessary to get the meeting_id from parent etc., it is set in payload.

jsangmeister: But this way, it's redundant information. We get the meeting id from the parent in all other scenarios, so we should keep it this way.

jsangmeister (approved):

PR #1618, opened 2023-02-06
Improve email error wording

jsangmeister:


openslides_backend/action/actions/user/send_invitation_email.py
```diff
@@ -148,31 +148,27 @@ def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
...
         from_email: Union[str, Address]
         if users_email_sender := mail_data.get("users_email_sender", "").strip():
             blacklist = ("[", "]", "\\")
             if any(x in users_email_sender for x in blacklist):
-                result["message"] = (
-                    f'Invalid characters in the sender name configuration of meeting_id "{meeting_id}". Not allowed chars: "'
-                    + '", "'.join(blacklist)
-                    + '"'
-                )
+                result[
```
r-peschke, line 169: Why you want to change this message? IMO the existing one is better, because more expressive

jsangmeister: ok, I'll change it back to integrate the meeting name and the forbidden characters. I had the client in mind which has trouble translating strings with variables in it, but with https://github.com/OpenSlides/openslides-backend/pull/1502 that should no longer be a problem

r-peschke (approved):

PR #1622, opened 2023-02-06
Add migration to remove user projections

jsangmeister:


dev/docker-compose.dev.yml
```diff
@@ -25,6 +25,7 @@ services:
             - AUTH_HOST=auth
             - MESSAGE_BUS_HOST=redis
             - DATASTORE_DATABASE_HOST=postgres
+            - DATASTORE_LOG_LEVEL=CRITICAL
```
jsangmeister, line 28: Remove uninteresting log ouput of the datastore reader when running tests with the `-s` flag

global/data/example-data.json
```diff
@@ -1740,7 +1740,6 @@
             "weight": 6,
             "css_class": "lightblue",
             "set_number": false,
-            "allow_support": true,
```
jsangmeister, line 1743: Some duplicate keys in the example data

openslides_backend/shared/interfaces/event.py
```diff
@@ -9,6 +9,9 @@ class EventType(str, Enum):
     Update = "update"
     Delete = "delete"
 
+    def __repr__(self) -> str:
+        return repr(self.value)
```
jsangmeister, line 13: For easier copying of the printed write requests - should not interfere with anything

jsangmeister:
test with https://github.com/OpenSlides/openslides-datastore-service/pull/229

jsangmeister:
datastore commit has to be updated before merging

tests/system/migrations/conftest.py
```diff
@@ -144,15 +152,18 @@ def _read_model(fqid, position=None):
...
-            # get max position
-            read_database: ReadDatabase = injector.get(ReadDatabase)
-            with read_database.get_context():
-                position = read_database.get_max_position()
-
-        # build model and check
-        assert read_model(fqid, position=position) == expected
+        try:
+            if position is None:
+                assert read_model(fqid) == expected
```
r-peschke, line 157: Sorry, but what is "expected"?

jsangmeister: The expected value of the model, see a few lines higher. What do you wish to be changed here? I could introduce a type annotation if that's what you mean, but we don't have any of them in any other migration tests. I could also add an explaining comment, but I think the name and usage is verbose enough to be self-explanatory.

r-peschke (approved):

r-peschke (changes requested):
After rebasing to the current there is an error:
E           AssertionError: assert {'meta_delete... ['motion/1']} == {'state_exten... ['motion/1']}
E             Omitting 2 identical items, use -vv to show
E             Left contains 2 more items:
E             {'meta_deleted': False, 'meta_position': 2}
The error occurs in 0035_fix_motion_extension_fields.py and not in this one 0037...!

emanuelschuetze:
This PR is important for a new update.

jsangmeister:
@r-peschke you missed some changes in the `assert_model` method while rebasing, which is why the tests failed. I fixed it and also cleaned up the redundant setting of the dev env var.

r-peschke (approved):

PR #1627, opened 2023-02-09
Update dependencies

jsangmeister:
Datastore had to be updated and needed some newer packages as dependencies, so I used the opportunity to update all backend dependencies.

Changes:
- `black` introduced some formatting changes, mostyl removal of empty lines
- `mypy' requires abstract methods now to be explicitely marked as abstract
- `mypy` changed default of `no_implicit_optional`, so I set it explicitely to `false` - although we could talk about leaving it at `true`
- `PyPDF2` was discontinued and renamed back to `pypdf`, so I migrated accordingly
- `bleach` was updated and required some migration
- removed obsolete migration requirements file
- ordered requirement files by category and then alphabetically
- updated datastore to fix failing test
- updated auth commit hash, although there should not be any changes from that


setup.cfg
```diff
@@ -12,6 +12,7 @@ extend-ignore = E203,E501
 disallow_untyped_defs = true
 exclude = tests/system/migrations
 namespace_packages = true
+no_implicit_optional = false
```
r-peschke, line 15: didn't found that option

r-peschke:
> Datastore had to be updated and needed some newer packages as dependencies, so I used the opportunity to update all backend dependencies.
> 
> Changes:
> 
>     * `black` introduced some formatting changes, mostyl removal of empty lines
> 
>     * `mypy' requires abstract methods now to be explicitely marked as abstract
> 
>     * `mypy` changed default of `no_implicit_optional`, so I set it explicitely to `false` - although we could talk about leaving it at `true`

I would prefer to use `strict_optional` with its default=True. `ìmplicit_optional` may be nice to set to True, but we also could leave on it's default False. In [documentation](https://mypy.readthedocs.io/en/stable/config_file.html) I can't find `no_implicit_optional`.
 
> 
>     * `PyPDF2` was discontinued and renamed back to `pypdf`, so I migrated accordingly
> 
>     * `bleach` was updated and required some migration
> 
>     * removed obsolete migration requirements file
> 
>     * ordered requirement files by category and then alphabetically
> 
>     * updated datastore to fix failing test
> 
>     * updated auth commit hash, although there should not be any changes from that

setup.cfg
jsangmeister: Funny that it still works, seems like it is just deprecated. But I will change our code base to use strict optional then.

jsangmeister:
@r-peschke it seems that the newest `aiostmpd` has resolved whatever issues we've had with it - at least the tests now run through without using your fork.

r-peschke (approved):

r-peschke:
> @r-peschke it seems that the newest `aiostmpd` has resolved whatever issues we've had with it - at least the tests now run through without using your fork.

very good:-)

PR #1634, opened 2023-02-13
Add body to repository dispatch to autoupdate

jsangmeister:


ostcar:
It looks like the link to the commit. Would it be possible to link the PR instead?

jsangmeister:
> It looks like the link to the commit. Would it be possible to link the PR instead?

Not easily. This workflow in the backend is triggered on push onto the main branch, so it only triggers when changes were actually merged, so the reference to the PR only accessible indirectly via the commit, and not in a standard, uniform way. I could go ahead and try parsing the commit message for the PR, but I don't think that's worth it. Instead, the PR is one additional click away, since the commit has a link to the PR in the GitHub interface, which I think suffices.

jsangmeister:
Merging it now to fix the workflow in the AU, which breaks as soon as it tries to access the non-existing body

PR #1641, opened 2023-02-16
Fix user projection migration

jsangmeister:
Two errors in this migration:
- some projections did not have a content object
- more tricky: on imports, the create event for the meeting appears before the create event for the projection. This lead to the removal of the projection from the meeting's `all_projection_ids` relation. I reordered the events to process the projection create events first to circumvent this error.


r-peschke (approved):

peb-adr (approved):
Fixed my issue :^)

PR #1643, opened 2023-02-20
Add migration for rm user default projector.

reiterl:
Add test.

### DIFF
--- a/openslides_backend/migrations/migrations/0038_remove_user_default_projector.py
+++ b/openslides_backend/migrations/migrations/0038_remove_user_default_projector.py
@@ -29,8 +29,10 @@ class Migration(BaseMigration):
 
     def remove_replacement(self, obj: Dict[str, Any]) -> None:
         for field in ("default_projector_$_ids", "used_as_default_$_in_meeting_id"):
-            if obj.get(field):
-                obj.pop(field, None)
+            if obj.get(field) and "user" in obj[field]:
+                obj[field].remove("user")
+                if not obj[field]:
+                    del obj[field]
 
     def migrate_event(
         self,
--- a/openslides_backend/migrations/migrations/0038_remove_user_default_projector.py
+++ b/openslides_backend/migrations/migrations/0038_remove_user_default_projector.py
@@ -1,17 +1,18 @@
 from typing import Any, Dict, List, Optional
 
-from datastore.migrations.core.base_migration import BaseMigration
 from datastore.migrations.core.events import (
     BaseEvent,
     CreateEvent,
-    DeleteFieldsEvent,
     ListUpdateEvent,
     UpdateEvent,
+)
+from datastore.migrations.migrations.remove_fields_migration import (
+    RemoveFieldsMigration,
 )
 from datastore.shared.util import collection_from_fqid
 
 
-class Migration(BaseMigration):
+class Migration(RemoveFieldsMigration):
     """
     This migration removes the field `meeting/default_projector_$user_ids`.
     """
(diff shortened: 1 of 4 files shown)


openslides_backend/migrations/migrations/0038_remove_user_default_projector.py
```diff
@@ -0,0 +1,10 @@
+from datastore.migrations import RemoveFieldsMigration
+
+
+class Migration(RemoveFieldsMigration):
+    """
+    This migration removes the field `meeting/default_projector_$user_ids`.
```
jsangmeister, line 6: Ah, I forgot: `user` must also be removed from `meeting/default_projector_$_ids`, if present. Please extend this migration by overwriting the `migrate` method.

jsangmeister (changes requested):

openslides_backend/migrations/migrations/0038_remove_user_default_projector.py
reiterl: Done.

reiterl:
The CI makes problems. Can someone look at it?

reiterl:
Ubunutu 18.04 is deprecated and will be removed on 1st April 2023 . Today is a brownout day. Error msg:
Build and test production docker image
This is a scheduled Ubuntu-18.04 brownout. The Ubuntu-18.04 environment is deprecated and will be removed on April 1st, 2023. For more details, see https://github.com/actions/runner-images/issues/6002

openslides_backend/migrations/migrations/0038_remove_user_default_projector.py
```diff
@@ -0,0 +1,60 @@
...
+        for field in ("default_projector_$_ids", "used_as_default_$_in_meeting_id"):
+            if obj.get(field) and "user" in obj[field]:
+                obj[field].remove("user")
+                if not obj[field]:
+                    del obj[field]
+
+    def migrate_event(
+        self,
+        event: BaseEvent,
+    ) -> Optional[List[BaseEvent]]:
```
jsangmeister, line 40: You can inherit from `RemoveFieldsMigration` and call `super().migrate_event(event)` here. Then, assert that the return value only contains one event and execute your `remove_replacement` code on that. This removes the code duplication.

openslides_backend/migrations/migrations/0038_remove_user_default_projector.py
```diff
@@ -0,0 +1,60 @@
...
+        "projector": ["used_as_default_$user_in_meeting_id"],
+    }
+
+    def remove_fields(self, obj: Dict[str, Any], fields: List[str]) -> None:
+        for field in fields:
+            if field in obj:
+                del obj[field]
+
+    def remove_replacement(self, obj: Dict[str, Any]) -> None:
+        for field in ("default_projector_$_ids", "used_as_default_$_in_meeting_id"):
```
jsangmeister, line 31: You can pass the collection to this method adn then iterate over `collection_fields_map[collection] instead. Removes unnecessary iterations and duplication fo the field names.

jsangmeister (changes requested):

openslides_backend/migrations/migrations/0038_remove_user_default_projector.py
reiterl: Nope, the fields in collection_fields_map are different.

openslides_backend/migrations/migrations/0038_remove_user_default_projector.py
reiterl: Done.

openslides_backend/migrations/migrations/0038_remove_user_default_projector.py
reiterl: I used a collection_replacement_fields_map for the other fields.

openslides_backend/migrations/migrations/0038_remove_user_default_projector.py
```diff
@@ -38,23 +38,17 @@ def migrate_event(
         self,
         event: BaseEvent,
     ) -> Optional[List[BaseEvent]]:
+        events = super().migrate_event(event)
+        if events is None:
+            return events
```
jsangmeister, line 43: This should never happen, but you should nevertheless still use the old event in this case to fully execute the migration.

jsangmeister (changes requested):

openslides_backend/migrations/migrations/0038_remove_user_default_projector.py
jsangmeister: Ah right, that's fine then.

openslides_backend/migrations/migrations/0038_remove_user_default_projector.py
reiterl: As far as I understand the code of RemoveFieldsMigration this case can happen, if collection doesn't match with one collection in the dict.

reiterl: Then I want to stop processing anymore.

jsangmeister: Right! My bad

jsangmeister (approved):

PR #1653, opened 2023-02-27
Rename last_email_send and default_100_percent.

reiterl:


r-peschke (changes requested):
Expand the test_all_migrations-test for at minimum 1 field rename-action. There is no one at the moment.

We should think about if we want to resolve such  unimportant changes in this phase of the project or wait until we have to restructure more in database.
For this one we have to migrate the database and **can merge this one only in 
synchronization with the client, autoupdate-service and I guess the vote-service**.

jsangmeister:
> We should think about if we want to resolve such unimportant changes in this phase of the project or wait until we have to restructure more in database.

I agree, which is why I tagged it as low - I just wanted to note that this is not the way it should be.

> For this one we have to migrate the database and **can merge this one only in synchronization with the client, autoupdate-service and I guess the vote-service**.

The vote service should not require any changes since the `default_*` fields are only evaluated in the client. The AU changes can be generated automatically. But yes, you are right, which I also noted in the original issue.

reiterl:
We want to do that *later*, because it is not important and includes a *migration*. Talked with @rrenkert and @r-peschke  .

jsangmeister:
We could think about merging this now since we will have a migration anyway with #1502. The changes in the client will only be some search-replaces, the autoupdate changes will be generated automatically and the vote service does not need changes since it does not fetch the default values and only these were changed. Or we close this completely and do it during/after the template fields removal - but since the code here is done, I think we could just merge it. What do you think @rrenkert @r-peschke?

rrenkert:
> We could think about merging this now since we will have a migration anyway with #1502.
Yes we can merge this.

> The changes in the client will only be some search-replaces
@jsangmeister Can you do this and reference this here?

r-peschke:
@rrenkert @jsangmeister  @reiterl :My review from 6.3.23 is still open: Please extend the meeting/test_import.py test_all_migrations test with data for renaming one field., there is no.

jsangmeister:
> @jsangmeister Can you do this and reference this here?

https://github.com/OpenSlides/openslides-client/pull/2245

reiterl:
Add the missing test data in test_all_migrations and rebase on current main.

r-peschke (approved):
Don't merge! This can only be merged together with clients https://github.com/OpenSlides/openslides-client/pull/2245

PR #1655, opened 2023-02-28
Improve action_worker logging, add action_names to log msgs.

reiterl:


m-schieder (changes requested):
Thanks for the changes. I am still missing the fqid in this log:
`Request was successful. Send response now.`

I think this is created [here](https://github.com/OpenSlides/openslides-backend/blob/768af7b01d59f4d9e17d98fe537f934be267edc7/openslides_backend/action/action_handler.py#L139).

reiterl:
It should not be needed there, if you enable DEBUG mode. As far as I understand the action worker code, it gives a debug message with the fqid and the action names in positive case. Would that be enough? Or should I change the MAIN info log message?

m-schieder:
I am sorry that I did not describe clearly what I wanted.

We need the logs for the analysis on the productive systems. There we will not activate the DEBUG log. Therefore this must be an INFO log.

Also, I don't want to have unnecessary additional log lines. That's why I would suggest to use this [logline](https://github.com/OpenSlides/openslides-backend/blob/768af7b01d59f4d9e17d98fe537f934be267edc7/openslides_backend/action/action_handler.py#L139).

reiterl:
What fqid are you looking for? The fqid of the action worker? The inner `ActionHandler` doesn't have a way to access the `ActionWorker` . I don't see a way to get the fqid there.
I could update a debug msg in the action worker to `ìnfo`. So that the information should be given.

m-schieder (approved):

m-schieder:
The solution is okay for me. It won't generate that much more logs. I have tested this. But did not look at the code.

r-peschke:
> I am sorry that I did not describe clearly what I wanted.
> 
> We need the logs for the analysis on the productive systems. There we will not activate the DEBUG log. Therefore this must be an INFO log.
> 
> Also, I don't want to have unnecessary additional log lines. That's why I would suggest to use this [logline](https://github.com/OpenSlides/openslides-backend/blob/768af7b01d59f4d9e17d98fe537f934be267edc7/openslides_backend/action/action_handler.py#L139).

In most cases there is no fqid for the action worker, because this will only be created, if it is a long lasting ation (> 1sec.)

PR #1658, opened 2023-03-02
Allow iframe and add sandbox to it. Update tests.

reiterl:
Update HTML Permissive fields (meeting/welcome_text, topic/text). Allow a way to insert iframes to have videos there.
Update the tests.

### DIFF
--- a/tests/system/action/meeting/test_settings.py
+++ b/tests/system/action/meeting/test_settings.py
@@ -34,4 +34,7 @@ class MeetingSettingsSystemTest(BaseActionTestCase):
         response = self.request("meeting.update", {"id": 1, "welcome_text": "<iframe>"})
         self.assert_status_code(response, 200)
         meeting = self.get_model("meeting/1")
-        self.assertEqual(meeting["welcome_text"], "&lt;iframe&gt;")
+        self.assertEqual(
+            meeting["welcome_text"],
+            '<iframe sandbox="allow-scripts allow-same-origin"></iframe>',
+        )
--- a/tests/system/action/meeting/test_settings.py
+++ b/tests/system/action/meeting/test_settings.py
@@ -31,12 +31,32 @@ class MeetingSettingsSystemTest(BaseActionTestCase):
         self.create_model(
             "meeting/1", {"welcome_text": "Hi", "is_active_in_organization_id": 1}
         )
-        response = self.request("meeting.update", {"id": 1, "welcome_text": "<iframe>"})
+        response = self.request(
+            "meeting.update", {"id": 1, "welcome_text": '<iframe allow="yes">'}
+        )
         self.assert_status_code(response, 200)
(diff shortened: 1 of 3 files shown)


bastianjoel:
I'm not completely sure if it is a good idea to not restrict some attributes. For example we might not want `allowfullscreen` and `allow` as they enable the iframe content to swtich into full screen mode.
I've also noticed that the referrer is sent to the iframes source. If this is a problem it could be omitted by adding `referrerpolicy="no-referrer"`. 

@rrenkert @emanuelschuetze maybe you could give some input on how restricted that feature can and should be.

r-peschke (approved):

tests/system/action/meeting/test_settings.py
```diff
@@ -34,4 +34,7 @@ def test_html_field(self) -> None:
         response = self.request("meeting.update", {"id": 1, "welcome_text": "<iframe>"})
         self.assert_status_code(response, 200)
         meeting = self.get_model("meeting/1")
-        self.assertEqual(meeting["welcome_text"], "&lt;iframe&gt;")
```
jsangmeister, line 37: Please find a tag which is still forbidden and add a test to check that forbidden tags are escaped correctly.

jsangmeister (changes requested):

jsangmeister:
@bastianjoel I think these are good suggestions. Maybe also forbid `allowpaymentrequest`, while we're at it?

tests/system/action/meeting/test_settings.py
reiterl: I used the "script" tag to test the escaping. Add this new test and rename the tests.

jsangmeister:
@rrenkert some input from you would be appreciated

rrenkert:
Yes, I think we should forbid the fields `allow`, `allowfullscreen` and `allowpaymentrequest`. And it is a good idea to set the  `referrerpolicy="no-referrer"` attribute.

bastianjoel:
When checking dependency updates for the client I stumbled over a security patch for angular regarding iframes [here](https://github.com/angular/angular/commit/54814c8e9ba6b82a7f4748ef2b2f47507efd758b#diff-14140d5e083fd72d69d738ee8c5161666cd7af0dc568e7bc7ad93ccaa477b67d). 

They forbid `sandbox`, `allow`, `allowfullscreen`, `referrerpolicy`, `csp` and `fetchpriority`. 

**So to sum this up:**
`referrerpolicy="no-referrer"` and `sandbox="allow-scripts allow-same-origin"` should be always set and not changeable. 
`allow`, `allowfullscreen`, `allowpaymentrequest`, `csp` and `fetchpriority` should not be settable by the user.

reiterl:
@jsangmeister  Can you review this?

jsangmeister (approved):

PR #1664, opened 2023-03-09
I1572 motion create forwarded2

r-peschke:
now works with forwarding multiple motions in one transaction


reiterl (approved):
It should be okay.

PR #1669, opened 2023-03-14
Fix user update permissions

### LINKED ISSUE #1668: An organization admin can demote a superadmin.
An organization admin can edit the organization_management_level of superadmins. If he does this, the superadmin loses his superadmin rights. I think in general no organization admin should be allowed to edit the organization_management_level of a superadmin.

jsangmeister:
fixes #1668


jsangmeister:
@r-peschke feel free to merge if you have no comments. We should include this ASAP.

r-peschke (approved):
I don't like the solution very much, building the user scope without applied instance, but I don't see a better solution at the moment.

PR #1670, opened 2023-03-14
Fix schema usages

jsangmeister:
follow-up to https://github.com/OpenSlides/openslides-backend/pull/1669: The error in the test for the `get_user_scope` presenter happened (partially) because we do not use our exsting schemas enough, so I replaced all places I could find with our predefined schemas. Also, I'm very confused that `fastjsonschema` does not throw an error in case of invalid keywords (like `item` instead of `items`) - if anyone knows if there exists a simple solution for that, I would love to fix this.

### DIFF
--- a/openslides_backend/presenter/get_users.py
+++ b/openslides_backend/presenter/get_users.py
@@ -11,30 +11,6 @@ from ..shared.schema import schema_version
 from .base import BasePresenter
 from .presenter import register_presenter
 
-get_users_schema = fastjsonschema.compile(
-    {
-        "$schema": schema_version,
-        "type": "object",
-        "title": "get_users",
-        "description": "get user ids data",
-        "properties": {
-            "start_index": {"type": "integer"},
-            "entries": {"type": "integer"},
-            "sort_criteria": {
-                "type": "array",
-                "item": {
-                    "type": "string",
-                    "enum": ["username", "first_name", "last_name"],
-                },
-            },
-            "reverse": {"type": "boolean"},
-            "filter": {"type": ["string", "null"]},
-        },
-        "required": [],
-        "additionalProperties": False,
-    }
-)
-
 # The values are the default values for None.
 ALLOWED = {
     "first_name": "",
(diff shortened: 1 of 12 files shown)


openslides_backend/presenter/get_users.py
```diff
@@ -22,9 +38,9 @@
             "entries": {"type": "integer"},
             "sort_criteria": {
                 "type": "array",
-                "item": {
+                "items": {
                     "type": "string",
-                    "enum": ["username", "first_name", "last_name"],
+                    "enum": list(ALLOWED.keys()),
```
jsangmeister, line 43: Simple fix since the enum beforehand was incorrect - there even exists a test which tests that sorting by `title` works, the enum just wasn't enforced. I'm open to improving this (e.g. removing the now-redundant check of the sort criteria below), but since the presenter is not used AFAIK, I would not put any work into this. We could even consider removing it altogethere.

tests/system/presenter/test_get_user_scope.py
```diff
@@ -84,9 +84,7 @@ def test_good(self) -> None:
     def test_without_user_None(self) -> None:
         status_code, data = self.request("get_user_scope", {"user_ids": [None]})
         self.assertEqual(status_code, 400)
-        self.assertIn(
-            "There is no user_id given to get the user_scope!", data["message"]
-        )
+        self.assertIn("data.user_ids[0] must be integer", data["message"])
```
r-peschke, line 87: I would prefer the old message. It is a lot easier to understand and find for anyone

r-peschke (changes requested):

tests/system/presenter/test_get_user_scope.py
jsangmeister: Yes, but this can be said for all the fastjsonschema messages and we use them everywhere. Implementing additional code for a custom error message instead of just using the schema like everywhere else should IMO only be done in very special cases where it's expected that the error message will often be displayed to the user, which is not the case here.

r-peschke (approved):

PR #1674, opened 2023-03-17
implement ListUpdateEvent for instance-recreation in import

r-peschke:


tests/system/action/meeting/test_import.py
```diff
@@ -1845,6 +1845,58 @@ def test_check_missing_admin_group_in_meeting(self) -> None:
...
+                    "speaker_ids": [],
+                    "projection_ids": [],
+                },
+            },
+        })
+        data["meeting"]["meeting"]["1"]["motion_ids"] = [5, 6]
+        data["meeting"]["meeting"]["1"]["list_of_speakers_ids"] = [1, 2]
+        data["meeting"]["motion_state"]["1"]["motion_ids"] = [5, 6]
+        data["meeting"]["_migration_index"] = 35
+        assert(data["meeting"]["motion"]["5"]["referenced_in_motion_state_extension_ids"] == [])
```
jsangmeister: unnecessary assert since you just set the data

jsangmeister (changes requested):
looks good besides one minor remark! can be merged directly afterwards

tests/system/action/meeting/test_import.py
r-peschke: Sure, but very useful to understand the test

jsangmeister: If the test itself is not clear enough (which this one is IMO), comments should be the way to go to describe the behaviour.

Issue #1661, opened 2023-03-08
[imports] Relocate csv imports to the backend: topic


r-peschke:
The steps to go:
1. Client reads the import csv-file and converts it to json format
2. New backend-action topic.json_upload will be called with file content
3. The backend creates data for preview and send the preview data with the response and an action-worker number for starting the import or abort it. The data will be stored in these action-worker instance. Preview data format will be defined by the client, but should be sufficient to do the import. For existing preview-format we need additionally fields per line check-result (ok or not ok) and in case of not ok need an errortext. There also remarks for the whole import and the number of lines. Error checks see above.
4. The client presents the preview and calls the 2nd new action topic.json_import. The payload needs the action_number, where the data is stored and the command "import" or "abort".
5. The backend will import or delete the json-data from specified action-worker instance.

Special themes:
1. we want to use the action worker, because there is no need to create events and other stuff with file content on an extra route.
2. Converting from csv to json in the client should be made without knowing specified formats just from a standard lib. Json format is the standard format of all backend-payloads we will keep.
3. We should think about dangling "import"-action-workers with content to delete.

Error checks:
- Title has to be unique in one agenda
-  Hint: Missing columns will be reported as missing and imported as empty column
-  required fields are define in import
-  @luisa-beerboom : Are there more checks?

Existing import can be seen in meeting under agenda, menu button .

luisa-beerboom:
>     Preview data format will be defined by the client, but should be sufficient to do the import.

I've taken a vague look over the clients `ImportListComponent`, I'm not quite sure on the exact form of the data that the client uses to generate the preview list (although that exact format would of course be the most helpful): But I believe it would look vaguely like this: We need to know:

- per column
    - What type of data is saved here (boolean, number, string, ...)
    - What is the column property
    - What is the column label
- per row
    - What will happen with this row? will it be...
        - imported
        - skipped because it is a duplicate
        - referenced
        - skipped because of an error
- per entry
    - what is the content
    - is there an error/warning with this fields content? Which one?
    - if the column is a reference to another model, will it be generated anew from this field

The client data seems to be mainly made up of two interfaces:
```
interface HeaderDefinition<ToCreate> {
    label: string;
    property: key of ToCreate;
    type?: `boolean` | `string` | `number`;
    isRequired?: boolean;         // Probably not necessary
    isTableColumn?: boolean;  // Probably not necessary
    width?: number;                   // This is the css column width, should maybe be done by the backend
}
```
[...]

PR #1678, opened 2023-03-22
Add topic.json_upload/import

reiterl:
Add some tests.

### DIFF
--- a/openslides_backend/action/actions/topic/import_.py
+++ b/openslides_backend/action/actions/topic/import_.py
@@ -1,5 +1,4 @@
-from time import time
-from typing import Any, Dict
+from typing import Any, Callable, Dict
 
 from ....models.models import ActionWorker
 from ....permissions.permissions import Permissions
@@ -11,6 +10,7 @@ from ....shared.schema import required_id_schema
 from ...action import Action
 from ...util.default_schema import DefaultSchema
 from ...util.register import register_action
+from ...util.typing import ActionData
 from .create import TopicCreate
 from .json_upload import ImportStatus
 from .mixins import DuplicateCheckMixin
@@ -44,24 +44,7 @@ class TopicImport(DuplicateCheckMixin, Action):
             and not self.check_for_duplicate(entry["data"]["title"])
         ]
         self.execute_other_action(TopicCreate, action_payload)
-        timestamp = int(time())
-        self.datastore.write_action_worker(
-            WriteRequest(
-                events=[
-                    Event(
-                        type=EventType.Update,
-                        fqid=fqid_from_collection_and_id("action_worker", store_id),
-                        fields={
-                            "timestamp": timestamp,
-                            "state": "end",
-                        },
-                    )
-                ],
-                user_id=self.user_id,
-                locked_fields={},
-            )
-        )
-        return {}
+        return instance
 
     def handle_relation_updates(self, instance: Dict[str, Any]) -> Any:
         return {}
@@ -77,3 +60,26 @@ class TopicImport(DuplicateCheckMixin, Action):
         if worker.get("result", {}).get("import") == "topic":
             return next(iter(worker["result"]["rows"]))["data"]["meeting_id"]
         raise ActionException("Import data cannot be found.")
+
+    def get_on_success(self, action_data: ActionData) -> Callable[[], None]:
+        def on_success() -> None:
+            for instance in action_data:
+                store_id = instance["id"]
+                self.datastore.write_action_worker(
+                    WriteRequest(
+                        events=[
+                            Event(
+                                type=EventType.Update,
+                                fqid=fqid_from_collection_and_id(
+                                    "action_worker", store_id
+                                ),
+                                fields={
+                                    "result": None,
+                                },
+                            )
+                        ],
+                        user_id=self.user_id,
+                        locked_fields={},
+                    )
+                )
+        return on_success
--- a/openslides_backend/action/actions/topic/import_.py
+++ b/openslides_backend/action/actions/topic/import_.py
@@ -26,30 +26,32 @@ class TopicImport(DuplicateCheckMixin, Action):
     schema = DefaultSchema(ActionWorker()).get_default_schema(
         additional_required_fields={
             "id": required_id_schema,
+            "import": {"type": "boolean"},
         }
     )
     permission = Permissions.AgendaItem.CAN_MANAGE
 
     def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
         store_id = instance["id"]
-        meeting_id = self.get_meeting_id(instance)
-        self.init_duplicate_set(meeting_id)
-        worker = self.datastore.get(
-            fqid_from_collection_and_id("action_worker", store_id),
-            ["result"],
-            lock_result=False,
-        )
-        action_payload = [
-            entry["data"]
-            for entry in worker.get("result", {}).get("rows", [])
-            if (
-                entry["status"] == ImportStatus.NEW
-                or entry["status"] == ImportStatus.ERROR
-                and entry["error"] == ["Duplicate"]
+        if instance["import"]:
+            meeting_id = self.get_meeting_id(instance)
+            self.init_duplicate_set(meeting_id)
+            worker = self.datastore.get(
+                fqid_from_collection_and_id("action_worker", store_id),
+                ["result"],
+                lock_result=False,
             )
-            and not self.check_for_duplicate(entry["data"]["title"])
-        ]
-        self.execute_other_action(TopicCreate, action_payload)
+            action_payload = [
+                entry["data"]
+                for entry in worker.get("result", {}).get("rows", [])
+                if (
+                    entry["status"] == ImportStatus.NEW
+                    or entry["status"] == ImportStatus.ERROR
+                    and entry["error"] == ["Duplicate"]
+                )
+                and not self.check_for_duplicate(entry["data"]["title"])
+            ]
+            self.execute_other_action(TopicCreate, action_payload)
         return instance
 
     def handle_relation_updates(self, instance: Dict[str, Any]) -> Any:
--- a/openslides_backend/action/actions/topic/import_.py
+++ b/openslides_backend/action/actions/topic/import_.py
@@ -1,8 +1,11 @@
+from time import time
 from typing import Any, Dict
 
 from ....models.models import ActionWorker
 from ....permissions.permissions import Permissions
 from ....shared.exceptions import ActionException
+from ....shared.interfaces.event import Event, EventType
+from ....shared.interfaces.write_request import WriteRequest
 from ....shared.patterns import fqid_from_collection_and_id
 from ....shared.schema import required_id_schema
 from ...action import Action
@@ -41,6 +44,23 @@ class TopicImport(DuplicateCheckMixin, Action):
             and not self.check_for_duplicate(entry["data"]["title"])
         ]
         self.execute_other_action(TopicCreate, action_payload)
+        timestamp = int(time())
+        self.datastore.write_action_worker(
+            WriteRequest(
+                events=[
+                    Event(
+                        type=EventType.Update,
+                        fqid=fqid_from_collection_and_id("action_worker", store_id),
+                        fields={
+                            "timestamp": timestamp,
+                            "state": "end",
+                        },
+                    )
+                ],
+                user_id=self.user_id,
+                locked_fields={},
+            )
+        )
         return {}
 
     def handle_relation_updates(self, instance: Dict[str, Any]) -> Any:
--- a/openslides_backend/action/actions/topic/import_.py
+++ b/openslides_backend/action/actions/topic/import_.py
@@ -9,6 +9,7 @@ from ...action import Action
 from ...util.default_schema import DefaultSchema
 from ...util.register import register_action
 from .create import TopicCreate
+from .json_upload import ImportStatus
 from .mixins import DuplicateCheckMixin
 
 
@@ -36,7 +37,7 @@ class TopicImport(DuplicateCheckMixin, Action):
         action_payload = [
             entry["data"]
             for entry in worker.get("result", {}).get("rows", [])
-            if entry["status"] == "new"
+            if entry["status"] == ImportStatus.NEW
             and not self.check_for_duplicate(entry["data"]["title"])
         ]
         self.execute_other_action(TopicCreate, action_payload)
--- a/openslides_backend/action/actions/topic/import_.py
+++ b/openslides_backend/action/actions/topic/import_.py
@@ -3,8 +3,6 @@ from typing import Any, Dict
 from ....models.models import ActionWorker
 from ....permissions.permissions import Permissions
 from ....shared.exceptions import ActionException
-from ....shared.interfaces.event import Event, EventType
-from ....shared.interfaces.write_request import WriteRequest
 from ....shared.patterns import fqid_from_collection_and_id
 from ....shared.schema import required_id_schema
 from ...action import Action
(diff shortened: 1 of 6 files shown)


reiterl:
The cov of the feature/remove-template-fields is too low.

openslides_backend/action/actions/topic/mixins.py
```diff
@@ -0,0 +1,20 @@
+from ....shared.filters import FilterOperator
+from ...action import Action
+
+
+class DuplicateCheckMixin(Action):
+    def init_duplicate_set(self, meeting_id: int) -> None:
+        self.all_titles_in_meeting = set(
+            [
```
jsangmeister, line 8: remove the array brackets

openslides_backend/action/actions/topic/json_upload.py
```diff
@@ -0,0 +1,138 @@
...
+
+    def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
+        data = instance.pop("data")
+
+        # enrich data with meeting_id
+        for entry in data:
+            entry["meeting_id"] = instance["meeting_id"]
+
+        # generate headers
+        self.headers = [
```
jsangmeister, line 63: Since these are constant, define them as a class variable

tests/system/action/topic/test_json_upload.py
```diff
@@ -0,0 +1,137 @@
...
+        assert result["headers"][0] == {
+            "property": "title",
+            "type": "string",
+        }
+        assert result["rows"][0] == {
+            "status": "new",
+            "error": [],
+            "data": {"title": "test", "meeting_id": 22},
+        }
+        assert result["statistics"] == {"itemCount": 1, "Created": 1, "Error": 0}
```
jsangmeister, line 97: Just assert the full result at once to make sure no unwanted data is included

tests/system/action/topic/test_json_upload.py
```diff
@@ -0,0 +1,137 @@
...
+        response = self.request(
+            "topic.json_upload",
+            {
+                "meeting_id": 22,
+                "data": [{"title": "test"}, {"title": "bla"}, {"title": "test"}],
+            },
+        )
+        self.assert_status_code(response, 200)
+        result = response.json["results"][0][0]
+        assert result["rows"][2]["error"] == ["Duplicate"]
```
jsangmeister, line 124: assert that the status is `error`. Also assert that the other 2 entries are imported correctly

jsangmeister (changes requested):

openslides_backend/action/actions/topic/json_upload.py
reiterl: Done.

openslides_backend/action/actions/topic/import_.py
```diff
@@ -0,0 +1,88 @@
...
+                        fields={
+                            "id": store_id,
+                            "result": None,
+                        },
+                    )
+                ],
+                user_id=self.user_id,
+                locked_fields={},
+            )
+        )
```
jsangmeister, line 65: Remove this: if the import fails for some reason, we need to have this data still available.

tests/system/action/topic/test_json_upload.py
reiterl: Done.

tests/system/action/topic/test_json_upload.py
reiterl: Done. Check the status and check the rows in the action worker.

openslides_backend/action/actions/topic/json_upload.py
```diff
@@ -0,0 +1,136 @@
...
+        for entry in self.rows:
+            itemCount += 1
+            if entry["status"] == "new":
+                itemNew += 1
+            if entry["status"] == "error":
+                itemError += 1
+        self.statistics = {
+            "itemCount": itemCount,
+            "Created": itemNew,
+            "Error": itemError,
```
jsangmeister, line 84: Do I understand it correctly that these strings are to be displayed as-is in the client? If so, we should choose more suitable ones.

openslides_backend/action/actions/topic/json_upload.py
```diff
@@ -0,0 +1,138 @@
...
+        fqid = fqid_from_collection_and_id("action_worker", self.new_store_id)
+        self.datastore.write_action_worker(
+            WriteRequest(
+                events=[
+                    Event(
+                        type=EventType.Create,
+                        fqid=fqid,
+                        fields={
+                            "id": self.new_store_id,
+                            "result": self.rows,
```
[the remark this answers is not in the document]
r-peschke: It is not necessary, because there is a type-specific action, i.e. topic.import, for this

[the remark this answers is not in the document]
jsangmeister: Yes, and `topic.import` only gets an action worker id. If in the action worker it is saved to which import it belongs, the validation is easier that no invalid id was provided.

openslides_backend/action/actions/topic/import_.py
reiterl: Mmh, this also handles the abort. Should an abort remove the data or not?

openslides_backend/action/actions/topic/import_.py
```diff
@@ -0,0 +1,88 @@
...
+class TopicImport(DuplicateCheckMixin, Action):
+    """
+    Action to import a result from the action_worker.
+    """
+
+    model = ActionWorker()
+    schema = DefaultSchema(ActionWorker()).get_default_schema(
+        additional_required_fields={
+            "id": required_id_schema,
+            "command": {"type": "string", "enum": ["import", "abort"]},
```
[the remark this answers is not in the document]
reiterl: Okay, removed the command and the abort.

openslides_backend/action/actions/topic/import_.py
reiterl: As I see in the later comments, we want to remove abort.

openslides_backend/action/actions/topic/import_.py
```diff
@@ -0,0 +1,88 @@
...
+        if command == "import":
+            meeting_id = self.get_meeting_id(instance)
+            self.init_duplicate_set(meeting_id)
+            worker = self.datastore.get(
+                fqid_from_collection_and_id("action_worker", store_id), ["result"]
+            )
+            action_payload = [
+                entry["data"]
+                for entry in worker.get("result", [])
+                if entry["status"] == "new"
```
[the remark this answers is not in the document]
reiterl: That will lead to such errors:
TypeError: Object of type ImportStatus is not JSON serializable

openslides_backend/action/actions/topic/import_.py
```diff
@@ -0,0 +1,88 @@
...
+        )
+        if worker.get("result") and isinstance(worker["result"], list):
+            meeting_id = None
+            okay = True
+            for entry in worker["result"]:
+                if "status" not in entry or "error" not in entry or "data" not in entry:
+                    okay = False
+                meeting_id = entry["data"].get("meeting_id")
+            if okay and meeting_id:
+                return meeting_id
```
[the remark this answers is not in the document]
reiterl: Updated.

openslides_backend/action/actions/topic/import_.py
```diff
@@ -0,0 +1,88 @@
...
+        raise ActionException("Topic import is aborted or done.")
```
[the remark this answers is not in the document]
reiterl: Use a "import": "topic" flag, to see, it's data are right.

openslides_backend/action/actions/topic/json_upload.py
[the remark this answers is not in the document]
reiterl: I think it is a good idea, to include an import: "topic" in the action_worker, so that the import action can look, that has correct data. I include it.

openslides_backend/action/actions/topic/import_.py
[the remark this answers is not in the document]
reiterl: Add an enum `ImportStatus` and use it in the actions and the tests.

openslides_backend/action/actions/topic/json_upload.py
reiterl: Use Total/New/Error now. Or do you have a better idea?

openslides_backend/action/actions/topic/import_.py
r-peschke: > Remove this: if the import fails for some reason, we need to have this data still available.

IMO we should remove the data, if the import is successful. The topic import never will be large, but others will.

openslides_backend/action/actions/topic/json_upload.py
r-peschke: Use Luisa's, see https://github.com/OpenSlides/openslides-backend/issues/1661#issuecomment-1466109237, like
* total: 10
* created: 5
* omitted: 5

tests/system/action/topic/test_json_upload.py
```diff
@@ -0,0 +1,173 @@
...
+    def test_json_upload_results(self) -> None:
+        response = self.request(
+            "topic.json_upload",
+            {"meeting_id": 22, "data": [{"title": "test"}]},
+        )
+        self.assert_status_code(response, 200)
+        self.assert_model_exists(
+            "action_worker/1",
+            {
+                "result": {
```
r-peschke, line 71: Missing the data for headers and statistics

reiterl: Why that? headers and statistics are not needed in the action worker.

reiterl: I don't think, that we should include headers and statistics in the action_worker, they are just wasted load. The topic import action doesn't need them. It just needs the `rows`.

openslides_backend/action/actions/topic/import_.py
reiterl: @jsangmeister seems to have another opinion on this. This is a contradiction. @r-peschke  pls discuss this with @jsangmeister .

reiterl:
Open are:

- Discussion of the overwriting of the result in the action_worker after import.
- Discussion of the headers/statistics in action worker results.

r-peschke:
Missing:
- we should have a field `type` in the action_worker-model to differentiate between imports and real action_workers
- we should fill the `created` and `timestamp` fields on json_upload and update the `timestamp`field on import (or abort, if exists)
- on import or abort set the `state`to ènd` 
- setting state and actions still depends the open discussion points above

openslides_backend/action/actions/topic/import_.py
jsangmeister: Yes, but we already discussed that a regular (e.g. daily) cleanup of all "old" import data (e.g. older than two weeks) would be the way to go here. The problem is, if we delete it during the action and the datastore request fails because of locked fields or something else, we cannot retry it. A better solution would be to include it in the `on_success` routine, but even that is not perfect because if something else goes wrong during the `import` request, the user has to regenerate the preview first before he can import again. So, for a better UX, we should keep the data here.

openslides_backend/action/actions/topic/json_upload.py
```diff
@@ -0,0 +1,143 @@
...
+        fqid = fqid_from_collection_and_id("action_worker", self.new_store_id)
+        self.datastore.write_action_worker(
+            WriteRequest(
+                events=[
+                    Event(
+                        type=EventType.Create,
+                        fqid=fqid,
+                        fields={
+                            "id": self.new_store_id,
+                            "result": {"import": "topic", "rows": self.rows},
```
r-peschke, line 105: set the `timestamp` field of action worker to current timestamp and the status-field to `running

r-peschke (changes requested):

openslides_backend/action/actions/topic/json_upload.py
reiterl: Add `created`, `timestamp` and `state` running to the json_upload and `timestamp` and `state` to import_
I rebased the PR.

r-peschke (changes requested):

openslides_backend/action/actions/topic/mixins.py
```diff
@@ -0,0 +1,18 @@
...
+        self.all_titles_in_meeting = set(
+            values.get("title")
+            for values in self.datastore.filter(
+                "topic", FilterOperator("meeting_id", "=", meeting_id), ["title"]
+            ).values()
+            if values.get("title")
+        )
+
+    def check_for_duplicate(self, title: str) -> bool:
+        result = title in self.all_titles_in_meeting
```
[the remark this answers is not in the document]
reiterl: I think, your code is broken. In all titles in a meeting should be every title, so I can check for it.

openslides_backend/action/actions/topic/mixins.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+class DuplicateCheckMixin(Action):
+    def init_duplicate_set(self, meeting_id: int) -> None:
+        self.all_titles_in_meeting = set(
+            values.get("title")
+            for values in self.datastore.filter(
+                "topic", FilterOperator("meeting_id", "=", meeting_id), ["title"]
+            ).values()
+            if values.get("title")
```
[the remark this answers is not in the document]
reiterl: Okay, removed.

openslides_backend/action/actions/topic/mixins.py
```diff
@@ -0,0 +1,18 @@
+from ....shared.filters import FilterOperator
+from ...action import Action
+
+
+class DuplicateCheckMixin(Action):
+    def init_duplicate_set(self, meeting_id: int) -> None:
+        self.all_titles_in_meeting = set(
```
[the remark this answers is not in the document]
reiterl: I think a set is the right datastruct here.

openslides_backend/action/actions/topic/json_upload.py
```diff
@@ -0,0 +1,148 @@
...
+            entry["meeting_id"] = instance["meeting_id"]
+
+        # validate and check for duplicates
+        self.init_duplicate_set(instance["meeting_id"])
+        self.rows = [self.validate_entry(entry) for entry in data]
+
+        # generate statistics
+        itemCount, itemNew, itemError = 0, 0, 0
+        for entry in self.rows:
+            itemCount += 1
```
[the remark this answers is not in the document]
reiterl: Yes, that would be possible. I like more my way.

openslides_backend/action/actions/topic/import_.py
```diff
@@ -0,0 +1,96 @@
...
+
+    def get_on_success(self, action_data: ActionData) -> Callable[[], None]:
+        def on_success() -> None:
+            for instance in action_data:
+                store_id = instance["id"]
+                self.datastore.write_action_worker(
+                    WriteRequest(
+                        events=[
+                            Event(
+                                type=EventType.Update,
```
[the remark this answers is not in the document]
reiterl: Done.

openslides_backend/action/actions/topic/json_upload.py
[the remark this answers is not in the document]
reiterl: Done. I use `len` to calculate itemCount.

openslides_backend/action/actions/topic/mixins.py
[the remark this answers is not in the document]
reiterl: I do *add* and *in* operations at the data structure.  *In* should be faster in a set and *add* is the same.

openslides_backend/action/actions/topic/mixins.py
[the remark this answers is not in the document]
reiterl: Remember `result` is True, if a title is in the all_title_meeting. Then I don't need to add it., but I need to add it if it is **not** in all_titles_in_meeting.

r-peschke (changes requested):

reiterl:
Okay, add the title only if result is false. It is senseless.

r-peschke (approved):

openslides_backend/action/actions/topic/import_.py
```diff
@@ -0,0 +1,93 @@
...
+                ["result"],
+                lock_result=False,
+            )
+            action_payload = [
+                entry["data"]
+                for entry in worker.get("result", {}).get("rows", [])
+                if (
+                    entry["status"] == ImportStatus.NEW
+                    or entry["status"] == ImportStatus.ERROR
+                    and entry["error"] == ["Duplicate"]
```
jsangmeister, line 50: Don't you import the duplicates this way too? Is this intended?

tests/system/action/topic/test_import.py
jsangmeister, line 1: I'm missing a test here with duplicates in the json upload

jsangmeister (changes requested):

tests/system/action/topic/test_import.py
reiterl: I have added such a test.

jsangmeister (approved):

openslides_backend/action/actions/topic/import_.py
jsangmeister: okay

Issue #1649, opened 2023-02-23
[remove-template-fields] Some required changes


jsangmeister:
- [x] The `meeting_user` should bundle all information about the user in this meeting, so the following fields should also be moved to there, since they are meeting-specific: (The first 3 were previously also template fields, the latter should have been one, if we were not in the process of removing them) -> leaving them in `user`
  - `poll_voted_ids`
  - `option_ids`
  - `vote_ids`
  - `poll_candidate_ids`
- [x] The test method `create_user_for_meeting` should be adjusted in the tests to automatically create a meeting user. Update tests which use it. (made via set_user_groups #1563)
- [x] The checks removed from the `user` methods (e.g. `check_meeting_and_users` from the `user_mixin`) must be added to the `meeting_user`.
- [x] The fields `projector/used_as_default_*_in_meeting_id` should be renamed to `used_as_default_projector_*_in_meeting_id` or, even more verbose, `used_as_default_projector_for_*_in_meeting_id` to better reflect their meaning (e.g. `used_as_default_polls_in_meeting_id` confuses at first sight in the projector model - how can a projector be a default poll?) 
- [x] Rename `meeting_user/submitted_motion_ids` to `motion_submitter_ids` to better reflect the meaning.
- [x] Remove redundant `required: false` from the `models.yml` - although this may also be done in https://github.com/OpenSlides/openslides-backend/issues/1648

jsangmeister:
@r-peschke any opinions?

r-peschke:
> @r-peschke any opinions?

Yes, but in the first phase we just want to remove the template fields

jsangmeister:
Yes, but I would argue that these changes are all part of it. Let's talk about it.

jsangmeister:
> This fields and also user.vote_delegated_vote_ids (renamed to user.delegated_vote_ids) should remain in user-collection, because they handle the user-object on the other side of the relation. Luisa confirmed that they are not used in the client/autoupdate-service.

The reverse relations are meeting-specific, so these fields belong in the `meeting_user` collection. All fields which were previously template fields (or should have been, in the case of `poll_candidate_ids`) should be moved to the meeting user. This has nothing to do with whether they are used or not.

> They are, but we are using the meeting_user - methods also for the meeting-dependent part of user.create/update

[...]

PR #1689, opened 2023-03-31
I1649 some required changes

r-peschke:
IMO that's all we should do for [Issue1649](https://github.com/OpenSlides/openslides-backend/issues/1649), but have a look there


jsangmeister:
see https://github.com/OpenSlides/openslides-backend/issues/1649#issuecomment-1493820068

r-peschke:
> see [#1649 (comment)](https://github.com/OpenSlides/openslides-backend/issues/1649#issuecomment-1493820068)

Talked about with @jsangmeister and @rrenkert : it is okay

PR #1690, opened 2023-04-03
[feature/remove-template-fields] Reactivate user history

### LINKED ISSUE #1650: [remove-template-fields] Reactivate user history
After https://github.com/OpenSlides/openslides-backend/pull/1563 is merged, the hotfixes implemented to deactivate the user history should be removed and the history should be fixed. The changes were introduced because it does not make much sense to fix the history behaviour as long as the action payloads aren't final.

Also look for commented get_history_information-methods

jsangmeister:
fixes #1650

### DIFF
--- a/openslides_backend/action/actions/meeting_user/create.py
+++ b/openslides_backend/action/actions/meeting_user/create.py
@@ -1,6 +1,8 @@
-from typing import Any, Dict
+from typing import Any, Dict, Optional
 
 from openslides_backend.shared.exceptions import ActionException
+from openslides_backend.shared.patterns import fqid_from_collection_and_id
+from openslides_backend.shared.typing import HistoryInformation
 
 from ....models.models import MeetingUser
 from ....permissions.permissions import Permissions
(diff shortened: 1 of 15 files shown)


openslides_backend/action/actions/meeting_user/create.py
```diff
@@ -38,3 +40,33 @@ def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
...
+                            ),
+                        ]
+                    )
+                else:
+                    instance_information.append(
+                        "Participant added to multiple groups in meeting {}",
+                    )
+            else:
+                instance_information.append(
+                    "Participant added to meeting {}",
```
r-peschke, line 64: Our definition of user in meeting is the membership in a group of the meeting. 
This case here has no description as far as I know

openslides_backend/action/actions/meeting_user/create.py
```diff
@@ -38,3 +40,33 @@ def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
...
+                    instance_information.extend(
+                        [
+                            "Participant added to group {} in meeting {}",
+                            fqid_from_collection_and_id(
+                                "group", instance["group_ids"][0]
+                            ),
+                        ]
+                    )
+                else:
+                    instance_information.append(
```
r-peschke, line 59: This kind of information hides the fact that someone did not allowed things, i.e. could participate on elections, where he is not allowed to vote.

openslides_backend/action/actions/meeting_user/create.py
```diff
@@ -38,3 +40,33 @@ def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
...
+                        "Participant added to multiple groups in meeting {}",
+                    )
+            else:
+                instance_information.append(
+                    "Participant added to meeting {}",
+                )
+            instance_information.append(
```
r-peschke, line 66: Why don't you use a format, where the placeholders are replaced by their respective values?

r-peschke (commented):
I don't agree with the content, see remark, nor the format, also see remark.
If you think it's okay, I assume it's okay to merge!

openslides_backend/action/actions/meeting_user/create.py
jsangmeister: Yes, that stems form the fact that you can create a meeting user without passing any groups. This means he was (somewhat) added to the meeting, but without groups. In practice, this does not make sense - maybe `group_ids` should be required? This would ofc require new changes...

openslides_backend/action/actions/meeting_user/create.py
jsangmeister: What would be your suggestion? Display all groups? That could potentially be a lot - the history information string should only provide a gist of what happened. Details aboout the changed data must be looked up in history mode or, in the future, in the diff view. The history information is not the right place to provide all data.

openslides_backend/action/actions/meeting_user/create.py
jsangmeister: I do, but the replacement is done in the client: every `{}` in the "main string" refers to a following replacement, which is interpreted by the client as an Fqid of which the title is then inserted.

PR #1691, opened 2023-04-04
Issue1680 write_action_worker with delete

r-peschke:
Needs https://github.com/OpenSlides/openslides-datastore-service/pull/238 from datastore


reiterl (approved):
Okay.

PR #1692, opened 2023-04-05
Add minimum to projector/scroll.

reiterl:


openslides_backend/action/actions/projector/control_view.py
```diff
@@ -50,6 +50,7 @@ def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
             new_value = projector.get(field, 0) - step
         else:
             raise ActionException(f"Unknown direction {direction}")
-
+        if field == "scroll" and new_value < 0:
```
r-peschke, line 53: Move this into the `elif direction == "down"`- block. It's the only place where it is needed.
What's about the scale? I asked @luisa-beerboom

r-peschke (changes requested):

openslides_backend/action/actions/projector/control_view.py
reiterl: Okay, done.

r-peschke (approved):

PR #1696, opened 2023-04-12
Add user.create_saml_account and user.update_saml_account

reiterl:


ManniLoe:
@reiterl Shouldn't we at least define the first and last name as a mandatory field?

openslides_backend/action/actions/user/create_saml_account.py
```diff
@@ -0,0 +1,58 @@
...
+    skip_archived_meeting_check = True
+
+    def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
+        instance["username"] = instance["saml_id"]
+        if self.datastore.exists(
+            "user", FilterOperator("saml_id", "=", instance["saml_id"])
+        ):
+            raise ActionException("Saml_id already exists.")
+
+        filter_operators: List[FilterOperator] = []
```
r-peschke, line 45: Remove the check of uniqueness for `first_name`, `last_name` and `email`.

tests/system/action/user/test_create_saml_account.py
```diff
@@ -0,0 +1,84 @@
+from tests.system.action.base import BaseActionTestCase
+
+
+class UserCreateSamlAccount(BaseActionTestCase):
+    def test_create_saml_account_correct(self) -> None:
```
r-peschke, line 5: This test is not needed, if we have the second one

tests/system/action/user/test_create_saml_account.py
```diff
@@ -0,0 +1,84 @@
...
+    def test_create_saml_account_saml_id_exists(self) -> None:
+        self.set_models({"user/78": {"username": "111222333", "saml_id": "111222333"}})
+        response = self.request(
+            "user.create_saml_account",
+            {"saml_id": "111222333", "first_name": "Max", "last_name": "Mustermann"},
+        )
+        self.assert_status_code(response, 400)
+        assert "Saml_id already exists." in response.json["message"]
+
+    def test_create_saml_account_user_exists(self) -> None:
```
r-peschke, line 62: We don't need this test anymore.
But we need a test to check for duplicate `username`.

tests/system/action/user/test_update_saml_account.py
```diff
@@ -0,0 +1,70 @@
+from tests.system.action.base import BaseActionTestCase
+
+
+class UserUpdateSamlAccount(BaseActionTestCase):
+    def test_update_saml_account_correct(self) -> None:
+        self.set_models({"user/78": {"username": "111222333", "saml_id": "111222333"}})
```
r-peschke, line 6: Set a different `username` and check that it doesn't changed

r-peschke (changes requested):
Sorry, but changes in action description led to some changes

r-peschke:
> @reiterl Shouldn't we at least define the first and last name as a mandatory field?

Open Slides don't need this, therefore we don't want to make them required. But the customer could and should fill the fields.

openslides_backend/action/actions/user/create_saml_account.py
reiterl: Removed.

global/meta/models.yml
```diff
@@ -200,6 +200,12 @@ user:
     type: string
     required: true
     restriction_mode: A
+  saml_id:
+    type: string
+    restriction_mode: A
+  single_sign_on_only:
```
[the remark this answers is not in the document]
reiterl: Removed.

openslides_backend/action/actions/user/create_saml_account.py
```diff
@@ -0,0 +1,58 @@
...
+        for field in ("first_name", "last_name", "email"):
+            if instance.get(field):
+                filter_operators.append(FilterOperator(field, "=", instance.get(field)))
+        if self.datastore.exists("user", And(*filter_operators)):
+            raise ActionException("User with name and email already exists.")
+
+        instance["single_sign_on_only"] = True
```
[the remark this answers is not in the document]
reiterl: Removed.

openslides_backend/action/actions/user/create_saml_account.py
```diff
@@ -0,0 +1,58 @@
...
+            "pronoun",
+            "is_active",
+            "is_physical_person",
+        ],
+        title="create saml account schema",
+    )
+    skip_archived_meeting_check = True
+
+    def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
+        instance["username"] = instance["saml_id"]
```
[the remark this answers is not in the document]
reiterl: I use UsernameMixin here.

tests/system/action/user/test_create_saml_account.py
reiterl: Removed then.

tests/system/action/user/test_update_saml_account.py
reiterl: Okay, done.

openslides_backend/action/actions/user/create_saml_account.py
```diff
@@ -0,0 +1,53 @@
...
+            "pronoun",
+            "is_active",
+            "is_physical_person",
+        ],
+        title="create saml account schema",
+    )
+    skip_archived_meeting_check = True
+
+    def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
+        instance["username"] = next(
```
r-peschke, line 40: For one instance an iterator is a little bit oversized I think, a `self.generate_usernames([instance["saml_id"]])[0]` as used in user/create.py would be sufficient

r-peschke (approved):
IMO you should use the simpler and in this case more efficient expression, see below, instead the `next(iter(`, but if you like, you could stay with this

openslides_backend/action/actions/user/create_saml_account.py
reiterl: Okay, I use `self.generate_usernames([instance["saml_id"]])[0]`.

PR #1697, opened 2023-04-12
Add account import

reiterl:
Add user.json_upload and user.import.
Add tests.

### DIFF
--- a/openslides_backend/action/action.py
+++ b/openslides_backend/action/action.py
@@ -715,6 +715,17 @@ class Action(BaseServiceProvider, metaclass=SchemaProvider):
         """
         return None
 
+    def execute_presenter(self, PresenterClass: Any, payload: Any) -> Any:
+        presenter_instance = PresenterClass(
+            payload,
+            self.services,
+            self.datastore,
+            self.logging,
+            self.user_id,
+        )
+        presenter_instance.validate()
+        return presenter_instance.get_result()
+
 
 def merge_history_informations(
     a: Optional[HistoryInformation], *other: Optional[HistoryInformation]
--- a/openslides_backend/action/action.py
+++ b/openslides_backend/action/action.py
@@ -26,6 +26,7 @@ from ..permissions.management_levels import (
 )
 from ..permissions.permission_helper import has_organization_management_level, has_perm
 from ..permissions.permissions import Permission
+from ..presenter.base import BasePresenter
 from ..services.datastore.commands import GetManyRequest
 from ..services.datastore.interface import DatastoreService
 from ..shared.exceptions import (
@@ -715,7 +716,9 @@ class Action(BaseServiceProvider, metaclass=SchemaProvider):
         """
         return None
 
-    def execute_presenter(self, PresenterClass: Any, payload: Any) -> Any:
+    def execute_presenter(
+        self, PresenterClass: Type[BasePresenter], payload: Any
+    ) -> Any:
         presenter_instance = PresenterClass(
             payload,
             self.services,
--- a/openslides_backend/action/actions/topic/json_upload.py
+++ b/openslides_backend/action/actions/topic/json_upload.py
@@ -78,7 +78,7 @@ class TopicJsonUpload(DuplicateCheckMixin, JsonUploadMixin):
                 status = ImportStatus.ERROR
                 error.append("Duplicate")
             else:
-                status = ImportStatus.CREATE
+                status = ImportStatus.NEW
         except fastjsonschema.JsonSchemaException as exception:
             status = ImportStatus.ERROR
             error.append(exception.message)
--- a/openslides_backend/action/actions/topic/json_upload.py
+++ b/openslides_backend/action/actions/topic/json_upload.py
@@ -64,9 +64,25 @@ class TopicJsonUpload(DuplicateCheckMixin, JsonUploadMixin):
 
         # validate and check for duplicates
         self.init_duplicate_set(instance["meeting_id"])
-        rows = [self.validate_entry(entry) for entry in data]
+        self.rows = [self.validate_entry(entry) for entry in data]
 
-        self.init_rows(rows)
+        # generate statistics
+        itemCount = len(self.rows)
+        status_to_count = {status: 0 for status in ImportStatus}
+        for entry in self.rows:
+            status_to_count[entry["status"]] += 1
+
+        raw_statistics = (
+            ("total", itemCount),
+            ("created", status_to_count[ImportStatus.NEW]),
+            ("updated", status_to_count[ImportStatus.DONE]),
+            ("error", status_to_count[ImportStatus.ERROR]),
+            ("warning", status_to_count[ImportStatus.WARNING]),
+        )
+        self.statistics = [
+            {"name": name, "value": value} for name, value in raw_statistics
+        ]
+
         self.store_rows_in_the_action_worker("topic")
         return {}
 
--- a/openslides_backend/action/actions/topic/json_upload.py
+++ b/openslides_backend/action/actions/topic/json_upload.py
@@ -1,4 +1,3 @@
-from enum import Enum
 from time import time
 from typing import Any, Dict, Optional
 
@@ -11,18 +10,13 @@ from ....shared.interfaces.write_request import WriteRequest
 from ....shared.patterns import fqid_from_collection_and_id
 from ....shared.schema import required_id_schema
 from ...action import Action
+from ...mixins.import_mixins import ImportStatus
 from ...util.default_schema import DefaultSchema
 from ...util.register import register_action
 from ...util.typing import ActionResultElement
 from ..agenda_item.agenda_creation import agenda_creation_properties
 from .create import TopicCreate
 from .mixins import DuplicateCheckMixin
-
-
-class ImportStatus(str, Enum):
-    NEW = "new"
-    ERROR = "error"
-    DONE = "done"
 
 
 @register_action("topic.json_upload")
@@ -81,7 +75,7 @@ class TopicJsonUpload(DuplicateCheckMixin, Action):
         # generate statistics
         itemCount, itemNew, itemError = len(self.rows), 0, 0
         for entry in self.rows:
-            if entry["status"] == ImportStatus.NEW:
+            if entry["status"] == ImportStatus.CREATE:
                 itemNew += 1
             if entry["status"] == ImportStatus.ERROR:
                 itemError += 1
@@ -124,7 +118,7 @@ class TopicJsonUpload(DuplicateCheckMixin, Action):
                 status = ImportStatus.ERROR
                 error.append("Duplicate")
             else:
-                status = ImportStatus.NEW
+                status = ImportStatus.CREATE
         except fastjsonschema.JsonSchemaException as exception:
             status = ImportStatus.ERROR
             error.append(exception.message)
--- a/openslides_backend/action/actions/topic/json_upload.py
+++ b/openslides_backend/action/actions/topic/json_upload.py
@@ -52,7 +52,7 @@ class TopicJsonUpload(DuplicateCheckMixin, JsonUploadMixin):
         {"property": "text", "type": "string"},
         {"property": "agenda_comment", "type": "string"},
         {"property": "agenda_type", "type": "string"},
-        {"proptery": "agenda_duration", "type": "number"},
+        {"property": "agenda_duration", "type": "number"},
     ]
 
     def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
--- a/openslides_backend/action/actions/topic/json_upload.py
+++ b/openslides_backend/action/actions/topic/json_upload.py
@@ -5,7 +5,7 @@ import fastjsonschema
 from ....models.models import Topic
 from ....permissions.permissions import Permissions
 from ....shared.schema import required_id_schema
-from ...mixins.import_mixins import ImportStatus, JsonUploadMixin
+from ...mixins.import_mixins import ImportState, JsonUploadMixin
 from ...util.default_schema import DefaultSchema
 from ...util.register import register_action
 from ..agenda_item.agenda_creation import agenda_creation_properties
@@ -68,33 +68,33 @@ class TopicJsonUpload(DuplicateCheckMixin, JsonUploadMixin):
 
         # generate statistics
         itemCount = len(self.rows)
-        status_to_count = {status: 0 for status in ImportStatus}
+        state_to_count = {state: 0 for state in ImportState}
         for entry in self.rows:
-            status_to_count[entry["status"]] += 1
+            state_to_count[entry["state"]] += 1
 
         self.statistics = [
             {"name": "total", "value": itemCount},
-            {"name": "created", "value": status_to_count[ImportStatus.NEW]},
-            {"name": "updated", "value": status_to_count[ImportStatus.DONE]},
-            {"name": "error", "value": status_to_count[ImportStatus.ERROR]},
-            {"name": "warning", "value": status_to_count[ImportStatus.WARNING]},
+            {"name": "created", "value": state_to_count[ImportState.NEW]},
+            {"name": "updated", "value": state_to_count[ImportState.DONE]},
+            {"name": "error", "value": state_to_count[ImportState.ERROR]},
+            {"name": "warning", "value": state_to_count[ImportState.WARNING]},
         ]
 
-        self.set_status(
-            status_to_count[ImportStatus.ERROR], status_to_count[ImportStatus.WARNING]
+        self.set_state(
+            state_to_count[ImportState.ERROR], state_to_count[ImportState.WARNING]
         )
         self.store_rows_in_the_action_worker("topic")
         return {}
 
     def validate_entry(self, entry: Dict[str, Any]) -> Dict[str, Any]:
-        status, error = None, []
+        state, error = None, []
         try:
             TopicCreate.schema_validator(entry)
             if self.check_for_duplicate(entry["title"]):
-                status = ImportStatus.WARNING
+                state = ImportState.WARNING
             else:
-                status = ImportStatus.NEW
+                state = ImportState.NEW
         except fastjsonschema.JsonSchemaException as exception:
-            status = ImportStatus.ERROR
+            state = ImportState.ERROR
             error.append(exception.message)
-        return {"status": status, "error": error, "data": entry}
+        return {"state": state, "error": error, "data": entry}
--- a/openslides_backend/action/actions/topic/json_upload.py
+++ b/openslides_backend/action/actions/topic/json_upload.py
@@ -72,15 +72,12 @@ class TopicJsonUpload(DuplicateCheckMixin, JsonUploadMixin):
         for entry in self.rows:
             status_to_count[entry["status"]] += 1
 
-        raw_statistics = (
-            ("total", itemCount),
-            ("created", status_to_count[ImportStatus.NEW]),
-            ("updated", status_to_count[ImportStatus.DONE]),
-            ("error", status_to_count[ImportStatus.ERROR]),
-            ("warning", status_to_count[ImportStatus.WARNING]),
-        )
         self.statistics = [
-            {"name": name, "value": value} for name, value in raw_statistics
+            {"name": "total", "value": itemCount},
+            {"name": "created", "value": status_to_count[ImportStatus.NEW]},
+            {"name": "updated", "value": status_to_count[ImportStatus.DONE]},
+            {"name": "error", "value": status_to_count[ImportStatus.ERROR]},
+            {"name": "warning", "value": status_to_count[ImportStatus.WARNING]},
         ]
 
         self.set_status(
--- a/openslides_backend/action/actions/topic/json_upload.py
+++ b/openslides_backend/action/actions/topic/json_upload.py
@@ -83,6 +83,9 @@ class TopicJsonUpload(DuplicateCheckMixin, JsonUploadMixin):
             {"name": name, "value": value} for name, value in raw_statistics
         ]
 
+        self.set_status(
+            status_to_count[ImportStatus.ERROR], status_to_count[ImportStatus.WARNING]
+        )
         self.store_rows_in_the_action_worker("topic")
         return {}
 
--- a/openslides_backend/action/actions/topic/json_upload.py
+++ b/openslides_backend/action/actions/topic/json_upload.py
@@ -52,7 +52,7 @@ class TopicJsonUpload(DuplicateCheckMixin, JsonUploadMixin):
         {"property": "text", "type": "string"},
         {"property": "agenda_comment", "type": "string"},
         {"property": "agenda_type", "type": "string"},
-        {"property": "agenda_duration", "type": "number"},
+        {"property": "agenda_duration", "type": "integer"},
     ]
 
     def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
--- a/openslides_backend/action/actions/topic/json_upload.py
+++ b/openslides_backend/action/actions/topic/json_upload.py
@@ -75,8 +75,7 @@ class TopicJsonUpload(DuplicateCheckMixin, JsonUploadMixin):
         try:
             TopicCreate.schema_validator(entry)
             if self.check_for_duplicate(entry["title"]):
-                status = ImportStatus.ERROR
-                error.append("Duplicate")
+                status = ImportStatus.WARNING
             else:
                 status = ImportStatus.NEW
         except fastjsonschema.JsonSchemaException as exception:
--- a/openslides_backend/action/actions/topic/json_upload.py
+++ b/openslides_backend/action/actions/topic/json_upload.py
@@ -87,14 +87,15 @@ class TopicJsonUpload(DuplicateCheckMixin, JsonUploadMixin):
         return {}
 
     def validate_entry(self, entry: Dict[str, Any]) -> Dict[str, Any]:
-        state, error = None, []
+        state, messages = None, []
         try:
             TopicCreate.schema_validator(entry)
             if self.check_for_duplicate(entry["title"]):
                 state = ImportState.WARNING
+                messages.append("Duplicate")
             else:
                 state = ImportState.NEW
         except fastjsonschema.JsonSchemaException as exception:
             state = ImportState.ERROR
-            error.append(exception.message)
-        return {"state": state, "error": error, "data": entry}
+            messages.append(exception.message)
+        return {"state": state, "messages": messages, "data": entry}
--- a/openslides_backend/action/actions/topic/json_upload.py
+++ b/openslides_backend/action/actions/topic/json_upload.py
@@ -1,26 +1,20 @@
-from time import time
-from typing import Any, Dict, Optional
+from typing import Any, Dict
 
 import fastjsonschema
 
 from ....models.models import Topic
 from ....permissions.permissions import Permissions
-from ....shared.interfaces.event import Event, EventType
-from ....shared.interfaces.write_request import WriteRequest
-from ....shared.patterns import fqid_from_collection_and_id
 from ....shared.schema import required_id_schema
-from ...action import Action
-from ...mixins.import_mixins import ImportStatus
+from ...mixins.import_mixins import ImportStatus, JsonUploadMixin
 from ...util.default_schema import DefaultSchema
 from ...util.register import register_action
-from ...util.typing import ActionResultElement
 from ..agenda_item.agenda_creation import agenda_creation_properties
 from .create import TopicCreate
 from .mixins import DuplicateCheckMixin
 
 
 @register_action("topic.json_upload")
-class TopicJsonUpload(DuplicateCheckMixin, Action):
+class TopicJsonUpload(DuplicateCheckMixin, JsonUploadMixin):
     """
     Action to allow to upload a json. It is used as first step of an import.
     """
@@ -70,44 +64,10 @@ class TopicJsonUpload(DuplicateCheckMixin, Action):
 
         # validate and check for duplicates
         self.init_duplicate_set(instance["meeting_id"])
-        self.rows = [self.validate_entry(entry) for entry in data]
+        rows = [self.validate_entry(entry) for entry in data]
 
-        # generate statistics
-        itemCount, itemNew, itemError = len(self.rows), 0, 0
-        for entry in self.rows:
-            if entry["status"] == ImportStatus.CREATE:
-                itemNew += 1
-            if entry["status"] == ImportStatus.ERROR:
-                itemError += 1
-        self.statistics = {
-            "total": itemCount,
-            "created": itemNew,
-            "omitted": itemError,
-        }
-
-        # store rows in the action_worker
-        self.new_store_id = self.datastore.reserve_id(collection="action_worker")
-        fqid = fqid_from_collection_and_id("action_worker", self.new_store_id)
-        time_created = int(time())
-        self.datastore.write_action_worker(
-            WriteRequest(
-                events=[
-                    Event(
-                        type=EventType.Create,
-                        fqid=fqid,
-                        fields={
-                            "id": self.new_store_id,
-                            "result": {"import": "topic", "rows": self.rows},
-                            "created": time_created,
-                            "timestamp": time_created,
-                            "state": "running",
-                        },
-                    )
-                ],
-                user_id=self.user_id,
-                locked_fields={},
-            )
-        )
+        self.init_rows(rows)
+        self.store_rows_in_the_action_worker("topic")
         return {}
 
     def validate_entry(self, entry: Dict[str, Any]) -> Dict[str, Any]:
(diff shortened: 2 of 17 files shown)


reiterl:
This draft version uses a lot of memory. It `get_all` users and saves them. One alternative is to use many `filter` calls. That would be slow. I try to stick with the get_all version for now.

jsangmeister:
> This draft version uses a lot of memory. It `get_all` users and saves them. One alternative is to use many `filter` calls. That would be slow. I try to stick with the get_all version for now.

This is not feasible, since there might be too many users to fetch them all! `get_all` should generally not be used in production. You should be able to just call the `search_users` presenters internally and use the result or, if you need to customize the code, orient yourself on the structure there, which uses only a single filter call.

reiterl:
Updated the duplicate mixin, it uses a large filter with Or and And and `filter` to get the data. So I don't need to get every
user.

openslides_backend/action/actions/user/user_mixin.py
```diff
@@ -173,3 +173,55 @@ def get_history_information(self) -> Optional[HistoryInformation]:
...
+    ) -> None:
+        # we don't want to call "get_all", so need to construct this large
+        # filter here.
+        filter_ = Or(
+            *[FilterOperator("username", "=", un) for un in usernames],
+            *[
+                And(
+                    FilterOperator("first_name", "=", entry[0]),
+                    FilterOperator("last_name", "=", entry[1]),
+                    FilterOperator("email", "=", entry[2]),
```
jsangmeister, line 190: The `search_users` presenter matches case-insensitive. For the import I could see that we do not want this, but this would mean the matching algorithm differs between the presenter and the import. @rrenkert what do you think?

openslides_backend/action/actions/user/user_mixin.py
```diff
@@ -173,3 +173,55 @@ def get_history_information(self) -> Optional[HistoryInformation]:
...
+                for entry in names_and_emails
+            ],
+        )
+
+        users = self.datastore.filter(
+            "user",
+            filter_,
+            ["id", "username", "first_name", "last_name", "email"],
+            lock_result=False,
+        ).values()
```
jsangmeister, line 201: Is there a reason you don't call the `search_users` presenter here? Even if we want to filter case-sensitive here, adding a flag to the presenter instead of duplicating the code would be the way to go.

openslides_backend/action/actions/user/user_mixin.py
```diff
@@ -173,3 +173,55 @@ def get_history_information(self) -> Optional[HistoryInformation]:
...
+        }
+
+    def check_username_for_duplicate(self, username: str) -> bool:
+        result = username in self.username_to_id
+        if not result:
+            self.username_to_id[username] = None
+        return result
+
+    def check_name_and_email_for_duplicate(
+        self, first_name: Optional[str], last_name: Optional[str], email: Optional[str]
```
jsangmeister, line 221: If `username` is not provided, all other three fields are required (see https://github.com/OpenSlides/OpenSlides/wiki/without_templ.account.json_upload#user-matching), so these fields are not optional. This should also be checked somewhere in the code above.

jsangmeister (changes requested):

openslides_backend/action/actions/user/import_.py
```diff
@@ -0,0 +1,145 @@
...
+                                    "action_worker", store_id
+                                ),
+                            )
+                        ],
+                        user_id=self.user_id,
+                        locked_fields={},
+                    )
+                )
+
+        return on_success
```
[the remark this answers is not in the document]
reiterl: Move these import methods into a `ImportMixin`

openslides_backend/action/actions/user/json_upload.py
```diff
@@ -0,0 +1,175 @@
...
+from ...util.register import register_action
+from ...util.typing import ActionResultElement
+from .create import UserCreate
+from .user_mixin import DuplicateCheckMixin
+
+
+class ImportStatus(str, Enum):
+    CREATE = "create"
+    UPDATE = "update"
+    ERROR = "error"
```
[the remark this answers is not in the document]
reiterl: Done.

openslides_backend/action/actions/user/json_upload.py
```diff
@@ -0,0 +1,175 @@
...
+
+    def create_action_result_element(
+        self, instance: Dict[str, Any]
+    ) -> Optional[ActionResultElement]:
+        return {
+            "id": self.new_store_id,
+            "headers": self.headers,
+            "rows": self.rows,
+            "statistics": self.statistics,
+        }
```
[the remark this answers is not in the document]
reiterl: It is moved.

openslides_backend/action/actions/user/user_mixin.py
reiterl: Done. I have added a check in json_upload and removed the Optional.

openslides_backend/action/actions/user/user_mixin.py
reiterl: I don't know, how to call a presenter in an action. I don't see the need to use the `search_users` here.

jsangmeister: Add a helper method analogous to `execute_other_action` to the base action class, but simplified since you don't need to care about created write requests and the like. You can call the presenter class like this: https://github.com/OpenSlides/openslides-backend/blob/88b6ea61d081e7c3e55256ec954f18fbd30c91fa/openslides_backend/presenter/presenter.py#L125.

There are two reasons to call the presenter, which go hand in hand:
- code deduplication
- one single place where the user matching algorithm is defined. We don't want the import to differ from the presenter in terms of user matching and would need to change both places everytime something about the algorithm changes.

reiterl: Okay, I added a helper `execute_presenter` in the action base class and use it to do the filtering.
I use the `SearchUsers` presenter in the context.

openslides_backend/action/actions/user/user_mixin.py
reiterl: I use the `search_users` presenter to get the possible duplicates.

r-peschke:
On working for the import committee action there have been some new ideas and I changed the action description for the account import. See there and let's talk about.

jsangmeister:
I updated the formatting, please try to keep the wiki entries as clean and readable as possible and use javascript for the code parts, since this is the default everywhere else. Two questions arised:
- What is the meaing of `is_list` and when should it be used? It is set to `true` in your example, but I don't see any relevance to it.
- Do we no longer want to differentiate between "updated" and "done"/"generated"? That means that a row which has some changed fields can only be marked as "done" and the client cannot mark it explicitely for the user to check; or that a field which was not generated, but updated for some reason, can only be marked as "done" or "generated", which both do not reflect the situation appropriately.

r-peschke:
> I updated the formatting, please try to keep the wiki entries as clean and readable as possible and use javascript for the code parts, since this is the default everywhere else. 

What are you talking about?

> Two questions arised:
> 
>     * What is the meaing of `is_list` and when should it be used? It is set to `true` in your example, but I don't see any relevance to it.

In the payload there are field types like `string[]` and @luisa-beerboom and I didn't wanted to duplicate all field types.
 
> 
>     * Do we no longer want to differentiate between "updated" and "done"/"generated"? That means that a row which has some changed fields can only be marked as "done" and the client cannot mark it explicitely for the user to check; or that a field which was not generated, but updated for some reason, can only be marked as "done" or "generated", which both do not reflect the situation appropriately.

@luisa-beerboom and I just defined the existing and used states of the existing client imports. If needed somewhere we could and will introduce new state/info-tags.

reiterl:
I don't understand the username/info stuff. It seems like we want to clutter the payload and clean that in the import again. I think there are four cases username: yes/no, found duplicate: yes/no. What should the code do in these 4 cases?

reiterl:
I also don't understand the default_password stuff. What should the code do there?

luisa-beerboom:
> I don't understand the username/info stuff. It seems like we want to clutter the payload and clean that in the import again. I think there are four cases username: yes/no, found duplicate: yes/no. What should the code do in these 4 cases?


> I also don't understand the default_password stuff. What should the code do there?

I'm also wondering: Does that mean that the backend will display the generated values for these fields in the preview?
If so that would certainly be a very user friendly new feature.
If it wasn't that, I would also be wondering why these fields have the object type, since they don't exactly represent models.
Could you clear that up @r-peschke ?

r-peschke:
> I'm also wondering: Does that mean that the backend will display the generated values for these fields in the preview? If so that would certainly be a very user friendly new feature.

Yes, it should show the generated values. Otherwise it is not possible for the user to preview the data.

>  If it wasn't that, I would also be wondering why these fields have the object type, since they don't exactly represent models.

The object type is needed to give more information in a column or an item inside a column with lists. This maybe useful regularly for models, but also for non model-fields.
The `username` is a model. As it is the only search-criterion for the user, it represents the user instance.

> Could you clear that up @r-peschke ?

Hope I did:-)

r-peschke:
> I also don't understand the default_password stuff. What should the code do there?

It should generate a default_password if:
- there is no username in payload and the 3 fields (first, last, email) are given, but no user was found or
- there is no username in data and minimum 1 of the 3 fields is not in data

reiterl:
When should the username value be `{"value": v_1, "info": i_1}` ? What should the `i_1` information include?

luisa-beerboom:
> Yes, it should show the generated values. Otherwise it is not possible for the user to preview the data.

That is a good idea.
 
> The object type is needed to give more information in a column or an item inside a column with lists. This maybe useful regularly for models, but also for non model-fields. The `username` is a model. As it is the only search-criterion for the user, it represents the user instance.

I generally agree with the sentiment, it makes sense for these two fields to be object type.
But please, let's not call `username` a model. It isn't. The entire row represents a model. If there was a reference to a meeting that would be representing a model. But, even if we search via the username, it's still just a field. If we defined "model" like that, the title field in the topic import would also be a "model".

r-peschke:
> I don't understand the username/info stuff. It seems like we want to clutter the payload and clean that in the import again. I think there are four cases username: yes/no, found duplicate: yes/no. What should the code do in these 4 cases?

In the following list you see the regular cases. It misses the error cases like duplicate username in db, 3 fields (first-name, last-name, email) found more than 1 user, missing one of username, first-name, last-name...

Conditions and actions
- username in data and found in db: set row state and username-info to done, store id of username in data
- username in data and not found in db: set row state to new(=marked for user creation) and username-info to done
- username not in data, but the 3 fields are and 1 user was found: set row state and username-info to done, store id and username in data
- username not in data, but the 3 fields are and no user was found: set row state to new, generate username and set username-info to generated, store generated username in data
- username not in data and not all of the 3 fields in data: set row state to new, generate username and set username-info to generated, store generated username in data

Error cases, not complete:
- duplicate username in db including import data
- 3 fields (first-name, last-name, email) found more than 1 user
- missing at least one of username, first-name, last-name in data (usual user create condition)

> > Yes, it should show the generated values. Otherwise it is not possible for the user to preview the data.
> 
> That is a good idea.
> 
> > The object type is needed to give more information in a column or an item inside a column with lists. This maybe useful regularly for models, but also for non model-fields. The `username` is a model. As it is the only search-criterion for the user, it represents the user instance.
> 
> I generally agree with the sentiment, it makes sense for these two fields to be object type. But please, let's not call `username` a model. It isn't. The entire row represents a model. If there was a reference to a meeting that would be representing a model. But, even if we search via the username, it's still just a field. If we defined "model" like that, the title field in the topic import would also be a "model".

`username` used in this way is a reference to a user(-model) and we should use it as a user with an id, see my response to @reiterl . The reference to  meeting by name is the same.
There is a small difference in the topic, because the topic for us is just a string and we don't try to identify it as a unique object.

r-peschke:
> When should the username value be `{"value": v_1, "info": i_1}` ? What should the `i_1` information include?

see https://github.com/OpenSlides/OpenSlides/wiki/without_templ.preface_special_imports, especially the object type in data, where the `info` is of type `ImportState`.

r-peschke (changes requested):
Change in topic/json_upload
- topic/json_upload.py Line 55: proptery => property, also in test
- set ImportStatus.WARNING instead of ERROR on duplicated topic. Set ERROR only, if the data should not be imported at all and prevent importing in the common base class of import.

openslides_backend/action/action.py
```diff
@@ -715,6 +715,17 @@ def get_on_failure(self, action_data: ActionData) -> Optional[Callable[[], None]
         """
         return None
 
+    def execute_presenter(self, PresenterClass: Any, payload: Any) -> Any:
```
jsangmeister, line 718:
```suggestion
    def execute_presenter(self, PresenterClass: Type[BasePresenter], payload: Any) -> Any:
```

jsangmeister (changes requested):
Only minor typing issue, did not get to look through everything (will do the rest tomorrow).

r-peschke:
> @r-peschke I just saw in the documentation that the structure for the preview is
> 
> ```
> rows: {
>             status: ImportState;
>             error: string[];
>             data: {
>                 // property name and type must match an entry in the given `headers`
>                 [property: string]: (
>                     | null
>                     | boolean
>                     | number
>                     | string
>                     | {
>                           value: boolean | number | string;
>                           info: ImportState;
>                           type: `boolean` | `number` | `string` | `date`;
>                       }
>                 )[]; // if is_list is set in corresponding header column, we need here also a list
>             }[];
>         }[];
> ```
> 
> Is it really necessary for `data` to be an array? I mean, it is a dictionary, so one object should be able to hold all of the data for one row, right?

I changed the definition one more time. If the column is just of a simple type like string or even object it is no array. But for something like `committee_managers` in https://github.com/OpenSlides/OpenSlides/wiki/without_templ.committee.json_upload, payload type `string[]` we need an array of objects.

openslides_backend/action/action.py
reiterl: I cannot import `BasePresenter` in the Action class, because of circular imports. So it won't work.

reiterl: I get:
`ImportError: cannot import name 'Action' from partially initialized module 'openslides_backend.action.action' (most likely due to a circular import) (/app/openslides_backend/action/action.py)`

reiterl:
I have updated the statistics in the json-upload.

openslides_backend/action/action.py
jsangmeister: Move the `export_helper.py` file into the `shared` folder - it is used in actions and presenters and does not belong in the action folder. This should resolve the circular import.

openslides_backend/action/actions/user/import_.py
```diff
@@ -0,0 +1,108 @@
...
+        store_id = instance["id"]
+        worker = self.datastore.get(
+            fqid_from_collection_and_id("action_worker", store_id), ["result"]
+        )
+        if (worker.get("result") or {}).get("import") != "account":
+            raise ActionException("Wrong id doesn't point on account import data.")
+
+        # handle abort in on_success
+        if not instance["import"]:
+            return {}
```
jsangmeister, line 42: You can move this up to return before fetching anything

openslides_backend/action/actions/user/json_upload.py
```diff
@@ -0,0 +1,167 @@
...
+    def generate_username(self, entry: Dict[str, Any]) -> str:
+        return self.generate_usernames(
+            [
+                re.sub(
+                    r"\W",
+                    "",
+                    entry.get("first_name", "") + entry.get("last_name", ""),
+                )
+            ]
+        )[0]
```
jsangmeister, line 143: Merge with method in `user.create` e.g. by moving it to the username mixin as `generate_username(first_name: str, last_name: str)`

openslides_backend/action/actions/user/json_upload.py
```diff
@@ -0,0 +1,167 @@
...
+            entry["default_password"] = {"value": value, "info": info}
+        elif status in (ImportStatus.DONE, ImportStatus.ERROR):
+            if "default_password" in entry:
+                entry["default_password"] = {
+                    "value": entry["default_password"],
+                    "info": "done",
+                }
+
+
+def _names_and_email(entry: Dict[str, Any]) -> Tuple[str, str, str]:
```
jsangmeister, line 162: Move into class

openslides_backend/action/actions/user/json_upload.py
```diff
@@ -0,0 +1,167 @@
...
+                if self.check_username_for_duplicate(entry["username"], payload_index):
+                    status = ImportStatus.DONE
+                    if searchdata := self.get_search_data(payload_index):
+                        entry["id"] = searchdata["id"]
+                    else:
+                        status = ImportStatus.ERROR
+                        error.append(f"Duplicate in csv list index: {payload_index}")
+                else:
+                    status = ImportStatus.NEW
+                entry["username"] = {"value": entry["username"], "info": "done"}
```
jsangmeister, line 104:
```suggestion
                entry["username"] = {"value": entry["username"], "info": ImportStatus.DONE}
```
Also use the enum everywhere else

openslides_backend/action/mixins/import_mixins.py
```diff
@@ -0,0 +1,129 @@
...
+from ...shared.patterns import fqid_from_collection_and_id
+from ..action import Action
+from ..util.typing import ActionData, ActionResultElement
+
+
+class ImportStatus(str, Enum):
+    ERROR = "error"
+    NEW = "new"
+    WARNING = "warning"
+    DONE = "done"
```
jsangmeister, line 16: Add `generated` here and use that as well instead of the literal string.

jsangmeister (changes requested):
Logic looks solid. Did not check the tests thoroughly. @r-peschke I think it would still be good to double check the logic, since you wrote the specs.

openslides_backend/action/action.py
reiterl: Moved export_helper into the shared folder. Add type BasePresenter to PresenterClass.

openslides_backend/action/mixins/import_mixins.py
```diff
@@ -0,0 +1,129 @@
...
+                        user_id=self.user_id,
+                        locked_fields={},
+                    )
+                )
+
+        return on_success
+
+
+class JsonUploadMixin(Action):
+    headers: Any
```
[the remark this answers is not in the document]
reiterl: Done.

openslides_backend/action/actions/user/import_.py
```diff
@@ -0,0 +1,108 @@
...
+    )
+    permission = OrganizationManagementLevel.CAN_MANAGE_USERS
+    skip_archived_meeting_check = True
+
+    def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
+        store_id = instance["id"]
+        worker = self.datastore.get(
+            fqid_from_collection_and_id("action_worker", store_id), ["result"]
+        )
+        if (worker.get("result") or {}).get("import") != "account":
```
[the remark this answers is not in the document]
reiterl: I want the or here. Reason:
```
d = {"test": None}
d.get("test", {}) -> None
d.get("test") or {} -> {}
```
I want the "or" behavior.

openslides_backend/action/actions/user/import_.py
reiterl: No, we must fetch the action worker to see, if this is a account import action worker. Else we have a way to remove
every action_worker.

openslides_backend/action/actions/user/import_.py
```diff
@@ -0,0 +1,108 @@
...
+                elif self.check_username_for_duplicate(
+                    entry["data"]["username"], payload_index
+                ):
+                    raise ActionException(
+                        "Error: want to create a new user, but username already exists."
+                    )
+                else:
+                    create_action_payload.append(entry["data"])
+            elif entry["status"] == ImportStatus.DONE:
+                sdata = self.get_search_data(payload_index)
```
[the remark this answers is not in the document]
reiterl: Done.

openslides_backend/action/actions/user/json_upload.py
reiterl: Done.

openslides_backend/action/mixins/import_mixins.py
reiterl: Done.

openslides_backend/action/actions/user/json_upload.py
reiterl: Done.

openslides_backend/action/actions/user/json_upload.py
reiterl: Done.

r-peschke (changes requested):
Review still not finished, but there are some request changes

openslides_backend/action/mixins/import_mixins.py
```diff
@@ -0,0 +1,156 @@
...
+            header["property"]: header["type"] for header in self.headers
+        }
+        for entry in list(instance.get("data", [])):
+            for field in dict(entry):
+                if field not in property_to_type:
+                    del entry[field]
+                else:
+                    type_ = property_to_type[field]
+                    if type_ == "integer":
+                        if str.isdigit(entry[field]):
```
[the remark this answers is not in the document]
reiterl: I changed to try/except ValueError.

openslides_backend/action/actions/user/json_upload.py
```diff
@@ -0,0 +1,154 @@
...
+                info = ImportStatus.GENERATED
+            entry["default_password"] = {"value": value, "info": info}
+        elif status in (ImportStatus.DONE, ImportStatus.ERROR):
+            if "default_password" in entry:
+                entry["default_password"] = {
+                    "value": entry["default_password"],
+                    "info": ImportStatus.DONE,
+                }
+
+    def _names_and_email(self, entry: Dict[str, Any]) -> Tuple[str, str, str]:
```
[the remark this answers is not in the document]
reiterl: Done.

openslides_backend/action/mixins/import_mixins.py
```diff
@@ -0,0 +1,156 @@
...
+class JsonUploadMixin(Action):
+    headers: List[HeaderEntry]
+
+    def init_rows(self, rows: List[Dict[str, Any]]) -> None:
+        self.rows = rows
+
+        # generate statistics
+        itemCount, itemNew, itemDone, itemError = len(self.rows), 0, 0, 0
+        itemWarning = 0
+        for entry in self.rows:
```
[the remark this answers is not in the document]
reiterl: Make the generate statistics individual again. I moved it into the json_upload actions.

openslides_backend/action/mixins/import_mixins.py
```diff
@@ -0,0 +1,156 @@
...
+        return {}
+
+    def create_events(self, instance: Dict[str, Any]) -> Any:
+        return []
+
+    def create_action_result_element(
+        self, instance: Dict[str, Any]
+    ) -> Optional[ActionResultElement]:
+        return {
+            "id": self.new_store_id,
```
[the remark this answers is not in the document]
reiterl: Done.

openslides_backend/action/mixins/import_mixins.py
```diff
@@ -0,0 +1,156 @@
...
+                events=[
+                    Event(
+                        type=EventType.Create,
+                        fqid=fqid,
+                        fields={
+                            "id": self.new_store_id,
+                            "result": {"import": import_name, "rows": self.rows},
+                            "created": time_created,
+                            "timestamp": time_created,
+                            "state": "running",
```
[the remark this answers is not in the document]
reiterl: Done.

openslides_backend/action/actions/user/import_.py
```diff
@@ -0,0 +1,108 @@
...
+            return {}
+
+        # init duplicate mixin
+        data = worker.get("result", {}).get("rows", [])
+        for entry in data:
+            # Revert username-info and default-password-info
+            for field in ("username", "default_password"):
+                if field in entry["data"]:
+                    entry["data"][field] = entry["data"][field]["value"]
+            if entry["status"] == ImportStatus.ERROR:
```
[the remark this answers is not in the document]
reiterl: Removed.

openslides_backend/action/actions/user/import_.py
```diff
@@ -0,0 +1,108 @@
...
+            "import": {"type": "boolean"},
+        }
+    )
+    permission = OrganizationManagementLevel.CAN_MANAGE_USERS
+    skip_archived_meeting_check = True
+
+    def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
+        store_id = instance["id"]
+        worker = self.datastore.get(
+            fqid_from_collection_and_id("action_worker", store_id), ["result"]
```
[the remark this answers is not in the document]
reiterl: Added.

openslides_backend/action/actions/user/user_mixin.py
```diff
@@ -173,3 +186,42 @@ def get_history_information(self) -> Optional[HistoryInformation]:
...
+                "permission_id": 1,
+                "search": data,
+            },
+        )
+        self.used_usernames: List[str] = []
+        self.used_names_and_email: List[Any] = []
+
+    def check_username_for_duplicate(self, username: str, payload_index: int) -> bool:
+        result = (
+            self.users_in_double_lists[payload_index] or username in self.used_usernames
```
r-peschke, line 206: the type of result is list, if there are elements in the self.users_in_double_lists[payload_index

openslides_backend/action/actions/user/user_mixin.py
```diff
@@ -173,3 +186,42 @@ def get_history_information(self) -> Optional[HistoryInformation]:
...
+        if username not in self.used_usernames:
+            self.used_usernames.append(username)
+        return result
+
+    def check_name_and_email_for_duplicate(
+        self, first_name: str, last_name: str, email: str, payload_index: int
+    ) -> bool:
+        entry = (first_name, last_name, email)
+        result = (
+            self.users_in_double_lists[payload_index]
```
r-peschke, line 217: what happens if you'll find 2 entries with different usernames (unique). Add a test for this case, even if it's difficult to appear in real life

openslides_backend/action/actions/user/json_upload.py
```diff
@@ -0,0 +1,174 @@
...
+                elif self.check_name_and_email_for_duplicate(
+                    *UserJsonUpload._names_and_email(entry), payload_index
+                ):
+                    status = ImportStatus.DONE
+                    if searchdata := self.get_search_data(payload_index):
+                        entry["username"] = {
+                            "value": searchdata["username"],
+                            "info": ImportStatus.DONE,
+                        }
+                        entry["id"] = searchdata["id"]
```
r-peschke, line 136: I thought we will store this in the "username" object. This way there is an unknown column for the client and in committee-import you'll have several fields with there user_id and have to store the id in its object. Changed it in wiki

r-peschke (changes requested):
I'm not quite sure if it is all, but after you changes will review again.

openslides_backend/action/actions/user/import_.py
jsangmeister: right

jsangmeister (changes requested):

openslides_backend/action/actions/topic/json_upload.py
```diff
@@ -64,9 +64,28 @@ def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
...
+        raw_statistics = (
+            ("total", itemCount),
+            ("created", status_to_count[ImportStatus.NEW]),
+            ("updated", status_to_count[ImportStatus.DONE]),
+            ("error", status_to_count[ImportStatus.ERROR]),
+            ("warning", status_to_count[ImportStatus.WARNING]),
+        )
+        self.statistics = [
+            {"name": name, "value": value} for name, value in raw_statistics
+        ]
```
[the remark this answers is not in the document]
reiterl: I don't need it. Removed.

openslides_backend/action/mixins/import_mixins.py
```diff
@@ -116,14 +108,35 @@ def create_action_result_element(
...
+                        try:
+                            entry[field] = int(entry[field])
+                        except ValueError:
+                            raise ActionException(
+                                f"Could not parse {entry[field]} expect integer"
+                            )
+                    elif type_ == "boolean":
+                        if entry[field] in ("1", "true", "True", "T", "t"):
+                            entry[field] = True
+                        elif entry[field] in ("0", "false", "False", "F", "f"):
```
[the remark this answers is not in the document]
reiterl: Done.

tests/system/action/user/test_json_upload.py
```diff
@@ -0,0 +1,315 @@
...
+                                },
+                            },
+                        }
+                    ],
+                }
+            },
+        )
+        result = response.json["results"][0][0]
+        assert result == {
+            "id": 1,
```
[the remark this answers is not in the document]
reiterl: Renamed the status into state.

openslides_backend/action/actions/user/user_mixin.py
reiterl: It should be bool, I add a check.

openslides_backend/action/actions/user/import_.py
```diff
@@ -0,0 +1,108 @@
...
+        additional_required_fields={
+            "id": required_id_schema,
+            "import": {"type": "boolean"},
+        }
+    )
+    permission = OrganizationManagementLevel.CAN_MANAGE_USERS
+    skip_archived_meeting_check = True
+
+    def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
+        store_id = instance["id"]
```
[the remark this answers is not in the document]
reiterl: Done.

openslides_backend/action/actions/user/user_mixin.py
reiterl: Two entries in the `users_in_double_lists[payload_index]` leads to not get the search_data and an Error in the state.

luisa-beerboom:
The row format for the topic.json_upload result isn't yet quite as is described in the wiki: The state is returned correctly, but the `error` string array has not yet been renamed to `message`.

I would also expect that `message` array to contain the messages for all the warnings contained in that row (not only the error messages).

Could you do that soon?

reiterl:
I added a version with `messages` instead of `error` and added `Duplicate` message to it.

luisa-beerboom:
> I added a version with `messages` instead of `error` and added `Duplicate` message to it.

Thank you

tests/system/action/user/test_import.py
```diff
@@ -0,0 +1,148 @@
...
+    def test_import_no_permission(self) -> None:
+        self.base_permission_test({}, "user.import", {"id": 2, "import": True})
+
+    def test_import_permission(self) -> None:
+        self.base_permission_test(
+            {},
+            "user.import",
+            {"id": 2, "import": True},
+            OrganizationManagementLevel.CAN_MANAGE_USERS,
+        )
```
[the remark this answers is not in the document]
reiterl: Send data back in error case and update the tests, to check the errors in the messages.

openslides_backend/action/actions/user/import_.py
```diff
@@ -0,0 +1,118 @@
...
+        # Recheck and update data, update needs "id"
+        create_action_payload: List[Dict[str, Any]] = []
+        update_action_payload: List[Dict[str, Any]] = []
+        self.error = False
+        for payload_index, entry in enumerate(data):
+            if entry["state"] == ImportState.NEW:
+                if not entry["data"].get("username"):
+                    self.error = True
+                    entry["state"] = ImportState.ERROR
+                    entry["messages"].append("Error: could not find username")
```
r-peschke, line 66: Better something like "Error: Want to create user, but missing username in import data"

openslides_backend/action/actions/user/import_.py
```diff
@@ -0,0 +1,118 @@
...
+                        "Error: want to create a new user, but username already exists."
+                    )
+                else:
+                    create_action_payload.append(entry["data"])
+            elif entry["state"] == ImportState.DONE:
+                search_data = self.get_search_data(payload_index)
+                if not entry["data"].get("username"):
+                    self.error = True
+                    entry["state"] = ImportState.ERROR
+                    entry["messages"].append("Error: could not find username")
```
r-peschke, line 82: Better something like "Error: Want to update user, but missing username in import data"

r-peschke (changes requested):
Just some smaller text changes, the rest is okay for me

openslides_backend/action/actions/user/import_.py
reiterl: Done.

openslides_backend/action/actions/user/import_.py
reiterl: Done.

openslides_backend/action/mixins/import_mixins.py
```diff
@@ -0,0 +1,179 @@
...
+
+    def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
+        store_id = instance["id"]
+        worker = self.datastore.get(
+            fqid_from_collection_and_id("action_worker", store_id),
+            ["result", "state"],
+            lock_result=False,
+        )
+        if (worker.get("result") or {}).get("import") != self.import_name:
+            raise ActionException("Wrong id doesn't point on account import data.")
```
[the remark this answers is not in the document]
reiterl: Done.

reiterl:
Add a check for multiple search data and update error msgs.

r-peschke (approved):

jsangmeister (approved):

Issue #1702, opened 2023-04-17
`STACK_INTERNAL` actions can be executed from outside the stack


jsangmeister:
The route `internal/handle_request` is intented for `STACK_INTERNAL` actions, but they can currently also be executed via the public `handle_request` route.

PR #1703, opened 2023-04-17
Ensure stack-internal actions can only be executed stack-internally

jsangmeister:
fixes https://github.com/OpenSlides/openslides-backend/issues/1702


tests/system/action/test_internal_actions.py
```diff
@@ -113,37 +118,35 @@ def tearDown(self) -> None:
...
         response = self.internal_request(
             "option.create",
             {"meeting_id": 1, "text": "test"},
             self.internal_auth_password,
         )
         self.assert_status_code(response, 400)
         self.assert_model_not_exists("option/1")
+
+    @disable_dev_mode
+    def test_internal_try_access_via_public_route(self) -> None:
```
r-peschke, line 148: IMO better name: `test_internal_try_access_stack_internal_via_public_route`

tests/system/action/test_internal_actions.py
```diff
@@ -113,37 +118,35 @@ def tearDown(self) -> None:
...
+        self.datastore.truncate_db()
+        response = self.request("organization.initial_import", {"data": {}})
+        self.assert_status_code(response, 400)
+        self.assert_model_not_exists("organization/1")
```
r-peschke, line 152: Show success with internal rout in a new test: `test_internal_try_access_stack_internal_via_internal_route`

r-peschke (changes requested):
I wrote some documentation on https://github.com/OpenSlides/OpenSlides/wiki/Action-Service#routes ff., please have a look

openslides_backend/action/action_handler.py
```diff
@@ -216,8 +216,17 @@ def perform_action(
         action_name = action_payload_element["action"]
         ActionClass = actions_map.get(action_name)
         if ActionClass is None or (
-            ActionClass.action_type == ActionType.BACKEND_INTERNAL
-            and not self.env.is_dev_mode()
+            not self.env.is_dev_mode()
+            and (
+                (
+                    self.internal
```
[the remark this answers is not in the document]
jsangmeister: agreed

tests/system/action/test_internal_actions.py
jsangmeister: That test already exists: https://github.com/OpenSlides/openslides-backend/blob/1ed1197d524f5c6f03db7fb7e6ca7ab14ec6f8cb/tests/system/action/test_internal_actions.py#L68

jsangmeister:
Update the documentation a little bit, but mostly formatting. Thanks for that!

tests/system/action/test_internal_actions.py
r-peschke: I wanted to document in the tests, that an action of type STACK_INTERNAL works on internal route, but not on public route. 
Why do all the ordinary actions should work without user permission, when they were called on the internal route?

jsangmeister: But exactly that is asserted by the linked test: `organization.initial_import` is a stack internal action and can be accessed via the internal route.

The normal actions must work without user permissions because there is no user whose permissions we can check - the `user_id` comes from the auth service and is based on the provided auth data. The internal route is not called with any auth data, hence we cannot check any permissions.

jsangmeister: > Proposal: Rename the tests to something with `stack_internal` and internal_route in it's name and/or put it besides the other test concerning the stack_internal-actions. There is to ,much `internal`!

In these tests, `internal == stack_internal` if not said otherwise, since these are supposed to test the stack internal route and actions. I think the current naming is fine.

> Why do we want this? If there are special cases like the SSO-Login with creating/updating users we have specialized actions. To allow any service inside the stack to use all actions without permission-check and user I think this is not a good idea

The manage service may execute arbitrary actions, e.g. to manage users. Also, this is not a new feature introduced with this PR, but it's existed for a rather long time - I don't think we should be having this discussion here.

jsangmeister:
@r-peschke I got a little lost in cleaning up the tests, but I hope they are clearer now. Also merged some functions with the migration route tests.

r-peschke (approved):

PR #1707, opened 2023-04-19
Add a check for only one start_time and end time.

reiterl:
Add tests, update meeting.create and meeting.update.

### DIFF
--- a/openslides_backend/action/actions/meeting/create.py
+++ b/openslides_backend/action/actions/meeting/create.py
@@ -67,6 +67,14 @@ class MeetingCreate(CreateActionWithDependencies, MeetingPermissionMixin):
                 f"You cannot create a new meeting, because you reached your limit of {limit_of_meetings} active meetings."
             )
 
+        if (
+            instance.get("start_time")
+            and not instance.get("end_time")
+            or not instance.get("start_time")
+            and instance.get("end_time")
+        ):
+            raise ActionException("Only one of start_time and end_time is not allowed.")
+
         instance["is_active_in_organization_id"] = committee["organization_id"]
         self.apply_instance(instance)
         action_data = [
--- a/openslides_backend/action/actions/meeting/create.py
+++ b/openslides_backend/action/actions/meeting/create.py
@@ -18,11 +18,13 @@ from ..motion_workflow.create import (
 from ..projector.create import ProjectorCreateAction
 from ..projector_countdown.create import ProjectorCountdownCreate
 from ..user.update import UserUpdate
-from .mixins import MeetingPermissionMixin
+from .mixins import MeetingCheckTimesMixin, MeetingPermissionMixin
 
 
 @register_action("meeting.create")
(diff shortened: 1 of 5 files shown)


r-peschke (approved):

openslides_backend/action/actions/meeting/create.py
```diff
@@ -67,6 +67,14 @@ def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
                 f"You cannot create a new meeting, because you reached your limit of {limit_of_meetings} active meetings."
             )
 
+        if (
+            instance.get("start_time")
+            and not instance.get("end_time")
+            or not instance.get("start_time")
+            and instance.get("end_time")
+        ):
+            raise ActionException("Only one of start_time and end_time is not allowed.")
```
jsangmeister, line 76: Move into a helper method in `MeetingPermissionMixin` (which then needs to be renamed to `MeetingMixin`) to avoid the code duplication.

jsangmeister (changes requested):

openslides_backend/action/actions/meeting/update.py
```diff
@@ -203,6 +203,15 @@ def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
                 )
             if instance["jitsi_domain"].strip().endswith("/"):
                 raise ActionException("It is not allowed to end jitsi_domain with '/'.")
+
+        if (
+            instance.get("start_time")
+            and not instance.get("end_time")
+            or not instance.get("start_time")
+            and instance.get("end_time")
```
jsangmeister, line 211: You also need to check with the database here. Add a test with the following: `start_time` and `end_time` are set to some values. A `meeting.update` sets `start_time` to `None` and does not touch `end_time`. I suppose the test will go through, but shouldn't.

reiterl: Add the test, it failed. Updated the mixin to handle with database cases here.

tests/system/action/meeting/test_update.py
```diff
@@ -325,6 +325,69 @@ def test_update_topic_poll_default_group(self) -> None:
...
+                    "default_group_id": 1,
+                    "projector_ids": [1],
+                    "reference_projector_id": 1,
+                    "default_projector_$_ids": Meeting.default_projector__ids.replacement_enum,
+                    **{
+                        f"default_projector_${name}_ids": [1]
+                        for name in cast(
+                            List[str], Meeting.default_projector__ids.replacement_enum
+                        )
+                    },
```
jsangmeister, line 360: Are the projector fields really needed for this test?

jsangmeister (changes requested):

tests/system/action/meeting/test_update.py
reiterl: No, not really, I removed some stuff from the test.

jsangmeister (approved):

PR #1723, opened 2023-05-08
Add ShortCharField and use it for pronoun.

reiterl:
Add a test in user/update.


jsangmeister (approved):
Still not happy with the name `short_string`, but LGTM otherwise

PR #1728, opened 2023-05-09
Replace short_string type with maxLength attribute

### LINKED ISSUE #1726: Document string types in `models.yml`
Please add the new type `short_string` to the documentation of the file: https://github.com/OpenSlides/openslides-backend/blob/ae8df2ed57e592e77356d3eb9750159b936adaeb/global/meta/models.yml#L3

Now we have "string", "text", and "short_string". There is no documentation, that the difference between this types are. As far as I understand, "string" had a max length and "text" had no max length. Like the difference between `CHAR` and `TEXT` in postgres. So what is the difference between "string" and "short_string"?

I would propose, that we remove "short_string" and add an attribute to the "string" type like "max_size". This attribute could be set to 255 as default. This would be similar to the attribute "minium" of the "number" type. I think this is more flexible and less confusing than adding a new type for each length that we need.

_Originally posted by @ostcar in https://github.com/OpenSlides/openslides-backend/issues/1724#issuecomment-1538080199_

jsangmeister:
fixes #1726 

While wanting to document the feature, I realized that the functionality to replace the `short_string` field was already present, so I went ahead and replaced it already. Sorry for the unecessary overhead.

### DIFF
--- a/cli/generate_models.py
+++ b/cli/generate_models.py
@@ -26,7 +26,6 @@ DESTINATION = os.path.abspath(
 
 COMMON_FIELD_CLASSES = {
     "string": "CharField",
-    "short_string": "ShortCharField",
     "number": "IntegerField",
     "boolean": "BooleanField",
     "JSON": "JSONField",
--- a/cli/modelsvalidator/validate.py
+++ b/cli/modelsvalidator/validate.py
@@ -28,7 +28,6 @@ RELATION_TYPES = (
 
 DATA_TYPES = (
     "string",
-    "short_string",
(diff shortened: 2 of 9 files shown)


jsangmeister:
`maxLength` did not work previously at all. I fixed this as well.

r-peschke (approved):

ostcar:
Does this mean, I can revert the change in the autoupdate service?

jsangmeister:
> Does this mean, I can revert the change in the autoupdate service?

yes

PR #1735, opened 2023-05-15
Implement structure for model migrations

jsangmeister:
For now, these are just the structural changes necessary to make the backend work with https://github.com/OpenSlides/openslides-datastore-service/pull/240. The migration for the template fields removal will follow.


jsangmeister:
Checks are obviously failing until the datastore commit is updated. @r-peschke you may review already if you like - optimally, only two additional files will be added and nothing changed when implementing the template fields removal migration.

PR #1754, opened 2023-06-08
[feature/remove-template-fields] Cleanup

jsangmeister:
Some cleanup of places I noticed while going through the changed of the feature branch. Feel free to look through it. Each commit mostly encapsulates one change and explains a bit about what was done, so it might be preferable to go through each commit instead of viewing all changes.


openslides_backend/action/actions/user/update.py
```diff
@@ -84,18 +84,3 @@ def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
             self.check_limit_of_user(1)
 
         return instance
-
-    def apply_instance(
```
jsangmeister, line 88: This looks like it was missed during a rebase, but it seems no longer necessary - at least the test introduced in https://github.com/OpenSlides/openslides-backend/pull/1664 together with this change works without this code, so I removed it. Maybe you have more insight about this @r-peschke

r-peschke: seems okay

r-peschke:
Made a review based on the commits:
- **Move additional model code to mixins** : We should revert that. Extracting parts from the generator splits the code. We may do this after decisions for a new repo structure. The new model mixins are used in the models.py, which will stay in the backend, I suppose. The generator itself may be moved to a separate repo.
- **Remove $ from BaseSetMediafileAction** okay
- **Remove field_prefix_map** okay
- **Use schema_version constant in checker** okay
- **Rename short variables** okay
- **Deduplicate meeting user filters & refactor check_mediafile_id tests ** okay
- **Clean up remnants of template field code ** okay

Please check whether changes in action documentation are necessary

jsangmeister:
> * **Move additional model code to mixins** : We should revert that. Extracting parts from the generator splits the code. We may do this after decisions for a new repo structure. The new model mixins are used in the models.py, which will stay in the backend, I suppose. The generator itself may be moved to a separate repo.

I agree that the code splitting is not optimal, but I don't think it's so bad, since it's only a few constants and utility functions. But we should not put so much code in literal strings - it hinders code checking, linting and formatting. If you have a better suggestion, feel free to add it here, but I think having two files side-by-side, one generated, one filled manually, is the best solution here.

I don't this has anyhting to do with the restructure, because the problem itself stays the same. Wherever the generator and the generated code lives, this solution (and any other we may implement) should of course be able to handle both situations.

r-peschke:
> > * **Move additional model code to mixins** : We should revert that. Extracting parts from the generator splits the code. We may do this after decisions for a new repo structure. The new model mixins are used in the models.py, which will stay in the backend, I suppose. The generator itself may be moved to a separate repo.
> 
> I agree that the code splitting is not optimal, but I don't think it's so bad, since it's only a few constants and utility functions. But we should not put so much code in literal strings - it hinders code checking, linting and formatting. If you have a better suggestion, feel free to add it here, but I think having two files side-by-side, one generated, one filled manually, is the best solution here.
> 
> I don't this has anyhting to do with the restructure, because the problem itself stays the same. Wherever the generator and the generated code lives, this solution (and any other we may implement) should of course be able to handle both situations.

Sorry, but split code in this place IMO is not helpful any way!

What's about documentation in the wiki: Is there something to do or done?

r-peschke (changes requested):
please remove the first commit.
It is not helpful to change these places, because of the massive changes we have to do related to the relational models. Ans splitting code parts may force us, to create the new `meta repo` in a specific way to keep things together.

jsangmeister:
> What's about documentation in the wiki: Is there something to do or done?

No, this is only cleanup and should not change documentation.

> It is not helpful to change these places, because of the massive changes we have to do related to the relational models.

This is completely independent of the relational rework. Also, we cannot postpone all changes until after the relational rework which will take at least a few months until it's live.

> Ans splitting code parts may force us, to create the new `meta repo` in a specific way to keep things together.

The generator is python-specific and will therefore probably stay in this repo. And even if not, how the generator is structured internally should have no impact on the structure of the meta repository. Please name an example where the changes from the first commit would lead to problems.

And the arguments for it still stand: It is generally bad behaviour to put code into string literals, especially huge amounts. A few constants as it was previously may be fine (although also not optimal), but with the helper methods introduced here we cannot just leave it as a string.

jsangmeister:
Removed the first commit and moved it to https://github.com/OpenSlides/openslides-backend/pull/1808 to be able to merge the other changes here. The discussion can be continued there.

r-peschke (approved):

r-peschke:
Approved, but see failing test

jsangmeister:
I don't know why it failed, a re-run fixed it - I guess some kind of ordering problem in the test?

PR #1755, opened 2023-06-12
Generate readable default passwords

### LINKED ISSUE #1734: Make generated passwords more readable
Description: Using 0 and O or I and l in a password is quite though. Most users can't recognize the difference. 

What should happen:
If it is possible we should avoid this letters and numbers in passwords. This would make it a lot easier for most users to enter the passwort without copying.

jsangmeister:
fixes #1734 

Removed the following chars from the list of possible password chars: `lIOQ01`

See also https://www.grc.com/ppp.htm

### DIFF
--- a/openslides_backend/action/util/crypto.py
+++ b/openslides_backend/action/util/crypto.py
@@ -1,6 +1,7 @@
 import secrets
 
 RANDOM_STRING_CHARS = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"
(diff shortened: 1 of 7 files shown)


openslides_backend/action/util/crypto.py
```diff
@@ -1,10 +1,18 @@
...
 def get_random_string(length: int, allowed_chars: str = RANDOM_STRING_CHARS) -> str:
     """
     Return a securely generated random string.
     """
     return "".join(secrets.choice(allowed_chars) for i in range(length))
+
+
+def get_random_password(length: int = 10, allowed_chars: str = PASSWORD_CHARS) -> str:
+    """
+    Return a securely generated random password which uses only uses easily identifiable characters.
```
r-peschke, line 16: typo

r-peschke (approved):
see remark with typo

PR #1756, opened 2023-06-12
Prevent deletion of group with users

### LINKED ISSUE #1743: group deletion deletes also associated meeting participants without a warning
**Description:** If a user deletes a group in a meeting, all of the associated participants will be deleted with it from the meeting. The deletion happens without any warning. The users are still available in accounts on the organisations level. The crucial point is, that the deletions happens without any hint. Most admins won't recognize this or will be irritated where the participants have gone. 

**Reproduction:**
1. open a meeting
2. open _participants_
3. create a new participant A
4. open the three-dot-menu in participants and choose _groups_. create a new group.
5. head back to the participant overview and user A to the new group.
6. open _groups_ menu again and delete the group
7. go back to the participant menu and search for the user. User A is deleted. 

**What should happen:**
The deletion of a group should not be possible, if the group contains one or more users. Also a failure dialog should appear which contains the information, that a deletion is not possible, because the group has users.

**Possible dialog:** The group can't be deleted, because it contains participants. Please change participant assignments to delete the group.

jsangmeister:
fixes #1743 

Also fixed some other group deletion tests - they requests were failing because the meeting was archived and not because  of the actual reason.

### DIFF
--- a/openslides_backend/action/actions/group/delete.py
+++ b/openslides_backend/action/actions/group/delete.py
@@ -1,6 +1,8 @@
 from collections import defaultdict
 from typing import Any, Dict, Iterable, List, Optional, Set, Union, cast
 
+from openslides_backend.shared.exceptions import ActionException
+
 from ....models.models import Group
 from ....permissions.permissions import Permissions
 from ....services.datastore.commands import GetManyRequest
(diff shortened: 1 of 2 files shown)


r-peschke (commented):
Where was this solution discussed? This was the way to remove a lot of users from a meeting!

r-peschke (commented):
Missing action documentation in wiki

jsangmeister:
> Missing action documentation in wiki

Thanks, done: https://github.com/OpenSlides/OpenSlides/wiki/group.delete

r-peschke (approved):

PR #1764, opened 2023-06-15
Add uniqueness checks to `external_id` fields

r-peschke:
Original PR was merged, but there should be made some checks for uniqueness of external_id in scope of meeting for group, committee for meeting and organization for committee.external_id

### DIFF
--- a/openslides_backend/action/actions/group/group_mixin.py
+++ b/openslides_backend/action/actions/group/group_mixin.py
@@ -1,12 +1,30 @@
 from typing import Any, Dict
 
+from openslides_backend.action.mixins.check_unique_name_mixin import (
+    CheckUniqueInContextMixin,
+)
 from openslides_backend.shared.exceptions import PermissionDenied
 
 from ....permissions.permission_helper import is_admin
 from ...action import Action
 
 
-class GroupMixin(Action):
+class GroupMixin(CheckUniqueInContextMixin, Action):
+    check_unique_error_text = (
+        "The external_id of the group is not unique in the meeting scope."
+    )
+
+    def validate_instance(self, instance: Dict[str, Any]) -> None:
+        super().validate_instance(instance)
+        if "external_id" in instance:
+            self.check_unique_in_context(
+                "external_id",
+                instance["external_id"],
+                self.__class__.check_unique_error_text,
+                "meeting_id",
+                self.get_meeting_id(instance),
+            )
+
     def check_permissions(self, instance: Dict[str, Any]) -> None:
         super().check_permissions(instance)
         # external id is only allowed for admins
--- a/openslides_backend/action/actions/group/group_mixin.py
+++ b/openslides_backend/action/actions/group/group_mixin.py
@@ -10,17 +10,14 @@ from ...action import Action
 
 
 class GroupMixin(CheckUniqueInContextMixin, Action):
-    check_unique_error_text = (
-        "The external_id of the group is not unique in the meeting scope."
-    )
-
     def validate_instance(self, instance: Dict[str, Any]) -> None:
         super().validate_instance(instance)
         if "external_id" in instance:
             self.check_unique_in_context(
                 "external_id",
                 instance["external_id"],
-                self.__class__.check_unique_error_text,
+                "The external_id of the group is not unique in the meeting scope.",
+                instance.get("id"),
                 "meeting_id",
                 self.get_meeting_id(instance),
             )
--- a/openslides_backend/action/actions/meeting/update.py
+++ b/openslides_backend/action/actions/meeting/update.py
@@ -309,10 +309,10 @@ class MeetingUpdate(
                 raise MissingPermission(OrganizationManagementLevel.SUPERADMIN)
 
     def get_committee_id(self, meeting_id: int) -> int:
-        if not hasattr(self, "__committee_id"):
-            self.__committee_id = self.datastore.get(
+        if not hasattr(self, "_committee_id"):
+            self._committee_id = self.datastore.get(
                 fqid_from_collection_and_id(self.model.collection, meeting_id),
                 ["committee_id"],
                 lock_result=False,
             )["committee_id"]
-        return self.__committee_id
+        return self._committee_id
--- a/openslides_backend/action/actions/meeting/update.py
+++ b/openslides_backend/action/actions/meeting/update.py
@@ -177,9 +177,6 @@ class MeetingUpdate(
         },
     )
     check_email_field = "users_email_replyto"
-    check_unique_error_text = (
-        "The external_id of the meeting is not unique in the committee scope."
-    )
 
     def validate_instance(self, instance: Dict[str, Any]) -> None:
         super().validate_instance(instance)
@@ -187,7 +184,8 @@ class MeetingUpdate(
             self.check_unique_in_context(
                 "external_id",
                 instance["external_id"],
-                self.__class__.check_unique_error_text,
+                "The external_id of the meeting is not unique in the committee scope.",
+                instance["id"],
                 "committee_id",
                 self.get_committee_id(instance["id"]),
             )
@@ -316,5 +314,5 @@ class MeetingUpdate(
                 fqid_from_collection_and_id(self.model.collection, meeting_id),
                 ["committee_id"],
                 lock_result=False,
-            ).get("committee_id", 0)
+            )["committee_id"]
         return self.__committee_id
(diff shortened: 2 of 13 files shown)


reiterl (commented):
I'm not sure, if the update case works with this.
Will the client send the old external_id of a group/committee/meeting, if sth else is changed?
The `self.__class__` construct looks strange. But okay. It seems to work.

openslides_backend/action/actions/group/group_mixin.py
```diff
@@ -1,12 +1,30 @@
...
+        "The external_id of the group is not unique in the meeting scope."
+    )
+
+    def validate_instance(self, instance: Dict[str, Any]) -> None:
+        super().validate_instance(instance)
+        if "external_id" in instance:
+            self.check_unique_in_context(
+                "external_id",
+                instance["external_id"],
+                self.__class__.check_unique_error_text,
```
jsangmeister, line 23:
```suggestion
                self.check_unique_error_text,
```
or just use the string directly

jsangmeister: same for all other places below

openslides_backend/action/actions/meeting/update.py
```diff
@@ -295,3 +309,12 @@ def check_permissions(self, instance: Dict[str, Any]) -> None:
...
                 raise MissingPermission(OrganizationManagementLevel.SUPERADMIN)
+
+    def get_committee_id(self, meeting_id: int) -> int:
+        if not hasattr(self, "__committee_id"):
+            self.__committee_id = self.datastore.get(
+                fqid_from_collection_and_id(self.model.collection, meeting_id),
+                ["committee_id"],
+                lock_result=False,
+            ).get("committee_id", 0)
+        return self.__committee_id
```
jsangmeister, line 320: The datastore already has caching so that the actions don't need to implement it themselves. I'm against doing that here.

Also, access the `committee_id` via `[]` - there should always be one present and if it doesn't, it's better to raise a 500 error instead of using `0` and generating undefined behaviour afterwards.

openslides_backend/action/mixins/check_unique_name_mixin.py
```diff
@@ -0,0 +1,34 @@
...
+                And(
+                    FilterOperator(unique_name, "=", unique_value),
+                    FilterOperator(context_name, "=", context_id),
+                ),
+            )
+        else:
+            name_exists = self.datastore.exists(
+                self.model.collection,
+                FilterOperator(unique_name, "=", unique_value),
+            )
```
jsangmeister, line 31:
```suggestion
        filter: Filter = FilterOperator(unique_name, "=", unique_value)
        if context_name:
            filter = And(
                filter,
                FilterOperator(context_name, "=", context_id),
            )
        name_exists = self.datastore.exists(
            self.model.collection,
            filter,
        )
```
is shorter and IMO better to read

jsangmeister (changes requested):
Also, as @reiterl already noted, test what happens when you update something with the same `external_id` again.

And it would be nice if you chose a more suitable title ;)

r-peschke:
> I'm not sure, if the update case works with this. Will the client send the old external_id of a group/committee/meeting, if sth else is changed? The `self.__class__` construct looks strange. But okay. It seems to work.

Very good that you found this one.
self.__class__ refers to the class, self refers to the instance. Therefore the construct self.__class__ technical is the correct one. But it is not necessary, because self uses the class-variable, if there is no instance variable. 
Finally I removed the expression and use the string directly.

openslides_backend/action/actions/meeting/update.py
jsangmeister: This is still open

jsangmeister: > I don't want to call 2 times an external service

But you don't, it's an internal call - the datastore adapter caches the result and returns it before you access the database again

openslides_backend/action/actions/meeting/update.py
```diff
@@ -295,3 +307,12 @@ def check_permissions(self, instance: Dict[str, Any]) -> None:
             )
             if not is_superadmin:
                 raise MissingPermission(OrganizationManagementLevel.SUPERADMIN)
+
+    def get_committee_id(self, meeting_id: int) -> int:
+        if not hasattr(self, "__committee_id"):
```
reiterl, line 312: Why this `__committee_id` cache var? I think it is not needed, as the ds should handle this.

openslides_backend/action/mixins/check_unique_name_mixin.py
```diff
@@ -0,0 +1,37 @@
...
+        unique_name: str,
+        unique_value: str,
+        exception_message: str,
+        self_id: Optional[int] = None,
+        context_name: str = "",
+        context_id: int = 0,
+    ) -> None:
+        """
+        checks uniqueness of a string_value in context, i.e. in meeting, committee or organisation
+        Leave context_name empty to look without context
```
reiterl, line 20: Why is not the `self_id` described?

openslides_backend/action/mixins/check_unique_name_mixin.py
```diff
@@ -0,0 +1,37 @@
...
+        """
+        filter: Filter = FilterOperator(unique_name, "=", unique_value)
+        if context_name:
+            filter = And(
+                filter,
+                FilterOperator(context_name, "=", context_id),
+            )
+        if self_id:
+            filter = And(filter, Not(FilterOperator("id", "=", self_id)))
```
reiterl, line 29: Better use `!=` and not *Not*.

reiterl (commented):
Some small remarks. I think the update case is now handled (with the self_id).
I don't have a good feeling about the `__committee_id` save var.

openslides_backend/action/actions/meeting/update.py
r-peschke: No, the adapter only caches, if the database access is made with lock_result = True, see https://github.com/OpenSlides/openslides-backend/blob/499cef34dad17578ee0c5141bb164df2be40e709/openslides_backend/services/datastore/cache_adapter.py#L62

r-peschke:
> Some small remarks. I think the update case is now handled (with the self_id). I don't have a good feeling about the `__committee_id` save var.

changed __committee to _committee

openslides_backend/action/actions/meeting/update.py
r-peschke: > Why this `__committee_id` cache var? I think it is not needed, as the ds should handle this.

The datastore adapter don't cache it, because lock_result = False, see https://github.com/OpenSlides/openslides-backend/blob/499cef34dad17578ee0c5141bb164df2be40e709/openslides_backend/services/datastore/cache_adapter.py#L62
The datastore itself is an external service

openslides_backend/action/mixins/check_unique_name_mixin.py
r-peschke: done

openslides_backend/action/mixins/check_unique_name_mixin.py
r-peschke: done

openslides_backend/action/mixins/check_unique_name_mixin.py
```diff
@@ -0,0 +1,38 @@
...
+        unique_value: str,
+        exception_message: str,
+        self_id: Optional[int] = None,
+        context_name: str = "",
+        context_id: int = 0,
+    ) -> None:
+        """
+        checks uniqueness of a string_value in context, i.e. in meeting, committee or organisation
+        Leave context_name empty to look without context. On case of update put the id of your object
+        to self_id to excude the object itself from search.
```
reiterl, line 21: typo "excude" -> "exclude"

reiterl (approved):
Typo in one description, Check of the style failed.
I approve it here. I think this is good to be merged. After the style check is finished.

openslides_backend/action/actions/meeting/update.py
jsangmeister: Then just use `lock_result=True`, locking an fqfield is virtually no overhead. Or, even better, put it in the `prefetch` method (without `lock_result=True`, of course), since that's exactly what it's for.

jsangmeister (approved):
I think the extra caching layer is redundant and an arbitrary point - we have many, many places where we use the internal caching, i.e. all action with a `prefetch` method. Implementing the own caching to maybe save a couple of milliseconds (if even) is unnecessary and inconsistent with other actions.

I'm approving it anyway to get this merged, since it's only unnecessary code.

openslides_backend/action/actions/meeting/update.py
r-peschke: We have problems with locked-fields error and locking such a field is not a good idea. It's a defect in our "action workflow", that our double cache is not able to cache regular data. And the advice to set lock_result = True is even worse!

Issue #1762, opened 2023-06-14
Wrong user count for Active accounts in imprint


MSoeb:
Changed requirements after talk with @emanuelschuetze @rrenkert :
2 models of counting:
- all users with is_active set true
- all users with is_active set true and member of an active meeting

First one is the standard. Choice  by Environment variable `User_Account_Model` **standard** or **no_archived_meetings**
Being organization- or committee-manager has no impact

Actions, where a user will be added to the number of users and we have to check the limit **(if it is not set to 0)**
- **unarchive a meeting**
- user.update/create: **adding to group, committee or orga-rights, activate an user**
- user.create:**create an user without giving him a group**
- **on data import not using the standard actions**
- **on saml/sso user addition**
- undelete actions on meeting, group, user, committee

**Remove checks in case of changed orga- or committee-rights**

rrenkert:
Counted users are (positive rules):
The user is active AND (
  The user is in one or more group(s) of one or more non archived meeting(s) OR
  The user is an orga or committee admin OR
  The user is in no group of any meeting)

@emanuelschuetze or @MSoeb can give further feedback but thats what I guess we should implement.

r-peschke:
> **Description:** The user count in imprint currently includes all users. This is not wanted. Some users must be excluded.
> 
> **What should be counted:**
> 
> The count for active users should not include users which belong to archived meetings. These user should not be counted in "Active accounts".
> 
Question: What's about a user, who belongs to an archived and an active meeting?
I assume, this user should be counted

> The count should include all users which are in an active meeting and all user that don't belong to a meeting (including organizationsadmin, committee admins, account admins, meeting admins).

1. meeting admins are member of a meeting and will be counted anyway
2. account admins I don't know what it is
3. If we count all user's that are in an active meeting and all that are not in an active meeting (don't belong to any active meeting), we will count all users as we do.

Proposal:
A user will we counted as active user, if the user has the flag is_active set to True and one of the following:
 - the user is member of a group in an active meeting. The meeting.user_ids should reflect them on a per meeting base
 - the user has a level for organization management
 - the user is committee-manager in at least one committee

reiterl:
Mmh, I don't know, how to access the `UserAccountModel` environment variable in the count_users_for_limit function. Etc.
I don't like this issue. I won't continue here.

luisa-beerboom:
@emanuelschuetze @MSoeb @rrenkert  what's the status of this?

[...]

PR #1765, opened 2023-06-15
Update count users for check against the limit of users. 

reiterl:
Draft version. In the first commit, I have moved the counting into one function and in the second commit I have implemented a possible(?) solution. As far as I understand this. 
Here some more time must be invested into the places where to check for limit of users. See user.update, groups,...

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,6 +143,9 @@ The action component listens to port 9002. The presenter component listens to po
 
   Implicitly used by the authlib to get the endpoint for the auth-service
 
+* USER_COUNT_MODE
+
+  Determines the way, active users are counted. Options: standard and no_archived_meetings. Default: standard
 
 # Some curl examples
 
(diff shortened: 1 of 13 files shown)


r-peschke:
Actions, where a user will be added to the number of users:
- **unarchive a meeting**
- user.update/create: **adding to group, committee or orga-rights, activate an user**
- user.create:**create an user without giving him a group**
- **on data import not using the standard actions**
- **on saml/sso user addition**
- undelete actions on meeting, group, user, committee

@rrenkert @MSoeb @emanuelschuetze could you have a look on these, maybe there are more

reiterl (commented):
Looks good. The orga management user is user/1.  So all cases there.
If we split the places of the limit check to a *different* issue, I approve this.
But I'm the creator of this pr, so I cannot approve this.

jsangmeister:
I assume that since https://github.com/OpenSlides/openslides-backend/issues/1762 was changed this is not ready for review anymore?

r-peschke:
> I assume that since #1762 was changed this is not ready for review anymore?

yes, the requirements changed

jsangmeister:
Currently too complex to implement. Closing for now.

Issue #1769, opened 2023-06-19
SSO/Saml problems


r-peschke:
- [X] Save boolean values explicitly, true and false, on saml_create if not given. Default: is_active: True, is_physically_person: True. Problem: autoupdate don't read null-Values
Annotation: is_present is not adressed, because currently doesn't exist. We only have a list `is_present_in_meeting_ids`.
- [x]  check boolean values with tests: Test all value variants plus empty, null as False and boolean values itself. See also schema definition. I assume, that the "true" value don't pass here
- [x]  `external_id` for groups should get the english standard names by default on creation of a meeting. In case of self created groups the user has to set the external_id.
- [x]  If a `saml_id` in user account is filled don't let the user set a password in any way, identified actions:   
  - [x] user.forget_password
  - [x] user.forget_password_confirm
  - [x] user.generate_new_password
  - [x] user.reset_password_to_default
  - [x] user.set_password
  - [x] user.set_password_self
  - [x] user.update(can_change_own_password and default_password may not be set)
- [x] On setting a saml_id (action save_saml_account create) set user.can_change_own_password to False and password and default_password to None) 
- [x] If we have an user import with saml_id set can_change_own_password to False and clear password and default_password

**Removing the late method call validate_fields**
After the schema validation we want to sanitize the payload fields immediately, because boolean fields now pass with values in string or int format. We want them to be sanitized (=transformed to type boolean) immediately, because other methods could try to access them as pure boolean fields.
The same call was made later in a method called validate_fields with identical code. For don't calling it twice a test was made:
- Running all tests with old code, asserting every code change by field.validate-methods. We had 17 errors
- Running the same code with additional early sanitizing code. There were no more assert errors in late validating with code changes by field.validate-methods.
 Conclusion: We could remove the late call. There are no more changes by field.validate methods. The coverage rate over 97% helps a lot!

rrenkert:
> 3. `external_id` for groups should get the english standard names by default. In case of self created groups external_id keeps empty   @rrenkert  Shouldn't we use the groupname here?

Since the dialog to create a group has only two fields, the user can copy paste the group name. I think this is not to much efford.

PR #1771, opened 2023-06-21
I1769 saml problems

r-peschke:
see  Issue #1769


r-peschke:
> We should split validation and sanitize!

The validation takes mainly place in schema validation and this is splitted. Sanitize with field.validate runs later, like it is since years, but now on an earlier point in process.

PR #1773, opened 2023-06-22
Fix clone error

jsangmeister:
When cloning a meeting with forwarded motions, the cloning sometimes failed with datastore relation errors. This was because of a wrong configuration of the checker in the clone.

Also cleaned up some other clone-related places.


openslides_backend/action/actions/meeting/clone.py
```diff
@@ -118,28 +112,8 @@ def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
...
-            fields_to_remove={
-                "motion": [
-                    "origin_id",
-                    "derived_motion_ids",
-                    "all_origin_id",
-                    "all_derived_motion_ids",
-                ]
-            },
-        )
+        checker = InternalChecker(meeting_json)
```
r-peschke, line 116: keep the code readable and add here the missing field `origin_meeting_id` instead complicating the code for creating a new class. Parameters are for configuring the instance

openslides_backend/models/checker.py
```diff
@@ -46,6 +46,14 @@
     }
 )
 
+external_motion_fields = [
```
r-peschke, line 49: remove this, see above

openslides_backend/models/checker.py
```diff
@@ -831,3 +839,18 @@ def get_to_generic_case(
         raise CheckException(
             f"The collection {foreign_collection} is not supported as a reverse relation in {collection}/{field}"
         )
+
+
+class InternalChecker(Checker):
```
r-peschke, line 844: remove this also

r-peschke (changes requested):
IMO no good idea building a new class, where the bunch of parameters are hidden. I agree that the construction is far to complicate with all these parameters, but using the checker, the developer should see, what will be done

openslides_backend/models/checker.py
jsangmeister: But this is the main point of the PR: The error occured because the `check_database` presenter and `meeting.clone` needed the same functionality, but implemented them both manually. I fixed in an earlier PR the field list of the presenter, but of course did not think about other places which might hold the same list. By having the field list in one central place, such errors can no longer occur.

jsangmeister:
Yes, parameters are for configuration, but if the configuration is complex and multiple places need exactly the same configuration, they should be unified somehow. I'm not fixed on the class approach if you don't like that, an alternative would be a helper function or a dict which holds the parameters and can be passed to the checker. But we need to find one solution against the code duplication to prevent such errors in the future.

openslides_backend/action/actions/meeting/clone.py
jsangmeister: But the code is simplified: Instead of having to manually pass all the arguments, the tasks are instead divided and `meeting.clone` does no longer need any knowledge of the parameters of the checker.

r-peschke:
> Yes, parameters are for configuration, but if the configuration is complex and multiple places need exactly the same configuration, they should be unified somehow. I'm not fixed on the class approach if you don't like that, an alternative would be a helper function or a dict which holds the parameters and can be passed to the checker. But we need to find one solution against the code duplication to prevent such errors in the future.

Are you sure the presenter should have set repair to True? Like I wrote to much parameters and hiding is no solution. IMO the best would be constant field list used on both places for this parameter.

rrenkert (approved):

jsangmeister:
> Are you sure the presenter should have set repair to True?

This was already this way before hand and is necessary to remove the fields. I'm not sure if it has other repercussions, but I don't think we should discuss this in this issue.

openslides_backend/models/checker.py
r-peschke: The class is not used anymore and you could remove it now

r-peschke (changes requested):
Just a small thing: remove the new class, not used anymore and it's okay for me

openslides_backend/models/checker.py
jsangmeister: My bad, missed it somehow

PR #1779, opened 2023-06-26
Add search_for_id_by_external_id presenter and saml_id to user.create

reiterl:


tests/system/action/user/test_create.py
```diff
@@ -377,6 +377,21 @@ def test_create_strip_spaces(self) -> None:
             },
         )
 
+    def test_create_smal_id_and_default_pasword(self) -> None:
```
r-peschke, line 380: saml and not smal_id:-)

r-peschke, line 380: write a test with saml_id set and empty default_password. Assert an empty_passwort, can_change_own_password False and no password set.

r-peschke (changes requested):

r-peschke:
Raimund wanted to have this in todays update and therefore I fixed my code review remarks. Then they decided to make no update, so we could review this one together.

r-peschke (approved):

Issue #1213, opened 2022-02-15
Use `get_many` instead of `get`/`filter` where possible


jsangmeister:
Gathering some places where `get_many` should be used instead of `get` or `filter`:
- https://github.com/OpenSlides/openslides-backend/blob/b9c48f6900a77664989164b6c994774d7fd1688f/openslides_backend/presenter/get_user_related_models.py#L148-L159
- https://github.com/OpenSlides/openslides-backend/blob/e1f83b6807027ffac64134ee30740b8b4558443e/openslides_backend/action/actions/meeting/export_helper.py#L19-L23

(to be extended)

reiterl:
resource.upload will be deleted in https://github.com/OpenSlides/openslides-backend/pull/1200

PR #1793, opened 2023-07-03
Fix user deletion with archived meetings

### LINKED ISSUE #1792: Issue with participant deletion
**Description:** When an participant is deleted from an active meeting, he or she is also incorrectly deleted from an archived meeting in which he or she participated. With this deletion, all references to the participant disappear in the archived meeting and in the account list.  

**Reproduction:**
Precondition: Two meetings are needed. One is archived.  Participant list on both meetings should include at least one identical account. 
1. Open active meeting
2. go to Participants > delete the user who is part of the archived meeting
3. open archived meeting and check the participant list. 

**What should happen:**
If a user is part in an archived meeting, the deletion should only takes part in the active meeting, where he or she is deleted from and not global. The deletion rule for archived meeting should be the same like for active meetings.

jsangmeister:
fixes #1792 

Changed the user scope calculation to also take archived meetings into account, since meeting admins should not be allowed to delete users which are in their meeting plus in other archived meetings.

### DIFF
--- a/tests/system/presenter/test_get_user_related_models.py
+++ b/tests/system/presenter/test_get_user_related_models.py
@@ -11,7 +11,11 @@ class TestGetUserRelatedModels(BasePresenterTestCase):
     def test_get_user_related_models_simple(self) -> None:
         status_code, data = self.request("get_user_related_models", {"user_ids": [1]})
         self.assertEqual(status_code, 200)
-        assert data == {"1": {}}
+        assert data == {
+            "1": {
+                "organization_management_level": OrganizationManagementLevel.SUPERADMIN,
+            }
+        }
 
     def test_get_user_related_models_committee(self) -> None:
         self.set_models(
@@ -29,7 +33,10 @@ class TestGetUserRelatedModels(BasePresenterTestCase):
         status_code, data = self.request("get_user_related_models", {"user_ids": [1]})
         self.assertEqual(status_code, 200)
         assert data == {
-            "1": {"committees": [{"id": 1, "cml": "can_manage", "name": "test"}]}
+            "1": {
+                "organization_management_level": OrganizationManagementLevel.SUPERADMIN,
+                "committees": [{"id": 1, "cml": "can_manage", "name": "test"}],
+            }
         }
 
     def test_get_user_related_models_committee_more_user(self) -> None:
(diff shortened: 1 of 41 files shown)


r-peschke (changes requested):
- I'm not convinced that we should change the user scope in this case, but that's to discuss.
- Please fix the checks
- if we do it this way, change the doc in wiki

jsangmeister:
Updated the wiki: https://github.com/OpenSlides/OpenSlides/wiki/Users did not mention archived meetings anyway and https://github.com/OpenSlides/OpenSlides/wiki/Archived-Meetings had one hint, which I removed. Also fixed the tests.

jsangmeister:
@r-peschke if you update the priority of the PR, you also have to update the priority of the issue. The `critical` evaluation came from @MSoeb, I just copied it from the issue

r-peschke:
Wrote comment into the issue as answer to the problem description.

jsangmeister:
Reverted the wiki changes. Will update this PR with the changes discussed in https://github.com/OpenSlides/openslides-backend/issues/1795 and https://github.com/OpenSlides/openslides-backend/issues/1792

jsangmeister:
Updated the PR with the presenter changes. Also finally implemented the improvement suggested in https://github.com/OpenSlides/openslides-backend/issues/1213 to reduce the amount of filter requests (3 instead of 3 times the amount of meetings for each user).

tests/system/presenter/test_get_user_scope.py
```diff
@@ -11,22 +11,22 @@ def test_good(self) -> None:
...
             {
                 "meeting/1": {
-                    "name": "meeting1",
                     "committee_id": 2,
                     "is_active_in_organization_id": 1,
                 },
+                # archived meeting
+                "meeting/2": {
+                    "committee_id": 2,
+                    "is_active_in_organization_id": None,
```
r-peschke, line 20: To be correct `is_archived_in_organization_id` is necessary to determine an archived meeting

tests/system/presenter/test_get_user_related_models.py
```diff
@@ -173,24 +183,72 @@ def test_get_user_related_models_no_permissions(self) -> None:
...
+            "2": {
+                "meetings": [
+                    {
+                        "id": 1,
+                        "name": "test",
+                        "is_active_in_organization_id": None,
+                    }
+                ]
+            },
+        }
```
r-peschke, line 251: The result still misses the `organization_management_level: OML-String,` info, which is mentioned in the presenter doc. The existing test **test_get_user_related_models_permissions** , beginning at line 279, should have such a result entry, but it is missed and not checked.

r-peschke (approved):
Just some smaller changes

r-peschke (changes requested):
See 3 small changes in the comments

tests/system/presenter/test_get_user_related_models.py
jsangmeister: Somehow I thought that the presenter already returned the OML. Added it now

jsangmeister:
Added the OML, fixed the tests & comments and reduced the amount of datastore requests in the presenter. Review again @r-peschke

r-peschke (approved):

jsangmeister:
This still needs changes in the client as described in https://github.com/OpenSlides/openslides-backend/issues/1792#issuecomment-1623203037

PR #1799, opened 2023-07-06
Fix weight assignment in agenda_item.create

### LINKED ISSUE #1797: Agenda: Numbering problem
**Description:** Incorrect numbering occurs sometimes when numbering agenda items. The problem seems to occur when agenda items are newly created and no editing of the items has taken place. Also i think it occurs on fresh meetings more often. The problem is, that I can't reproduce it. 

Talked with @bastianjoel about it. He has a guess as to what the problem might be. @bastianjoel Could you pleae add a comment about your clue?

Addtionally @Elblinator seems to have a clue how to trigger it.

**Possible Reproduction:**
1. create new meeting
2. open meeting
3. open agenda
4. add four agenda items. Just add a name and save.
5. open agenda menu again > click on the three-dot-menu > choose numbering 
6. if it doesn't happen in first instance try to repeat step 5

**What should happen?**
The numbering should alwasy number the the agenda items in the correct order.

jsangmeister:
fixes #1797

### DIFF
--- a/global/meta/models.yml
+++ b/global/meta/models.yml
@@ -1944,7 +1944,6 @@ agenda_item:
     restriction_mode: A
   weight:
     type: number
-    default: 10000
     restriction_mode: A
 
   content_object_id:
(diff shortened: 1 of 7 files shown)


r-peschke (changes requested):
There is an error in the implementation, 1 test fails, see .github/tests

r-peschke (approved):

PR #1803, opened 2023-07-07
0044 migrate nothing

r-peschke:


reiterl:
What is the reason for this migration, which does nothing?

r-peschke:
Just temporary for customer

PR #1807, opened 2023-07-10
Add saml_id to search_users presenter

reiterl:


openslides_backend/presenter/search_users.py
```diff
@@ -100,9 +100,11 @@ def get_result(self) -> List[List[Dict[str, Any]]]:
                     if all(
                         (instance.get(field) or "").lower() == search[field]
                         for field in search_def
-                    ):
+                    ) and any(search[field] for field in search_def):
                         current_result.append(instance)
                         break
+                    elif any(search[field] for field in search_def):
+                        break
```
jsangmeister, line 107: Deduplicate the if-clause:
```suggestion
                    if any(search[field] for field in search_def):
                        if all(
                            (instance.get(field) or "").lower() == search[field]
                            for field in search_def
                        ):
                            current_result.append(instance)
                        break
```
Also maybe add a comment explaining the any/all statements - I found them quite confusing to understand on the first read.

jsangmeister (changes requested):

openslides_backend/presenter/search_users.py
reiterl: Deduplicate the any clause. Done.

jsangmeister (approved):

r-peschke (approved):
Just a small thing: Add the saml_id to the comment of the presenter class. Then you can merge without reviewing again:-)

PR #1810, opened 2023-07-11
Add `external_id` to `meeting.clone`

### LINKED ISSUE #1809: Add `external_id` to `meeting.clone`
Analoguous to `meeting.create` and analoguous to the other fields in the payload of `meeting.clone`, it should also be able to set  an `external_id`. Currently, it is not possible in the client to clone a meeting since the form is the same as the one for `meeting.create`.

jsangmeister:
fixes #1809

### DIFF
--- a/global/meta/models.yml
+++ b/global/meta/models.yml
@@ -1944,7 +1944,6 @@ agenda_item:
     restriction_mode: A
   weight:
     type: number
-    default: 10000
     restriction_mode: A
 
   content_object_id:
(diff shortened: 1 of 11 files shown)


emanuelschuetze (approved):

r-peschke (changes requested):
The external_id must be unique. This will not be checked in clone code and there is also no test, which would, fail, for this case

jsangmeister:
Also unified the behaviour of some meeting actions regarding unique check and permissions.

Issue #1801, opened 2023-07-06
Error in relation handling with multiple action data items


jsangmeister:
While fixing #1799 I came across another bug which seems to have been present for a long time:
https://github.com/OpenSlides/openslides-backend/blob/e782cd20918134d84748457e2492166aa5037e1f/tests/system/action/agenda_item/test_create.py#L58-L85

In the test above, two agenda items are created which both point to the same topic (I checked that this is really the case, as the test does not assert this). But the relation is a 1-1 relation, the topic afterwards points only to `agenda_item/44`. That means the relation field `agenda_item/43/content_object_id` points to `topic/1`, but without a valid back relation.

jsangmeister:
This will be extremely complex to fix. The problem is the the `changed_models` are not used consistently throughout the code base. The initial though was that _after_ an action is finished, the results are written to the `changed_models` so that they are available for following actions. This is, however, not always done consequently and so the `changed_models` are sometimes used as a cache or change propagation system. This results in the relation handling accessing the datastore directly instead of via the changed models when calculating the relation diffs:
https://github.com/OpenSlides/openslides-backend/blob/8074bd211e7824944e7870900e3f55c70ee455b8/openslides_backend/action/relations/single_relation_handler.py#L271-L276
This leads to the bug described above.

I'm not sure if and how to proceed here...

jsangmeister:
Closing this as it does not seem to affect the productive instances and this will be solved with the relational db.

PR #1816, opened 2023-07-18
Some cleanup & typing improvements

jsangmeister:
Improvements done while investigating https://github.com/OpenSlides/openslides-backend/issues/1801


jsangmeister:
@r-peschke ping

openslides_backend/action/actions/meeting/import_.py
```diff
@@ -575,7 +575,7 @@ def create_events(
         )
 
         # add meetings to organization if set in meeting
-        adder: Dict[str, List[Union[int, str]]] = {}
+        adder: ListFieldsValue = {}
```
r-peschke, line 578: Please revert this file. A type ListFieldsValue for a dict is the opposite of readable

r-peschke (changes requested):
Just one thing

openslides_backend/action/actions/meeting/import_.py
r-peschke: I still think the best is to revert it.
I asked myself, why it is a union? One would expect the meeting_id to be always int.

r-peschke (changes requested):

openslides_backend/action/actions/meeting/import_.py
jsangmeister: `Dict` is [invariant](https://mypy.readthedocs.io/en/stable/common_issues.html#variance) and the `list_fields` can potentially be used with integers and strings, therefore, the type used here must be a union as well. But this is actually an argument *for* a type alias: You see a union of strings and integers and are confused at first. If we just use some type alias, this confusion can be avoided. Additionally, this rather verbose type is used in 3 separate places, so it also goes against the DRY principle to repeat the type. As I said, I'm open for another name suggestion if `ListFieldsValue` does not suit you.

jsangmeister: renamed to `ListFieldsDict`

r-peschke (approved):

PR #1819, opened 2023-07-18
Add tag_ids to topic.create

reiterl:
If you add `attachment_ids` or `tag_ids` it will be checked, that they are from the same meeting as the new created topic.
I add a test for tag_ids to check this.


jsangmeister (changes requested):
Also add a successful test where you set the `tag_ids`.

PR #1626, opened 2023-02-09
Add an unique check for motion_category.prefix.

reiterl:
And add tests.

The first part of the linked issue.

PR #1828, opened 2023-07-24
Revert "Add an unique check for motion_category.prefix. (#1626)"

reiterl:
This reverts commit 787555a3fc37b741469ae2badf0d3ae63d41b66a.


jsangmeister (changes requested):
Leave the tests in and change them to check that the request is successful.

jsangmeister (approved):

PR #1833, opened 2023-07-26
Add two options to workflow import

reiterl:
Add the options `set_workflow_timestamp` and `allow_motion_forwarding` to the motion_workflow.import.

### DIFF
--- a/tests/system/action/motion_workflow/test_import.py
+++ b/tests/system/action/motion_workflow/test_import.py
@@ -39,7 +39,13 @@ class MotionWorkflowImport(BaseActionTestCase):
                 "name": "test_Xcdfgee",
                 "meeting_id": 42,
                 "first_state_name": "begin",
-                "states": [self.get_state("begin", [], [])],
+                "states": [
+                    {
+                        **self.get_state("begin", [], []),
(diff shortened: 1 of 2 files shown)


tests/system/action/motion_workflow/test_import.py
```diff
@@ -39,7 +39,13 @@ def test_import_simple_case(self) -> None:
                 "name": "test_Xcdfgee",
                 "meeting_id": 42,
                 "first_state_name": "begin",
-                "states": [self.get_state("begin", [], [])],
+                "states": [
+                    {
+                        **self.get_state("begin", [], []),
+                        "set_workflow_timestamp": True,
+                        "allow_motion_forwarding": True,
```
jsangmeister, line 46: Add these to `get_state` instead of only this payload

jsangmeister (changes requested):

tests/system/action/motion_workflow/test_import.py
reiterl: Okay, I have added the two options to `get_state`.

jsangmeister (approved):

PR #1839, opened 2023-08-02
Add test case for participant delete error

reiterl:
This test case should be used for further debugging. So I provide it here.


jsangmeister:
The "cleaner" solution would be to rework the committee user calculated fields handler, but this would be much more work and this solution also works.

PR #1859, opened 2023-08-09
Fix of a problem with calculated user/committee_ids

reiterl:
Add a test for the problem and a possible fix.


jsangmeister (approved):
also change the docstring of the method

r-peschke (approved):

PR #1860, opened 2023-08-09
change saml_private_key type to text

peb-adr:
In order to allow adequate saml private keys to be set, we need more than 256 chars (which ist `string`s limt)

This was rebased to main and models newly generated for correct checksum.


peb-adr:
autoupdate and client do not need a PR for the change in models, because their generated code doesn't distinguish between `string` and `text` types

rrenkert (approved):

jsangmeister (approved):

PR #1861, opened 2023-08-10
Fix agenda item update

reiterl:
Add a test to show the problem.

### DIFF
--- a/openslides_backend/action/actions/agenda_item/update.py
+++ b/openslides_backend/action/actions/agenda_item/update.py
@@ -1,3 +1,4 @@
+from operator import itemgetter
 from typing import Any, Dict, Optional
 
 from ....models.models import AgendaItem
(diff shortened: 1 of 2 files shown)


jsangmeister (dismissed):
Two remarks:
- You don't need to implement the parent cache manually. Just call `apply_instance` everytime you add an instance to the list (so in lines 77 and 119) and use the `get` method as before.
- This solution does not work if the order of the items is changed in the payload. A possible fix is to sort the items at the start of the `get_updated_instances` method by their `level`, so that lower level items are handled first (meaning the new children values will override the calculated values from the parents). This still means there is some redundancy becase if multiple items high up in the chain are updated, the last child item will get updated with every other item, but I don't think this poses a real performance problem which we should tackle.

reiterl:
* Use apply instances to cache the data.
 * Add a level sorting for the instances, instances with a changed parent are delayed.

r-peschke (approved):
Ludwigs changes concerning the apply_instances were correct and solved the problem.
The sort isn't necessary, because the 2 triggers (change type with parent consideration and child handling) don't interfere each other in their main competence. IMO this is covered with more complex testcases in a sufficient way.

Issue #1886, opened 2023-08-25
Error in get_user_scope called from /user/create_update_permissions_mixin.py", line 201, in check_permissions


r-peschke:
Excerpts from  monitoring openslides traceback-mon from 25.8.23 11:00:
```
Traceback (most recent call last):
  [...]
  File "/app/openslides_backend/action/action_handler.py", line 115, in handle_request
    results = self.execute_write_requests(self.parse_actions, payload)
  File "/app/openslides_backend/action/action_handler.py", line 156, in execute_write_requests
    write_requests, data = get_write_requests(*args)
  File "/app/openslides_backend/action/action_handler.py", line 191, in parse_actions
    write_request, results = self.perform_action(
  File "/app/openslides_backend/action/action_handler.py", line 238, in perform_action
    write_request, results = action.perform(
  File "/app/openslides_backend/action/action.py", line 161, in perform
    self.check_permissions(instance)
  File "/app/openslides_backend/action/actions/user/create_update_permissions_mixin.py", line 201, in check_permissions
    ) = self.get_user_scope(instance.get("id") or instance)
  File "/app/openslides_backend/shared/mixins/user_scope_mixin.py", line 48, in get_user_scope
    "meeting_user", id_or_instance["id"]
KeyError: 'id'
```

Reason: Error during transformation "remove template fields"
create_update_permissions_mixin.py handles user-instances, in this case I assume it's a user user.create action, because of missing the `id`. In get_user_scope this instance is assumed of being a meeting_user instance (The `id` will be used to read meeting_user instances)
Error 1: There is no `id` in instance
Error 2: If there is an `id` in instance, it is treated as a meeting user id which leads to wrong results of user scope checking, but usually no exception

PR #1887, opened 2023-08-25
fix key error in get_user_scope

r-peschke:
see #1886


jsangmeister (changes requested):
Can you add a test for this case? Otherwise looks fine

jsangmeister (approved):

PR #1889, opened 2023-08-29
Issue1880 small import changes

r-peschke:
Uses the Lookup class from import_mixins.py to check for duplicate sin account_json_upload


global/meta/models.yml
```diff
@@ -3825,6 +3825,9 @@ action_worker:
       - running
       - end
       - aborted
+      - warning
+      - error
+      - done
```
jsangmeister, line 3830: Are you sure you want to add these to the general action worker? What's e.g. the difference between `end` and `done`? `warning` and `error` are really import-specific and do not convey meaning about the status of the action worker.

openslides_backend/action/mixins/import_mixins.py
```diff
@@ -1,27 +1,147 @@
...
+    FOUND_MORE_IDS = 2
+    NOT_FOUND = 3
+
+
+class Lookup:
+    def __init__(
+        self,
+        datastore: DatastoreService,
+        collection: str,
+        name_entries: List[Tuple[Union[str, Tuple[str, ...]], Dict[str, Any]]],
```
jsangmeister, line 43: Please use type aliases for types which you use multiple times

openslides_backend/action/mixins/import_mixins.py
```diff
@@ -1,27 +1,147 @@
...
+        if len(values := self.name_to_ids.get(name, [])) == 1:
+            if values[0].get("id"):
+                return ResultType.FOUND_ID
+            else:
+                return ResultType.NOT_FOUND
+        elif len(values) > 1:
+            return ResultType.FOUND_MORE_IDS
+        raise ActionException("Logical Error in Lookup Class")
+
+    def get_x_by_name(
```
jsangmeister, line 104:
```suggestion
    def get_field_by_name(
```
x does not convey any meaning of the intended function of the method

openslides_backend/action/mixins/import_mixins.py
```diff
@@ -43,8 +163,12 @@ def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
             raise ActionException(
                 f"Wrong id doesn't point on {self.import_name} import data."
             )
+        if worker.get("state") not in ImportState.__members__.values():
```
jsangmeister, line 166:
```suggestion
        if worker.get("state") not in iter(ImportState):
```
should work as well. I would not use `__*__` methods if not absolutely necessary.

openslides_backend/action/mixins/import_mixins.py
```diff
@@ -85,29 +209,27 @@ def on_success() -> None:
...
-
 class StatisticEntry(TypedDict):
     name: str
     value: int
 
 
-class JsonUploadMixin(Action):
-    headers: List[HeaderEntry]
+class JsonUploadMixin(BaseImportJsonUpload):
+    headers: List[Dict[str, Any]]
```
jsangmeister, line 218: Why did you remove the typing here?

openslides_backend/action/mixins/import_mixins.py
```diff
@@ -147,21 +269,41 @@ def create_action_result_element(
...
+            for header in self.headers
         }
         for entry in list(instance.get("data", [])):
             for field in dict(entry):
                 if field not in property_to_type:
                     del entry[field]
                 else:
-                    type_ = property_to_type[field]
-                    if type_ == "integer":
+                    type_, is_object, is_list = property_to_type[field]
```
jsangmeister, line 295: `is_object` is not used if I see it correctly

jsangmeister (changes requested):

global/meta/models.yml
r-peschke: I agree and we knew, that this is a misuse of the action_worker. We did it with the idea that this is the faster way to develop the imports. It should be a separate collection, but also to be written in a separate transaction in parallel to the data. We will separate action_worker and its functionality from preview_import-worker on moving to relational db-design.
The meaning of the action_worker states is not compatible with that of the preview_import.
`end` in action_worker means, that the action_worker ended and can be deleted from database. The last point is, why we didn't wanted to use this state for the preview_import. `abort` may be similar.
We could use always `running`, maybe with some statistical problems or deleting long running action workers.

`done` in preview_import means, that everything is okay without warnings (could be imported) or error (whole import may not be imported). A program, cleaning or statistic job can easily decide by the state if it is an action worker or a preview_import.

openslides_backend/action/mixins/import_mixins.py
r-peschke: done

openslides_backend/action/mixins/import_mixins.py
r-peschke: It is only used here and Dict[str, Any] is easy to understand

openslides_backend/action/mixins/import_mixins.py
r-peschke: it is used in the ActionException to descripe an unknown type

openslides_backend/action/mixins/import_mixins.py
r-peschke: Your proposal ends in a mypy error:

global/meta/models.yml
jsangmeister: Yes, but I don't think we should mix the stati of action workers and imports. I would suggest to leave the state of an import inside the data saved in the action worker and leave the state at running until the import is either sucessful or aborted. If we implement some cleanup tool for action workers, it can, as you said, easily decide for every action worker still running whether it is an actual action worker or an import and then decide to keep or delete it. To summarize: We already have a spearate import state inside the data, I would prefer not to mix it with the action worker's state.

openslides_backend/action/mixins/import_mixins.py
jsangmeister: That's unfortunate. Enums and mypy do not seem to get along very well... Then, I would suggest to either use `# type: ignore`/`cast(Iterator[str], iter(ImportState))` or use the following construct:
```
try:
    ImportState(worker.get("state"))
except ValueError:
    raise ActionException(...)
```
although I agree that that's barely better than accessing `__members__`...

r-peschke: I think, the __members__ variant is the clearest and shortest and I want you to accept my opinion

global/meta/models.yml
r-peschke: There is no separate state inside the data, there is only a row state. Lets talk about this, I'm not willing to change

openslides_backend/action/mixins/import_mixins.py
jsangmeister: The mypy error is a bug in mypy: https://github.com/python/mypy/issues/16089. If you use e.g. `list(ImportState)`, it works without errors. I think that this is the clearest and most concise solution.

r-peschke: Changed to `list(ImportState)`

global/meta/models.yml
r-peschke: New collection import_preview introduced and not using the action_worker for this anymore.
I'll close this in favor of https://github.com/OpenSlides/openslides-backend/pull/1912

r-peschke:
Closing this one in favor of https://github.com/OpenSlides/openslides-backend/pull/1912.
The last 2 items, reusing HeaderEntry and separating the import_preview data from action_worker-collection to new collection.

PR #1893, opened 2023-09-05
Catch all exceptions during migration

jsangmeister:
If exceptions other than `MigrationException` happen during a migration, they are currently not logged anywhere. They should instead be logged.

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -18,7 +18,7 @@ jobs:
         working-directory: .github/docker-compose
 
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v4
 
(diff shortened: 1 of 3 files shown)


jsangmeister:
> What's about the yml changes. Please write an issue if you change there.

It's a simple version upgrade, I don't think a separate issue is needed here. If you wish, I could extract it into its own PR, but I think that's overkill. We should think about adding a `dependabot.yml` file analogous to e.g. the autoupdate service: https://github.com/OpenSlides/openslides-autoupdate-service/blob/main/.github/dependabot.yml

> For the migration part: I miss the reason why you do this and/or a test.

This was not tested before and is really hard to test, so I didn't add a test. The reason why I did this is linked earlier in this thread.

Issue #1888, opened 2023-08-29
Make `meeting_user` actions backend internal


jsangmeister:
The meeting user actions should not be accessible from outside the backend. The client does not and should not use them as the user actions encapsulate them for the client.

Also, the payload of `meeting_user.{create|update}` should be reduced and cleaned up. I don't see any necessity to be able to set the various relations via these actions, especially not together with the action currently being public.

reiterl:
Which relations should be excluded from the meeting_user create and update actions?
I think, we need a way to update them. So why excluding them?

jsangmeister:
> Which relations should be excluded from the meeting_user create and update actions? I think, we need a way to update them. So why excluding them?

Most are updated through the reverse relation. Here is a list of all places where the actions are used and which fields are sent (besides the required fields):
- `meeting_user.create`:
  - https://github.com/OpenSlides/openslides-backend/blob/6195994174e549f3c7f16ae648027e557eda5fcc/openslides_backend/action/actions/meeting/create.py#L199 `group_ids`
  - https://github.com/OpenSlides/openslides-backend/blob/6195994174e549f3c7f16ae648027e557eda5fcc/openslides_backend/action/actions/meeting/create.py#L212 `group_ids`
  - https://github.com/OpenSlides/openslides-backend/blob/6195994174e549f3c7f16ae648027e557eda5fcc/openslides_backend/action/actions/meeting_user/helper_mixin.py#L16 none
  - https://github.com/OpenSlides/openslides-backend/blob/6195994174e549f3c7f16ae648027e557eda5fcc/openslides_backend/action/actions/motion/create_forwarded.py#L100 `group_ids`
  - https://github.com/OpenSlides/openslides-backend/blob/6195994174e549f3c7f16ae648027e557eda5fcc/openslides_backend/action/actions/motion/create_forwarded.py#L136 `group_ids`
  - https://github.com/OpenSlides/openslides-backend/blob/6195994174e549f3c7f16ae648027e557eda5fcc/openslides_backend/action/actions/personal_note/create.py#L62 `personal_note_ids`
- `meeting_user.update`:
  - https://github.com/OpenSlides/openslides-backend/blob/6195994174e549f3c7f16ae648027e557eda5fcc/openslides_backend/action/actions/motion/create_forwarded.py#L111 `group_ids`
[...]

PR #1895, opened 2023-09-06
Make meeting_user actions backend internal

reiterl:
First part fix  of the https://github.com/OpenSlides/openslides-backend/issues/1888


jsangmeister (changes requested):
see https://github.com/OpenSlides/openslides-backend/issues/1888#issuecomment-1709726596

reiterl:
meeting_user delete is backend internal. Remove most relation fields form the meeting_user create and update actions.
Remove the delegate tests, they are no longer needed.

jsangmeister:
> meeting_user delete is backend internal

My bad, missed that

> Remove the delegate tests, they are no longer needed.

Yes they are! Move the code to `meeting_user.set_data` and the test to the according test file. Currently it seems like the relation can just be set without any checks... This should be caught by the user tests. We should check that no other such checks were broken with the template field removal.

jsangmeister (commented):
Not sure about the removed permission checks in the mixin

jsangmeister (approved):

reiterl:
I talked with @jsangmeister about the check_permissions. I will merge this next week.

PR #1898, opened 2023-09-07
Revert "Support point of order speakers in re-add-speaker"

peb-adr:
Reverts OpenSlides/openslides-backend#1894


peb-adr:
The client does not yet have support for this. We should merge it at a later point again.

PR #1901, opened 2023-09-11
Remove all topic.tag_ids from database

r-peschke:
This migrations enables meeting clone, because otherwise the existing topic.tag_ids are throwing validation exception

### DIFF
--- a/openslides_backend/migrations/migrations/0046_fix_topic_tag_ids_error.py
+++ b/openslides_backend/migrations/migrations/0046_fix_topic_tag_ids_error.py
@@ -1,0 +1,29 @@
+from typing import Any, Dict, List, Optional
+
+from datastore.migrations import BaseModelMigration
+from datastore.writer.core import BaseRequestEvent, RequestUpdateEvent
+
+from openslides_backend.shared.patterns import fqid_from_collection_and_id
+
+
+class Migration(BaseModelMigration):
+    """
+    Remove all topic.tag_ids from database
+    """
+
+    target_migration_index = 47
+
+    field = "tag_ids"
+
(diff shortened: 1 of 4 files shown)


openslides_backend/migrations/migrations/0046_fix_topic_tag_ids_error.py
```diff
@@ -0,0 +1,27 @@
...
+    Remove all topic.tag_ids from database
+    """
+
+    target_migration_index = 47
+
+    field = "tag_ids"
+
+    def migrate_models(self) -> Optional[List[BaseRequestEvent]]:
+        events: List[BaseRequestEvent] = []
+        db_models = self.reader.get_all("topic")
```
jsangmeister, line 20: Actually, using `filter()` here should be more performant than `get_all()`. The migration reader uses the normal reader methods, so they are translated into SQL, so we filter by query instead of by data. We could also update migration 45 to use this.

jsangmeister (changes requested):

openslides_backend/migrations/migrations/0046_fix_topic_tag_ids_error.py
r-peschke: Filter like 
```
        db_models = self.reader.filter(
            collection="topic",
            filter=FilterOperator("tag_ids", "=", []),
            mapped_fields=["id", "tag_ids"],
        )

```
doesn't find anything. Tried with a value set in the array, but still find nothing.

reiterl:
The get_all is *okay* for now.
Your assumption is, that the empty list in the topic/tag_ids is the problem and that no filled list is in the data.

PR #1915, opened 2023-09-21
Fix opentelemetry in backend

r-peschke:
The problem is the tracer_provider, which can be set globally in a service only one time.
The backend with the imported datastore-reader is doing this more times:
in main.py line 9 this will be done for the datastore-reader, in line 83 this is done for the backend.
The change prevents the backend to use it's own, therefore all data is written with reader-server.

2nd problem: the reader as lib in backend can only be enabled/disabled together with the backend. We need a second env var for enabled/disabling it as a lib. Could be done in datastore code environment_service this way:
```
    def is_otel_enabled(self) -> bool:
        value = self.try_get("OPENTELEMETRY_DATASTORE_LIB_ENABLED")
        if value is not None:
            return self.is_truthy(value)
        value = self.try_get(OTEL_ENABLED_ENVIRONMENT_VAR)
        return self.is_truthy(value)
```

3rd problem: we want a trace span for more than 1 service. See a possible solution [here](https://stackoverflow.com/questions/66743389/how-to-use-opentelmetry-with-multiple-microservices-python)


jsangmeister (changes requested):
I would rather solve the problem by making the datastore configurable so that it does not init OTEL if not needed, e.g. with an additional boolean parameter to `register_services()`. This way, we do not need to touch the internal gloabl OTEL variables and do not need a second env var for the datastore. This requires to rewrite https://github.com/OpenSlides/openslides-datastore-service/blob/main/datastore/reader/app.py as well, but it should not be too hard.

peb-adr (commented):
I don't think setting the `DATASTORE_COMMIT_HASH` here make sense yet.
Once the PR in datastore is (squash-)merged it will get a new hash on main, which then should be set here.
Other than that LGTM.

r-peschke:
> I don't think setting the `DATASTORE_COMMIT_HASH` here make sense yet. Once the PR in datastore is (squash-)merged it will get a new hash on main, which then should be set here. Other than that LGTM.

done

r-peschke:
> I would rather solve the problem by making the datastore configurable so that it does not init OTEL if not needed, e.g. with an additional boolean parameter to `register_services()`. This way, we do not need to touch the internal gloabl OTEL variables and do not need a second env var for the datastore. This requires to rewrite https://github.com/OpenSlides/openslides-datastore-service/blob/main/datastore/reader/app.py as well, but it should not be too hard.

done

PR #1921, opened 2023-10-05
Introduce speaker deletion logic to user.update and user.delete

### LINKED ISSUE #1917: Don't delete past speakers upon user.delete
For background see OpenSlides/openslides-client#2639

Upon deletion of a user, any speaker instance related to him is currently also deleted.
This is fine for speakers who are still in the waiting list, but those who are either currently speaking (i.e. `start_time` is defined) or finished (i.e. `end_time` is defined) should not be deleted for the sake of keeping a coherent protocol.

So:
If a user is deleted:
- All speakers of this user with neither a defined `start_time` or `end_time` should be deleted.
- All speakers of this user with either a defined `start_time` or `end_time` should have the user removed, but should not be deleted.

**Also:**
While testing we've noticed that removing a user from a meeting via removal from the relevant groups doesn't remove him from the meetings speaker lists either.
This should show a similar behaviour to the deletion:

If a user is removed from a meeting:
- All speakers of this user in this meeting with neither a defined `start_time` or `end_time` should be deleted.
- All speakers of this user in this meeting with either a defined `start_time` or `end_time` should not be deleted. _In this case the user should NOT be removed from the speaker though._

luisa-beerboom:
Closes #1917

### DIFF
--- a/openslides_backend/action/actions/user/conditional_speaker_cascade_mixin.py
+++ b/openslides_backend/action/actions/user/conditional_speaker_cascade_mixin.py
@@ -1,4 +1,4 @@
-from typing import Any, Dict, List
+from typing import Any, Dict
 
 from ....services.datastore.commands import GetManyRequest
 from ....shared.patterns import fqid_from_collection_and_id
@@ -12,9 +12,9 @@ class ConditionalSpeakerCascadeMixin(Action):
     """
 
     def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
-        removed_meeting_ids = self.get_meetings_removed(instance)
-        if len(removed_meeting_ids) != 0:
-            delete_all = 0 in removed_meeting_ids
+        removed_meeting_id = self.get_removed_meeting_id(instance)
+        if removed_meeting_id is not None:
+            delete_all = removed_meeting_id == 0
             user = self.datastore.get(
                 fqid_from_collection_and_id("user", instance["id"]),
                 [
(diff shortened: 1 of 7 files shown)


tests/system/action/user/test_update.py
```diff
@@ -2068,3 +2068,67 @@ def test_update_saml_id_default_password_error(self) -> None:
...
+        self.assert_status_code(response, 200)
+        self.assert_model_exists(
+            "user/1234",
+            {
+                "username": "username_abcdefgh123",
+                "meeting_user_ids": [2345, 2346],
+            },
+        )
+        self.assert_model_exists(
+            "meeting_user/2345", {"group_ids": [], "meta_deleted": False}
```
r-peschke, line 2126: also include the speaker_ids in this assert

openslides_backend/action/actions/user/conditional_speaker_cascade_mixin.py
```diff
@@ -0,0 +1,78 @@
...
+class ConditionalSpeakerCascadeMixin(Action):
+    """
+    Mixin for user actions that deletes unstarted speeches of users that were either deleted, or removed from a meeting
+    """
+
+    def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
+        removed_meeting_id = self.get_removed_meeting_id(instance)
+        if removed_meeting_id is not None:
+            delete_all = removed_meeting_id == 0
+            user = self.datastore.get(
```
r-peschke, line 18: You could read in1 call and only the needed ones with:
```python
filter_ = FilterOperator("user_id", "=", instance["id"])
if removed_meeting_id:
    filter_ = And(filter_, FilterOperator("meeting_id", "=", removed_meeting_id)
meeting_user_ids = self.datastore.filter("meeting_user", filter_, ["speaker_ids"])
```
and you read only the needed meeting_user_ids in 1 call

r-peschke (changes requested):

r-peschke (approved):
LGTM

PR #1924, opened 2023-10-17
Add `dependabot.yml`

### LINKED ISSUE #1899: Add `dependabot.yml`
E.g. the autoupdate service has the following file: https://github.com/OpenSlides/openslides-autoupdate-service/blob/main/.github/dependabot.yml. I think this would be a good addition to the backend and would take over work which we currently do manually. Also, it eases the update workflow because each updated is applied for itself and not in a batch with all other updates.

jsangmeister:
resolves #1899 

Could not test this but https://github.com/dependabot/dependabot-core/issues/3940 assured me that non-standardly names requirement files are discovered as well, so this should work as intended.


r-peschke (approved):

r-peschke:
Approved, but fix the failing check

PR #1928, opened 2023-10-17
I1927 Accelerate poll.stop by reducing locked fields

r-peschke:
Removes unnecessary lock_fields for memory and performance issues. For not being punished with performance penalty and error prune changes in manual optimization of database calls, the caching ability is still used with locking calls.

Waiting for approval and decision, if this change should be taken to main branch or  only be used for special cases


jsangmeister:
This solution is a little too specific and hacky for this situation, I think. But the general idea is good. How about we mark each field that will never change its value as `constant` in the `models.yml` and then never lock those fields in the datatstore adapter of the backend? This way, the solution would be nice and generic and also benefit all other actions instead of only `poll.stop`.

The "cleaner" solution in some sense would of course be to not even request those fields to be locked, but that would be very cumbersome to implement.

jsangmeister:
Opened https://github.com/OpenSlides/openslides-backend/issues/2172 to implement a cleaner solution of this. Closing this PR for now.

r-peschke:
approximately 50% acceleration reported

PR #1935, opened 2023-10-26
Fix speaker.delete and speaker.update for deleted users

### LINKED ISSUE #1934: Account deletion: speech in LoS cannot be deleted
**Description:**
Follow-up to #1917 
If a user is deleted their Metadata stays intact (yay)  but cannot be deleted anymore.

**Wanted:**
If the contribution (of a deleted user) is removed from a LoS then the contrubution should actually be deleted

Console if the contribution is deleted:
012 A

Payload:
```[{action: "speaker.delete", data: [{id: 830}]}]```

luisa-beerboom:
Closes #1934 

Wrote tests and fixed the action.

Found the same problem with `speaker.update`, so I did the same there

### DIFF
--- a/openslides_backend/action/actions/speaker/delete.py
+++ b/openslides_backend/action/actions/speaker/delete.py
@@ -20,11 +20,12 @@ class SpeakerDeleteAction(DeleteAction):
             ["meeting_user_id"],
             lock_result=False,
         )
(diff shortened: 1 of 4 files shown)


r-peschke (approved):
LGTM:-)

PR #1937, opened 2023-10-27
I1936_fix_import_lookup_duplicates

r-peschke:
Error with 2 records in import with same search value, i.e. title in topic or username in account, that will find an existing object. This should be identified as duplicate

### DIFF
--- a/openslides_backend/action/mixins/import_mixins.py
+++ b/openslides_backend/action/mixins/import_mixins.py
@@ -75,7 +75,7 @@ class Lookup:
         or_filters: List[Filter] = []
         if "id" not in mapped_fields:
             mapped_fields.append("id")
-        if type(field) == str:
+        if type(field) is str:
             if field not in mapped_fields:
                 mapped_fields.append(field)
             if name_entries:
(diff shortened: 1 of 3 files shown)


openslides_backend/action/mixins/import_mixins.py
```diff
@@ -66,8 +66,11 @@ def __init__(
         self.name_to_ids: Dict[SearchFieldType, List[Dict[str, Any]]] = defaultdict(
             list
         )
-        for name, _ in name_entries:
-            self.name_to_ids[name] = []
+        for name, name_entry in name_entries:
+            if name in self.name_to_ids:
+                self.name_to_ids[name].append(name_entry)
+            else:
+                self.name_to_ids[name] = []
```
luisa-beerboom, line 73: does this not cause problems in comparison lines 107 - 116? It should at the very least make the else clause in line 111 obsolete. Might as well delete it, if that's the intention.

luisa-beerboom (changes requested):

openslides_backend/action/mixins/import_mixins.py
r-peschke: No, that works quite good together. 
If line 71 will be reached, it is a duplicate and line 110 will append the second entry to recognize the duplicate.

reiterl, line 73: I don't like init a defaultdict with empty lists. I don't like the type of name_to_ids. I would expect
`Dict[str, List[int]]`.
Yes, the tests works. As far as I understand the code, it seems to be correct and I don't see the 
problems with lines 107-116. So I approve it.

reiterl (approved):
Looks good enough. It seems to work. See my comment for possible improvements.

PR #1938, opened 2023-11-02
Prevent vote weight of 0

### LINKED ISSUE #1926: Prevent vote weight of 0
A vote weight of 0 should be forbidden. Currently, the minimum is defined in the `models.yml` as 0. This should be changed to `0.000001` (the lowest possible decimal number). For this, the minimum for decimal values in the `models.yml` should be enhanced to allow decimal values and evaluate them accordingly.

jsangmeister:
fixes #1926

also some cleanup of the models validator

### DIFF
--- a/openslides_backend/models/fields.py
+++ b/openslides_backend/models/fields.py
@@ -216,6 +216,8 @@ class DecimalField(Field):
         schema = self.extend_schema(super().get_schema(), **decimal_schema)
         if not self.required:
             schema["type"] = ["string", "null"]
+        # remove minimum since it is checked in the validate method
+        schema.pop("minimum", None)
         return schema
 
     def validate(self, value: Any, payload: Dict[str, Any] = {}) -> Any:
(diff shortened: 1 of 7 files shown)


openslides_backend/models/fields.py
```diff
@@ -216,6 +216,8 @@ def get_schema(self) -> Schema:
         schema = self.extend_schema(super().get_schema(), **decimal_schema)
         if not self.required:
             schema["type"] = ["string", "null"]
+        # remove minimum since it is checked in the validate method
+        schema.pop("minimum", None)
```
r-peschke, line 220: IMO the minimum value is added to be checked!??? Why this should be removed?
Maybe the check has to be suppressed in case of a non required decimal. But not required with minimum value seems curious IMO.

r-peschke (changes requested):

openslides_backend/models/fields.py
jsangmeister: As the comment states, the minimum is checked manually in the `validate` method right below that. jsonschema is not capable of handling decimals as string, therefore this has to be done manually and the `minimum` constraint has to be removed from the schema.

r-peschke (approved):

PR #1946, opened 2023-11-07
Fix dependency structure for dependabot

### LINKED ISSUE #1945: Fix `dependabot.yml`
Contrary to the statement in https://github.com/OpenSlides/openslides-backend/pull/1924, it seems like dependabot does not pick up on the non-standard requirement files (e.g., our mypy version is 1.4.1, while the current version is 1.6.1). This should be fixed.

jsangmeister:
fixes #1945

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -13,6 +13,9 @@ updates:
       interval: "daily"
     assignees:
       - "jsangmeister"
+    ignore:
+      - dependency-name: "python"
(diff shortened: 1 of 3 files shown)


jsangmeister:
Seems like dependabot breaks because we use an URL as a dependency in one of our files: https://github.com/OpenSlides/openslides-backend/network/updates/745973222. I restructured the dependencies so dependabot does not scan this line. This should hopefully fix the problem.

PR #1975, opened 2023-11-08
Update datastore & improve dep tests

jsangmeister:
- update datastore & related dependencies
- add CI test for the local setup, so it does not break with random dependabot updates
- fix dependabot python version updates


jsangmeister:
@r-peschke the new workflow ensures a working local setup, but at the same time introduces some overhead to the CI, since building the images takes time. Since we don't execute the tests again in the local setup, it is still faster than the other job which executes the tests, therefore it shouldn't make a difference for us. What's your opinion on including the new workflow?

jsangmeister:
TODO: looks like the new fastjsonschema version sorts missing properties alphabetically. I will adjust the tests accordingly.

r-peschke:
> @r-peschke the new workflow ensures a working local setup, but at the same time introduces some overhead to the CI, since building the images takes time. Since we don't execute the tests again in the local setup, it is still faster than the other job which executes the tests, therefore it shouldn't make a difference for us. What's your opinion on including the new workflow?

It is good to ensure the local dev:-)
Without the test it is worth the additional time.

r-peschke (approved):

PR #1978, opened 2023-11-09
Improve migration test with SQL dump

jsangmeister:
To make it handle dumps with `COPY` statements correctly.

### DIFF
--- a/tests/system/migrations/test_with_sql_dump.py
+++ b/tests/system/migrations/test_with_sql_dump.py
@@ -1,4 +1,5 @@
 import os
+from io import StringIO
 
 import pytest
 from datastore.migrations.core.migration_handler import MigrationHandler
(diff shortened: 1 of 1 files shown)


tests/system/migrations/test_with_sql_dump.py
```diff
@@ -16,7 +17,23 @@ def test_with_sql_dump():
...
+                            continue
+                        assert line.startswith("COPY")
+                        data = StringIO()
+                        while (record := lines.pop(0).strip()) != "\\.":
+                            data.write(record + "\n")
+                        data.seek(0)
+                        cursor.copy_expert(line, data)
+                    cursor.execute("SET session_replication_role = 'origin'")
+                else:
+                    cursor.execute(line, [])
```
r-peschke, line 36: this `line` is always undefined

r-peschke (changes requested):

tests/system/migrations/test_with_sql_dump.py
jsangmeister: right, my bad

r-peschke (approved):

Issue #1977, opened 2023-11-09
Prevent setting `saml_id` via `user.update` from the client


jsangmeister:
The import needs this field, but it should not be possible to set it from outside the backend. This should be included in the permission check. It suffices if this is respected with the new account/participant import version.

r-peschke:
This will be made in #1714

r-peschke:
Requirements from #1983, which will be closed as similar to this one:
New users with saml_id can be created via login with their saml_id.
A new user can also be created or an existing user can be updated by importing the data via participant or account import

PR #1980, opened 2023-11-09
I1714 Participant Import

r-peschke:


rrenkert:
will also close #1977

openslides_backend/action/mixins/import_mixins.py
```diff
@@ -384,3 +423,15 @@ def validate_instance(self, instance: Dict[str, Any]) -> None:
...
+            id_ = instance["meeting_id"]
+            meeting = self.datastore.get(
+                fqid_from_collection_and_id("meeting", id_),
+                ["id"],
+                lock_result=False,
+                raise_exception=False,
+            )
+            if not meeting:
+                raise ActionException(
+                    f"Participant import try to use not existing meeting {id_}"
```
jsangmeister, line 436:
```suggestion
                    f"Participant import tries to use not existing meeting {id_}"
```

openslides_backend/action/actions/user/create_update_permissions_mixin.py
```diff
@@ -171,14 +183,22 @@ class CreateUpdatePermissionsMixin(UserMixin, UserScopeMixin, Action):
...
         "F": ["default_password"],
         "G": ["is_demo_user"],
+        "H": ["saml_id"],
     }
 
+    def perform(
+        self, action_data: ActionData, user_id: int, internal: bool = False
+    ) -> Tuple[Optional[WriteRequest], Optional[ActionResults]]:
+        self.internal = internal
+        return super().perform(action_data, user_id, internal)
```
jsangmeister, line 200: Ah, I had a similar change in https://github.com/OpenSlides/openslides-backend/pull/2022. I prefer the global approach of solving this in the `action.py` instead of each action implementing their own solution.

openslides_backend/action/actions/user/create_update_permissions_mixin.py
```diff
@@ -190,9 +210,13 @@ def check_permissions(self, instance: Dict[str, Any]) -> None:
         if "forwarding_committee_ids" in instance:
             raise PermissionDenied("forwarding_committee_ids is not allowed.")
 
-        permstore = PermissionVarStore(self.datastore, self.user_id)
-        if permstore.user_oml == OrganizationManagementLevel.SUPERADMIN:
-            return
+        if not hasattr(self, "permstore"):
```
jsangmeister, line 213: `hasattr` is always kind of hacky. How about an optional constructor parameter which you can check normally and create the perm store if it is `None`?

Then you could also remove the verbose constructor in `CreateUpdatePermissionsFailingFields`.

openslides_backend/action/actions/user/create_update_permissions_mixin.py
```diff
@@ -434,3 +462,84 @@ def _meetings_from_group_B_fields_from_instance(
...
+        services: Services,
+        datastore: DatastoreService,
+        relation_manager: RelationManager,
+        logging: LoggingModule,
+        env: Env,
+        skip_archived_meeting_check: Optional[bool] = None,
+        use_meeting_ids_for_archived_meeting_check: Optional[bool] = None,
+    ) -> None:
+        self.permstore = permstore
+        self.user_id = self.permstore.user_id
```
jsangmeister, line 480: The `user_id` is set first thing in the `perform` method. Where do you need it earlier?

openslides_backend/action/actions/user/participant_common.py
```diff
@@ -0,0 +1,50 @@
...
+            )
+            raise MissingPermission(
+                {
+                    Permissions.User.CAN_MANAGE: self.meeting_id,
+                    OrganizationManagementLevel.CAN_MANAGE_ORGANIZATION: 1,
+                    CommitteeManagementLevel.CAN_MANAGE: meeting["committee_id"],
+                }
+            )
+
+        self.permission_check = CreateUpdatePermissionsFailingFields(
```
jsangmeister, line 41: Mixins should not be directly instanstiated, they should be mixed in. Does anything speak against simply inheriting from it?

jsangmeister (changes requested):
I did not check the logic of the import, but rather looked at the structure. I think @luisa-beerboom might be more fit to check the logic

luisa-beerboom (approved):
I didn't see anything that struck me as wrong in the functionality. I suppose it would be still be possible that @Elblinator might find something when he's testing the client, though

openslides_backend/action/actions/user/create_update_permissions_mixin.py
r-peschke: This is made, because in participant import I need the `permstore` for each import line, always with the same user. No need to integrate the `permstore` in all actions.

openslides_backend/action/actions/user/create_update_permissions_mixin.py
r-peschke: done

openslides_backend/action/actions/user/create_update_permissions_mixin.py
r-peschke: We decided to use the `hasattr`, because the other variant expects 19 type casts by mypy

Issue #1948, opened 2023-11-07
Add structure level management


jsangmeister:
Details see Nextcloud. Branch: `feature/los-extension`

TODOS:
- [x] Add CUD actions for structure levels
- [x] Add meeting setting to enable structure level countdowns
- [x] Add create/delete actions for `structure_level_list_of_speakers`
- [x] Add `structure_level_id` to `speaker.create` 
- [x] Add `structure_level_list_of_speakers.add_time`
- [x] Add coupling between speaker and structure level: When a speaker starts/stops speaking for a structure level, the related `structure_level_list_of_speakers` model has to be updated accordingly
- [x] Remove old structure level functionality
- [x] Write migration for old structure levels

jsangmeister:
Should all be done now

PR #1990, opened 2023-11-14
Remove old structure level functionality

jsangmeister:
sets up on https://github.com/OpenSlides/openslides-backend/pull/1989

part of https://github.com/OpenSlides/openslides-backend/issues/1948

TODO: add migration


openslides_backend/action/actions/user/update.py
```diff
@@ -47,7 +47,6 @@ class UserUpdate(
             "gender",
             "email",
             "default_number",
-            "default_structure_level",
```
r-peschke, line 50: Why didn't you add structure_level_ids? You did in create_update_permissions.

openslides_backend/action/actions/user/create.py
```diff
@@ -44,7 +44,6 @@ class UserCreate(
             "gender",
             "email",
             "default_number",
-            "default_structure_level",
```
r-peschke, line 47: Why didn't you add structure_level_ids? You did in create_update_permissions.

r-peschke (commented):
See remarks on user.update/create.
Also you should write some notes in action wiki including the hint for the feature branch (or not?). This concerns the account/participant.json_upload, and the user actions.

openslides_backend/action/actions/user/update.py
jsangmeister: `default_structure_level` was removed completely without replacement. `structure_level` was replaced with `structure_level_ids`, which are present a few lines below in the `UserMixin.transfer_list`.

r-peschke (commented):
Please change the action wiki concerning user.create/update, account- and participant.json_upload
Will you add the migration to this PR or write a new one?

jsangmeister:
> Please change the action wiki concerning user.create/update, account- and participant.json_upload

Sure, what do you want added there? Just a hint that the imports will need changes after this is merged?

> Will you add the migration to this PR or write a new one?

A new one, see https://github.com/OpenSlides/openslides-backend/pull/2000

r-peschke:
> > Please change the action wiki concerning user.create/update, account- and participant.json_upload
> 
> Sure, what do you want added there? Just a hint that the imports will need changes after this is merged?
> 
You could mark the old fields as deprecated with the new branch and the new ones with the feature branch name, which will introduce them. On merging we just had to remove the comments.

> > Will you add the migration to this PR or write a new one?
> 
> A new one, see #2000

jsangmeister:
> You could mark the old fields as deprecated with the new branch and the new ones with the feature branch name, which will introduce them. On merging we just had to remove the comments.

I updated `user.{create|update}` and `account.json_upload` with a hint just like the other structure level actions. Since `participant.json_upload` should stay the same from a payload perspectuve

jsangmeister:
Merged in https://github.com/OpenSlides/openslides-backend/pull/2000

PR #1996, opened 2023-11-15
List of speakers extension

jsangmeister:
PR to merge the feature branch into main. Will only be merged after the feature is fully completed. In the meantime, the PR is useful for rebasing and automated CI.


global/meta/permission.yml
```diff
@@ -10,6 +10,9 @@ agenda_item:
     can_manage:
         can_see_internal:
             can_see:
+    can_manage_moderator_notes:
+        can_see_moderator_notes:
```
ostcar, line 14: Are you sure, that a user with `agenda_item.can_manage` can not see the moderator notes?

global/meta/models.yml
```diff
@@ -1741,6 +1762,38 @@ meeting:
...
+    type: color
+    restriction_mode: A
+  default_time:
+    type: number
+    minimum: 0
+    restriction_mode: A
+  meeting_user_ids:
+    type: relation-list
+    equal_fields: meeting_id
+    restriction_mode: B
```
ostcar, line 1784: This has to be restruction_mode: A

global/meta/models.yml
```diff
@@ -1741,6 +1762,38 @@ meeting:
...
+    to: meeting_user/structure_level_ids
+  structure_level_list_of_speakers_ids:
+    type: relation-list
+    equal_fields: meeting_id
+    restriction_mode: B
```
ostcar, line 1789: This has to be restruction_mode: A

global/meta/permission.yml
jsangmeister: Not by default, yes. Discussed with @rrenkert that there might be cases where the permissions for the agenda item and the moderator notes might be required to be split up. If necessary, the user can just give both permissions to every group to archieve the wanted behaviour.

global/meta/models.yml
jsangmeister: Yes, thank you

PR #2035, opened 2023-12-04
Change password reset email contents

### LINKED ISSUE #1979: Change E-Mail-subject if reset-mail is send out
**Current behaviour:**
If one E-Mail address is connected to several accounts and a password reset request is send out, then one E-Mail is send out for every account (correct behaviour).  besides the link and the username the Mails are indistinguishable.

**Reproduction:**
1. create two account with the same Mail-address
2. logout < press forgot password 
3. reset password for the Mail-address

**Wanted behaviour:**
The Mails should be distinguishable just by the subject. 
the subject of the Mail should change from "Reset your OpenSlides password" to "Reset your OpenSlides password: username"

luisa-beerboom:
Closes #1979


openslides_backend/action/actions/user/forget_password.py
```diff
@@ -41,6 +41,8 @@ def get_updated_instances(self, action_data: ActionData) -> ActionData:
 Please open the following link and choose a new password:
 {url}/login/forget-password-confirm?user_id={user_id}&token={token}
 
+The link will only be valid for 10 minutes.
+
 For completeness your username: {username}"""
```
jsangmeister, line 46: I think this line can be removed then

jsangmeister (changes requested):

openslides_backend/action/actions/user/forget_password.py
luisa-beerboom: I added that because of https://github.com/OpenSlides/openslides-backend/issues/1979#issuecomment-1829528618 is that retracted?

jsangmeister: I meant the last line, "For completeness your username: {username}", as that information is now in the subject

luisa-beerboom: ah okay, I'll do that then

jsangmeister (approved):

PR #2046, opened 2023-12-06
Make text-type mimetype recognition more lenient

### LINKED ISSUE #2043: Make file upload mimetype recognition more lenient on text types
We've had problems where we weren't able to upload csv files after #2031 
That's because python magic recognizes csv as `text/plain`, but mimetype as `text/csv`.
Given that files recognized by python-magic as `text/plain` are pretty safe in our context, we should be more lenient there:

If python-magic returns the mimetype `text/plain` do the filename verification via `mimetypes.guess_type` and accept anything of the form `text/...`, instead of explicitly checking the extensions.

luisa-beerboom:
Closes #2043


Elblinator (approved):
function

openslides_backend/action/actions/mediafile/upload.py
```diff
@@ -90,18 +90,24 @@ def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
         file_ = instance.pop("file")
         decoded_file = base64.b64decode(file_)
         mimetype_ = python_magic.from_buffer(decoded_file, mime=True)
+        use_mimetype: Optional[str] = mimetype_
```
r-peschke, line 93: Change sequence of lines 93 and 94. Then you could use type `str` for the `use_mimetype` and there is no need for casting in line 125

r-peschke (changes requested):

openslides_backend/action/actions/mediafile/upload.py
luisa-beerboom: Won't work because it may be filled with None in line 97

r-peschke (approved):

PR #2047, opened 2023-12-06
Forbid whitespaces in usernames

### LINKED ISSUE #2002: Forbid whitespaces in usernames
Currently, it is possible to create two users with the following names:

`Max Mustermann` and `Max  Mustermann`.

The create user method should find and remove double whitespaces

jsangmeister:
fixes #2002

### DIFF
--- a/openslides_backend/action/actions/user/create.py
+++ b/openslides_backend/action/actions/user/create.py
@@ -75,6 +75,8 @@ class UserCreate(
                 if not (instance.get("first_name") or instance.get("last_name")):
                     raise ActionException("Need username or first_name or last_name")
                 instance["username"] = self.generate_username(instance)
+        elif " " in instance["username"]:
+            raise ActionException("Username may not contain spaces")
         instance = super().update_instance(instance)
         if saml_id:
             instance["can_change_own_password"] = False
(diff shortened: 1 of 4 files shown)


openslides_backend/action/actions/user/create.py
```diff
@@ -75,6 +75,8 @@ def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
                 if not (instance.get("first_name") or instance.get("last_name")):
                     raise ActionException("Need username or first_name or last_name")
                 instance["username"] = self.generate_username(instance)
+        elif " " in instance["username"]:
```
r-peschke, line 78: Then you should look with a regex for \W. That's how the generate_username limits the allowed signs and IMO would be more convenient.

openslides_backend/action/actions/user/update.py
```diff
@@ -83,6 +83,9 @@ def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
             instance["default_password"] = ""
             instance["password"] = ""
 
+        if instance.get("username") and " " in instance["username"]:
```
r-peschke, line 86: s.o.

r-peschke (changes requested):

openslides_backend/action/actions/user/create.py
jsangmeister: done

jsangmeister:
`\W` matches any non-word character, which I think is not what we want. I replaced it with `\s` which matches only whitespace characters.

PR #2050, opened 2023-12-08
Add enum to `import_preview/name`

### LINKED ISSUE #1920: Add enum to `import_preview/name`
As suggested in https://github.com/OpenSlides/openslides-autoupdate-service/pull/781, we should constrain the values of this field with an enum to ensure only valid import names are entered. Also, I think that `type` would be more fitting as a field name.

jsangmeister:
resolves #1920

### DIFF
--- a/global/meta/models.yml
+++ b/global/meta/models.yml
@@ -3846,6 +3846,10 @@ import_preview:
   name:
     type: string
     required: true
+    enum:
+      - account
+      - participant
+      - topic
     restriction_mode: A
   state:
     type: string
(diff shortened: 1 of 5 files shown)


global/meta/models.yml
```diff
@@ -3846,6 +3846,10 @@ import_preview:
   name:
     type: string
     required: true
+    enum:
```
r-peschke, line 3849: Please add also motion and committee

r-peschke (changes requested):

global/meta/models.yml
jsangmeister: This should be done in the relevant PRs - the current state of the main does not support motion and committee imports.

r-peschke: okay

PR #2057, opened 2023-12-11
svg and json as html/plain working

r-peschke:


openslides_backend/action/actions/mediafile/upload.py
```diff
@@ -89,21 +91,36 @@ def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
         filename_ = instance.get("filename", "")
         file_ = instance.pop("file")
         decoded_file = base64.b64decode(file_)
-        mimetype_ = python_magic.from_buffer(decoded_file, mime=True)
-        if mimetype_ is None:
+        use_mimetype, _ = mimetypes.guess_type(filename_)
+        if use_mimetype is None:
             raise ActionException(f"Cannot guess mimetype for {filename_}.")
```
luisa-beerboom, line 96: There's no test for this case, could you write it? All that needs to happen is that an unknown file extension needs to be used.

luisa-beerboom (changes requested):

luisa-beerboom (approved):
nice

rrenkert (approved):

luisa-beerboom (approved):

PR #2063, opened 2023-12-12
Use postgres 15 images + debian bookworm images

peb-adr:
Apart from debian being sensical on it's own, buster packages include postgres software for v11 - bookworm does for v15.


jsangmeister (approved):

jsangmeister:
> Service 'backend' failed to build : manifest for python:3.10.3-slim-bookworm not found: manifest unknown: manifest unknown

Are you sure that the image exists?

PR #2069, opened 2023-12-13
Try to fix failing auth token test

jsangmeister:
- set `MAX_RETRY` to 1 in tests in case that the token is used because for some reason the action has to be repeated
- flush the redis database before each test just to be sure

I, too, cannot reproduce the failure locally, so these are just guesses what might help.


r-peschke (approved):
Good guess:-)

PR #2072, opened 2023-12-16
Fix CI for local setup

jsangmeister:
Looks ugly, but is the official approach to setting env variables in github actions: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-environment-variable and is better than the previous solution (which additionally did not work correctly and would have required to fetch the complete repository instead of only the required commit).

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -70,7 +70,7 @@ jobs:
     
     - name: Checkout datastore commit
       working-directory: openslides-datastore-service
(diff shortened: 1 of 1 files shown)


jsangmeister:
needed for other PRs (e.g. https://github.com/OpenSlides/openslides-backend/pull/2071), so merging this now.

jsangmeister:
needed for other PRs (e.g. https://github.com/OpenSlides/openslides-backend/pull/2071), so merging this now.

PR #2073, opened 2023-12-18
Support truncation of gender list

### LINKED ISSUE #2070: Emptying `genders` field in `organization.update` causes genders to not be deleted from the users
See [this comment in the client](https://github.com/OpenSlides/openslides-client/pull/3101#pullrequestreview-1779842540)

If the genders field is set to [] in the payload the users will not have their genders removed.

There are also lock break errors that sometimes occur, idk how to trigger them though

jsangmeister:
fixes #2070 

Also improve datastore adapter to not include duplicate locked fields.


jsangmeister:
test only works with https://github.com/OpenSlides/openslides-datastore-service/pull/295

luisa-beerboom (approved):
Code looks good

PR #2079, opened 2023-12-21
Add structure level to `speaker.update`

jsangmeister:
see changes in https://github.com/OpenSlides/OpenSlides/wiki/speaker.update


bastianjoel (changes requested):
It should also be possible to clear the current selection by for example sending `null` as `structure_level_id`.

bastianjoel (approved):

PR #2093, opened 2024-01-08
Issue2080 Accept ttf-file-extension for mimetype font/sfnt

r-peschke:


jsangmeister (changes requested):
Please add some comments on what is done and why as the procedure for mimetype checking is becoming increasingly complex and difficult to understand.

openslides_backend/action/actions/mediafile/upload.py
```diff
@@ -112,10 +112,21 @@ def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
...
-            mismatched = not any(
-                [filename_.endswith(extension) for extension in possible_extensions]
-            )
+            mismatched = check_extension(filename_, possible_extensions)
+            if mismatched:
+                possible_extensions = (
+                    python_magic.Magic(extension=True)  # type: ignore
+                    .from_buffer(decoded_file)
+                    .split("/")
+                )
```
luisa-beerboom, line 128: What does this code do?

luisa-beerboom (commented):
I just noticed in ln 102 of `upload.py` it checks a mimetype against "application.json". Wouldn't that have to be "application/json" to have any effect?

r-peschke:
> I just noticed in ln 102 of `upload.py` it checks a mimetype against "application.json". Wouldn't that have to be "application/json" to have any effect?

Yes you are right!
This case now is resolved with the pygments.lexers

r-peschke:
> Please add some comments on what is done and why as the procedure for mimetype checking is becoming increasingly complex and difficult to understand.

I think I guess what you are asking for. Let's talk about tomorrow!

jsangmeister (approved):

PR #2116, opened 2024-01-17
Make structure_level_list_of_speakers.update set remaining_time upon …

### LINKED ISSUE #2114: Fix structure_level_list_of_speaker update action
In the structure_level_list_of_speaker update action the remaining_time is not set to initial_time in case it is allowed to set the initial_time. So the remaining time can not be changed and has a potential wrong value.

luisa-beerboom:
…initial_time change

Closes #2114 

Also makes sure that initial time can't be set if remaining time has already been diminished

### DIFF
--- a/openslides_backend/action/actions/structure_level_list_of_speakers/update.py
+++ b/openslides_backend/action/actions/structure_level_list_of_speakers/update.py
@@ -32,7 +32,7 @@ class StructureLevelListOfSpeakersUpdateAction(UpdateAction):
     def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
         db_instance = self.datastore.get(
             fqid_from_collection_and_id(self.model.collection, instance["id"]),
-            ["list_of_speakers_id", "meeting_id", "remaining_time"],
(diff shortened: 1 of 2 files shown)


openslides_backend/action/actions/structure_level_list_of_speakers/update.py
```diff
@@ -32,27 +32,33 @@ def validate_instance(self, instance: Dict[str, Any]) -> None:
...
+                        FilterOperator("meeting_id", "=", db_instance["meeting_id"]),
+                        FilterOperator(
+                            "list_of_speakers_id",
+                            "=",
+                            db_instance["list_of_speakers_id"],
+                        ),
                     ),
-                ),
+                )
+                or db_instance["initial_time"] != db_instance["remaining_time"]
```
r-peschke, line 56: Use this as first position in if, because if is resolved as True, there is no need to ask the database (much more expensive) anymore

r-peschke (changes requested):
Just a small thing

r-peschke (approved):

Issue #2123, opened 2024-01-18
Only use intervention time for interventions


Elblinator:
If a structure level and the type intervention is selected, then the timer from the structure level is displayed and counts down. If intervention is selected/activated then the timer from the intervention should be used and the structure time countdown should remain unchanged

see https://github.com/OpenSlides/openslides-client/pull/3131 for context

PR #2126, opened 2024-01-18
Don't use structure level timer for interposed questions and interventions

jsangmeister:
fixes https://github.com/OpenSlides/openslides-backend/issues/2123

### DIFF
--- a/openslides_backend/action/actions/projector_countdown/mixins.py
+++ b/openslides_backend/action/actions/projector_countdown/mixins.py
@@ -2,6 +2,12 @@ from enum import Enum, auto
 from time import time
 from typing import Optional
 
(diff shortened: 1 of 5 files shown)


r-peschke (approved):

Elblinator (changes requested):
the intervention time is still being subtracted from the (remaining) displayed time

If an intervention is selected, the correct time is displayed and only the intervention timer counts down (correct behaviour)
But if you stop the intervention the time from the intervention is subtracted from the displayed timer from the structure level

jsangmeister:
Fixed this and the analogous problem when pausing an interposed question/intervention with a structure level.

Elblinator (approved):

PR #2139, opened 2024-01-22
End interposed questions with their parents

### LINKED ISSUE #2136: End interposed question when parent is ended
see https://github.com/OpenSlides/openslides-client/issues/3169:

- if the child does not have a start time yet, then it should be dismissed
- if the child has a start time then the child should be saved and the end time should be the same as the parents (if no speaker is selected then it should be saved as Deleted User)

jsangmeister:
fixes #2136


Elblinator (approved):
function

luisa-beerboom (approved):

PR #2146, opened 2024-01-23
Add text hash and identical motions calculation

### LINKED ISSUE #2121: Automatically check motions for equality
Upon saving a motion (either via `motion.create` or `motion.update`), if the `motion/text` changed, all other motions of the meeting should be checked if they have the same text [1] [2]. If that's the case, the new relation `identical_motion_ids` should point to the motion(s) with the same text.

[1] "same text" is defined as having the same textual content, ignoring all html markup. All spaces, line breaks, hyphenation etc. is kept and may lead to two motions not being identified as identical.

[2] The check is implemented via hashes: Upon save, the hash for the pure text of the motion is calculated and saved in the new field `text_hash`. Then, it can be checked via a simple `filter()` if any other motion has the same hash.

jsangmeister:
closes #2121


luisa-beerboom (approved):

jsangmeister:
waiting for https://github.com/OpenSlides/openslides-client/pull/3205

PR #2148, opened 2024-01-24
Make participant import create structure levels

### LINKED ISSUE #2147: Make participant.import create structure levels
The participant import should lookup the new structure level model and if any structure levels named in the import don't exist, they should be created during the import process

The preview fields should be marked with the 'new' state and there should be entries in the statistics

luisa-beerboom:
Closes #2147 

For now I covered the edge case of "What if the corresponding structure level is created between json_upload and import?" by letting the import fail (i.e. forcing the user to start anew) , I could probably make it so that it succeeds instead with the field being switched over to an update field, what do you think @rrenkert ?

### DIFF
--- a/openslides_backend/action/actions/user/base_import.py
+++ b/openslides_backend/action/actions/user/base_import.py
@@ -33,6 +33,7 @@ class BaseUserImport(ImportMixin):
 
         self.rows = [self.validate_entry(row) for row in self.rows]
 
+        self.handle_create_relations()
         if self.import_state != ImportState.ERROR:
             rows = self.flatten_copied_object_fields(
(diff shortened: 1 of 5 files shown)


jsangmeister (changes requested):
Code looks fine, but IMO we should not fail with an error if a structure level was created in the meantime but just use it, as you already proposed

luisa-beerboom:
> Code looks fine, but IMO we should not fail with an error if a structure level was created in the meantime but just use it, as you already proposed

Done

jsangmeister (changes requested):

jsangmeister (approved):

PR #2167, opened 2024-01-31
Implement email warnings in user imports

### LINKED ISSUE #2138: Account/Participant import: fix missing warning for wrong email
**Current:** 
If a wrong email (lara@@gmx.de or lara@gmx,de ) is in the CSV-File no warning/block is given. 

**Wanted:**
Parallel behaviour: if the vote weight is too small in the CSV-File then the import of this account/participant is blocked

If a wrong email is given then the same blocking behaviour as a too small vote weight should occur

luisa-beerboom:
Closes #2138

### DIFF
--- a/openslides_backend/action/actions/user/base_import.py
+++ b/openslides_backend/action/actions/user/base_import.py
@@ -21,15 +21,13 @@ class BaseUserImport(BaseImportAction):
 
         if self.import_state != ImportState.ERROR:
             rows = self.flatten_copied_object_fields(
-                self.handle_remove_email_and_group_fields
+                self.handle_remove_and_group_fields
             )
             self.create_other_actions(rows)
 
         return {}
 
-    def handle_remove_email_and_group_fields(
-        self, entry: Dict[str, Any]
-    ) -> Dict[str, Any]:
+    def handle_remove_and_group_fields(self, entry: Dict[str, Any]) -> Dict[str, Any]:
         if (groups := entry.pop("groups", None)) is not None:
             entry["group_ids"] = [id_ for group in groups if (id_ := group.get("id"))]
 
@@ -104,8 +102,6 @@ class BaseUserImport(BaseImportAction):
         self.validate_with_lookup(row, self.saml_id_lookup, "saml_id", False, id)
         if row["state"] == ImportState.ERROR and self.import_state == ImportState.DONE:
             self.import_state = ImportState.ERROR
-
-        return row
 
     def setup_lookups(self) -> None:
         self.username_lookup = Lookup(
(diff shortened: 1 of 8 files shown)


openslides_backend/action/actions/user/base_import.py
```diff
@@ -100,6 +105,8 @@ def validate_entry(self, row: ImportRow) -> None:
         if row["state"] == ImportState.ERROR and self.import_state == ImportState.DONE:
             self.import_state = ImportState.ERROR
 
+        return row
```
r-peschke, line 108: The method has no return value

openslides_backend/action/actions/user/base_json_upload.py
```diff
@@ -246,6 +247,15 @@ def validate_entry(self, entry: Dict[str, Any]) -> Dict[str, Any]:
                 entry["gender"] = {"info": ImportState.WARNING, "value": gender}
                 messages.append(f"Gender '{gender}' is not in the allowed gender list.")
 
+        if email := entry.get("email"):
+            if EmailUtils.check_email(email):
+                entry["email"] = {"info": ImportState.DONE, "value": email}
+            else:
+                entry["email"] = {"info": ImportState.WARNING, "value": email}
+                messages.append(
+                    f"'{email}' is not a valid email address and will be skipped. This may have caused problems with user recognition."
```
r-peschke, line 256: IMO something like
"... and will not be imported. This may cause problems with user recognition"
is clearer. Concerning the 2nd phrase: why you are using past tense?
But see first my final comment

openslides_backend/action/actions/user/base_import.py
```diff
@@ -21,13 +21,15 @@ def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
...
             rows = self.flatten_copied_object_fields(
-                self.handle_remove_and_group_fields
+                self.handle_remove_email_and_group_fields
             )
             self.create_other_actions(rows)
 
         return {}
 
-    def handle_remove_and_group_fields(self, entry: Dict[str, Any]) -> Dict[str, Any]:
+    def handle_remove_email_and_group_fields(
```
r-peschke, line 30: Don't change the method name. "email" is removed/poped like gender and other fields

r-peschke (changes requested):
A wrong "email" should prevent the whole import, this is what the vote weight does in participant and account import. Also see my note in the related issue.
But if we want this, it makes no sense using the state "warning" on both of the fields, it should be "error", if they prevent the import.

r-peschke (approved):

PR #2168, opened 2024-01-31
Move conditional speaker deletion from user- to meeting_user.delete

luisa-beerboom:
Noticed that this may otherwise cause problems when using the meeting_user.delete elsewhere (as is planned in #2111 f.E.).
Also checked for deletion status before carrying out the deletion to avoid breaking the cascades in meeting.delete and wrote a new test in meeting.delete


r-peschke (commented):
Could you please write a test for user.delete, with 1 speaker with and 1 without `begin`?
I assume one will be deleted and the other stays unchanged.

r-peschke (approved):

PR #2175, opened 2024-02-03
Fix update past interposed questions

jsangmeister:


openslides_backend/action/actions/speaker/update.py
```diff
@@ -46,13 +46,14 @@ def update_instance(self, instance: Dict[str, Any]) -> Dict[str, Any]:
...
                 "You cannot change the speech state of an interposed_question."
             )
-        if "meeting_user_id" in instance and (
-            instance["meeting_user_id"] is None
-            or speaker.get("meeting_user_id")
-            or speaker.get("speech_state") != SpeechState.INTERPOSED_QUESTION
-        ):
-            raise ActionException("You cannot set the meeting_user_id.")
-        if "structure_level_id" in instance and speaker.get("begin_time"):
+        if "meeting_user_id" in instance:
```
r-peschke, line 49: This way you can change the structure_level_id for all not waiting users, if meeting_user_id is in instance, even just confirming it. I'm not able to see all the consequences of this change. Could you explain the conditions, where this is allowed? I would like to have a decision table or another tool to change this kind of related criteria.

r-peschke (commented):
See comment

openslides_backend/action/actions/speaker/update.py
jsangmeister: This is ok: You can only set the meeting user once. The only case where a speaker can have an empty `meeting_user_id` is in the case of an interposed question. As soon as the meeting user is set once (where a structure level can be set simultaniously as well, as intended), if a user tries to send another `meeting_user_id` again, an error will be thrown in the lines below (specifically line 52).

r-peschke (approved):

PR #2182, opened 2024-02-03
Inject version.txt in via docker build

peb-adr:


peb-adr:
Adding the `--chown` to `COPY` was necessary to later be able to write to the filesystem, since the `COPY` command is not done as appuser, see https://docs.docker.com/engine/reference/builder/#user.

jsangmeister (approved):

PR #2189, opened 2024-02-06
Build group creation into participant import

### LINKED ISSUE #2180: Participant JsonUpload/Import with group to be created
Not existing groups were created without any rights
user.update: Don't include field username in update action, because of existing usernames with spaces, which should not be changed
user.create: im json upload Prüfung auf Leerstellen im Username

luisa-beerboom:
Closes #2180


r-peschke (changes requested):
see comment

PR #2193, opened 2024-02-08
Enhance meta repo

jsangmeister:
see https://github.com/OpenSlides/openslides-meta/pull/18

### DIFF
--- a/openslides_backend/action/actions/committee/json_upload.py
+++ b/openslides_backend/action/actions/committee/json_upload.py
@@ -51,7 +51,7 @@ class CommitteeJsonUpload(BaseJsonUploadAction, MeetingCheckTimesMixin):
         }
     )
     headers = [
-        {"property": "name", "type": "string"},
+        {"property": "name", "type": "string", "is_object": True},
         {"property": "description", "type": "string"},
         {
             "property": "forward_to_committees",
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -41,7 +40,7 @@ test-unit-integration:
 check-all: validate-models-yml check-models check-initial-data-json check-example-data-json check-permissions
 
 validate-models-yml:
-	python global/meta/models_validator/validate.py global/meta/models.yml
+	make -C global/meta/dev validate-models
```
r-peschke, line 43: IMO we should remove this test in backend. The integrity of model.yml should be the concern of the meta repo

r-peschke (changes requested):

Makefile
jsangmeister: We should keep it in the Makefile as an easy-to-access way of executing the check (otherwise you have to manually go into the `global/meta` folder). But if you want, we can remove it from the CI. I agree that it's redundant.

r-peschke: I agree with the easy-to-access way of checking from backend during dev.
What do you think from removing the models validation from .github/workflows/continuous_integration.yml? That would be sufficient for me

jsangmeister: done

r-peschke (approved):

PR #2203, opened 2024-02-09
Vote decrypt with merged main

r-peschke:
- [x] Merge main into branch and fix errors
- [x] Add prefix-length byte with fixed value 32 as prefixx to encrypted data


jsangmeister:
merged manually

PR #2208, opened 2024-02-13
Build group creation into participant import

### LINKED ISSUE #2180: Participant JsonUpload/Import with group to be created

luisa-beerboom:
Closes #2180


openslides_backend/action/actions/user/participant_import.py
```diff
@@ -141,17 +155,15 @@ def validate_entry(self, row: ImportRow) -> None:
             if field in entry:
                 singular_field = field.rstrip("s")
                 for instance in (instances := entry[field]):
-                    if not (instance_id := instance.get("id")):
```
r-peschke, line 144: Removing this condition let's the code continue with Line 156 (left side) or 166/168 on right side. Why did you removed it?

PR #2211, opened 2024-02-13
Allow import for existing users with legacy usernames

### LINKED ISSUE #2204: Participant csv import: Import of existing users with space in username not possible
**Description:** If a user wants to import existing users with usernames that contain a space between surname and name, an error appears. Import ist not possible. 

**Reproduction:**
A reproduction in a new meeting instance is not possible. An old instance with an existing user data base and usernames with spaces must be used. - For reproduction simply try to import an existing user with space in username via csv file in a meeting. 

**What should happen:**
It should be possible to import old, existing usernames that contain a space between surname and name. The import should check if the username still exists and reference it with the space. Additionally it should not be possible to generate new users with usernames that contain spaces.

jsangmeister:
fixes #2204


tests/system/action/base.py
```diff
@@ -290,7 +290,9 @@ def _get_user_data(
             "meeting_ids": list(partitioned_groups.keys()),
         }
 
-    def create_user_for_meeting(self, meeting_id: int) -> int:
+    def create_user_for_meeting(
+        self, meeting_id: int, username: str | None = None
```
luisa-beerboom, line 294: username parameter isn't used, remove it

luisa-beerboom (changes requested):

r-peschke (approved):

luisa-beerboom (approved):

PR #2217, opened 2024-02-19
Add migration command to clear collectionfield tables

jsangmeister:
@m-schieder the command is reachable via the migration route with the command `clear-collectionfield-tables`.


m-schieder (changes requested):
It works. But can you please update the [documentation](https://github.com/OpenSlides/openslides-backend/blob/main/docs/migration_route.md).

jsangmeister:
done

m-schieder (approved):

PR #2231, opened 2024-02-20
Enhance speaker creation

### LINKED ISSUE #2229: Allow several simultaneous contributions from the same participant on a LoS
see for context: https://github.com/OpenSlides/openslides-client/issues/3313

1. We need a new settings field which allows the activate/deactivate the possibility to be on a LoS several times simultaneously 
2. The backend need to allow / not block this behaviour if this setting is activated

### LINKED ISSUE #2230: Allow to set point of orders for other users
see for context: https://github.com/OpenSlides/openslides-client/issues/3314

1. We need a new settings field which allows to activate/deactivate the possibility to set a PoO for pther participants
2. The backend needs to allow / not block this behaviour if this setting is activated

jsangmeister:
closes #2229 
closes #2230


r-peschke (approved):

Elblinator (changes requested):
When I try to add a point of order for some one else the following Error snackbar error message is thrown: 
The requesting user 1 is not the user 3 the point-of-order is filed for.

Elblinator (approved):

PR #2237, opened 2024-02-22
Implement readding of interposed questions

### LINKED ISSUE #2233: List of Speakers - Interposed questions: Re-Add without parent
**Description:** If a user wants to re-add interposed questions to the list of speakers, the re-adding is done without the parent speech. The interposed questions will be added above all other speeches. If this occurs, it is also not possible to change the posistions of the speeches in the list. A snackbar with an error appears. To solve this, the re-added interposed questions needs to be deleted first. - Is guess, that the follow-up problem will be solved, if the main issue is fixed.

**List of Speakers: Before all have spoken**
image

**List of Speakers: After Re-Adding** 
image

**Reproduction:** 
1. open a meeting
2. open settings > List of Speakers > enable "Enable interposed questions"
3. open Agenda > Create a topic > Project the topic
4. open Autopilot
5. Add two person to the List of Speakers
6. Start speech of person 1 on the LoS
7. Add two interposed questions to the LoS -> let them both speak to create time and stop them
8. Stop the speech of person 1
9. Now Re-Add all persons to the LoS -> LoS should look now like the second image
10. Try now to re-arrange the speaker positions. snackbar with "Error: speaker sorting failed, because some elements were not included in the call." will appear

What should happen:
Re-Adding persons which are associated with interposed questions should re-build the old parent/child-list-behavior. Additionally the re-arranging of the LoS should be possible after Re-Add.

luisa-beerboom:
Closes #2233 

@jsangmeister I implemented it this far.
No idea if this is the wanted behaviour, I couldn't seem to find any meaning behind the weight distribution that gets created as a result of the various speaker actions. As a result I can only make guesses as to what logic I should be implementing here.

The action wiki entry is no real help either as the action entry for los.re_add_last doesn't seem to be up-to-date with the feature implementation (f.E. "The new waiting speaker gets the weight of `min-1` of all waiting speakers or 1, if there are no waiting speakers." wherein the "or 1" part doesn't seem to currently be implemented).

If this doesn't work in the client, please tell me what exactly the result of the request should be


jsangmeister (changes requested):
Seems fine code-wise. Please enhance the docs to incldue the weight calculation for interposed question (that you do not decrement the weight). Not tested functionally.

jsangmeister (changes requested):
Please revert the changes to the weight calculation. Re-adding an interposed question should only be allowed if there is a currently speaking speaker (`being_time != null && end_time == null`). In this case, the previous weight calculation was already correct.

jsangmeister:
> > Please revert the changes to the weight calculation. Re-adding an interposed question should only be allowed if there is a currently speaking speaker (`being_time != null && end_time == null`). In this case, the previous weight calculation was already correct. With these changes, the rebuilding of the parent-child structure should work out-of-the-box.
> 
> Why would it be correct? That sort of weight calculation would place the interposed question above the parent motion, wouldn't it?

No, they get the same weight, so the interposed question is sorted after its parent. You can try it with the current main: If you stop a speaker with an interposed question, then re-add the speaker and start it and then re-add the interposed question, everything works as intended. The only thing that must be prohibited is re-adding the interposed question if no one is speaking since then the UI breaks.

jsangmeister:
> The previous functionality literally sets everything to `"weight": lowest_weight - 1`. That's _never_ going to be the same as the parent motion

The weight is irrelevant for interposed questions, as they are always displayed directly after the current speaker. Only the relative weight between them is relevant, and for that purpose it is correct.

jsangmeister (approved):

luisa-beerboom:
@jsangmeister This is constantly failing it's checks. I don't know why. They don't seem to fail when I run the tests locally

PR #2256, opened 2024-02-28
Automatically create PRs for commits on staging branches

jsangmeister:
@peb-adr If you approve this, I will add this to all repos to automatically keep all main branches up-to-date.

### DIFF
--- a/.github/workflows/staging-to-main.yml
+++ b/.github/workflows/staging-to-main.yml
@@ -3,7 +3,7 @@ name: Copy staging commits to main
 on:
   push:
     branches:
-      - 'staging/**'
+      - 'staging/4*'
       
 
 jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/staging-to-main.yml
```diff
@@ -0,0 +1,53 @@
+name: Copy staging commits to main
+
+on:
+  push:
+    branches:
+      - 'staging/**'
```
peb-adr, line 6: https://github.com/OpenSlides/OpenSlides/blob/main/.github/workflows/build-images.yml#L7
runs `on: push: branches: - 'stable/4*'`
And will soon match the `staging` branches as well.
We should keep these matchings uniform. If you're fine with changing this to `'staging/4*'` please do so. Otherwise I will change the matching in `build-images.yml` to `'stable/**'` (and also `'staging/**'` later)

peb-adr (approved):
Functionally I approve.

.github/workflows/staging-to-main.yml
jsangmeister: Sure, will change

PR #2257, opened 2024-02-28
Add merge command

jsangmeister:
resolves https://github.com/OpenSlides/openslides-meta/issues/36

Tested up until the point of pushing the changes in the meta repo as I could not do this without the app in my fork. Fine-tuning can be done after merging.

### DIFF
--- a/.github/workflows/commands.yml
+++ b/.github/workflows/commands.yml
@@ -4,78 +4,83 @@ on:
     types: [created]
 
 jobs:
-  merge-meta-repo:
+  update-meta-repo:
     name: Recursively merge the main branch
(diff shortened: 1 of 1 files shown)


peb-adr (approved):
Talked it through together.
Seems like it sensibly merges changes from main into the branch active PR.
We should rename the command to something more concise IMO, but then we can test it in the wild.

jsangmeister:
Renamed the command to `\update-meta` and allowed alternative case styles to not have to remember that.

jsangmeister:
I will test this in the backend as soon as the chance opens up and roll it out to the other repos if it works.

PR #2278, opened 2024-03-06
main merged into Feature/vote decrypt

r-peschke:


jsangmeister:
Tip for the next time: If you run `make run-dev` without the `-local` option, it will output a more concise error:
```
The conflict is caused by:
     datastore 1.0.0 depends on redis==5.0.2
     authlib 1.0.0 depends on redis==5.0.1
```

r-peschke:
> Tip for the next time: If you run `make run-dev` without the `-local` option, it will output a more concise error:
> 
> ```
> The conflict is caused by:
>      datastore 1.0.0 depends on redis==5.0.2
>      authlib 1.0.0 depends on redis==5.0.1
> ```

Thank you for the tip:-).  I saw it in the github checks.  My idea was to build it locally and could see the problem. But didn't find any redis 5.0.1. The reason: auth service will not be build from local data.

PR #2295, opened 2024-03-12
Add initial data creation

jsangmeister:
fixes https://github.com/OpenSlides/OpenSlides/issues/6745

See https://github.com/OpenSlides/openslides-datastore-service/issues/245#issuecomment-1991882731 for additional context.

Also fixed some types and interface usages.

During startup of the action container, it checks now if the env var `OPENSLIDES_BACKEND_CREATE_INITIAL_DATA` ist set. In that case, the initial data file is read and send to the action `organization.initial_import`.


r-peschke (approved):
LGTM

peb-adr:
This will leave the superadmin (`user/1`) unusable as no password is defined in https://github.com/OpenSlides/openslides-backend/blob/main/global/data/initial-data.json.\
There shouldn't be one defined either IMO as it would pose a security risk as people may start using the same default password for the superadmin.

The `initial-data` manage command currently will fire a `user.set_password` action after the initial import to set the generated password from the `secrets/superadmin` file.\
So the code should be expanded to check for `$SUPERADMIN_PASSWORD_FILE` (default: `/run/secrets/superadmin`) and if it exists set superadmin's password to it's content.

openslides_backend/http/application.py
```diff
@@ -18,20 +25,56 @@
...
+                        {
+                            "action": "organization.initial_import",
+                            "data": [{"data": initial_data}],
+                        }
+                    ],
+                    -1,
+                    internal=True,
+                )
+            except ActionException as e:
+                self.logger.error(f"Initial data creation failed: {e}")
```
peb-adr, line 77: Call `user.set_password`.
See my above comment.

peb-adr (changes requested):

jsangmeister:
@peb-adr I added the password setting as described. I renamed the variable to `OPENSLIDES_BACKEND_SUPERADMIN_PASSWORD_FILE` to conform with the other backend variables, I hope that's okay.

I also changed the setup slightly to load the example data file in test mode instead of the initial data. This is currently also the case and should not change, since it would make life harder for client developers.

jsangmeister:
@r-peschke I adjusted the test setup slightly to make it easier to access application class specific attributes, e.g. the logger. Maybe you can use the same adjustments for your tests in https://github.com/OpenSlides/openslides-backend/pull/2335

peb-adr:
> @peb-adr I added the password setting as described. I renamed the variable to `OPENSLIDES_BACKEND_SUPERADMIN_PASSWORD_FILE` to conform with the other backend variables, I hope that's okay.

Tbh I would prefer it stays as is.\
I think should not conform with the other backend variables but rather the other secret path variables.\
See https://github.com/OpenSlides/openslides-manage-service/blob/ac17d372845d464740b9c9ac43a42bbfc898faf8/pkg/config/default-config.yml#L76

jsangmeister:
> Tbh I would prefer it stays as is. I think should not conform with the other backend variables but rather the other secret path variables. See https://github.com/OpenSlides/openslides-manage-service/blob/ac17d372845d464740b9c9ac43a42bbfc898faf8/pkg/config/default-config.yml#L76

Sure, I also see that, renamed it

peb-adr (approved):

PR #2176, opened 2024-02-03
Fix entitled users

jsangmeister:
fixes #2156

PR #2314, opened 2024-03-18
Fix `KeyError` in entitled users calculation

jsangmeister:
fixes https://github.com/OpenSlides/openslides-backend/issues/2309

reverts part of https://github.com/OpenSlides/openslides-backend/pull/2176/ which led to the error


r-peschke (commented):
LGTM

PR #2331, opened 2024-03-22
Adapt adapter to read directly with sql

luisa-beerboom:
For now this is just a copy of the backend processes and not yet attuned to the new db schema (it's also based on @jsangmeister 's backend-setup PR)

TODO:
- [x] Make get_many work with the db schema
- [x] Implement the other functions in `read_adapter.py`
- [x] Use the `ReadAdapter` in `adapter.py`
   - [ ] Remove unnecessary parameters from `adapter.py` payloads
- [x] What about locking? => ` for update` behind every select query when fields are supposed to be locked
   - [ ] What about min/max/count/exists -> those can't use `for_update` => Will need further deliberation
   - [ ] Somehow test this locking mechanism => As discussed, will be done by @r-peschke 
- [ ] Test base class update to proper system
- [ ] Don't just create a new transaction for every get call, use one passed down by the calling class
- [x] Use `@retry_on_db_failure`?


dev/entrypoint.sh
```diff
@@ -9,4 +9,6 @@ printf "\nMigrations:\n"
 python openslides_backend/migrations/migrate.py finalize
 printf "\n"
 
+global/meta/dev/scripts/apply_db_schema.sh
+
```
r-peschke, line 13: This one we don't want. Sometimes it would leave your database corrupt, sometimes not. 
Finally all changes on an existing database have to be made with migrations, never with applying structural changes without data migration.
But we should talk about this one, soon.

jsangmeister: This is a remainder from a previous version of my PR. If you rebase this PR onto the current `feature/relational-db` branch, this should be gone

openslides_backend/services/datastore/read_adapter.py
```diff
@@ -0,0 +1,160 @@
...
+from datastore.shared.postgresql_backend.connection_handler import ConnectionHandler
+from datastore.shared.postgresql_backend.sql_query_helper import SqlQueryHelper
+from datastore.shared.typing import Collection, Id
+
+Model = dict[str, Any]
+
+
+class ReadAdapter:
+    connection: (
+        ConnectionHandler  # TODO use PgConnectionHandlerService from datastore ?
```
r-peschke, line 15: This theme we should talk about with @jsangmeister. Perhaps we could easily use that from datastore, but the  process/thread behavior is different.

dev/entrypoint.sh
luisa-beerboom: Done

openslides_backend/services/datastore/read_adapter.py
luisa-beerboom: I think I am currently getting that connection handler anyway. As far as I understand it the datastore injector that I use to initialize this variable in ln 20 delivers the `PgConnectionHandlerService` when the ConnectionHandler is injected (as per [this file](https://github.com/OpenSlides/openslides-datastore-service/blob/main/datastore/shared/postgresql_backend/__init__.py)).

requirements/partial/requirements_development.txt
```diff
@@ -11,6 +11,7 @@ pytest-cov==5.0.0
 pytest-profiling==1.7.0
 pyupgrade==3.15.2
 pyyaml==6.0.1
+psycopg[binary]==3.1.18  # only dev, in production use psycopg
```
r-peschke, line 14: May be should move this to the production requirement file, which should be used for all the common base packages.

tests/system/db_adapter/base_relational_db_test.py
```diff
@@ -0,0 +1,49 @@
+import os
+from typing import TypedDict
+from unittest import TestCase
+
+from psycopg2 import connect
```
r-peschke, line 5: IMO using psycopg2 and psycopg(3) in parallel is no good idea

jsangmeister: I agree, but for the transition, it is easiest to keep both. As more and more datstore code rotates out, we can update the code to version 3

tests/system/db_adapter/base_relational_db_test.py
```diff
@@ -0,0 +1,49 @@
...
+    def setUp(self) -> None:
+        self._filled_tables_cache: list[str] = []
+
+    def tearDown(self) -> None:
+        self._clearDB()
+        return super().tearDown()
+
+    def write_data(self, payloads: list[WritePayload]) -> None:
+        env = os.environ
+        connect_data = f"dbname='{env['DATABASE_NAME']}' user='{env['DATABASE_USER']}' host='{env['DATABASE_HOST']}' password='{env['PGPASSWORD']}'"
```
r-peschke, line 24: One connection should serve for the whole session, read and write together

luisa-beerboom:
@r-peschke This is absolutely not done yet (see list of points in description), but I have the basic functionality of the read_adapter done and I changed things to the extent where it is actually used in the proper `Adapter` now.
Could you take a look at my code and see if I implemented the already implemented basic processes correctly?

luisa-beerboom:
Will be handled differently

PR #2335, opened 2024-03-25
Add user to group in case of saml login

r-peschke:
Adding login user to meeting/group given in saml_id field-mapping description

### DIFF
--- a/openslides_backend/action/actions/user/save_saml_account.py
+++ b/openslides_backend/action/actions/user/save_saml_account.py
@@ -11,6 +11,7 @@ from ....shared.filters import And, FilterOperator
 from ....shared.interfaces.event import Event
 from ....shared.schema import schema_version
 from ....shared.typing import Schema
+from ...mixins.meeting_user_helper import get_meeting_user
 from ...mixins.send_email_mixin import EmailCheckMixin
 from ...mixins.singular_action_mixin import SingularActionMixin
 from ...util.action_type import ActionType
@@ -129,20 +130,13 @@ class UserSaveSamlAccount(
         )
         if len(users) == 1:
             self.user = next(iter(users.values()))
-            instance["id"] = self.user["id"]
-            if group_id:
-                meeting_users = self.datastore.filter(
-                    "meeting_user",
-                    And(
-                        [
-                            FilterOperator("user_id", "=", self.user["id"]),
-                            FilterOperator("meeting_id", "=", meeting_id),
-                        ]
-                    ),
-                    ["id", "group_ids"],
+            instance["id"] = (user_id := cast(int, self.user["id"]))
+            if meeting_id and group_id:
+                meeting_user = get_meeting_user(
+                    self.datastore, meeting_id, user_id, ["id", "group_ids"]
                 )
-                if len(meeting_users) == 1:
-                    old_group_ids = next(iter(meeting_users.values()))["group_ids"]
+                if meeting_user:
+                    old_group_ids = meeting_user["group_ids"]
                     if group_id not in old_group_ids:
                         instance["meeting_id"] = meeting_id
                         instance["group_ids"] = old_group_ids + [group_id]
@@ -204,7 +198,7 @@ class UserSaveSamlAccount(
             group_id = meeting["default_group_id"]
         else:
             self.logger.warning(
-                f"save_saml_account found {len(meetings)} meetings with external_id {external_id}"
+                f"save_saml_account found {len(meetings)} meetings with external_id '{external_id}'"
             )
             return NoneResult
         if external_group_id := meeting_info.get("external_group_id"):
@@ -219,10 +213,14 @@ class UserSaveSamlAccount(
                 mapped_fields=["id"],
             )
             if len(groups) == 1:
-                group_id = next(iter(groups.values()))["id"]
+                group_id = next(iter(groups.keys()))
+            else:
+                self.logger.warning(
+                    f"save_saml_account found no group in meeting '{external_id}' for '{external_group_id}', but use default_group of meeting"
+                )
         if not group_id:
             self.logger.warning(
-                f"save_saml_account found no group in meeting {external_id} for {external_group_id}"
+                f"save_saml_account found no group in meeting '{external_id}' for '{external_group_id}'"
             )
             return NoneResult
         return meeting.get("id"), group_id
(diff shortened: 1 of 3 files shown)


openslides_backend/action/actions/user/save_saml_account.py
```diff
@@ -124,27 +130,48 @@ def base_update_instance(self, instance: dict[str, Any]) -> dict[str, Any]:
...
+                meeting_users = self.datastore.filter(
+                    "meeting_user",
+                    And(
+                        [
+                            FilterOperator("user_id", "=", self.user["id"]),
+                            FilterOperator("meeting_id", "=", meeting_id),
+                        ]
+                    ),
+                    ["id", "group_ids"],
+                )
```
jsangmeister, line 143: Use https://github.com/OpenSlides/openslides-backend/blob/4aa7bf082f3bba8e29be765d82e364a49106848b/openslides_backend/action/mixins/meeting_user_helper.py#L15

openslides_backend/action/actions/user/save_saml_account.py
```diff
@@ -157,7 +184,45 @@ def set_defaults(self, instance: dict[str, Any]) -> dict[str, Any]:
...
+                filter=And(
+                    [
+                        FilterOperator("external_id", "=", external_group_id),
+                        FilterOperator("meeting_id", "=", meeting.get("id")),
+                    ]
+                ),
+                mapped_fields=["id"],
+            )
+            if len(groups) == 1:
+                group_id = next(iter(groups.values()))["id"]
```
jsangmeister, line 222:
```suggestion
                group_id = next(iter(groups.keys()))
```

openslides_backend/action/actions/user/save_saml_account.py
```diff
@@ -157,7 +184,45 @@ def set_defaults(self, instance: dict[str, Any]) -> dict[str, Any]:
...
+                collection="group",
+                filter=And(
+                    [
+                        FilterOperator("external_id", "=", external_group_id),
+                        FilterOperator("meeting_id", "=", meeting.get("id")),
+                    ]
+                ),
+                mapped_fields=["id"],
+            )
+            if len(groups) == 1:
```
jsangmeister, line 221: Maybe log a warning in the other case? (that the default group is used because the external group id was not found)

tests/system/action/user/test_save_saml_account.py
```diff
@@ -448,3 +448,185 @@ def test_create_saml_account_boolean_integer_types_error(self) -> None:
         )
         self.assert_status_code(response, 400)
         self.assertIn("Could not parse 2, expect boolean", response.json["message"])
+
+
+class UserAddToGroup(UserBaseSamlAccount):
+    def setUp(self) -> None:
+        super().setUp()
+        self.set_models(
```
jsangmeister, line 456: Move this to the other `set_models` call

tests/system/action/user/test_save_saml_account.py
```diff
@@ -448,3 +448,185 @@ def test_create_saml_account_boolean_integer_types_error(self) -> None:
...
+            "group/2", {"meeting_user_ids": [1], "external_id": "Delegates"}
+        )
+
+    def test_create_user_invalid_meeting(self) -> None:
+        """silent fail, user created and logged in"""
+        organization = self.get_model("organization/1")
+        for key in ("meta_deleted", "meta_position"):
+            organization.pop(key, None)
+        organization["saml_attr_mapping"]["meeting"]["external_id"] = "Kreistag"
+        self.set_models({"organization/1": organization})
```
jsangmeister, line 533: Cleaner solution: save the `saml_attr_mapping` into its own variable in the `setUp` and modify it here and simply update the respective field in the organization

(same for the other tests)

tests/system/action/user/test_save_saml_account.py
```diff
@@ -448,3 +448,185 @@ def test_create_saml_account_boolean_integer_types_error(self) -> None:
...
+        )
+        self.assert_model_exists(
+            "meeting_user/1", {"user_id": 1, "group_ids": [2], "meeting_id": 1}
+        )
+        self.assert_model_exists(
+            "group/2", {"meeting_user_ids": [1], "external_id": "Delegates"}
+        )
+
+    def test_create_user_invalid_meeting(self) -> None:
+        """silent fail, user created and logged in"""
```
jsangmeister, line 528: Maybe mock the logger to assert that a warning is logged

(same for the other error tests)

jsangmeister (changes requested):

tests/system/action/user/test_save_saml_account.py
r-peschke: > Maybe mock the logger to assert that a warning is logged
> 
> (same for the other error tests)

How is it  possible. The logger is mocked in the test application. That makes the "catlog" from pytest not available.

jsangmeister: You should be able to access the mock and assert that the warning method was called with the respective message

r-peschke: Then I have to access the application, created by factory, from inside a test. Don't know how to get the application.

jsangmeister: `self.app` contains the application and `self.services` is a shorthand for `self.app.services`

openslides_backend/action/actions/user/save_saml_account.py
r-peschke: done

tests/system/action/user/test_save_saml_account.py
r-peschke: done

tests/system/action/user/test_save_saml_account.py
r-peschke: done

tests/system/action/user/test_save_saml_account.py
r-peschke: done

tests/system/action/user/test_save_saml_account.py
```diff
@@ -526,13 +523,13 @@ def test_update_user_with_membership(self) -> None:
 
     def test_create_user_invalid_meeting(self) -> None:
         """silent fail, user created and logged in"""
-        organization = self.get_model("organization/1")
-        for key in ("meta_deleted", "meta_position"):
-            organization.pop(key, None)
-        organization["saml_attr_mapping"]["meeting"]["external_id"] = "Kreistag"
-        self.set_models({"organization/1": organization})
+        self.organization["saml_attr_mapping"]["meeting"]["external_id"] = "Kreistag"  # type: ignore
```
jsangmeister, line 526: Why do you need the `type: ignore` here? I would prefer this to be solved by assigning the correct type in the constructor (`dict[str, Any]` should suffice)

jsangmeister (approved):
I would prefer a better solution for the typing problem, but the code itself is fine

Issue #2311, opened 2024-03-18
Use relational database


jsangmeister:
I would make the following suggestions regarding the change to the relational DB in the backend:
- [ ] Create work on a feature branch `feature/relational-db`
- [ ] Revert some changes to the `models.yml`: The plan was to not change anything about the relation handling in the backend in the first step. With the current changes to the `models.yml`, this is no longer possible, as many `to` fields have been removed. I think we should keep the models file backwards-compatible as this will not only keep the relation handling intact, it will also simplify the update of the `generate-models` script and the general handling of the changes. This means re-introducing all removed relation meta data.
  - From this, it also follows that we do not want the `n:m` tables in the `models.py`, as all backend code is not fit to handle these. The changes in the write adapter must therefore get the required information from the additional fields in the `models.yml` as well as the helper functions supplied by the meta repo.
  - [ ] Afterwards, the `generate-models` script can be updated - in the first step, it should suffice to allow the new meta attributes and simply write them to the `models.py` as-is. If we later figure out a smarter way to handle these, we can still change it.
- [ ] Create exactly one transaction per backend request so that all queries (read and write) run inside this transaction
  - I don't think we need to implement a new postgres adapter, it should suffice to simple use the one from the datastore (and when the datastore is removed, the adapter can simply be moved to the backend). The adapter should probably be added as a new service to the service declaration container.
- [ ] Adjust the read methods to not use the datastore reader, but to directly access the database via the adapter. Since the views connection tables and calculated fields always have the same name as the collection, this should be rather easy. For all filter-related methods, the datastore already provides helper methods to translate the `Filter` classes to SQL queries, but they have to be updated to the new database definition. We have two options here:
  - Also create a feature branch in the datastore and implement the changes there
  - Copy all required methods to the backend (with the foresight that more may follow, so choosing a smart directory placement), since the datastore will be removed later anyway
- The above point includes the following methods:
  - [ ] `get`
  - [ ] `get_many`
  - [ ] `get_all`
  - [ ] `filter`
  - [ ] `exists`
  - [ ] `count`
  - [ ] `min`
  - [ ] `max`
- [ ] Adjust the write methods to write directly to the database. This includes the following methods:
  - [ ] `write`: The new procedure will be the following:
    - [ ] Extract all relevant data from the write request (i.e., all fields which are physically saved in the DB and not represented by a calculated view field) and drop the rest
[...]

PR #2337, opened 2024-03-26
Update model generation to relational models.yml

jsangmeister:
part of #2311

### DIFF
--- a/cli/generate_models.py
+++ b/cli/generate_models.py
@@ -199,7 +199,8 @@ class Attribute(Node):
     default: Any = None
     on_delete: OnDelete | None = None
     equal_fields: str | list[str] | None = None
-    contraints: dict[str, Any]
+    constraints: dict[str, Any]
+    is_view_field: bool = False
 
     FIELD_TEMPLATE = string.Template(
         "    ${field_name} = fields.${field_class}(${properties})\n"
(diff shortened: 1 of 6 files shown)


cli/generate_models.py
```diff
@@ -210,38 +211,39 @@ def __init__(self, value: str | dict) -> None:
...
+            self.type = value.pop("type")
             if self.type in RELATION_FIELD_CLASSES.keys():
-                self.to = To(value.get("to", {}))
-                self.on_delete = value.get("on_delete")
+                self.to = To(value.pop("to"))
+                self.on_delete = value.pop("on_delete", None)
+                if self.type in ("relation-list", "generic-relation-list") or value.get(
+                    "sql"
+                ):
+                    self.is_view_field = True
```
jsangmeister, line 226: @r-peschke this is the important section where it is determined whether a field is a view or table field. Please check if this conforms with the current/planned models schema or if there are other cases we need to account for

r-peschke: No, this is far away from current or planned case decision. I'll will present a table soon

r-peschke (changes requested):
As first step use the recommendation for checking field or not field. But I think we will replace this by functionality from the generation of relational db schema.

jsangmeister:
Tests are failing because `motion/identical_motion_ids` is still modelled as a `number[]` field because the current generate script cannot handle this case. This has to be changed before the tests here can succeed.

r-peschke (approved):

PR #2341, opened 2024-03-26
Fix `readd_last` with anonymous interposed questions

### LINKED ISSUE #2340: LoS: re-adding interposed questions not possible - hint improvement
**Description:** If an interprosed question needs to be re-added from a person who is on the interposed question list more than once, but multiple adding to LoS is not activated, a failure message appears which is not helpful. 

image

**Reproduction:**
1. open a meeting
2. open settings > LoS > active interposed questions
3. open agenda > create an item 
4. open participants > create two new users A and B
5. open the LoS of the created agenda item > place user A on the list and start it's contribution > add now two unknown speakers as interposed question > let first speak and stop it > try to re-add it > failure message appears. 

**What should happen:**
If the multiple adding to the LoS is not activated and an unknown users should be re-added twice to the list, the message should be "User is already on the list. Multiple additions are only possible if multiple adding is activated.".

jsangmeister:
fixes #2340

### DIFF
--- a/openslides_backend/action/actions/list_of_speakers/re_add_last.py
+++ b/openslides_backend/action/actions/list_of_speakers/re_add_last.py
@@ -86,7 +86,9 @@ class ListOfSpeakersReAddLastAction(UpdateAction):
             for speaker in speakers.values():
                 if (
                     speaker.get("end_time") is None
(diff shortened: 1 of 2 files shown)


Elblinator (approved):
function

r-peschke (approved):

Issue #2353, opened 2024-04-04
Remove `lxml` in favor of `bs4`


jsangmeister:
The breaking changes of `lxml` version 5.2 (seen in https://github.com/OpenSlides/openslides-backend/pull/2351) pointed me to the fact that we are currently using two tools (`lxml` and `bs4`) for the same purpose (extracting plain text from html content). We should conglomerate these and not use two tools without a good reason (which I don't see in this case). Since `bs4` seems to be the more versatile, forgiving and easier-to-use parser, I would opt to remove `lxml`.

PR #2354, opened 2024-04-04
Replace `lxml` with existing `bs4` methods

jsangmeister:
fixes https://github.com/OpenSlides/openslides-backend/issues/2353

moved the existing method using `bs4` into the `shared` folder and used it in the email mxin where `lxml` was previously used.

### DIFF
--- a/openslides_backend/action/actions/motion/mixins.py
+++ b/openslides_backend/action/actions/motion/mixins.py
@@ -2,12 +2,11 @@ from hashlib import md5
 from typing import Any
 
 import simplejson as json
-from bs4 import BeautifulSoup
-
-from openslides_backend.shared.filters import And, FilterOperator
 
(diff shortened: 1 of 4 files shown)


r-peschke:
May be we should use lxml, because bs4 can use other backend as (better) parser, including lxml. See https://dnmtechs.com/parsing-html-in-python-lxml-vs-beautifulsoup/

r-peschke (commented):
see comment

jsangmeister:
Yes, I know, but I don't think that we want a strict parser, but a reliable parser. The two places where we parse HTML are for identifying identical motions and generating the plaintext for emails and neither of those require strict syntax. Therefore, I think the current parser suffices and is even the best option. At the very least, it is not worth investing time identifying the best parser possible as probably all of them will work.

r-peschke (approved):

PR #2360, opened 2024-04-09
Move datastore code to backend

jsangmeister:


jsangmeister:
@r-peschke I looked into migrating psycopg, but it seems that the error handling changed completely (the `pgcode` and `pgerror` attributes of all errors were removed and the [`sqlstate`](https://www.psycopg.org/psycopg3/docs/api/errors.html#psycopg.Error.sqlstate) attribute was added, but I'm not sure how much of the functionality this replaces). Since the error handling of the connection handler class has grown a lot based an which errors occured and it is additionally very hard to test, I don't feel comfortable trying to migrate it. The propability that something breaks without us noticing seems pretty high. Therefore, I think we should work with version 2 for the moment.

r-peschke:
> @r-peschke I looked into migrating psycopg, but it seems that the error handling changed completely (the `pgcode` and `pgerror` attributes of all errors were removed and the [`sqlstate`](https://www.psycopg.org/psycopg3/docs/api/errors.html#psycopg.Error.sqlstate) attribute was added, but I'm not sure how much of the functionality this replaces). Since the error handling of the connection handler class has grown a lot based an which errors occured and it is additionally very hard to test, I don't feel comfortable trying to migrate it. The propability that something breaks without us noticing seems pretty high. Therefore, I think we should work with version 2 for the moment.

In current meeting we decided to migrate to psycopg now. Reasons:
- it will not be easier to migrate in some months, when we want to use the relDB-schema in production. Migrating now gives us more time to detect regressions
- the connection handler is not that much code

dev/dc.local.yml
```diff
@@ -11,26 +11,6 @@ services:
         environment:
             - PYTHONPATH=/app:/datastore-service:/authlib
             - MYPYPATH=/app:/datastore-service:/authlib
-    datastore-writer:
```
r-peschke, line 14: IMO the volume mapping in line 9 isn't needed anymore

dev/entrypoint.sh
```diff
@@ -5,8 +5,8 @@ set -e
 source scripts/export_datastore_variables.sh
 scripts/wait.sh $DATASTORE_WRITER_HOST $DATASTORE_WRITER_PORT
 
-printf "\nMigrations:\n"
-python openslides_backend/migrations/migrate.py finalize
-printf "\n"
+# printf "\nMigrations:\n"
```
r-peschke, line 8: Why did you comment/removed this? IMO this is still needed.

openslides_backend/datastore/reader/flask_frontend/flask_frontend.py
```diff
@@ -0,0 +1,15 @@
+from flask import Flask
```
r-peschke, line 1: Why do we need the reader application and the whole flask frontend?

r-peschke (commented):
There are some questions. Could you please answer them here for documentation. I'm not sure about routing, flask frontend and datastore service.

dev/dc.local.yml
jsangmeister: right, removed

dev/entrypoint.sh
jsangmeister: commented it out for developing purposes, added it back now

openslides_backend/datastore/reader/flask_frontend/flask_frontend.py
jsangmeister: It is used (at least) in the migration tests. The same goes for the writer. They can theoretically be removed when the adjusted reader/writer are implemented, but then the tests have to be adjusted as well. For now, it's easiest to keep them around.

r-peschke (approved):

PR #2385, opened 2024-04-24
Add get_forwarding_committee presenter

### LINKED ISSUE #2356: update to the ```get_forwarding_meetings``` presenter
see for context https://github.com/OpenSlides/openslides-client/issues/3429

> This will require an update to the get_forwarding_meetings presenter as it currently returns only the meetings which can be forwarded to.

Please update the presenter in a way that the correct committees can be displayed

reiterl:
Resolve #2356 

payload: A committee_id, returns: a list with committee names

@jsangmeister  I have found no matching perm. We will need to add meeting_id to the payload, if we want to use
the user.can_manage perm.

### DIFF
--- a/openslides_backend/presenter/__init__.py
+++ b/openslides_backend/presenter/__init__.py
@@ -4,6 +4,7 @@ from . import (  # noqa
     check_mediafile_id,
     export_meeting,
     get_active_users_amount,
+    get_forwarding_committees,
     get_forwarding_meetings,
     get_history_information,
     get_user_related_models,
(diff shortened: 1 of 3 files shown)


jsangmeister:
> @jsangmeister I have found no matching perm. We will need to add meeting_id to the payload, if we want to use the user.can_manage perm.

Good point. I would suggest to replace the committee id of the payload with a meeting id and use the meeting's committee in the presenter. Also, the required permission is `motion.can_manage`.

reiterl:
@jsangmeister  points me to Location.joinWithSlash (https://angular.io/api/common/Location#static-properties) 
which could help, I need to test it.

reiterl:
Updated, use now meeting_id as payload and use the committee of the meeting. 
Also add the perms check.

jsangmeister (approved):

PR #2399, opened 2024-04-26
New flags in motion.create_forwarded

### LINKED ISSUE #2382: Enhance `motion.create_forwarded`
- add boolean to control which submitter is used: The original one or the forwarding committee
- add boolean to control which number is used: The original one or none

### LINKED ISSUE #2419: Rework behaviour that a User is created if motions are forwarded
**Currently:**
If a motion is forwarded then a new User is created. 
The surname of this user is the committee name -it was forwarded from- and the username is the committee name without spaces. 
The User is a new participant in the meeting the motion was forwarded to and the user is assigned as a submitter in the forwarded motion.

**To discuss behaviour:**
Should the action to forward a motion maybe not create a new (inactive and no natural person) User who is then assigned as a submitter
But maybe the action to forward a motion should assign the Committee name as the extension for the submitter.
This would prevent a new automatically created (and very likely unused) account.
If the committee name is assigned in the extension-field then the name is still displayed everywhere the submitters are displayed. 
This would also align the automatic forwarding behaviour with new feature described in OpenSlides/openslides-client#3429

luisa-beerboom:
Closes #2382 
Closes #2419 

For now I handled the problem of "how to use original submitters if they aren't in the target meeting" by using the meeting_user in the other meeting (discussed this with @rrenkert) 
If we're going to do this, we should see if there's strange bugs in the client.

Things like:
- Can't see the foreign-meeting submitter names (i.e. missing fields in autoupdate subscriptions or restrictor problems)
- Origin motion submitter can see forwarded motions in origin meeting's motion list (unlikely)
- Origin motion submitter appears in participant list of forward meeting (may only happen after opening forwarded motion)

Removed `forwarding_user_id` relation from the committee, as it loses it's purpose beginning with this PR

### DIFF
--- a/openslides_backend/action/actions/user/merge_together.py
+++ b/openslides_backend/action/actions/user/merge_together.py
@@ -99,7 +99,6 @@ class UserMergeTogether(
                 ],
                 "error": [
                     "is_demo_user",
-                    "forwarding_committee_ids",
                 ],
                 "merge": [
                     "committee_ids",
--- a/tests/system/action/user/test_merge_together.py
+++ b/tests/system/action/user/test_merge_together.py
@@ -463,20 +463,6 @@ class UserMergeTogether(BaseVoteTestCase):
             response.json["message"],
         )
 
-    def test_merge_forwarding_committee_ids_error(self) -> None:
-        self.set_models(
-            {
-                "committee/3": {"forwarding_user_id": 3},
-                "user/3": {"forwarding_committee_ids": [3]},
-            }
-        )
-        response = self.request("user.merge_together", {"id": 2, "user_ids": [3, 4]})
-        self.assert_status_code(response, 400)
-        self.assertIn(
-            "Cannot merge user models that have forwarding_committee_ids set: Problem in user/3",
-            response.json["message"],
-        )
-
     def test_merge_saml_id_error(self) -> None:
         self.set_models({"user/3": {"saml_id": "SAML"}})
         response = self.request("user.merge_together", {"id": 2, "user_ids": [3, 4]})


openslides_backend/action/actions/motion/create_forwarded.py
```diff
@@ -97,73 +103,123 @@ def update_instance(self, instance: dict[str, Any]) -> dict[str, Any]:
...
-                [
-                    {
-                        "user_id": forwarding_user_id,
-                        "meeting_id": meeting_id,
-                        "group_ids": [target_meeting["default_group_id"]],
-                    }
-                ],
-            )
-        instance["submitter_ids"] = [forwarding_user_id]
+            if committee.get("forwarding_user_id"):
```
r-peschke, line 135: Proposal (get forwarding_user_id on checking and delete line 136):
```
if forwarding_user_id := committee.get("forwarding_user_id"):
    meeting_id = instance["meeting_id"]
    forwarding_user_groups = self.get_groups_from_meeting_user(
                    meeting_id, forwarding_user_id
                ) ...
```

docs/actions/motion.create_forwarded.md
```diff
@@ -25,6 +27,8 @@ The original motion must be updated as well (this is done by the automatic relat
 * The unique `id` of the newly created motion has to be linked to the _origin motion_s `derived_motion_ids` field.
   * Deleting the newly created motion has to ensure that the corresponding entry was removed from the _origin motion_s `derived_motion_ids` field
 
+The optional flags `use_original_submitter` and `use_original_number` will cause the original submitters and original numbers to be used in the new motion respectively. In case of the submitters, the action will use the origin meetings meeting_user for the `meeting_user_id` field of the new submitter instance in the target meeting.
```
r-peschke, line 30: IMO it is no good idea to use the meeting_user from other meeting in target meeting.
Alternatives:

1. Create meeting_user in target_meeting with empty group (could be used also for committee user)
2. motion_submitter should use `user`  instead of `meeting_user`. Disadvantage: getting forward motions per meeting and user only via filter

@rrenkert should have a look too

r-peschke (changes requested):

docs/actions/motion.create_forwarded.md
luisa-beerboom: WDYM by empty group?

r-peschke: IMO we shouldn't use meeting_user instances from meeting A in meeting B. The meeting_id is an integral part of an meeting_user instance. 
Empty group means a group without any permission. This could be part of a solution for alternative 1.

luisa-beerboom: @rrenkert what should happen here?

openslides_backend/action/actions/motion/create_forwarded.py
```diff
@@ -97,73 +103,123 @@ def update_instance(self, instance: dict[str, Any]) -> dict[str, Any]:
...
-                                "user_id": forwarding_user_id,
-                                "group_ids": [target_meeting["default_group_id"]],
-                            }
-                        ],
-                    )
-                else:
+            if submitters:
+                self.apply_instance(instance)
+                weight = 1
+                for submitter in submitters.values():
```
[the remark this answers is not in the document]
luisa-beerboom: Done

openslides_backend/action/actions/motion/create_forwarded.py
```diff
@@ -215,42 +40,8 @@ def check_permissions(self, instance: dict[str, Any]) -> None:
...
-                ]
-            )
-        return forwarded_entries | {
-            fqid_from_collection_and_id("motion", instance["id"]): [
-                "Motion created (forwarded)"
-            ]
-            for instance in self.instances
-        }
+    def should_forward_amendments(self, instance: dict[str, Any]) -> bool:
+        return instance.pop("with_amendments", False)
```
r-peschke, line 47: Dangerous combination of method name (seems to be idempotent like a "variable") and internal behavior (definitely not idempotent). Please change behavior or name (=> pop_with_amendment)

openslides_backend/action/actions/user/create.py
```diff
@@ -51,7 +51,7 @@ class UserCreate(
             "is_present_in_meeting_ids",
             "committee_management_ids",
             "is_demo_user",
-            "forwarding_committee_ids",
+            # "forwarding_committee_ids",
```
r-peschke, line 54: missing documentation change of action and if you really want it, remove the line

openslides_backend/action/actions/user/create_update_permissions_mixin.py
```diff
@@ -213,8 +213,8 @@ def check_permissions(self, instance: dict[str, Any]) -> None:
         """
         self.assert_not_anonymous()
 
-        if "forwarding_committee_ids" in instance:
-            raise PermissionDenied("forwarding_committee_ids is not allowed.")
+        # if "forwarding_committee_ids" in instance:
```
r-peschke, line 216: remove code and document in action descriptions, if you really want

openslides_backend/action/mixins/motion_meeting_user_create.py
```diff
@@ -18,7 +18,7 @@
 
 
 def build_motion_meeting_user_create_action(
```
r-peschke, line 20: Why do you want to create a class at run time?

openslides_backend/migrations/migrations/0054_remove_forwarding_user.py
```diff
@@ -0,0 +1,28 @@
...
+class Migration(BaseModelMigration):
+    """
+    This migration removes the forwarding user relation
+    """
+
+    target_migration_index = 55
+
+    def migrate_models(self) -> list[BaseRequestEvent] | None:
+        events: list[BaseRequestEvent] = []
+        for collection, field in [
```
r-peschke, line 15: no need to put the collection into loop head and where is the collection from?

r-peschke (changes requested):

docs/actions/motion.create_forwarded_amendment.md
```diff
@@ -0,0 +1,22 @@
...
+```
+{
+// Required
+    meeting_id: Id;
+    title: string;
+    lead_motion_id: Id;
+    origin_id: Id;
+
+// Optional
+    text: HTML;
```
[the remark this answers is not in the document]
luisa-beerboom: `text` is also in the normal motion payload. There it is required, since it is a required field in the motion.
amendments have either `text` or `amendment_paragraphs`, so these must be optional here.

The reason they are in the payload and not directly handled by the code is so that `create_forwarded_amendment` can re-use as much of the normal amendment code as possible.
It's an internal action anyway.

openslides_backend/action/mixins/motion_meeting_user_create.py
luisa-beerboom: This function is used to create base classes for about four actions for motion-related meeting_user-models.
Therefore I don't think it is called that often.

I am merely editing this, because I need some of these actions changed.

If you need further information on how this all works, why don't you look at [this commit](https://github.com/OpenSlides/openslides-backend/commit/218ca391bcd8c628429c1c58e90ced987a3733ed), _where @jsangmeister wrote this code_

openslides_backend/migrations/migrations/0054_remove_forwarding_user.py
luisa-beerboom: This loop goes through a list of tuples in a (collection, field) format.
That's where the collection comes from.

This is the case since this migration needs the exact same functionality for two different fields in two different collections. (i.e. "Delete the field from all instances that have it")

This was the least redundant way to write that code.

rrenkert:
@luisa-beerboom can you please resolve the conflicts and base the PR on the `feature/motion-forward` branch?

openslides_backend/action/actions/motion/create_forwarded.py
luisa-beerboom: Changed the function to fit the name instead

openslides_backend/action/actions/user/create.py
luisa-beerboom: Can't change the documentation, because this field was never mentioned there in the first place.

openslides_backend/action/actions/user/create_update_permissions_mixin.py
luisa-beerboom: see comment in user.create for why the documentation wasn't amended

openslides_backend/action/actions/motion/base_create_forwarded.py
```diff
@@ -0,0 +1,376 @@
...
+    def prefetch(self, action_data: ActionData) -> None:
+        self.datastore.get_many(
+            [
+                GetManyRequest(
+                    "meeting",
+                    list(
+                        {
+                            meeting_id
+                            for instance in action_data
+                            if (meeting_id := instance.get("meeting_id"))
```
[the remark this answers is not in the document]
luisa-beerboom: That is not quite so easy because the prefetch is called before the schema validation.
Not catching this here is going to lead to error messages unintelligible to the user

openslides_backend/action/actions/motion/base_create_forwarded.py
```diff
@@ -0,0 +1,376 @@
...
+                        "motion_ids",
+                    ],
+                ),
+                GetManyRequest(
+                    "motion",
+                    list(
+                        {
+                            origin_id
+                            for instance in action_data
+                            if (origin_id := instance.get("origin_id"))
```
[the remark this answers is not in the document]
luisa-beerboom: see above

openslides_backend/action/actions/motion/base_create_forwarded.py
```diff
@@ -0,0 +1,376 @@
...
+from .create_base import MotionCreateBase
+
+
+class BaseMotionCreateForwarded(TextHashMixin, MotionCreateBase):
+    """
+    Base create action for forwarded motions.
+    """
+
+    def prefetch(self, action_data: ActionData) -> None:
+        self.datastore.get_many(
```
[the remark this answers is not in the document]
luisa-beerboom: Idk, most of this class is pre-existing code that was simply transferred over from the normal motion forward.
I don't remember touching the prefetch at all, therefore I also didn't put too much thought into it either.

openslides_backend/action/actions/motion/create_forwarded.py
r-peschke: removed

r-peschke (commented):
Please resolve the remaining conversations

luisa-beerboom:
> Please resolve the remaining conversations

Then please reply to it. Usually, if I reply without changing the code, it is because I want further explanation.

luisa-beerboom:
@r-peschke I've pretty much removed all of the locks on the datastore calls in the forwarding methods, as none of the data called is actually overwritten, since this is, in fact, a type of create function.

docs/actions/motion.create_forwarded.md
r-peschke: outdated, this solution was removed

openslides_backend/action/actions/motion/base_create_forwarded.py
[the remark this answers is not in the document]
r-peschke: changed

r-peschke:
Please fix failing tests. Except this one and the remark from @rrenkert (basing on feature/motion_forward) thi sis okay for me

Elblinator:
I'll dismiss @r-peschke CR adn merge this pr given that this PR stops work on the issue https://github.com/OpenSlides/openslides-backend/issues/2519 @r-peschke comment "this is okay for me" and that the feature is merged into it's own feature branch

PR #2403, opened 2024-05-02
Fix LoS countdowns on speaker deletion

### LINKED ISSUE #2402: Fix LoS - with active structure level countdowns when user is cleared
See OpenSlides/openslides-client#3555 for context

luisa-beerboom:
Closes #2402 

@rrenkert Does this need a migration to remove previous instances of this bug, or is this fine as is?

### DIFF
--- a/openslides_backend/action/actions/speaker/delete.py
+++ b/openslides_backend/action/actions/speaker/delete.py
@@ -1,4 +1,7 @@
+from time import time
 from typing import Any
+
+from openslides_backend.action.util.typing import ActionData
 
 from ....models.models import Speaker
 from ....permissions.permissions import Permissions
@@ -6,11 +9,14 @@ from ....shared.patterns import fqid_from_collection_and_id
 from ...generics.delete import DeleteAction
 from ...util.default_schema import DefaultSchema
 from ...util.register import register_action
+from ..projector_countdown.mixins import CountdownCommand, CountdownControl
 from ..user.delegation_based_restriction_mixin import DelegationBasedRestrictionMixin
 
 
 @register_action("speaker.delete")
-class SpeakerDeleteAction(DeleteAction, DelegationBasedRestrictionMixin):
+class SpeakerDeleteAction(
+    DeleteAction, CountdownControl, DelegationBasedRestrictionMixin
+):
     model = Speaker()
     schema = DefaultSchema(Speaker()).get_delete_schema()
     permission = Permissions.ListOfSpeakers.CAN_MANAGE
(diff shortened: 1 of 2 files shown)


Elblinator (approved):

rrenkert:
I don't think we need a migration here. The error occurs dependent on the speech state that is likely short when it is the "speeking" state.

openslides_backend/action/actions/speaker/delete.py
```diff
@@ -34,3 +40,30 @@ def check_permissions(self, instance: dict[str, Any]) -> None:
             if meeting_user.get("user_id") == self.user_id and not len(restricted):
                 return
         super().check_permissions(instance)
+
+    def get_updated_instances(self, action_data: ActionData) -> ActionData:
+        self.end_time = round(time())
```
r-peschke, line 45: Seems that this set's always the end_time and will overwrite an already closed speaker

luisa-beerboom: 1. This is a delete function. Any speaker whose id is given will be deleted in the same step.
2. The end_time variable is looked up only in line 67, for the expressed purpose of being able to calculate the correct `remaining_time` when stopping the structure level countdown.
3. This mirrors the functionality in `speaker.end_speech`.

openslides_backend/action/actions/speaker/delete.py
```diff
@@ -34,3 +40,30 @@ def check_permissions(self, instance: dict[str, Any]) -> None:
...
+                "unpause_time",
+                "structure_level_list_of_speakers_id",
+            ],
+        )
+        if (
+            speaker.get("begin_time")
+            and not speaker.get("end_time")
+            and not speaker.get("pause_time")
+        ):
+            self.decrease_structure_level_countdown(self.end_time, speaker)
```
r-peschke, line 67: if the end_time is only looked up here, then you should get it here and use round(time()) as function call parameter. As consequence the changes in lines 43-47 are not necessary.

r-peschke (changes requested):

openslides_backend/action/actions/speaker/delete.py
luisa-beerboom: Then there may be discrepancies when multiple speakers are stopped at once

r-peschke: More than 1 speaker talking at the same would be no good idea at all;-)

openslides_backend/action/actions/speaker/delete.py
r-peschke: IMO you don't need to calculate the remaining time, if the end_time was set before.

PR #2434, opened 2024-05-21
Fix CheckMediafileId presenter for anonymous user

bastianjoel:


r-peschke (approved):
There are some differences between the existing code and the description in  https://github.com/OpenSlides/openslides-autoupdate-service/blob/main/internal/restrict/collection/mediafile.go. There are some paths to give an anonymous user access to mediafile, for instance with owner collection and mediafile with token or with disabled anonymous in meeting with mediafile.CAN_SEE and public mediafiles or some projector permissions.
But as a POC for anonymous user I'll approve it.

bastianjoel:
> There are some paths to give an anonymous user access to mediafile, for instance with owner collection and mediafile with token or with disabled anonymous in meeting with mediafile.CAN_SEE and public mediafiles or some projector permissions.

This has nothing to do with this PR. The changes made here are implemented the same in the restrictor. 
Please open an issue if you think that there are other cases which need to be adjusted.

PR #2464, opened 2024-06-04
Add option to project without side effects

### LINKED ISSUE #2467: Add projections without deleting same type projections from other projectors
With https://github.com/OpenSlides/openslides-client/pull/3691 we now use the `options` attribute to distinguish between chyron projections. 
For that we need to be able to project chyrons to new projectors without affecting the `options` attribute on other projectors. This is currently not possible as projections are identified identical by attributes `meeting_id`, `content_object_id`, `stable` and `type`. When calling `projector.project` all projections that are identified as the same are moved to history or get unset.

bastianjoel:
resolves #2467

needed for https://github.com/OpenSlides/openslides-client/pull/3691


emanuelschuetze (approved):

luisa-beerboom (changes requested):
Write tests with the new field.
Also: If the field has only two possible options, why not use a boolean?

luisa-beerboom (approved):

PR #2476, opened 2024-06-07
Removed member_number name generation from user imports

### LINKED ISSUE #2474: Participant csv import: Membership number and new created user
**Description:** If a membership number is used in the csv file that cannot be assigned to an existing account, a new account is created during import. The membership number is used as base for a username. This behavior is undesirable as it leads to accounts that either have to be deleted or updated immediately, because they are practically unusable.

Note: In the csv file, a value is only entered for the membership number, all other cells remain empty. 

**Reproduction:**
1. open a meeting
2. open participants > three dot menu > import
3. download the example file > delete all data > add a random value in the field member_number. save it.
4. import the csv file > A new user will be created.

**What should happen:**
If only the membership number is specified in the csv file, which cannot be referenced, no account may be created. The import must be blocked. An error message should appear stating that a membership number is not sufficient to create a new account.

luisa-beerboom:
Closes #2474

### DIFF
--- a/openslides_backend/action/actions/user/base_json_upload.py
+++ b/openslides_backend/action/actions/user/base_json_upload.py
@@ -155,14 +155,6 @@ class BaseUserJsonUpload(UsernameMixin, BaseJsonUploadAction):
             elif check_result == ResultType.NOT_FOUND or id_ == 0:
                 self.row_state = ImportState.NEW
         else:
-            # if not entry.get("username") or (
-            #     not isinstance(entry.get("username"), str)
-            #     and not entry.get("username", {}).get("value")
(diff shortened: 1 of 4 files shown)


MSoeb (approved):

Elblinator (changes requested):
Only the first Error is shown in the preview
In the following screenshot every line is a user I tried to create via only the member number
Screenshot_20240607_135218

r-peschke (approved):
Seems to be correct, but we really should refactor the validate_entry code!

r-peschke (approved):

Elblinator (approved):

PR #2487, opened 2024-06-24
Change scope calculation for committee members

### LINKED ISSUE #2448: Allow committee admins to change personal information
The counter part Issue from the client: https://github.com/OpenSlides/openslides-client/issues/3702

**Wanted behaviour:**
The committee admin should be allowed to edit all personal information from every user who is in **their** committee.

It is possible that one committee admin is not a committee admin in a different committee, and it is possible that a user from their committee is also in different committees: 
- A committee admin should ONLY be able to change data from users where they themselves are the committee admin
- If a user is in more committees, then the committee admin still should be able to change their personal data

luisa-beerboom:
Closes #2448 

Includes appropriate changes to get_user_scope presenter, should probably be used when making the appropriate amendments in the client


openslides_backend/permissions/permission_helper.py
```diff
@@ -119,6 +119,32 @@ def has_committee_management_level(
...
+    """Checks wether a user has the minimum necessary CommitteeManagementLevel"""
+    if user_id > 0:
+        cml_fields = ["committee_management_ids"]
+        user = datastore.get(
+            fqid_from_collection_and_id("user", user_id),
+            [*cml_fields],
+            lock_result=False,
+            use_changed_models=False,
+        )
+        if user.get("organization_management_level") in (
```
r-peschke, line 137: you forget to read the oml and additionally you should have a test to detect this

r-peschke (changes requested):

openslides_backend/permissions/permission_helper.py
luisa-beerboom: 1. The oml is read right in this line at `user.get("organization_management_level")
2. This code is in case the user is a super- or orga-admin because in this case he is implicitly also allowed to above any committee_admin for the sake of the user_scope

r-peschke (approved):

PR #2494, opened 2024-06-25
Fix overwriting superadmin password on container start

peb-adr:


peb-adr:
When the backend creates initial data on container start, the `initial_import` action will throw an exception if DS is not empty resulting in no organization data being overwritten.

However the follwing `set_password` action was still executed resulting in the superadmin password being reset.\
Returning when catching the DS-not-empty exception fixes this.

r-peschke (approved):

PR #2500, opened 2024-06-26
create small fix for issue 2365 + test

### LINKED ISSUE #2365: Can't change direction of delegation in one transaction
Something like
```
{
  "action": "user.update",
  "data": [
    {
      "id": 4,
      "meeting_id": 2,
      "vote_delegated_to_id": null,
      "vote_delegations_from_ids": [
        6
      ],
    }
  ]
}
```

Will throw error `User(s) [5] can't delegate their votes because they receive vote delegations.` if there was already a delegation between these users (pointing in the other direction) before the call.
Since that very delegation is supposed to be deleted through this action, there is no reason why this should raise an error.
This should be fixed.

It may also be worthwhile to look into whether this also happens if the same transaction is started from the other direction (from-delegation deleted, to delegation opened)

hjanott:
Closes #2365

### DIFF
--- a/openslides_backend/action/actions/meeting_user/mixin.py
+++ b/openslides_backend/action/actions/meeting_user/mixin.py
@@ -76,7 +76,9 @@ class MeetingUserMixin(MeetingUserHistoryMixin):
                 raise ActionException(
                     f"User {meeting_user_delegated_to.get('user_id')}'s delegation id don't belong to meeting {meeting_id_self}."
                 )
-            if meeting_user_delegated_to.get("vote_delegated_to_id"):
+            if instance.get("id") != meeting_user_delegated_to.get(
+                "vote_delegated_to_id"
+            ) and meeting_user_delegated_to.get("vote_delegated_to_id"):
                 raise ActionException(
                     f"User {user_id_self} cannot delegate his vote to user {meeting_user_delegated_to['user_id']}, because that user has delegated his vote himself."
                 )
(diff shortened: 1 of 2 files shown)


tests/system/action/meeting_user/test_set_data_delegation.py
```diff
@@ -159,6 +159,31 @@ def test_delegated_to_error_user_cannot_delegate_has_delegations_himself(
...
+        )
+        self.assert_status_code(response, 200)
+        self.assert_model_exists(
+            "meeting_user/14",
+            {"vote_delegated_to_id": 13, "vote_delegations_from_ids": []},
+        )
+        self.assert_model_exists(
+            "meeting_user/13",
+            {"vote_delegated_to_id": None, "vote_delegations_from_ids": [14]},
+        )
```
luisa-beerboom, line 185: Also add this test the other way around: With 14 delegating to 13 and the request payload being `{"vote_delegations_from_ids": [13], "vote_delegated_to_id": None}`
I am quite sure that that will have to be fixed as well

openslides_backend/action/actions/meeting_user/mixin.py
```diff
@@ -76,7 +76,9 @@ def check_vote_delegated_to_id(
                 raise ActionException(
                     f"User {meeting_user_delegated_to.get('user_id')}'s delegation id don't belong to meeting {meeting_id_self}."
                 )
-            if meeting_user_delegated_to.get("vote_delegated_to_id"):
+            if instance.get("id") != meeting_user_delegated_to.get(
+                "vote_delegated_to_id"
+            ) and meeting_user_delegated_to.get("vote_delegated_to_id"):
```
luisa-beerboom, line 81:
```suggestion
            if meeting_user_delegated_to.get("vote_delegated_to_id") 
            and instance["id"] != meeting_user_delegated_to["vote_delegated_to_id"]:
```

luisa-beerboom (changes requested):

luisa-beerboom (approved):

PR #2511, opened 2024-07-03
Fix removal recognition in get_user_related_models_presenter

### LINKED ISSUE #2449: Change presenter ```get_user_related_models```
**Current behaviour:**
The presenter currently gives back meeting ids even if the User is only in there as a remnant and not as a informational important User:
Example response if the User is not ANYMORE in the meeting (but was in there before) and has no other connections to the meeting: 
```
id: 5
is_active_in_organization_id: 1
name: "Meetings Name ABC"
```
Example response from the presenter if the User is a candidate, a submitter, a speaker and a manager in the same meeting:
```
{
  "id": 6,
  "name": "Meeting name",
  "is_active_in_organization_id": 1,
  "speaker_ids": [ 20 ],
  "motion_submitter_ids": [ 40 ],
  "assignment_candidate_ids": [ 7 ]
}
```

**Reproduction to get the first response:**
It is important that you add User C to (at least) two meetings before removing/deleting them from any
1. Create User C 
2. Create two meetings
3. Add C to both meetings 
4. In Meeting A delete User C
5. In meeting B delete User C
6. Navigate to the account list
7. C has no assigned Meetings
8. Delete User C from the account list
9. In the deletion dialog both meetings will be listed 

**Wanted behaviour:**
If a User only has a meeting_user_id but is not assigned to any group, or is in one of the other groups (candidate, submitter, speaker) then this meeting Id should not be returned by the presenter

luisa-beerboom:
Closes #2449

### DIFF
--- a/openslides_backend/presenter/get_user_related_models.py
+++ b/openslides_backend/presenter/get_user_related_models.py
@@ -7,7 +7,6 @@ from openslides_backend.shared.schema import id_list_schema
 
 from ..services.datastore.commands import GetManyRequest
 from ..shared.exceptions import PresenterException
-from ..shared.filters import And, FilterOperator
 from ..shared.schema import schema_version
 from .base import BasePresenter
 from .presenter import register_presenter
@@ -90,7 +89,7 @@ class GetUserRelatedModels(UserScopeMixin, BasePresenter):
         return committees_data
 
     def get_meetings_data(self, user: dict[str, Any]) -> list[dict[str, Any]]:
-        if not user.get("meeting_user_ids"):
+        if not (meeting_user_ids := user.get("meeting_user_ids")):
             return []
 
         result_fields = (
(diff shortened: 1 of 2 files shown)


openslides_backend/presenter/get_user_related_models.py
```diff
@@ -90,22 +89,41 @@ def get_committees_data(self, user: dict[str, Any]) -> list[dict[str, Any]]:
...
+                [
+                    GetManyRequest(
+                        "meeting_user",
+                        meeting_user_ids,
+                        [*result_fields, "group_ids", "meeting_id"],
+                    )
+                ]
+            )["meeting_user"].values()
+        )
+        meeting_users = [
```
r-peschke, line 111: This part i snot necessary. After line 110 the meeeting_users are read in in memory. Something like
`meeting_user = [meeting_user for meeting_user in meeting_users if meeting_user.pop("group_ids", None)]`
is sufficient.
Then continue with 125

r-peschke (changes requested):

r-peschke (changes requested):

openslides_backend/presenter/get_user_related_models.py
```diff
@@ -90,22 +89,41 @@ def get_committees_data(self, user: dict[str, Any]) -> list[dict[str, Any]]:
...
         result_fields = (
             "speaker_ids",
             "motion_submitter_ids",
             "assignment_candidate_ids",
         )
-        meeting_users = self.datastore.filter(
-            "meeting_user",
-            And(
-                FilterOperator("user_id", "=", user["id"]),
-                FilterOperator("group_ids", "!=", []),
```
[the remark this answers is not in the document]
luisa-beerboom: The new method exists precisely because this didn't work when I tried it.

openslides_backend/presenter/get_user_related_models.py
luisa-beerboom: Deleted the other part

r-peschke (approved):

PR #2526, opened 2024-07-15
Set openslides-meta URL to HTTPS instead of SSH

peb-adr:


peb-adr:
I'm assuming this was changed to SSH by accident in #1928.

For uniformity with the other services that use the meta submodule and compatibility I am changing it back to HTTPS.

m-schieder (approved):

PR #2531, opened 2024-07-22
Implement locking out participants

### LINKED ISSUE #2498: Add feature „lock participants out of a meeting“
See https://github.com/OpenSlides/OpenSlides/issues/6756#issuecomment-2159989520 for reference
**Wanted behaviour:**
- [x] If locked out the participant is still in the meeting
- [x] A participant with the permission of user.canManage can lock out users without user.canManage permission or cml or oml permissions.
 - [x] It should be possible to change is_locked_out and the group (with canManage rights) at the same time, as long as you try to activate a combination which is allowed 
- [x] What should happen you try to assign a group with canManage permission to a locked out participant?
     - [x] You are told that this is a locked out participant and you cannot give this participant canManage rights until you lift the lock out first 
     - [x] participant edit-dialog: you should be able to select any group but if you select a group with canManage rights and you try to save it an Error should be thrown: It should tell you that a locked out participant is not allowed to have user.canManage rights and you have to lift the lockout first

- [x] What should happen you try to lockout a participant with canManage permission?
     - [x] You are told that this is a participant with canManage rights and you cannot lock out such participants until they lose this right
- [x] Who can NOT be locked out: 
    - [x] You from yourself
    - [x] (Meeting-) Admins
    - [x] participant with user.canManage permission
    - [x] Superadmins
    - [x] Orgaadmins
    - [x] Accountadmins
    - [x] Committeeadmins (but only from their own committees)
- [x] Who can be locked out: 
    - [x] Every User in a meeting who does not  have the permission participant.canManage
    - [x] Committeeadmins (but only from meetings where they are not committeeadmins)

- [x] Imports: 
    - [x] If the import is faulty (you try to lock out <-> assign canManage rights) then the import should be stopped and every line which is faulty should have the error symbol

luisa-beerboom:
Closes #2498


openslides_backend/action/actions/meeting_user/mixin.py
```diff
@@ -9,9 +17,244 @@
...
+            final.get("group_ids", [])
+        ):
+            self._add_message(
+                f"Group(s) {', '.join([str(id_) for id_ in forbidden_groups])} have user.can_manage permissions and may therefore not be used by users who are locked out",
+                result,
+                raise_exception,
+                list(forbidden_groups),
+            )
+
+    def _check_setting_locked_for_oml_cml(
```
hjanott, line 148: Document different functions of these check methods.

openslides_backend/action/actions/meeting_user/mixin.py
```diff
@@ -9,9 +17,244 @@
...
+        ):
+            return
+        filters = [
+            FilterOperator("user_id", "=", user_id),
+            FilterOperator("locked_out", "=", True),
+        ]
+        if (new_locked := instance.get("locked_out")) is False and meeting_id:
+            filters.append(FilterOperator("meeting_id", "!=", meeting_id))
+        filter_: Filter = And(filters)
+        if new_locked and meeting_id:
```
hjanott, line 191: Think whether there is a better variable name.

hjanott (approved):

PR #2551, opened 2024-08-05
Implement anonymous group

### LINKED ISSUE #2544: Anonymous user
see for context: https://github.com/OpenSlides/OpenSlides/issues/6798

**Wanted behaviour**
We want to enable anonymous users for meetings

- if you are using OS as an anonymous then the backend should throw 400 responses

- A new setting inside of a meeting is needed (e.g. enable_anonymous_user)
- If that setting is activated for the first time then a new group, which is not deletable, should be created, it's name should be be "Anonymous"
- This group should then stay existing even if the setting is deactivated
- Another setting ```locked_from_inside``` and this new setting are not allowed to be true at the same time, the backend should thrown an Error if someone tries to activate them at the same time (sth along the lines of  ```Locking a meeting from the inside and activating anonymous user is not allowed at the same time, please deactivate one setting```)

luisa-beerboom:
Closes #2544 

- [x] Change wiki
- [x] Stop people from adding users to the anonymous group via `user.create`, `user.update`, `user.assign_meetings`, etc.
- [x] Also in participant import
- [x] Use anonymous group instead of default for anonymous perm checks
   - [x] Test that
- [x] Ignore non-read perms or stop them from being set in anonymous group?
- [x] Test
- [x] Forbid write permissions on polls, chat_groups, comment_sections, and anything related to polls


docs/actions/poll.create.md
```diff
@@ -62,6 +62,8 @@ If the `type` is `pseudoanonymous`, `is_pseudoanonymized` has to be set to `true
 
 If the `content_object_id` points to a `motion` and the `motion_state` of the motion misses `allow_create_poll`, it is forbidden to create a poll.
 
+The `entitled_user_ids` may not contain the meetings `anonymous_group_id`.
```
hjanott, line 65: I think you meant `entitled_group_ids`.

docs/actions/poll.update.md
```diff
@@ -37,6 +37,8 @@ For analog polls: If the state is created and at least one vote value is given (
 
 For electronic polls some fields can only be updated, if the state is *created*.
 
+The `entitled_user_ids` may not contain the meetings `anonymous_group_id`.
```
hjanott, line 40: I think you meant `entitled_group_ids`.

tests/system/action/meeting/test_update.py
```diff
@@ -877,3 +882,138 @@ def test_update_cant_lock_template_4(self) -> None:
...
+                    "enable_anonymous": True,
+                    "locked_from_inside": True,
+                    "admin_group_id": 2,
+                }
+            }
+        )
+        self.set_user_groups(1, [2])
+        response = self.request(
+            "meeting.update",
+            {"id": 1, "location": "Geneva"},
```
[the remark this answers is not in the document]
luisa-beerboom: You're right, this state cannot exist, I'll remove the test

tests/system/action/meeting_user/test_update.py
```diff
@@ -75,6 +75,28 @@ def test_update_merge_fields_correct(self) -> None:
...
+                "meeting/1": {"group_ids": [1, 2, 3, 4]},
+                "group/4": {"anonymous_group_for_meeting_id": 1},
+            }
+        )
+        self.create_user("dummy", [1])
+        response = self.request(
+            "meeting_user.update",
+            {
+                "id": 1,
+                "group_ids": [4],
```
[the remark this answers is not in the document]
luisa-beerboom: It's the same mixin controlling the functionality, I tested the `[1, 4]` case in set_data, here I'm testing if the mixin functionality is there, I don't think we need to add unnecessary bloat by re-testing the same mundane mixin functionality in multiple test classes

docs/actions/poll.create.md
luisa-beerboom: Done

docs/actions/poll.update.md
luisa-beerboom: Done

hjanott (approved):

bastianjoel:
Would it be possible to enable `meeting.can_see_frontpage` by default for the anonymous group?

PR #2576, opened 2024-08-15
Allow meeting admin user to update a non admin user that shares all his meetings with requesting user.

### LINKED ISSUE #2522: Change user.scope for Meeting admins if they are admins in several meetings
**Current behaviour:**
If a user (Ann) is a meeting admin (more precise when Ann is allowed to manage participants) in two meetings and another user (Ben) is only in these two meetings then Ann is currently not allowed to change the personal information from Ben with the backend reply:
```
Error: You are not allowed to perform action user.update. Missing permissions: OrganizationManagementLevel can_manage_users in organization # or CommitteeManagementLevel can_manage in committee #
```

**Wanted behaviour:**
As long as Ann is allowed to manage participants in every meeting Ben is a participant in then Ann is allowed to change Ben's personal information

hjanott:
Closes #2522

### DIFF
--- a/openslides_backend/action/actions/user/create_update_permissions_mixin.py
+++ b/openslides_backend/action/actions/user/create_update_permissions_mixin.py
@@ -213,7 +213,9 @@ class CreateUpdatePermissionsMixin(UserMixin, UserScopeMixin, Action):
     def check_permissions(self, instance: dict[str, Any]) -> None:
         """
         Checks the permissions on a per field and user.scope base, details see
-        https://github.com/OpenSlides/OpenSlides/wiki/user.update or user.create
+        https://github.com/OpenSlides/OpenSlides/wiki/Users
+        https://github.com/OpenSlides/OpenSlides/wiki/Permission-System
+        https://github.com/OpenSlides/OpenSlides/wiki/Restrictions-Overview
         The fields groups and their necessary permissions are also documented there.
         """
         self.assert_not_anonymous()
@@ -227,13 +229,14 @@ class CreateUpdatePermissionsMixin(UserMixin, UserScopeMixin, Action):
             )
         actual_group_fields = self._get_actual_grouping_from_instance(instance)
 
-        # store scope, id and OML-permission for requested user
+        # store id, scope, scope id and OML-permission for requested user
+        self.instance_id = instance.get("id", 0)
         (
             self.instance_user_scope,
             self.instance_user_scope_id,
             self.instance_user_oml_permission,
             self.instance_committee_ids,
-        ) = self.get_user_scope(instance.get("id") or instance)
+        ) = self.get_user_scope(self.instance_id or instance)
 
         if self.permstore.user_oml != OrganizationManagementLevel.SUPERADMIN:
             self._check_for_higher_OML(actual_group_fields, instance)
@@ -247,7 +250,7 @@ class CreateUpdatePermissionsMixin(UserMixin, UserScopeMixin, Action):
                 lock_result=False,
             ).get("locked_from_inside", False)
 
-        # Ordered by supposed velocity advantages. Changing order can only effect the sequence of detected errors for tests
+        # Ordered by supposed speed advantages. Changing order can only effect the sequence of detected errors for tests
         self.check_group_H(actual_group_fields["H"])
         self.check_group_E(actual_group_fields["E"], instance)
         self.check_group_D(actual_group_fields["D"], instance)
@@ -256,6 +259,70 @@ class CreateUpdatePermissionsMixin(UserMixin, UserScopeMixin, Action):
         self.check_group_A(actual_group_fields["A"])
         self.check_group_F(actual_group_fields["F"])
         self.check_group_G(actual_group_fields["G"])
+
+    def _meeting_admin_can_manage_non_admin(self) -> bool:
+        """
+        Checks if the requesting user has permissions to manage participants in all of requested users meetings.
+        Also checks if the requesting user has meeting admin rights and the requested user doesn't.
+        Returns true if permissions are given. False if not. Raises no Exceptions.
+        """
+        a_meeting_ids = self.permstore.user_meetings
+        b_meeting_ids = set(
+            self.datastore.get(
+                fqid_from_collection_and_id("user", self.instance_id),
+                ["meeting_ids"],
+                lock_result=False,
+            ).get("meeting_ids", [])
+        )
+        intersection_meeting_ids = a_meeting_ids.intersection(b_meeting_ids)
+        if not b_meeting_ids.issubset(intersection_meeting_ids):
+            return False
+        intersection_meetings = self.datastore.get_many(
+            [
+                GetManyRequest(
+                    "meeting",
+                    list(intersection_meeting_ids),
+                    ["meeting_user_ids", "admin_group_id"],
+                )
+            ],
+            lock_result=False,
+        ).get("meeting", {})
+        for meeting_id, meeting_dict in intersection_meetings.items():
+            # get meetings admins
+            admin_group = self.datastore.get(
+                fqid_from_collection_and_id(
+                    "group", meeting_dict.get("admin_group_id", 0)
+                ),
+                ["meeting_user_ids"],
+                lock_result=False,
+            )
+            admin_meeting_users = self.datastore.get_many(
+                [
+                    GetManyRequest(
+                        "meeting_user",
+                        admin_group.get("meeting_user_ids", []),
+                        ["user_id"],
+                    )
+                ],
+                lock_result=False,
+            ).get("meeting_user", {})
+            # if instance/requested user is a meeting admin in this meeting.
+            if [
+                admin_meeting_user
+                for admin_meeting_user in admin_meeting_users.values()
+                if admin_meeting_user.get("user_id") == self.instance_id
+            ] != []:
+                return False
+            # if requesting user is not a meeting admin in this meeting.
+            if not next(
+                iter(
+                    admin_meeting_user
+                    for admin_meeting_user in admin_meeting_users.values()
+                    if admin_meeting_user.get("user_id") == self.user_id
+                )
+            ):
+                return False
+        return True
 
     def check_group_A(
         self,
@@ -272,7 +339,10 @@ class CreateUpdatePermissionsMixin(UserMixin, UserScopeMixin, Action):
         if self.instance_user_scope == UserScope.Organization:
             if self.permstore.user_committees.intersection(self.instance_committee_ids):
                 return
-            raise MissingPermission({OrganizationManagementLevel.CAN_MANAGE_USERS: 1})
+            elif not self._meeting_admin_can_manage_non_admin():
+                raise MissingPermission(
+                    {OrganizationManagementLevel.CAN_MANAGE_USERS: 1}
+                )
         if self.instance_user_scope == UserScope.Committee:
             if self.instance_user_scope_id not in self.permstore.user_committees:
                 raise MissingPermission(
--- a/tests/system/action/user/test_update.py
+++ b/tests/system/action/user/test_update.py
@@ -860,6 +860,54 @@ class UserUpdateActionTest(BaseActionTestCase):
             },
         )
 
+    def test_perm_group_A_belongs_to_same_meetings(self) -> None:
+        """May update group A fields on any scope as long as admin user Ann belongs to all meetings user Ben belongs to. See issue 2522."""
+        self.permission_setup()  # meeting 1 + logged in test user + user 111
+        self.create_meeting(4)  # meeting 4
+        # Admin groups of meeting/1 and meeting/4 for test user
+        self.set_user_groups(self.user_id, [2, 5])
+        self.set_user_groups(111, [1, 4])  # 111 into both meetings
+        response = self.request(
+            "user.update",
+            {
+                "id": 111,
+                "pronoun": "I'm gonna get updated.",
+            },
+        )
+        self.assert_status_code(response, 200)
+        self.assert_model_exists(
+            "user/111",
+            {
+                "pronoun": "I'm gonna get updated.",
+            },
+        )
+
+    def test_perm_group_A_belongs_to_same_meetings_both_admin(self) -> None:
+        """May update group A fields on any scope as long as admin user Ann belongs to all meetings user Ben belongs to. See issue 2522."""
+        self.permission_setup()  # meeting 1 + logged in test user + user 111
+        self.create_meeting(4)  # meeting 4
+        # Admin groups of meeting/1 and meeting/4 for test user
+        self.set_user_groups(self.user_id, [2, 5])
+        self.set_user_groups(111, [1, 5])  # 111 into both meetings one admin group
+        response = self.request(
+            "user.update",
+            {
+                "id": 111,
+                "pronoun": "I'm gonna get updated.",
+            },
+        )
+        self.assert_status_code(response, 403)
+        self.assertIn(
+            "You are not allowed to perform action user.update. Missing permission: OrganizationManagementLevel can_manage_users in organization 1",
+            response.json["message"],
+        )
+        self.assert_model_exists(
+            "user/111",
+            {
+                "pronoun": None,
+            },
+        )
+
     def test_perm_group_A_meeting_manage_user_archived_meeting(self) -> None:
         self.perm_group_A_meeting_manage_user_archived_meeting(
             Permissions.User.CAN_UPDATE


Elblinator (commented):
A meeting admin should not be allowed to change the information of a committee admin. The following Error should be thrown
```
Error: You are not allowed to perform action user.update. Missing permission: OrganizationManagementLevel can_manage_users in organization 1
```

openslides_backend/action/actions/user/create_update_permissions_mixin.py
```diff
@@ -257,6 +262,75 @@ def check_permissions(self, instance: dict[str, Any]) -> None:
...
+    def _meeting_admin_can_manage_non_admin(self) -> bool:
+        """
+        Checks if the requesting user has permissions to manage participants in all of requested users meetings.
+        Also checks if the requesting user has meeting admin rights and the requested user doesn't.
+        Returns true if permissions are given. False if not. Raises no Exceptions.
+        """
+        if not self.instance_id:
+            return False
+        b_user = self.datastore.get(
+            fqid_from_collection_and_id("user", self.instance_id),
```
luisa-beerboom, line 274: I don't think `instance_id` needs to be an instance attribute, just pass it on as a parameter

openslides_backend/action/actions/user/create_update_permissions_mixin.py
```diff
@@ -247,7 +250,9 @@ def check_permissions(self, instance: dict[str, Any]) -> None:
                 lock_result=False,
             ).get("locked_from_inside", False)
 
-        # Ordered by supposed velocity advantages. Changing order can only effect the sequence of detected errors for tests
+        # Ordered by supposed speed advantages. Changing order can only effect the sequence of detected errors for tests
+        if self._meeting_admin_can_manage_non_admin():
```
luisa-beerboom, line 254: This should not skip the field group checks, but rather be part of them.
You should look into the action description to see what field groups need to be changed and then only change those.
You should especially not be skipping the checks for that have nothing to do with what our issue management people usually call `personal information`, f.E. groups `E`, `G`, and `H`.

Also think of maybe moving this functionality to the `UserScopeMixin` and using it for multiple classes.

luisa-beerboom (changes requested):
From the issue name I read that this is supposed to be a scope change. Given that, I would expect at least the `check_permissions_for_scope` function of `UserScopeMixin` to be changed, as well as for all actions using said mixin to have at least been looked at for the sake of considering potential necessary changes. `Scope` is after all something that is more generally used for user actions concerning "personal information".

Keeping with that theme, I'd also expect the password actions to probably have been changed as well and maybe user.delete (should probably discuss whether this is wanted with @MSoeb or @Elblinator )

I'd expect [this wiki entry](https://github.com/OpenSlides/OpenSlides/wiki/Users#permissions-for-altering-a-user) to have been changed accordingly, as well as maybe the related actions backend documentation files.
I can't really judge if the new functionality has been implemented, unless the requirements are documented in the wiki, after all.

There should be tests for the other user actions as well (maybe even one in create, just to see if the functionality actually stayed the same if current tests don't cover that).
There ought to be tests for the user imports as well: I.e. check if the participant import also newly permits what was previously forbidden and if the account import still can't be used by those who don't have organization user management rights, even if they have meeting perms for all of a users meetings.

openslides_backend/shared/mixins/user_scope_mixin.py
```diff
@@ -163,4 +163,90 @@ def check_permissions_for_scope(
                 committees,
             ):
                 return
-            raise MissingPermission({OrganizationManagementLevel.CAN_MANAGE_USERS: 1})
+            if not self.check_for_admin_in_all_meetings(instance_id):
+                raise MissingPermission(
+                    {OrganizationManagementLevel.CAN_MANAGE_USERS: 1}
+                )
```
luisa-beerboom, line 169: It is good that you have worked it into the scope, but if the way the scope checking works has been changed in the backend, the other services (mainly the client) need to be able to get the information required to calculate the same result via the scope presenter. This means meeting memberships and committee_management_ids need to somehow be part of what the scope presenter returns.

Look into how that presenters return value would optimally look after the new changes, change the wiki entry accordingly and change the get_user_scope_function in this presenter.
It is then only appropriate that the `check_for_admin_in_all_meetings` function should not calculate everything and get all data from scratch, but use the type of data calculated by `get_user_scope` in the previous part of the calling function.

luisa-beerboom (changes requested):

hjanott:
Discussed with @Elblinator that it should be sufficient for the client to only have a presenter that returns true if a user can be edited. An issue for the client needs to be created anyway.
Relevant group permissions are `user.can_update` and `user.can_manage`.

Elblinator:
The client counterpart is the following Issue: 
https://github.com/OpenSlides/openslides-client/issues/3841

openslides_backend/action/actions/user/create.py
```diff
@@ -67,6 +70,10 @@ class UserCreate(
     history_information = "Account created"
     own_history_information_first = True
 
+    def check_permissions(self, instance: dict[str, Any]) -> None:
+        self.assert_not_anonymous()
```
luisa-beerboom, line 74: Ensure there's a test that checks this line of code.
Then delete this line and see if it still works. Because AFAIK anonymous should be auto-locked out of action requests by now.

openslides_backend/action/actions/user/update.py
```diff
@@ -76,6 +79,10 @@ class UserUpdate(
     permission = Permissions.User.CAN_UPDATE
     check_email_field = "email"
 
+    def check_permissions(self, instance: dict[str, Any]) -> None:
+        self.assert_not_anonymous()
```
luisa-beerboom, line 83: See my comment in user/create

openslides_backend/shared/mixins/user_create_update_permissions_mixin.py
```diff
@@ -203,20 +206,24 @@ class CreateUpdatePermissionsMixin(UserMixin, UserScopeMixin, Action):
...
-        "D": ["committee_ids", "committee_management_ids"],
+        "D": ["committee_management_ids"],
         "E": ["organization_management_level"],
         "F": ["default_password"],
         "G": ["is_demo_user"],
         "H": ["saml_id"],
     }
 
-    def check_permissions(self, instance: dict[str, Any]) -> None:
+    def check_permissions(self, instance: dict[str, Any]) -> bool | None:
```
luisa-beerboom, line 216: This function always returns `True` or Error, never `False`.
Having it be a boolean return type is therefore unnecessary and only removes it from the action check_permissions functions which it is supposed to implement.
Remove the return type `bool`.

openslides_backend/presenter/get_user_editable.py
```diff
@@ -0,0 +1,65 @@
...
+            )
+        instance = {field_name: "" for field_name in self.data["fields"]}
+        for user_id in self.data["user_ids"]:
+            instance.update({"id": user_id})
+            result[str(user_id)] = {}
+            try:
+                if self.check_permissions(instance):
+                    result[str(user_id)]["editable"] = True
+                else:
+                    result[str(user_id)]["editable"] = False
```
luisa-beerboom, line 61:
```suggestion
                self.check_permissions(instance)
                result[str(user_id)]["editable"] = True
```

The way you've implemented it, this check_permissions function always either returns true or raises an error so:
Either the permissions are all there in which case `editable` must be `True`, or it enters the except branch anyway.

openslides_backend/shared/mixins/user_create_update_permissions_mixin.py
```diff
@@ -203,20 +206,24 @@ class CreateUpdatePermissionsMixin(UserMixin, UserScopeMixin, Action):
...
         Checks the permissions on a per field and user.scope base, details see
-        https://github.com/OpenSlides/OpenSlides/wiki/user.update or user.create
+        https://github.com/OpenSlides/OpenSlides/wiki/Users
+        https://github.com/OpenSlides/OpenSlides/wiki/Permission-System
+        https://github.com/OpenSlides/OpenSlides/wiki/Restrictions-Overview
         The fields groups and their necessary permissions are also documented there.
+        Returns true if permissions are given.
         """
-        self.assert_not_anonymous()
+        if "meta_position" in instance:  # TODO still needed with all the changes?
```
luisa-beerboom, line 225: Look into it then

openslides_backend/shared/mixins/user_create_update_permissions_mixin.py
```diff
@@ -247,34 +254,68 @@ def check_permissions(self, instance: dict[str, Any]) -> None:
...
+                    ) and has_committee_management_level(
+                        self.datastore,
+                        instance_id,
+                        CommitteeManagementLevel.CAN_MANAGE,
+                        committee_id,
+                    ):
+                        raise MissingPermission(
+                            {OrganizationManagementLevel.CAN_MANAGE_USERS: 1}
+                        )
                 return
```
luisa-beerboom, line 292: This return statement does nothing

openslides_backend/shared/mixins/user_create_update_permissions_mixin.py
```diff
@@ -247,34 +254,68 @@ def check_permissions(self, instance: dict[str, Any]) -> None:
...
+                        CommitteeManagementLevel.CAN_MANAGE,
+                        committee_id,
+                    ):
+                        raise MissingPermission(
+                            {
+                                OrganizationManagementLevel.CAN_MANAGE_USERS: 1,
+                                CommitteeManagementLevel.CAN_MANAGE: self.instance_user_scope_id,
+                            }
+                        )
+                return
```
luisa-beerboom, line 317: This return statement does nothing

openslides_backend/shared/mixins/user_create_update_permissions_mixin.py
```diff
@@ -307,7 +348,7 @@ def check_group_B(
             or locked_from_inside
         ) and fields:
             meeting_ids = self._meetings_from_group_B_fields_from_instance(instance)
-            if diff := meeting_ids - self.permstore.user_meetings:
+            if (diff := meeting_ids - self.permstore.user_meetings) and diff != {None}:
```
luisa-beerboom, line 351: Why is it possible for None to be in the set? Seems broken to me

openslides_backend/shared/mixins/user_scope_mixin.py
```diff
@@ -160,7 +175,125 @@ def check_permissions_for_scope(
...
+                    list(intersection_meeting_ids),
+                    ["admin_group_id", "group_ids", "meeting_user_ids"],
+                )
+            ],
+            lock_result=False,
+        ).get("meeting", {})
+        for meeting_id, meeting_dict in intersection_meetings.items():
+            # get meetings admins
+            admin_meeting_users = {}
+            # unnecessary "if" due to admin group always existant?
```
luisa-beerboom, line 240: What about this comment

openslides_backend/shared/mixins/user_scope_mixin.py
```diff
@@ -160,7 +175,125 @@ def check_permissions_for_scope(
...
+                    [
+                        GetManyRequest(
+                            "meeting_user",
+                            admin_group.get("meeting_user_ids", []),
+                            ["user_id"],
+                        )
+                    ],
+                    lock_result=False,
+                ).get("meeting_user", {})
+            # unnecessary "if" due to default group always existant?
```
luisa-beerboom, line 257: see above

tests/system/presenter/base.py
```diff
@@ -26,3 +27,82 @@ def request(
...
+        # meet_user_ids = self.datastore.get(f"user/{user1}", ["meeting_user_ids"]).get("meeting_user_ids")
+        # self.update_model(
+        #     f"user/{user1}",
+        #     {"meeting_user_ids": meet_user_ids + int(f"{base}{user1}")}
+        # )
+        # meet_user_ids = self.datastore.get(f"user/{user2}", ["meeting_user_ids"]).get("meeting_user_ids")
+        # self.update_model(
+        #     f"user/{user2}",
+        #     {"meeting_user_ids": meet_user_ids + int(f"{base}{user2}")}
+        # )
```
luisa-beerboom, line 87: delete

luisa-beerboom (changes requested):

openslides_backend/shared/mixins/user_scope_mixin.py
hjanott: Done and decided.

openslides_backend/shared/mixins/user_create_update_permissions_mixin.py
hjanott: Done

openslides_backend/shared/mixins/user_scope_mixin.py
hjanott: Done and decided.

openslides_backend/presenter/get_user_editable.py
hjanott: Done

openslides_backend/action/actions/user/create_update_permissions_mixin.py
```diff
@@ -272,15 +271,19 @@ def check_group_A(
...
                 )
+        if self.instance_user_scope == UserScope.Committee:
+            if self.instance_user_scope_id not in self.permstore.user_committees:
+                if not self.check_for_admin_in_all_meetings(instance.get("id", 0)):
+                    raise MissingPermission(
+                        {
+                            OrganizationManagementLevel.CAN_MANAGE_USERS: 1,
+                            CommitteeManagementLevel.CAN_MANAGE: self.instance_user_scope_id,
+                        }
+                    )
```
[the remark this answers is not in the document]
hjanott: Done

openslides_backend/action/actions/user/create_update_permissions_mixin.py
```diff
@@ -272,15 +271,19 @@ def check_group_A(
...
-        if self.instance_user_scope == UserScope.Committee:
-            if self.instance_user_scope_id not in self.permstore.user_committees:
+            elif not self.check_for_admin_in_all_meetings(instance.get("id", 0)):
                 raise MissingPermission(
-                    {
-                        OrganizationManagementLevel.CAN_MANAGE_USERS: 1,
-                        CommitteeManagementLevel.CAN_MANAGE: self.instance_user_scope_id,
-                    }
+                    {OrganizationManagementLevel.CAN_MANAGE_USERS: 1}
                 )
```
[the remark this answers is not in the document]
hjanott: Done

openslides_backend/action/actions/user/create_update_permissions_mixin.py
hjanott: Done

openslides_backend/shared/mixins/user_scope_mixin.py
```diff
@@ -160,7 +176,137 @@ def check_permissions_for_scope(
...
+            a_meeting_ids = (
+                self.permstore.user_meetings
+            )  # returns only admin level meetings
+        else:
+            a_user = self.datastore.get(
+                fqid_from_collection_and_id("user", instance_id),
+                ["meeting_ids"],
+                lock_result=False,
+            )
+            a_meeting_ids = set(a_user.get("meeting_ids", []))
```
luisa-beerboom, line 239: Same thing here, if `a_user` is only called here, just do the datastore call in the `a_meeting_ids` calculation.
Alternatively consider if it wouldn't be easier to do one datastore call at the beginning of the method.

openslides_backend/shared/mixins/user_scope_mixin.py
```diff
@@ -160,7 +176,137 @@ def check_permissions_for_scope(
...
+        if not a_meeting_ids:
+            return False
+        intersection_meeting_ids = a_meeting_ids.intersection(b_meeting_ids)
+        if not b_meeting_ids.issubset(intersection_meeting_ids):
+            return False
```
luisa-beerboom, line 244: You can also do this:
```suggestion
        if not a_meeting_ids or not b_meeting_ids.issubset(intersection_meeting_ids := a_meeting_ids.intersection(b_meeting_ids)):
            return False
```

openslides_backend/shared/mixins/user_scope_mixin.py
```diff
@@ -160,7 +176,137 @@ def check_permissions_for_scope(
...
+            admin_meeting_users = self.datastore.get_many(
+                [
+                    GetManyRequest(
+                        "meeting_user",
+                        admin_group.get("meeting_user_ids", []),
+                        ["user_id"],
+                    )
+                ],
+                lock_result=False,
+            ).get("meeting_user", {})
```
luisa-beerboom, line 272: May be a good idea to aggregate admin_group_ids across meetings, then load them all with _one_ `get_many` request outside of a loop (again: list comprehensions).
Bonus points for also loading the admin_meeting_users in bulk.

There is also the option of getting just the user with `mapped_fields=["meeting_user_ids"]` and looking for intersections with the `meeting_user_ids` in the admin groups.

openslides_backend/shared/mixins/user_scope_mixin.py
```diff
@@ -160,7 +176,137 @@ def check_permissions_for_scope(
...
+                    "meeting",
+                    list(intersection_meeting_ids),
+                    ["admin_group_id", "group_ids", "meeting_user_ids"],
+                )
+            ],
+            lock_result=False,
+        ).get("meeting", {})
+        for meeting_id, meeting_dict in intersection_meetings.items():
+            admin_meeting_users = {}
+            admin_group_id = meeting_dict.get("admin_group_id", 0)
```
luisa-beerboom, line 257: Every meeting has an `admin_group_id`. It is a required field. If the tests will fail with `meeting_dict["admin_group_id"]` that's a sign that the tests are wrong

openslides_backend/action/actions/user/create.py
```diff
@@ -15,13 +18,13 @@
...
 from ...util.typing import ActionResultElement
 from ..meeting_user.mixin import CheckLockOutPermissionMixin
-from .create_update_permissions_mixin import CreateUpdatePermissionsMixin
 from .password_mixins import SetPasswordMixin
 from .user_mixins import LimitOfUserMixin, UserMixin, UsernameMixin, check_gender_helper
 
 
 @register_action("user.create")
 class UserCreate(
+    UserMixin,
```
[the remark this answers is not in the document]
luisa-beerboom: Then try just `**transfer_field_list`

openslides_backend/shared/mixins/user_scope_mixin.py
hjanott: Yes that would be possible but it seems a little too convoluted to me. In my opinion my version is more readable. However, it has got more lines.

openslides_backend/action/actions/user/create.py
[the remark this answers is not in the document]
hjanott: I tried but it seems that a class attribute from another package is only visible by giving its class name.
Which makes sense because the `UserMixin` was also part of the inheritance hierarchy before moving it directly here.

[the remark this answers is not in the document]
hjanott: I also tried moving the attribute initialization to the __init__ method, but that would defeat the purpose of class attributes and leads to many other problems.

openslides_backend/shared/mixins/user_scope_mixin.py
```diff
@@ -189,15 +190,32 @@ def check_permissions_for_scope(
...
     def check_for_admin_in_all_meetings(
-        self, instance_id: int, b_meeting_ids: set[int] | None = None
+        self,
+        instance_or_id: int | dict[str, Any],
+        b_meeting_ids: set[int] | None = None,
     ) -> bool:
         """
-        Checks if the requesting user has permissions to manage participants in all of requested users meetings.
-        Also checks if the requesting user has meeting admin rights and the requested user doesn't.
+        Checks if the requesting user has permissions to manage participants in all of requested users meetings but requested user doesn't.
```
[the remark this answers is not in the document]
hjanott: I improved the method description but I could not think of a better name without becoming excessively verbose.
The helper functions are now defined and documented as protected functions below.

openslides_backend/shared/mixins/user_scope_mixin.py
```diff
@@ -189,15 +190,32 @@ def check_permissions_for_scope(
                 )
 
     def check_for_admin_in_all_meetings(
-        self, instance_id: int, b_meeting_ids: set[int] | None = None
+        self,
+        instance_or_id: int | dict[str, Any],
+        b_meeting_ids: set[int] | None = None,
     ) -> bool:
```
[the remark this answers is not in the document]
hjanott: I split it up.

openslides_backend/shared/mixins/user_scope_mixin.py
```diff
@@ -160,7 +176,137 @@ def check_permissions_for_scope(
...
+                    return False
+        else:
+            instance_id = instance_or_id
+        if not b_meeting_ids:
+            if not hasattr(self, "instance_committee_meeting_ids"):
+                return False
+            b_meeting_ids = set()
+            for m_ids in self.instance_committee_meeting_ids.values():
+                if m_ids:
+                    b_meeting_ids.update(m_ids)
```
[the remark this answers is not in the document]
hjanott: Done

openslides_backend/shared/mixins/user_scope_mixin.py
hjanott: Done and decided against doing one datastore call for two users.

openslides_backend/shared/mixins/user_scope_mixin.py
hjanott: Done
However I did not look deeper into the latter suggestion. I think, that would mean two n:m comparisons per meeting whereas now it is two 1:n comparisons per meeting.

openslides_backend/shared/mixins/user_scope_mixin.py
hjanott: Did so and updated the tests.

openslides_backend/action/actions/user/create.py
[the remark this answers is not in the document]
luisa-beerboom: Alright, then leave it

openslides_backend/shared/mixins/user_scope_mixin.py
```diff
@@ -179,134 +179,193 @@ def check_permissions_for_scope(
...
+                            "group", group_ids, ["meeting_user_ids", "permissions"]
+                        )
+                    ],
+                    lock_result=False,
+                )
+                .get("group", {})
+                .items()
+                if (
+                    "user.can_update" in group.get("permissions", [])
+                    or "user.can_manage" in group.get("permissions", [])
```
luisa-beerboom, line 317: Additionally check if "admin_group_for_meeting_id" is set and you can skip the other comprehension and the variable `admin_group_ids`

openslides_backend/shared/mixins/user_scope_mixin.py
```diff
@@ -179,134 +179,193 @@ def check_permissions_for_scope(
...
+                .items()
+                for mu_id in group.get("meeting_user_ids", [])
+            },
+        }
+
+    def _collect_admin_users(self, meeting_user_ids: set[int]) -> dict[int, set[int]]:
+        """
+        Helper function used in method check_for_admin_in_all_meetings.
+        Returns the corresponding users of the groups meeting_users in a defaultdict meeting_id: user_ids.
+        """
```
luisa-beerboom, line 337: Internally, this method doesn't care at all about admin status, the name is misleading.

luisa-beerboom: `collect_users` or `get_meeting_to_user_id_dict` would be a better name

openslides_backend/shared/mixins/user_scope_mixin.py
```diff
@@ -160,7 +176,196 @@ def check_permissions_for_scope(
...
+            return False
+
+        if not (
+            intersection_meetings := self._collect_intersected_meetings(b_meeting_ids)
+        ):
+            return False
+        assert isinstance(intersection_meetings, dict)
+        admin_meeting_users = self._collect_admin_meeting_users(intersection_meetings)
+        meeting_to_admin_users = self._collect_admin_users(admin_meeting_users)
+        return self._analyze_meetings(meeting_to_admin_users, instance_id)
```
luisa-beerboom, line 224:
```suggestion
        return self._analyze_meetings(meeting_to_admin_users, instance_id)
```
`_analyze_meetings` has very admin-related internal variable names, but no actual admin-specific logic. Moving the calculation of `meeting_to_admin_users` into `_analyze_meetings` would be more sensible.

openslides_backend/shared/mixins/user_scope_mixin.py
```diff
@@ -179,134 +179,193 @@ def check_permissions_for_scope(
...
+        Helper function used in method check_for_admin_in_all_meetings.
+        Compares the admin users of all meetings with the ids of requested user and requesting user.
+        Requesting user must be admin in all meetings. Requested user cannot be admin in any.
+        """
+        for meeting_id, admin_users in meeting_to_admin_users.items():
+            if instance_id in admin_users:
+                return False
+            if self.user_id not in admin_users:
                 return False
         return True
```
luisa-beerboom, line 371:
```suggestion
        return not any(
           instance_id in admin_users or self.user_id not in admin_users
           for meeting_id, admin_users in meeting_to_admin_users.items()
        )
```

openslides_backend/shared/mixins/user_scope_mixin.py
```diff
@@ -179,134 +179,193 @@ def check_permissions_for_scope(
...
+        return meeting_to_admin_user
+
+    def _analyze_meetings(
+        self, meeting_to_admin_users: dict[int, set[int]], instance_id: int
+    ) -> bool:
+        """
+        Helper function used in method check_for_admin_in_all_meetings.
+        Compares the admin users of all meetings with the ids of requested user and requesting user.
+        Requesting user must be admin in all meetings. Requested user cannot be admin in any.
+        """
```
luisa-beerboom, line 365: This method contains a very admin-related description, but actually the admin-context only exists because you assume that the data delivered will be admin-related, it would be better to move the calculation of this admin-related data into this function to make it more topical or rename the function to be more generic.

openslides_backend/shared/mixins/user_scope_mixin.py
```diff
@@ -179,134 +179,193 @@ def check_permissions_for_scope(
...
         if not b_meeting_ids:
-            if not hasattr(self, "instance_committee_meeting_ids"):
+            if not (
+                b_meeting_ids := {
+                    m_id
+                    for m_ids in self.instance_committee_meeting_ids.values()
+                    for m_id in m_ids
+                }
+            ):
                 return False
```
luisa-beerboom, line 256: Can this be shortened, there are two if clauses nested inside each other with no further content on the outer one

luisa-beerboom (changes requested):

openslides_backend/shared/mixins/user_scope_mixin.py
hjanott: Yes it can be handled with 'and'. The outer one is there for performance reasons. There is no need to calculate b_meeting_ids if they are given. If the calculation also delivers nothing, False should be returned.

openslides_backend/shared/mixins/user_create_update_permissions_mixin.py
```diff
@@ -382,12 +401,26 @@ def check_group_F(
...
+            ):
+                raise MissingPermission(
+                    {
+                        expected_oml_permission: 1,
+                        **{
+                            Permissions.User.CAN_UPDATE: meeting_id
+                            for meeting_ids in self.instance_committee_meeting_ids.values()
+                            if meeting_ids is not None
+                            for meeting_id in meeting_ids
+                            if meeting_id is not None
```
luisa-beerboom, line 420: This is just going to return the last meeting that was checked where the permission is missing.
Always using the same keyword means it will be overwritten every time, after all.
Also: Missing permission concatenates the permissions with 'or' and never 'and'.
If that is what you wanted, it's fine.
If it is not, you should probably looking into using something other that `MissingPermission` and generating your own response message. I've seen the `MissingPermission` base class used like that before.

luisa-beerboom (changes requested):

luisa-beerboom (approved):

Elblinator:
I tested the PR with different scenarios and sometimes the backend returns Errors even though the action should be permitted. 
What I did: 
- Create three meetings 
- Add A always as admin to all meetings
- Add B groups do not seem to matter, I could not find a connection at least
- Try changing B's personal information with A 
sometimes, and only sometimes, while trying to change the personal information B's personal information is not changed but an Error that A is missing the user.canUpdate permissions for he meetings {1,2,3}

Elblinator (approved):
I found no Errors
My previous Errors probably stem from a faulty setup

Elblinator (changes requested):
Situation: 
The "Admin" in question is in meeting A user.can_update permissions and in the other meeting user.can_manage permissions 
And the presenter throws the following answer if the admin tries to reach the presenter 
```
[{"4": {"first_name": [false, "Missing permissions: OrganizationManagementLevel can_manage_users in organization 1 or CommitteeManagementLevel can_manage in committee 1 or Permission user.can_update in meetings {1, 2}"], "default_password": [false, "Missing permissions: OrganizationManagementLevel can_manage_users in organization 1 or CommitteeManagementLevel can_manage in committee 1"]}}]
```

And the default password should be changeable if you have user.can_update rights. 
You currently need OrganizationManagementLevel or CommitteeManagementLevel

As far as I could see the payload and the presenter both block this right now

luisa-beerboom (approved):

reiterl (commented):
Some small remarks.

openslides_backend/presenter/get_user_editable.py
```diff
@@ -0,0 +1,85 @@
...
+            for group_fields in self.field_rights.values()
+            if field_name in group_fields
+        }
+        result: defaultdict[str, dict[str, tuple[bool, str]]] = defaultdict(dict)
+        for user_id in self.data["user_ids"]:
+            result[str(user_id)] = {}
+            groups_editable = {}
+            for field_name in one_field_per_group:
+                try:
+                    self.check_permissions({"id": user_id, field_name: ""})
```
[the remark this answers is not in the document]
hjanott: Only the keys are used by the 'check_permissions'. So there shouldn't be any problems. But I agree that 'None' is more beautiful.

openslides_backend/presenter/get_user_editable.py
```diff
@@ -0,0 +1,85 @@
...
+
+    schema = get_user_editable_schema
+    name = "get_user_editable"
+    permission = Permissions.User.CAN_MANAGE
+
+    def get_result(self) -> Any:
+        if "fields" not in self.data or not self.data["fields"]:
+            raise PresenterException(
+                "Need at least one field name to check editability."
+            )
```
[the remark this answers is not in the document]
hjanott: That's a good idea. Didn't think of it. So I removed that part of the check, but still checking for the list to have an entry.

Elblinator (changes requested):
Combinations that currently return the wrong value: 
```
{
  "4": {
    "first_name": [
      false,
      "Missing permissions: OrganizationManagementLevel can_manage_users in organization 1 or CommitteeManagementLevel can_manage in committee 1 or Permission user.can_manage in meeting 1"
    ],
    "default_password": [
      false,
      "Missing permissions: OrganizationManagementLevel can_manage_users in organization 1 or CommitteeManagementLevel can_manage in committee 1 or Permission user.can_manage in meeting 1"
    ]
  }
}
```
The Admin has user.can_update permissions in meeting 1 and the user with the id is only in meeting 1



Combinations which return the correct value:
- The Admin has user.can_update permissions in meeting 1 and user.can_manage in meeting 2 and the user with the id 4 is  in meeting 1 and 2. Meeting 1 and 2 belong to the same committee
- The Admin has user.can_update permissions in meeting 1 and user.can_manage in meeting 3 and the user with the id 6 is in meeting 1 and meeting 3. Meeting 1 and 3 are in different committees
- every other from my test cases returns the correct value

luisa-beerboom (approved):

Elblinator (approved):

PR #2590, opened 2024-09-04
Add motion_poll_default_method setting to meeting

### LINKED ISSUE #2578: Add 'Default poll method' for motions
Since https://github.com/OpenSlides/openslides-client/pull/3979 a motion poll can be method YN and YNA

**Wanted behaviour:**
Like the election "Default election method" the motions need a default setting for their poll method.

Please add a new setting with a similar key to "motion_poll_default_method"

Like the assignment_poll_method this new setting should save the default method a motion poll is created with

luisa-beerboom:
Closes #2578 
Needs OpenSlides/openslides-meta#160


hjanott (approved):
Looks good to me.

PR #2591, opened 2024-09-04
Add motion.can_see_origin permission

### LINKED ISSUE #2550: Enhance motion forwarded feature
see for context: https://github.com/OpenSlides/openslides-client/issues/3529

**Wanted:**
1. We need a new setting in the group matrix which allows a user to see the origin of a motion
e.g. can_see_origin
if activated the client should allow a different view and the autoupdate should give back some more information

luisa-beerboom:
Closes #2550 
Needs OpenSlides/openslides-meta#161

For now the permission has no backend-functionality other than being settable. I.e. it is not in the anonymous group permission whitelist and is not explicitly part of any of the pre-set groups for meetings.


hjanott (approved):
Looks good to me.

PR #2614, opened 2024-09-11
Let mediafile managers in meeting manage access to published orgafiles

### LINKED ISSUE #2613: Fix setting groups for published mediafiles
**Current behaviour:** 
If a meeting adminwithout orga management permissions tries to change the groups from a publizised mediafile then the following Error is thrown: 
```
Error: You are not allowed to perform action mediafile.update. Missing OrganizationManagementLevel: can_manage_organization
```

**Reproduction:**
1. in orga level upload file
2. publish file
3. login as a normal meeting admin
4. navigate to files and try to change the groups
5. Error is thrown

**Wanted behaviour:**
Anyone with the right to manage the mediafiles should be able to change the groups from the files (this includes published and meeting specific files)

luisa-beerboom:
Closes #2613

### DIFF
--- a/openslides_backend/action/actions/mediafile/mixins.py
+++ b/openslides_backend/action/actions/mediafile/mixins.py
@@ -21,6 +21,8 @@ class MediafileMixin(Action):
     Overwrite update_instance(), check_permissions() and get_meeting_id().
     """
 
+    meeting_fields: list[str] = ["meeting_id", "access_group_ids"]
+
     def update_instance(self, instance: dict[str, Any]) -> dict[str, Any]:
         collection, id_ = self.get_owner_data(instance)
         self.check_parent_is_dir_and_owner(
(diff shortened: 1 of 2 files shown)


openslides_backend/action/actions/mediafile/mixins.py
```diff
@@ -21,6 +21,9 @@ class MediafileMixin(Action):
     Overwrite update_instance(), check_permissions() and get_meeting_id().
     """
 
+    meeting_fields: list[str] = ["meeting_id", "access_group_ids"]
+    is_delete_action: bool = False
```
hjanott, line 25: Are the type annotations necessary? In the delete action they are omitted.

hjanott (approved):

openslides_backend/action/actions/mediafile/mixins.py
luisa-beerboom: because the delete action expands on the mixin. So the type annotation is already there

Elblinator (approved):

PR #2630, opened 2024-09-19
Fix migration 54

luisa-beerboom:
There was a problem that only occurs if there are projections without `content_object_id`s


hjanott (approved):
Thumbs up!

m-schieder (approved):

PR #2632, opened 2024-09-19
Allow to edit own delegation via user.update_self

### LINKED ISSUE #2503: Add allow_to_edit_own_delegation
See https://github.com/OpenSlides/OpenSlides/issues/6223 for context

**Desired behaviour:**

<s>We have a new setting which allows to delegate your own vote (allow_to_edit_own_delegation)</s>(decided againgst see comment below

And a group permission, which allows user from this group to delegate their own vote (`user.can_edit_own_delegation`)

A participant should be allowed to delegate their own delegation to whoever they want AND they should be allowed to remove the delegation from another person to themselves

`Can edit own delegation` needs the permission ‚Can see participants‘

luisa-beerboom:
Closes #2503


docs/actions/user.update_self.md
```diff
@@ -14,5 +16,9 @@ Updates the request user. Removes starting and trailing spaces from `username`.
 
 The given `gender` must be present in `organization/genders`.
 
+`meeting_id` is only for editing meeting-internal data, and the value will be thrown away afterwards.
```
hjanott, line 19: How about:
`meeting_id` is only used for processing meeting-internal data and will be omitted otherwise.

hjanott (approved):
LGTMAW.
If this will not lead to widespread changes you could overhaul the grammar of the existing error messages.

docs/actions/user.update_self.md
luisa-beerboom: The purpose of my description was to point out that `meeting_id` won't be written into the data anywhere and is just additional information given to aid action-internal calculations.
Your proposition doesn't do that, so I won't change it-

PR #2644, opened 2024-09-24
Fix migration 56

peb-adr:


peb-adr:
My change is to not try to iterate over unset 'access_group_ids'.
There may be a better way to do that. This is just my quick fix.

rrenkert (approved):

PR #2647, opened 2024-09-25
add new meeting specific setting hide metadata background for projector

### LINKED ISSUE #2642: Add new meeting specific setting
Context: https://github.com/OpenSlides/openslides-client/issues/4170 

**Wanted:**
A meeting needs a new setting to make the background from the motion-metadata transparent in the projector.

Recommendation for a name: `motions_hide_metadata_background`
It should be a boolean.

The backend only needs to enable/disable this setting. 

This setting should be optional, there should be no migration.
If the setting is none the client should pretend like the setting is false (I am not sure if that is relevant for the backend-issue).

hjanott:
Closes #2642


global/data/example-data.json
```diff
@@ -296,6 +296,7 @@
             "motions_enable_reason_on_projector": false,
             "motions_enable_sidebox_on_projector": false,
             "motions_enable_recommendation_on_projector": true,
+            "motions_hide_metadata_background": true,
```
luisa-beerboom, line 299: Why not use the default?

openslides_backend/action/actions/meeting/update.py
```diff
@@ -102,6 +102,7 @@
     "motions_enable_text_on_projector",
     "motions_enable_reason_on_projector",
     "motions_enable_sidebox_on_projector",
+    "motions_hide_metadata_background",
```
luisa-beerboom, line 105: Add the field to one of the tests

luisa-beerboom (changes requested):

global/data/example-data.json
hjanott: I changed the value to false.

openslides_backend/action/actions/meeting/update.py
hjanott: added

Elblinator (approved):

luisa-beerboom (approved):

PR #2657, opened 2024-10-02
Ensure speaker.delete only resets projector countdown with active speakers

### LINKED ISSUE #2655: List of speakers: Remove speaker stops countdown
**Describe the bug**
If a user is deleted from the list of speakers during an active speech with a countdown, the countdown is stopped and set back to the default start time.

**How to Reproduce**
1. open a meeting
2. open agenda, create an item and project it
3. open Projector > Detail view of standard projector > timers > activate countdown
4. open Autopilot > add a few users to the list of speakers. Start a speech. Countdown should now running in standard projector. Delete a user from the list. CD stops and resets. 

**Expected behavior**
Deleting a user from the list of speakers during a running speech should not reset and stop the countdown.

**Additional note:**
[{"action":"speaker.delete","data":[{"id":147}]}]

luisa-beerboom:
Closes #2655 

There were no projector_countdown-related tests in the speaker delete test file, so I added some.

### DIFF
--- a/tests/system/action/speaker/test_delete.py
+++ b/tests/system/action/speaker/test_delete.py
@@ -1,3 +1,5 @@
+from math import floor
+from time import time
 from typing import Any
 
 from openslides_backend.action.mixins.delegation_based_restriction_mixin import (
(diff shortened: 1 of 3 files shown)


tests/system/action/speaker/test_delete.py
```diff
@@ -399,3 +401,71 @@ def test_with_paused_structure_level_speaker(self) -> None:
...
+                    "meeting_id": 1,
+                },
+            }
+        )
+        self.add_coupled_countdown()
+        now = floor(time())
+        response = self.request("speaker.delete", {"id": 891})
+        self.assert_status_code(response, 200)
+        countdown = self.get_model("projector_countdown/75")
+        assert countdown.get("running") is True
```
hjanott, line 462: Isn't this redundant with the `assert_model_exists`?

tests/system/action/speaker/test_delete.py
```diff
@@ -399,3 +401,71 @@ def test_with_paused_structure_level_speaker(self) -> None:
...
+                },
+                "user/8": {
+                    "username": "test_username2",
+                    "meeting_user_ids": [8],
+                    "is_active": True,
+                    "default_password": DEFAULT_PASSWORD,
+                    "password": self.auth.hash(DEFAULT_PASSWORD),
+                },
+                "meeting_user/8": {"meeting_id": 1, "user_id": 8, "speaker_ids": [891]},
+                "list_of_speakers/23": {"speaker_ids": [890, 891], "meeting_id": 1},
```
hjanott, line 449: `meeting_id` doesn't need to be overwritten.

tests/system/action/speaker/test_delete.py
```diff
@@ -399,3 +401,71 @@ def test_with_paused_structure_level_speaker(self) -> None:
...
+        assert countdown.get("running") is True
+        self.assert_model_exists(
+            "projector_countdown/75",
+            {
+                "running": True,
+                "default_time": 200,
+                "meeting_id": 1,
+            },
+        )
+        self.assertAlmostEqual(countdown["countdown_time"], 100, delta=now)
```
hjanott, line 471: As 'countdown_time' has a timestamp while the countdown is running, we should compare to the timestamp 'now' with delta 100. Why don't you do this in the 'assert_model_exists'? Is it because this can have different results on different setups? Then how about passing `now` to 'add_coupled_countdown' to be using the same timestamp everywhere?

hjanott (approved):

tests/system/action/speaker/test_delete.py
hjanott: Change the two parameters around and this PR should be perfect.

hjanott (approved):

PR #2660, opened 2024-10-07
Global enable anonymous

### LINKED ISSUE #2656: Public Access: Add a new setting for global activation
As stated in https://github.com/OpenSlides/openslides-client/issues/4204 we need a new setting for the superadmin to activate/deactivate the public access globally on orga level.
The setting should prevent publishing a meeting if this option is set to false.
The default should be `false` and if toggled from true to false all published meetings should be unpublished.

luisa-beerboom:
Closes #2656

### DIFF
--- a/docs/actions/meeting.update.md
+++ b/docs/actions/meeting.update.md
@@ -174,6 +174,7 @@
 
 // Group D
     external_id: string;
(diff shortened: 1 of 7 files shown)


hjanott (approved):
LGTMAW.

PR #2662, opened 2024-10-07
Fix assignment_candidate cascade deletion error

### LINKED ISSUE #2651: Elections: deletion error
**Describe the bug**
Currently an election with at least one candidate and the final state 'finished' can't be deleted. Error message with 'Error: It is not permitted to remove a candidate from a finished assignment!' appears.

**How to Reproduce**
1. open a meeting
2. open elections
3. create an election
4. add a candidate and change the state to 'finished'. -> try to delete the election

**Expected behavior**
Deleting an election should always be possible to users with the permission to manage elections. State of election should not matter for this.

luisa-beerboom:
Closes #2651 

Found that this bug would actually affect _every_ cascading deletion of assignment_candidates, as well as the user merge.


hjanott (approved):
LGTMAW.

Elblinator (approved):

PR #2678, opened 2024-10-15
make meeting clone with gender in user possible again

### LINKED ISSUE #2677: Fix Error if meetings are cloned
**Current behaviour:**
if you try to clone a meeting where a user has a gender then an error is thrown

Thrown Error:
```
Error: user/1: Invalid fields gender (value: male) user/2: Invalid fields gender (value: female) user/3: Invalid fields gender (value: diverse)
```

Payload:
```
[
    {
        "action": "meeting.clone",
        "data": [
            {
                "meeting_id": 1
            }
        ]
    }
]
```

**Wanted behaviour:**
You should be able to clone a meeting even if there are users with a gender inside of them

hjanott:
Closes #2677 
add `internal_target` bool to `export_meeting` to prevent gender id conversion during meeting.clone


Elblinator (changes requested):
after once cloning a meeting the following error is thrown if you try to clone again
```
Error: user/1/is_present_in_meeting_ids: Relation Error: points to meeting/3/present_user_ids, but the reverse relation for it is corrupt
```
(based one the initial data)

Elblinator (approved):

PR #2683, opened 2024-10-16
Fix action worker test

luisa-beerboom:
Replaced the payload with one that

1. won't inevitably cause the action to fail
2. uses a faster action


hjanott (approved):
LGTM.

Issue #2695, opened 2024-09-03
password reset mail empty


selyod-ka:
**Describe the bug**
When using the password reset link on firefox, a mail is sent to the correct mail-address, but the mail is empty.

**How to Reproduce**
1. Go to 'https://openslides-domain.tld/login/forget-password
2. Enter mail-address, press button.
3. Mail is sent to the address.
4. The mail is empty; the subject reads as ": <username>"

**Expected behavior**
The mail should contain a password-reset-link.

**System information**
- OpenSlides version: 4.1.14
- Browser: Firefox 
- Browser version: 129.0.2

**Additional context**
With Brave (=Chromium 127) the mail sent contains a password reset link.

iiuc, the logs do not contain any useful information and they are the same for both password reset attempts (with firefox and with brave).

```
docker logs -f
[...]
backendAction-1     | INFO:openslides_backend.shared.handlers.base_handler:Performing action user.forget_password.
auth-1              | 2024-09-03T11:28:01.601Z http://auth:9004: POST -- /internal/auth/create-authorization-token
datastoreWriter-1   | [2024-09-03 11:28:01 +0000] [22] [INFO] Created position 22
datastoreWriter-1   | [2024-09-03 11:28:01 +0000] [22] [INFO] Events executed (1 UPDATE)
datastoreWriter-1   | [2024-09-03 11:28:01 +0000] [22] [INFO] UPDATE: ['user/3']
backendAction-1     | INFO:openslides_backend.shared.handlers.base_handler:Request was successful. Send response now.
[...]
```

selyod-ka:
sorry, somehow i did not receive any notification on this ticket's update.

i just tested again, with firefox 131 and https://demo.os4.openslides.com
the post-request to https://demo.os4.openslides.com/system/action/handle_request is:
`
[{"action":"user.forget_password","data":[{"email":"redacted@example.com"}]}]`

the mail i received now contains the username in the subject, but the mail body is still empty.

```
Return-Path: <noreply@openslides.com>
Delivered-To: user@example.com
Received: from proxy02.example.name ([127.0.0.1])
	by dovecot03.example.name (Dovecot) with LMTP id 6gnxMIxtG2dBxQIAVYRftg
	for <user@example.com>; Fri, 25 Oct 2024 12:11:05 +0200
Received: from proxy02.example.com ([127.0.0.1])
	by proxy02.example.name (Dovecot) with LMTP id Et2UM/I+G2e9iAEAGFAyLg
[...]

PR #2696, opened 2024-10-22
Fix wrong sorting of parsed language header

emanuelschuetze:
Fixed partly #2695 

Prefered language settings of browser (weighted by user, e.g. [de,en,cs]) should not sorted! In this example: [cs,de,en] -> cs translation is empty for password reset strings -> backend send empty email strings!

### DIFF
--- a/openslides_backend/i18n/translator.py
+++ b/openslides_backend/i18n/translator.py
@@ -56,7 +56,7 @@ class _Translator:
             code = code.split("-")[0]
             result.append((q, code))
         # sort by quality value and return only the codes
-        return [t[1] for t in sorted(result)]
+        return [t[1] for t in result]
 
 
 Translator = _Translator()


openslides_backend/i18n/translator.py
```diff
@@ -56,7 +56,7 @@ def parse_language_header(self, lang_header: str) -> list[str]:
             code = code.split("-")[0]
             result.append((q, code))
         # sort by quality value and return only the codes
-        return [t[1] for t in sorted(result)]
+        return [t[1] for t in result]
```
luisa-beerboom, line 59:
```suggestion
        return [t[1] for t in sorted(result, key=lambda tup: tup[0])]
```
The sort did not only sort the languages in alphabetical order: It sorted them by the weight value q first. If that wasn't given or equal to another value, it sorted them in alphabetical order.
I will agree that the latter is probably a bug (in this case caused by the way python sorts tuples), but the former definitely isn't.

The above code _should_ sort the languages by weight first and leave empty values in the same order as before.
If, during testing,  it turns out to wrongly prioritize or wrongly not-prioritize languages with no q-value over those with q-value, I'd suggest trying to set q to `"0.0"` instead of `"1.0"` in line 50.

luisa-beerboom (changes requested):
Apart from the code change request, this bug PR is missing a test.

I'd suggest using `openslides_backend/tests/system/action/meeting/test_create.py` lines 525 (just use the header_language string that causes the bug as a value) and 544 (`_(name)`) as a reference for how translation can be set up and used and writing a simple test in a completely new test file in a central place somewhere in the tests folder.

luisa-beerboom:
I moved the code to #2701 

This will happen as part of that PR.
Closing this.

PR #2698, opened 2024-10-22
Case insensitive email matching for forget password

### LINKED ISSUE #2687: Password reset: Email input field is case sensitive
**Describe the bug**
The 'Reset password' input field is case-sensitive. This is problematic if e-mails with capital letters have been entered in accounts. If users want to reset their password, the program requires exactly the same upper and lower case letters that were entered in the account. 

Additionally with have at the moment the situation that in many instances accounts are created with an e-mail that contains capital letters.

This field is meant:
image

**How to Reproduce**
1. open Accounts
2. Create an Account > Add a E-Mail with at least a capital letter.
3. Log out > open Password reset page > enter the E-Mail in lowercase. > You will not get mail. > Try it again and use the same upper and lower case letters as you've entered in step 2. > You will receive a mail.

**Expected behavior**
Three changes are necessary as a result. 

1. the password reset input field should ignore upper and lower case letters, as e-mail never has upper case letters.

2. create user: The single creation of an account as well as the creation of multiple accounts via csv file should ALWAYS ignore capital letters. Capital letters are always converted to lower case when creating or importing. This prevents e-mail addresses with upper and lower case letters from existing in the system. 

3. migration of existing data: All existing emails should be converted to lower case. If 3. does not work or requires consultation, please contact me.

luisa-beerboom:
Closes #2687


docs/actions/user.forget_password.md
```diff
@@ -9,11 +9,11 @@
...
+A user can change their password by this action without being authenticated. To perform a changing of their password the user enters their email-address (case insensitive). The client then sends a hard-coded reset email to the given email-address (as `email`). TODO: translations
 
 Regardless if the email-address is used by a user, the client shows the user a successful message (for example "An email was successfully sent to the given email-address"). This is necessary to avoid filtering which email-address is used by an OpenSlides-user.
 
-In the case that an email-address is used by a user, an email is sent to that email-address with the given text including a link to set a new password. If multiple users use the same email address, one email is sent per user.
+In the case that an email-address is used by a user, an email is sent to that email-address of that user including a link to set a new password. If multiple users use the same email address (case insensitive), one email is sent per user.
```
hjanott, line 16:
```suggestion
In the case that an email-address is used by a user, an email is sent to the email-address of that user including a link to set a new password. If multiple users use the same email address (case insensitive), one email is sent per user.
```

hjanott (approved):
LGTM.

Elblinator (approved):

PR #2701, opened 2024-10-24
Fix empty string translator

### LINKED ISSUE #2697: Fix empty string usage if translations are missing
**Describe the bug**
If the translations from a string are missing and the strings are delivered by the backend then empty strings are used to translate the string

**How to Reproduce**
Two examples: 
1. Set your browser default language to Spanish
2. In the es.po file remove the translated string from "You are receiving this email because you have requested a new password for your OpenSlides account..." and from "Email subject"
3. Send yourself a reset email the subject line and the content are empty

1. from the es.po remove the translation from "Delegates"
2. Create a new meeting in spanish
3. The following Error is thrown ```Error: data.name must be longer than or equal to 1 characters```

**Expected behavior**
If the translations in the translation file are missing (aka. are empty strings) then the english strings should be used

luisa-beerboom:
Closes #2697
Replaces #2696 
Fixed partly #2695

We need to test if this is a full fix for #2695

### DIFF
--- a/openslides_backend/i18n/translator.py
+++ b/openslides_backend/i18n/translator.py
@@ -22,7 +22,7 @@ class _Translator:
 
     def translate(self, msg: str) -> str:
         translation = self.translations[self.current_language].get(msg)
-        if translation:
+        if translation and translation.string:
             return translation.string
         else:
             return msg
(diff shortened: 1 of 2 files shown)


hjanott (approved):
LGTM.

MSoeb (approved):

PR #2717, opened 2024-11-06
Add user_id to action_worker

### LINKED ISSUE #2715: Add `user_id` to `action_worker`
To have an association from the action_worker to a user, a new field `user_id` should be added to the `action_worker`. The field should always contain the request user id.

Adding the field needs a migration. Since action_workers are used temporarily we can simply delete all old action_workers. @peb-adr before we merge this migration we should talk about the impact.

luisa-beerboom:
Closes #2715 
Needs OpenSlides/openslides-meta#191

### DIFF
--- a/global/data/example-data.json
+++ b/global/data/example-data.json
@@ -1,5 +1,5 @@
 {
-    "_migration_index": 61,
+    "_migration_index": 62,
     "gender":{
         "1":{
             "id": 1,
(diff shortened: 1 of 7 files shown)


rrenkert (changes requested):
We also need the migration of the old action_workers as described in https://github.com/OpenSlides/openslides-backend/issues/2715#issuecomment-2459112697

hjanott (approved):
LGTM.

rrenkert (approved):

PR #2745, opened 2024-11-28
saml-meeting-mapping: allow boolean attributes and vote weight as string

### LINKED ISSUE #2741: Mapping of boolean attributes in saml mapping
Mapping boolean attributes always saves `true` even if the the attribute is set to `False` (a string) in the IdP.
It not possible to set e.g. `is_active` to false.

### LINKED ISSUE #2743: Allow int in saml attribute mapping for vote weight
The regex for vote weight is to restrictive and only allows float with six decimals.
Allowed should be any float as a string with UP TO six decimals and int.

hjanott:
Closes #2741 
Closes #2743 

- [x] accept boolean as strings as defined in field validation
- [x] accept no float just int and decimal as a string for vote weight
- [x] complete json schema


openslides_backend/shared/schema.py
```diff
@@ -26,7 +26,7 @@
 }
 optional_str_schema: Schema = {
     "type": ["string", "null"],
-    "minLength": 1,
+    "minLength": 0,
```
luisa-beerboom, line 29: Why is it necessary to change one of the basic schemas?
The other optional schemas still have a `minLength` of 1. This is an inconsistency, that will cause confusion during future development. Either change them all (impractical, since that'll change the functionality of quite a few actions), or change none of them and write this code directly in the class that needs it.

luisa-beerboom (changes requested):
This PR needs a proper title and description

openslides_backend/shared/schema.py
hjanott: That's just a remnant of me experimenting on something. Thank you for pointing that out. :) I will revert that.

openslides_backend/action/actions/user/save_saml_account.py
```diff
@@ -465,10 +459,26 @@ def get_field_data(
...
+                        # Result must be string and have 6 digits after dot.
+                        if isinstance(value, int):
+                            value = f"{value:f}"
+                        elif isinstance(value, str) and re.compile(
+                            r"^[1-9]\d*\.?\d{0,6}$"
+                        ).match(value):
+                            if re.compile(r"^\d*$").match(value):
+                                value += ".000000"
+                            else:
+                                regex_condition = re.compile(DECIMAL_REGEX)
```
luisa-beerboom, line 471: `DECIMAL_REGEX` is already compiled, so no need to do it again ;-)

openslides_backend/action/actions/user/save_saml_account.py
```diff
@@ -465,10 +459,26 @@ def get_field_data(
...
                     elif saml_meeting_user_field == "vote_weight":
-                        # Result is string and has 6 digits.
-                        if isinstance(value, str):
-                            value = float(value)
-                        result = f"{value:f}"
+                        # Result must be string and have 6 digits after dot.
+                        if isinstance(value, int):
+                            value = f"{value:f}"
+                        elif isinstance(value, str) and re.compile(
+                            r"^[1-9]\d*\.?\d{0,6}$"
```
luisa-beerboom, line 466: I do not believe this will accept `0.***...` values (or negative values for that matter, but those don't make sense for the vote_weight anyway). I get that you need the fraction part to be variable length and therefore cannot use the `DECIMAL_REGEX` outright, but maybe you could replicate its format more closely

openslides_backend/action/actions/user/save_saml_account.py
```diff
@@ -465,10 +459,26 @@ def get_field_data(
...
+                        ).match(value):
```
luisa-beerboom, line 467: What is the rationale behind throwing numbers with 7 or more digits after the point away?

luisa-beerboom (changes requested):
You should test `save_saml_account.py` ln. 465-480 more thoroughly.
With more examples.

openslides_backend/action/actions/user/save_saml_account.py
hjanott: As I was told by @rrenkert that we only support numbers with up to 6 digits precision. Everything else must be handled as an error by the user/idp. This is also ensured by the field validation checks where exactly 6 digits after the dot are the accepted.

openslides_backend/action/actions/user/save_saml_account.py
hjanott: I used the 'DECIMAL_PATTERN' now.

openslides_backend/action/actions/user/save_saml_account.py
luisa-beerboom: Okay

luisa-beerboom (approved):

PR #2752, opened 2024-12-02
Kill all sessions of user set to inactive

### LINKED ISSUE #2742: Remove sessions if user is set to inactive
The action user.update with the attribute `is_active` set to false should invalidate all sessions of this user.

hjanott:
Closes #2742 
Waiting OpenSlides/openslides-auth-service#590


openslides_backend/action/actions/user/update.py
```diff
@@ -140,8 +140,11 @@ def update_instance(self, instance: dict[str, Any]) -> dict[str, Any]:
                 raise PermissionException(
                     "A superadmin is not allowed to set himself inactive."
                 )
-        if instance.get("is_active") and not user.get("is_active"):
-            self.check_limit_of_user(1)
+        if instance.get("is_active"):
+            if not user.get("is_active"):
+                self.check_limit_of_user(1)
+        elif user.get("is_active"):
+            self.auth.clear_sessions_by_user_id(instance["id"])
```
reiterl, line 147: "If is_action is not in payload (instance) and user is_active, then clear session", this doesn't seem right.
The is_active payload is **optional** as far as I know.

reiterl (changes requested):
Looks broken to me.

openslides_backend/action/actions/user/update.py
hjanott: That should be repaired now.

luisa-beerboom (approved):
You should give this PR a proper name

tests/system/action/user/test_update.py
```diff
@@ -2421,6 +2421,39 @@ def test_update_user_limit_okay(self) -> None:
...
+        response = self.request(
+            "user.update",
+            {
+                "id": 2,
+                "is_active": False,
+            },
+        )
+        self.assert_status_code(response, 200)
+        self.assert_model_exists("user/2", {"is_active": False})
+        self.assert_logged_out()
```
luisa-beerboom, line 2455: It would probably be better to just add the upper half of this test to another test, to ensure that we can differentiate what exactly is failing more easily

hjanott: Since we have a lot of tests and our current test setup creates a relatively big overhead for initializing and cleaning up after the tests, it would be better to keep those similar checks in one test. @rrenkert told me also that the situation is not going to change in the near future.

luisa-beerboom:
This PR still needs a proper name

luisa-beerboom (changes requested):
See above

tests/system/action/user/test_update.py
hjanott: Moved the test case to `test_update_no_OML_set`.

reiterl (approved):
Okay.

luisa-beerboom (approved):

PR #2766, opened 2024-12-12
Allow submitter ids and additional submitter on motion create

### LINKED ISSUE #2760: Fix motion.create with only permission 'motion.can_change_meta'
**Description**
Currently participants with 'can create motions' and 'can manage motion metadata' permissions (and not 'can manage motions') can only create motions without adding other submitters.

**How to Reproduce**
1. in settings > motions activate 'Enable text-based submitter in motion creation'
2. give a participants the permissions 'can create motions' and 'can manage motion metadata'
3. log in as that participant
4. create a motion
5. add any submitter from the dropdown menu or additonal user 
6. trying to save will results in this error: `You are not allowed to perform action motion.create. Forbidden fields: submitter_ids` or `You are not allowed to perform action motion.create. Forbidden fields: additional_submitter`

**Payload**
```
[
  {
    "action": "motion.create",
    "data": [
      {
        "meeting_id": 404,
        "title": "Test",
        "text": "<p>test</p>",
        "submitter_ids": [
          8249,
          8180
        ],
        "additional_submitter": "Vorstand",
        "workflow_id": 874
      }
    ]
  }
]
```

**Response**
```
{
    "success": false,
    "message": "You are not allowed to perform action motion.create. Forbidden fields: additional_submitter, submitter_ids"
}
```

**Expected Behaviour**
If participant has 'motion.can_change_meta' and 'user.can_see', then they should be allowed to assign submitter_ids.
If participant has 'motion.can_change_meta' and 'motions_create_enable_additional_submitter_text' is active, they should be allowed to assign additional_submitter

hjanott:
Closes #2760

### DIFF
--- a/openslides_backend/action/actions/motion/create.py
+++ b/openslides_backend/action/actions/motion/create.py
@@ -1,4 +1,5 @@
-from typing import Any
+from collections import defaultdict
+from typing import Any, cast
 
 from ....models.models import Motion
 from ....permissions.base_classes import Permission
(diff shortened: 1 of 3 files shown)


openslides_backend/action/actions/motion/create.py
```diff
@@ -146,19 +147,25 @@ def check_permissions(self, instance: dict[str, Any]) -> None:
...
+            Permissions.Mediafile.CAN_SEE: ["attachment_mediafile_ids"],
+            Permissions.Motion.CAN_MANAGE_METADATA: [
+                "additional_submitter",
+                "submitter_ids",
+            ],
+            Permissions.User.CAN_SEE: ["submitter_ids"],
+        }
+        for perm, fields in permission_to_fields.items():
+            if has_perm(self.datastore, self.user_id, perm, instance["meeting_id"]):
+                whitelist.extend(fields)
```
reiterl, line 164: You put submitter_ids on the whitelist, even when not UserCanSee, that is a problem.

reiterl (changes requested):
Missing tests for motion.create and motion can manage metadata with enabled setting.

reiterl (commented):
It seems to work, okay, but elegant is sth different. 
The permission_to_fields doesn't seem to be the right config as you want a list of perms and a list of fields be connected. Perhaps a list of pair of lists, or an own class?
That would be overkill. Why refactor the code in the first place? I don't know.

reiterl (approved):

openslides_backend/action/actions/motion/create.py
hjanott: This is now prevented.

Elblinator (approved):

PR #2769, opened 2024-12-16
Remove global folder and move folder  contents up

### LINKED ISSUE #2604: Renaming global directory
At the moment we use global as the name for the directory containing the meta repository and data intended for further use in other modules.

'global' is one of the [reserved keywords in python](https://docs.python.org/3/reference/lexical_analysis.html#keywords) and thus must not be used in this case as it will lead to a syntax error. There are work-arounds but they will lead to errors by the linting tools we use (e.g. mypy) that need to be suppressed.

example: 
`>>> import global.example_module`

work-around:
`>>> sys.path.append("global")`
`>>> import example_module`

work-around (suppressed linting):
`>>> sys.path.append("global")`
`>>> import example_module # type: ignore # noqa`

To tackle the problem at the root I suggest we rename the directory to something similar (e.g. universal, general, unbounded, planetary, comprehensive) and adjust all references.

hjanott:
Closes #2604


dev/docker-compose.dev.yml
```diff
@@ -15,7 +15,6 @@ services:
             - ../openslides_backend:/app/openslides_backend
             - ../tests:/app/tests
             - ../cli:/app/cli
-            - ../global:/app/global
             - ../scripts:/app/scripts
         environment:
```
reiterl, line 19: I would think, that the data and meta are missing in the docker compose part now.
Pls test.

reiterl (approved):
One open point, else it looks good.
Please don't forget to test the scripts e.g. generate_models.

dev/docker-compose.dev.yml
hjanott: Yes you are right. Everything was working but it should be included into the volumes. As well should requirements; to be more consistent.

PR #2793, opened 2025-01-08
Fix merge for saml users in case of password changing rights

### LINKED ISSUE #2789: Fix user merge if one of them is a saml user
**Describe the bug**
If two users are merged together and one if them is a saml user and the other one is not a saml user then the following Error is thrown: 
```
Error: user <string> is a Single Sign On user and may not set the local default_passwort or the right to change it locally.
```

**How To Reproduce**
1. In a meeting create user A and via import import a saml user (e.g. G)
2. On the Orga-level: Merge these two accounts together, merge the normal user into the saml user (saml user persists)
4. Error thornw

```
{
  "action": "user.merge_together",
  "data": [
    {
      "id": 7,
      "user_ids": [
        3
      ]
    }
  ]
}
```

**Response from backend**
```
{
    "success": false,
    "message": "user <string> is a Single Sign On user and may not set the local default_passwort or the right to change it locally.",
    "action_error_index": 0
}
```

**Expected behavior**
The two users should be merged into the one saml user

luisa-beerboom:
Closes #2789

@Elblinator Given that the exact data of the two users were not included, I could only guess what happened based on the error message. In the end I quickly found a way to reproduce it and this PR solves the problem in the case of that reproduction (I got to the conclusion that the secondary user must've had `can_change_own_password` set to `true`). You should check it out yourself, though, just to make sure that the entire problem is actually fixed from the client's perspective.

For now my solution is to throw password changing rights of users away if they are merged into a saml user, as this value is not settable on a saml user. This is technically a change to the way the intended merge functionality is defined in this case, so maybe it should be discussed if this is fine, or if its supposed to act another way.


openslides_backend/action/actions/user/merge_together.py
```diff
@@ -590,6 +588,19 @@ def handle_special_field(
...
+                        return None
+                    if len(
+                        comp_data := [
+                            date
+                            for model in [into_, *ranked_others]
+                            if (date := model.get("can_change_own_password"))
+                            is not None
+                        ]
+                    ):
+                        return max(comp_data)
```
hjanott, line 602: Is max faster than any?

hjanott (approved):
LGTM.

PR #2797, opened 2025-01-10
Build in poll.can_see_progress permission

luisa-beerboom:
No issue was made for this, but this is a read permission anyway, so the backend doesn't particularly have to do much


hjanott (approved):
LGTM.

PR #2820, opened 2025-01-31
Delete all moderator notes in agenda items

### LINKED ISSUE #2819: Migration doesn't delete all moderator notes from agenda item
**Describe the bug**
Migration 60 moves moderator notes from the agenda item to the list of speakers but doesn't delete those with empty string.

**How To Reproduce**
With a version prior to migration 60 generate an agenda and agenda item.
Save a moderator note with empty text. 
Migrate the database.

**Response from backend**
agenda_item/XXXX: Invalid fields moderator_notes (value: )

**Expected behavior**
The agenda item shouldn't even have a moderator note with empty string after the migration.

**Additional context**

hjanott:
Closes #2819

### DIFF
--- a/data/example-data.json
+++ b/data/example-data.json
@@ -1,5 +1,5 @@
 {
-    "_migration_index": 63,
+    "_migration_index": 64,
     "gender":{
         "1":{
             "id": 1,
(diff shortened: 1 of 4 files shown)


luisa-beerboom (changes requested):
You should test what happens if there's a deleted model that fulfils the requirements.
Otherwise the code looks fine

PR #2822, opened 2025-02-03
Make meeting user delete motion editor and working group speaker cascadingly

### LINKED ISSUE #2821: Meeting user with motion editor can not be deleted
**Describe the bug**
If a `users` `meeting_user` has a `motion_editor` it can not be deleted.

**How To Reproduce**
* Create a user.
* Make this user somehow a motion editor.
* Try to delete this user.

**Response from backend**
`openslides_backend.shared.handlers.base_handler:Error occured on index 0: Update of motion_editor/X: You try to set following required fields to an empty value: ['meeting_user_id']`

**Additional context**
Since deleting the user leads to deleting the meeting user, which then leads to removing the entry from  `motion_editor/X/meeting_user_id`, and `meeting_user_id` is required in this context, the error is thrown.

~It is to be discussed if the required field in the `motion_editor` and `motion_working_group_speaker` should not be required or if~ the `motion_editor` and `motion_working_group_speaker` should be deleted cascadingly like `motion_submitters` are ~or if someone else should be made the editor f.e. the meeting admin. If we just delete the `meeting_user_id` or `motion_editor` itself this would possibly require the client to display something like "Deleted User" or a string from a new field, which could contain the name of the deleted user.
We should also evaluate if there currently is a use case for the `motion_editor`.~

hjanott:
closes #2821
related OpenSlides/openslides-meta#210

### DIFF
--- a/docs/actions/user.merge_together.md
+++ b/docs/actions/user.merge_together.md
@@ -108,7 +108,7 @@ This is usually necessary in cases where the relation is set to cascade-delete
 Equivalence is determined via `meeting_id`: All meeting_users with the same `meeting_id` are grouped together.
 
 The primary model is updated/re-created with the information from the secondary models using the following rules:
-- `assignment_candidate_ids` are update-merged
+- `assignment_candidate_ids` is update-merged
 - `motion_editor_ids`, `motion_submitter_ids`, `motion_working_group_speaker_ids`, `personal_note_ids` and `speaker_ids` are create-merged
 - other relation-lists are set to the union of their content among all selected users
 - `comment`, `number`, `about_me`, `vote_weight`, `vote_delegated_to_id` are set to the value from the highest ranked model that has the field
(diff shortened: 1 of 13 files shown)


Elblinator (commented):
this is the current behaviour in the client:
Big Letter chad is only removed and the most down one is deleted:
Screenshot_20250203_131526

Is it possible to remove the current person completely ifthey are deleted? @hjanott 
-> Is that the wanted behaviour? @rrenkert

hjanott:
> Is it possible to remove the current person completely ifthey are deleted? @hjanott

Yes that is possible.

Elblinator (approved):

docs/actions/user.merge_together.md
```diff
@@ -108,8 +108,8 @@ This is usually necessary in cases where the relation is set to cascade-delete
 Equivalence is determined via `meeting_id`: All meeting_users with the same `meeting_id` are grouped together.
 
 The primary model is updated/re-created with the information from the secondary models using the following rules:
-- `assignment_candidate_ids`, `motion_editor_ids` and `motion_working_group_speaker_ids` are update-merged
-- `motion_submitter_ids`, `personal_note_ids` and `speaker_ids` are create-merged
+- `assignment_candidate_ids` are update-merged
```
luisa-beerboom, line 111:
```suggestion
- `assignment_candidate_ids` is update-merged
```

luisa-beerboom (changes requested):

luisa-beerboom (approved):

PR #2833, opened 2025-02-07
Implement new database reader and writer

hjanott:
This PR is meant for a more efficient communication and should not be merged as is.


tests/database/reader/system/get_many/test_get_many.py
```diff
@@ -0,0 +1,289 @@
...
+    assert "A field does not exist in model table: column " in e_info.value.msg
+    assert "does_not_exist" in e_info.value.msg
+
+
+def test_negative_id(
+    db_connection: Connection, db_cur: Cursor, reader: ExtendedDatabase
+) -> None:
+    request = [
+        GetManyRequest("committee", [-1], ["name"]),
+        GetManyRequest("committee", [1], ["forwarding_user_id"]),
```
luisa-beerboom, line 257: Why the second row?

tests/database/reader/system/get_many/test_get_many.py
```diff
@@ -0,0 +1,289 @@
...
+    with pytest.raises(InvalidFormat) as e_info:
+        reader.get_many(request, use_changed_models=False)
+    assert "Id must be positive." in e_info.value.msg
+
+
+def test_invalid_collection(
+    db_connection: Connection, db_cur: Cursor, reader: ExtendedDatabase
+) -> None:
+    request = [
+        GetManyRequest("committeee", [1], ["name"]),
```
luisa-beerboom, line 268: Why the second row

tests/database/reader/system/get/test_get.py
```diff
@@ -0,0 +1,138 @@
...
+    assert response == standard_response
+
+
+def test_mapped_fields_not_exists(
+    db_connection: Connection, db_cur: Cursor, reader: ExtendedDatabase
+) -> None:
+    setup_data(db_connection, db_cur, data)
+    with pytest.raises(InvalidFormat) as e_info:
+        reader.get(FQID, ["field_that_doesnt_exist"])
+    assert "A field does not exist in model table: " in e_info.value.msg
```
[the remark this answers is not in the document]
hjanott: In the first case we are testing for a number of fields that is over the maximum so that we report back the whole model. However, the current maximum of 1000 fields is currently not reached by any model we use.
In the second case we are querying for a field that doesn't exist. In such a case we would get an error by postgres, which we evaluate here. 
What could possibly still be improved, is to state the exact field that does not work.

tests/database/reader/system/get_many/test_get_many.py
hjanott: This assures that the second GetManyRequest is not somehow executed and the exception ignored.

PR #2841, opened 2025-02-12
Add delegation history messages

### LINKED ISSUE #2835: History doesn't list changes done to vote delegation
**Describe the bug**
Changes to vote delegations are not displayed in the history. 

I came across this error while testing a new feature (transfer your own voting rights). At first I thought that it only affected the new feature, but this is not the case. 

**How To Reproduce**
Precondition: Activate vote delegation in a meeting
1. open a meeting
2. open participants > create two participants as delegates
3. Delegate vote right from User A to User B
4. Open the history and check whether the change is displayed.
 
Image
-- Not sure, if this is the correct payload which could help. It is the payload from changing a vote delegation. --

**Response from backend**
Image

**Expected behavior**
Vote delegation changes should also be recorded in the history. This includes changes made by admins as well as participants (via the new feature).

The changes should appear as an information that a vote delegation change has been made, but no details of the change.

luisa-beerboom:
Closes #2835


tests/system/action/user/test_delegation_history.py
```diff
@@ -0,0 +1,364 @@
...
+            [
+                "Proxy voting rights removed in meeting {}",
+                "meeting/1",
+                "Proxy voting rights received in meeting {}",
+                "meeting/1",
+            ],
+        )
+        self.assert_history_information(
+            f"user/{self.alice_id}",
+            ["Vote delegation canceled in meeting {}", "meeting/1"],
```
hjanott, line 167: Wouldn't it be nice if the history stated which is the other affected user? Similar to "Vote delegated to {} in meeting {}"

tests/system/action/user/test_delegation_history.py
```diff
@@ -0,0 +1,364 @@
...
+            self.alice_id,
+            self.next_user_id,
+            prepend_to=[
+                "Account created",
+                "Participant added to meeting {}.",
+                "meeting/1",
+            ],
+        )
+
+    def setup_delegation(self) -> None:
```
hjanott, line 83: I think it's more orderly to have the setup methods at the beginning. What do you think?

hjanott (commented):
I have a few minor remarks.

tests/system/action/user/test_delegation_history.py
luisa-beerboom: If it were the test cases `setUp` function (thatis called before every test), sure, but I think its preferable to have helper functions like this above the first test that uses them, better for the flow of reading.

openslides_backend/action/actions/meeting_user/history_mixin.py
```diff
@@ -1,87 +1,486 @@
...
+                    )
+        if instance_information:
+            self.add_entries_to_history_information(
+                information, instance_information, for_user_id=user_id
+            )
 
-            if instance_information:
-                information[fqid_from_collection_and_id("user", user_id)] = (
-                    instance_information
+    def add_created_meeting_user_history_information(
```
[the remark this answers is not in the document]
hjanott: I've meant `add_updated_meeting_user_history_information`

tests/system/action/user/test_delegation_history.py
hjanott: That's the case if it is used only immediately after the definition. However, for longer files that flow of reading argument is not that relevant in my eyes anymore. Was this style the convention before? Then we can continue like this. Otherwise I would argue for not mixing helper methods with test methods. That's also obstructing the flow of reading in a way.

tests/system/action/user/test_delegation_history.py
hjanott: I'm going along with your suggestion.

openslides_backend/action/actions/meeting_user/history_mixin.py
```diff
@@ -1,87 +1,486 @@
...
+            meeting_user_ids.append(muser_id)
+        if "vote_delegated_to_id" in instance and db_instance.get(
+            "vote_delegated_to_id"
+        ):
+            meeting_user_ids.append(db_instance["vote_delegated_to_id"])
+        if meeting_user_ids:
+            user_ids: set[int] = {
+                muser["user_id"]
+                for muser in self.datastore.get_many(
+                    [GetManyRequest("meeting_user", meeting_user_ids, ["user_id"])],
```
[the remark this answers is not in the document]
luisa-beerboom: Done

openslides_backend/action/actions/meeting_user/history_mixin.py
```diff
@@ -1,87 +1,486 @@
...
+            user_id = for_user_id
+        fqid = fqid_from_collection_and_id("user", user_id)
+        if fqid not in information:
+            information[fqid] = entries
+        else:
+            for entry in entries:
+                if entry not in information[fqid]:
+                    information[fqid].append(entry)
+
+    def compose_history_string(
```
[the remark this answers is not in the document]
luisa-beerboom: The create history information function is fine where it is.
With the current build, deleting it would cause errors

tests/system/action/user/test_delegation_history.py
luisa-beerboom: reorganized the methods

tests/system/action/user/test_delegation_history.py
luisa-beerboom: Done

hjanott (approved):
lgtm

PR #2845, opened 2025-02-13
convert to str before condition check

### LINKED ISSUE #2843: Non string value in SAML data fails in backend action
**Describe the bug**
If a saml user sends a value other than a string which is then checked by a mapper condition during login, the backend fails.

**How To Reproduce**
Create a SAML account.
Set the meeting mapping to check a value that is not a general model field.
Set the SAML accounts data to have a to be checked value to be f.e. a number.
Login with said SAML account.

**Stacktrace from backend**
File "/app/openslides_backend/action/actions/user/save_saml_account.py", line 278, in apply_meeting_mapping
if self.validate_meeting_mapper(instance_old, meeting_mapper):
File "/app/openslides_backend/action/actions/user/save_saml_account.py", line 260, in validate_meeting_mapper
return all(
File "/app/openslides_backend/action/actions/user/save_saml_account.py", line 263, in <genexpr>
and regex_condition.search(instance_value)
TypeError: expected string or bytes-like object

**Expected behavior**
The SAML user should be mapped with this mapper and the value interpreted as string.

hjanott:
closes #2843

### DIFF
--- a/docs/actions/user.save_saml_account.md
+++ b/docs/actions/user.save_saml_account.md
@@ -38,7 +38,7 @@ Extras to do on creation:
     - A mapper can be given a 'name' for debugging purposes.
     - The 'external_id' maps to the meeting and is required (logged as warning if meeting does not exist). Multiple mappers can map to the same meeting.
     - If 'allow_update' is set to false, the mapper is only used if the user does not already exist. If it is not given it defaults to true.
-    - Mappers are only used if every condition in the list of 'conditions' resolves to true. For this the value for 'attribute' in the payload data has to be a string and match the string or regex given in 'condition'. If no condition is given this defaults to true.
+    - Mappers are only used if every condition in the list of 'conditions' resolves to true. For this the value for 'attribute' in the payload data has to be a string and match the string or regex given in 'condition'. If no condition is given this defaults to true. If the data is not a string, then it will be cast to a string for this purpose. F.e. "True", "4", "[False, 42, 'Text']" and so on.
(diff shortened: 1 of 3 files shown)


luisa-beerboom (approved):

docs/actions/user.save_saml_account.md
```diff
@@ -38,7 +38,7 @@ Extras to do on creation:
     - A mapper can be given a 'name' for debugging purposes.
     - The 'external_id' maps to the meeting and is required (logged as warning if meeting does not exist). Multiple mappers can map to the same meeting.
     - If 'allow_update' is set to false, the mapper is only used if the user does not already exist. If it is not given it defaults to true.
-    - Mappers are only used if every condition in the list of 'conditions' resolves to true. For this the value for 'attribute' in the payload data has to be a string and match the string or regex given in 'condition'. If no condition is given this defaults to true.
+    - Mappers are only used if every condition in the list of 'conditions' resolves to true. For this the value for 'attribute' in the payload data has to be a string and match the string or regex given in 'condition'. If no condition is given this defaults to true. If the data is not a string, then it will be cast to a string for this purpose. F.e. "True", "4", "[False, 42, 'Text']" and so on.
```
peb-adr, line 41: "For example" is usually abbreviated with "e.g." (meaning "exempli gratia"), not "f.e.".

hjanott: That's correct. I like the non-Latin variant variant more, though.

hjanott:
Closing this and reopening again, because of repository mishap.

PR #2847, opened 2025-02-14
Make meeting/external_id unique in org

### LINKED ISSUE #2806: Make meeting external ids unique in orga context.
Right now meeting external ids are unique in context of committees. This creates problems with the SAML meeting mapping which expects not to use committee information. Using that would bloat the json configuration of the SAML attribute mapping.
We suggest to make meeting external ids to be unique in orga context.

hjanott:
Closes #2806


hjanott:
Closing this because of repository mishap.

PR #2855, opened 2025-02-17
Make meeting/external_id unique in org

### LINKED ISSUE #2806: Make meeting external ids unique in orga context.

hjanott:
Closes #2806

### DIFF
--- a/tests/system/action/meeting/test_import.py
+++ b/tests/system/action/meeting/test_import.py
@@ -32,6 +32,7 @@ class MeetingImport(BaseActionTestCase):
                 "meeting/1": {
                     "committee_id": 1,
                     "group_ids": [1],
+                    "external_id": "ext_id",
                     "is_active_in_organization_id": ONE_ORGANIZATION_ID,
                 },
                 "group/1": {"meeting_id": 1, "name": "group1_m1"},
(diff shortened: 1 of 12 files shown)


tests/system/action/meeting/test_import.py
```diff
@@ -1566,6 +1567,16 @@ def test_missing_required_field(self) -> None:
...
 
+    def test_duplicate_external_id(self) -> None:
+        request_data = self.create_request_data({})
+        request_data["meeting"]["meeting"]["1"]["external_id"] = "ext_id"
+        response = self.request("meeting.import", request_data)
+        self.assert_status_code(response, 400)
+        assert (
+            "The external id of the meeting is not unique in the organization scope. Send a differing external id with this request."
+            in response.json["message"]
+        )
```
luisa-beerboom, line 1578: We can't really expect the user to edit the import file. Perhaps the import action should also have an `external_id` option.

luisa-beerboom (approved):
Generally LGTM, one remark that I think should be discussed.
Also merging this should wait until the client is using the new `external_id` action parameters.

tests/system/action/meeting/test_import.py
hjanott: @rrenkert and I discussed this. 
We came to the conclusion that this is nearly just a find and replace which should be doable for someone being superadmin.
As we are now removing the external_id, if no new external id was given, this can be merged without the client.

luisa-beerboom (approved):

PR #2914, opened 2025-03-26
Update csv import warning message

### LINKED ISSUE #2886: Participant csv import: information change
**Current behavior**
If a meeting admin wants to import participants for whom he does not have the permission to edit, a technically correct but unhelpful message is displayed in the preview during the csv import. The message suggests problems, despite the fact that there are no problems. The wording of the note should therefore be improved.

Current wording: Following fields were removed from payload, because the user has no permissions to change them: member_number, username

-> Technical the wording is good, but a common user won't understand this.

Information can be found by hovering of the Info-icon
Image

**Reproduction**
preconditions: One Meeting. A user (Max Mustermann) which is admin of the meeting AND committee of the associated meeting. One additional user (Maxim Musterfrau) with a member number, which is not in the scope of Max Mustermann. Max can only reference the user.
1. login as max
2. open participants > import > download exmpale csv file > delete all content, except of the column names, in it > add the member number of Maxim Musterfrau. Other columns should be empty. save file > upload the file and check the info icon. 

**Wanted behavior**
Change info text for this case to: "Account is added to the meeting, but changes to the following field(s) are not possible: member_number, username"

vkrasnovyd:
Closes #2886

### DIFF
--- a/openslides_backend/action/actions/user/participant_json_upload.py
+++ b/openslides_backend/action/actions/user/participant_json_upload.py
@@ -118,7 +118,7 @@ class ParticipantJsonUpload(BaseUserJsonUpload, ParticipantCommon):
                 failing_fields.remove("member_number")
         if failing_fields:
             messages.append(
-                f"Following fields were removed from payload, because the user has no permissions to change them: {', '.join(failing_fields)}"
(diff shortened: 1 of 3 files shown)


luisa-beerboom (approved):
LGTM

PR #2917, opened 2025-03-26
Prevent archiving of meeting with running processes

### LINKED ISSUE #2836: Meeting archiving: improvement for running process needed
**Current behavior**
When archiving meetings, there is no check as to whether there are ongoing processes, such as voting. This leads to practical problems. For example, accounts cannot be merged if they are in an ongoing voting process. 

If a meeting is archived, the merge cannot simply be carried out. The current solution is to unarchive the meeting and end the voting. However, this (unarchiving) can only be done by a superadmin. This in turn leads to practical problems, as the superadmin is usually not immediately available.

**Wanted behavior**
To solve this problem, a check should be carried out during archiving to ensure that there are running processes. This includes votes on motions, elections and active speeches. If one of these processes is active, a dialog box should appear indicating that the following processes are active and should be terminated in order to perform archiving.

vkrasnovyd:
Closes #2836

### DIFF
--- a/docs/actions/meeting.archive.md
+++ b/docs/actions/meeting.archive.md
@@ -9,7 +9,7 @@
 ## Action
 Archives the meeting by setting the is_active_in_organization_id to `None`.
 
-This fails if meeting has ongoing polls or an active speaker.
+This fails if meeting has ongoing polls or active speakers.
 
 ## Permissions
 archive a meeting is allowed with CML can_manage or OML can_manage_organization rights
--- a/docs/actions/meeting.archive.md
+++ b/docs/actions/meeting.archive.md
@@ -9,5 +9,7 @@
 ## Action
 Archives the meeting by setting the is_active_in_organization_id to `None`.
 
+This fails if meeting has ongoing polls or an active speaker.
+
 ## Permissions
 archive a meeting is allowed with CML can_manage or OML can_manage_organization rights
--- a/openslides_backend/action/actions/meeting/archive.py
+++ b/openslides_backend/action/actions/meeting/archive.py
@@ -9,7 +9,8 @@ from ....permissions.permission_helper import (
     has_committee_management_level,
     has_organization_management_level,
 )
-from ....shared.exceptions import PermissionDenied
+from ....shared.exceptions import ActionException, PermissionDenied
+from ....shared.filters import And, FilterOperator
 from ....shared.patterns import fqid_from_collection_and_id
 from ...generics.update import UpdateAction
 from ...util.default_schema import DefaultSchema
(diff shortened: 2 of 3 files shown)


openslides_backend/action/actions/meeting/archive.py
```diff
@@ -22,6 +23,41 @@ class MeetingArchive(UpdateAction, GetMeetingIdFromIdMixin):
...
+        if active_speaker_exists:
+            messages.append("speaker")
+        if len(active_polls):
+            messages.append(
+                f"polls ({', '.join([model['title'] for model in active_polls.values()])})"
+            )
+        if len(messages):
+            raise ActionException(
+                f"Cannot archieve meeting with active {' and '.join(messages)}."
+            )
```
luisa-beerboom, line 59: `validate_instance` is for payload format checking and should not include any datastore calls, please move this to `update_instance`.

openslides_backend/action/actions/meeting/archive.py
```diff
@@ -22,6 +23,41 @@ class MeetingArchive(UpdateAction, GetMeetingIdFromIdMixin):
...
+            ["title"],
+            lock_result=False,
+        )
+
+        if active_speaker_exists:
+            messages.append("speaker")
+        if len(active_polls):
+            messages.append(
+                f"polls ({', '.join([model['title'] for model in active_polls.values()])})"
+            )
```
luisa-beerboom, line 55: If you want to list the specific models, I'd say using the model ids would be better since names can be non-unique (and it's also how we usually do things if it needs to be this specific), but in this case (since the state of all polls in the meeting can actually be viewed in the client) it'd probably be enough to just say `"Cannot archive meeting with active polls."`

openslides_backend/action/actions/meeting/archive.py
```diff
@@ -22,6 +23,41 @@ class MeetingArchive(UpdateAction, GetMeetingIdFromIdMixin):
...
+        )
+        active_polls = self.datastore.filter(
+            "poll",
+            And(
+                FilterOperator("meeting_id", "=", instance["id"]),
+                FilterOperator("state", "=", "started"),
+            ),
+            ["title"],
+            lock_result=False,
+        )
```
luisa-beerboom, line 48: No need to treat speakers and polls differently, just use `datastore.exists` for both

tests/system/action/meeting/test_archive.py
```diff
@@ -100,3 +104,113 @@ def test_archive_wrong_meeting(self) -> None:
...
+                    "list_of_speakers_id": 1,
+                    "meeting_id": 1,
+                    "begin_time": floor(time()) - 100,
+                },
+            }
+        )
+        response = self.request("meeting.archive", {"id": 1})
+        self.assert_status_code(response, 400)
+        assert (
+            response.json["message"] == "Cannot archieve meeting with active speaker."
```
luisa-beerboom, line 168:
```suggestion
            response.json["message"] == "Cannot archive meeting with active speakers."
```
Some spelling

docs/actions/meeting.archive.md
```diff
@@ -9,5 +9,7 @@
 ## Action
 Archives the meeting by setting the is_active_in_organization_id to `None`.
 
+This fails if meeting has ongoing polls or an active speaker.
```
luisa-beerboom, line 12:
```suggestion
This fails if meeting has ongoing polls or active speakers.
```
There can be multiple active speakers across multiple list_of_speakers.

luisa-beerboom (changes requested):

luisa-beerboom (approved):

PR #2923, opened 2025-04-03
Extend user scope with archived meetings for cmo and oml

### LINKED ISSUE #2901: Allow committee admin to user.update if user is only in archived meetings
**Describe the bug**
If a user is in only archived meeting(s) then the user cannot be updated by an committee admin

**How To Reproduce**
1. Create two new users (A and B)
2. Create a new committee add A as committee admin
3. Create a meeting 
4. Add B as participant to this meeting
5. Archive this meeting
6. Login as A 
7. Try  update B

```
{
	"0": {
		"action": "user.update",
		"data": [
			{
				"email": "",
				"id": 1096,
				"pronoun": "jj"
			}
		]
	}
}
```

**Response from backend**
```
success | false
message | "You  are not allowed to perform action user.update. Missing permission:  OrganizationManagementLevel can_manage_users in organization 1"
```

**Expected behavior**
That you can update a user even if this user is only in (an) archived meeting(s)

**Additional Information:**
I only tested it for one archived meeting. 
So I cannot say for sure that this is an issue for several meetings but I highly suspect that

vkrasnovyd:
Closes #2901 

* Moved checking 'is_active_in_organization_id' flag to the beginning of 'CheckForArchivedMeetingMixin.check_for_archived_meeting' to improve performance;
* Removed filtering of 'committee_meetings' by 'is_active_in_organization_id' if user has olm or cmo rights;
* Made check_for_archived_meeting dependend on the user permission level in the committee of the meeting;


tests/system/action/user/test_update.py
```diff
@@ -982,6 +982,67 @@ def test_perm_group_A_cml_manage_user_archived_meeting_in_other_committee(
...
+        )
+        self.assert_status_code(response, 200)
+        self.assert_model_exists(
+            "user/111",
+            {
+                "username": "new_username",
+                "meeting_ids": [1],
+                "committee_ids": [60],
+            },
+        )
```
luisa-beerboom, line 1044: You should also check what happens if a user is in one archived _and_ one active meeting.
Both in the same and in different committees.

luisa-beerboom (changes requested):

openslides_backend/shared/mixins/user_create_update_permissions_mixin.py
```diff
@@ -292,21 +293,25 @@ def check_group_A(self, fields: list[str], instance: dict[str, Any]) -> None:
...
+                            if meeting_id is not None
+                        },
+                    }
+                )
+                raise MissingPermission(missing_permissions)
+            elif (
+                self.check_for_admin_in_all_meetings(instance.get("id", 0))
+                and self.user_in_archived_meetings_only
+            ):
+                raise MissingPermission(missing_permissions)
```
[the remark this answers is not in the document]
vkrasnovyd: Moved the logic to the class method _check_missing_permissions_groups_AF for clarity and called it for both groups

tests/system/action/user/test_update.py
vkrasnovyd: 1 archived and 1 active meetings in different committees are checked in the test_perm_group_A_meeting_manage_user_archived_meeting and test_perm_group_A_meeting_manage_user_archived_meeting_with_parent_permission.

Below them I've dded similar tests for one archived and one active meeting in the same committee.

tests/system/action/user/test_update.py
```diff
@@ -1331,6 +1331,58 @@ def perm_group_A_meeting_manage_user_archived_meeting(
...
+                "id": 111,
+                "username": "new_username",
+            },
+        )
+        self.assert_status_code(response, 200)
+        self.assert_model_exists(
+            "user/111",
+            {
+                "username": "new_username",
+                "committee_ids": None,
```
luisa-beerboom, line 1380: `committee_ids` should not be None if the user is in the committees meetings. Please make sure you have a valid state of this field before the action is called and then check if it is filled correctly afterwards here.

tests/system/action/user/test_update.py
```diff
@@ -1331,6 +1331,58 @@ def perm_group_A_meeting_manage_user_archived_meeting(
...
+    ) -> None:
+        self.perm_group_A_meeting_manage_user_active_and_archived_meetings_in_same_committee(
+            Permissions.User.CAN_MANAGE
+        )
+
+    def perm_group_A_meeting_manage_user_active_and_archived_meetings_in_same_committee(
+        self, permission: Permission
+    ) -> None:
+        """
+        May update group A fields on meeting scope. User belongs to 1 meeting without being part of a committee
```
luisa-beerboom, line 1352: A user is considered "part of a committee" if he is in a meeting in that committee or is a committee manager (with committee hierarchy feature this will also include if they have the committee set as `home_committee_id`), so belonging to a meeting without being part of a committee is not a realistic situation. I don't think this even interacts with the archived state of a meeting.

Please update this description and ensure you are testing a realistic case

luisa-beerboom (changes requested):

tests/system/action/user/test_update.py
vkrasnovyd: I've updated this test and the description of the other one added by me.

While doing that I've found some more tests with the similar setup - with user being a part of a meeting but not a committee (base test perm_group_A_meeting_manage_user_archived_meeting and test_perm_group_A_meeting_manage_user). They should probably be adjusted too, but not within this pull request, right?

luisa-beerboom (approved):

PR #2943, opened 2025-05-14
Change generate-models for models.yml split

luisa-beerboom:
See https://github.com/OpenSlides/openslides-meta/issues/236


hjanott (approved):
Code looks good.
Just want to mention that #2950 is related. An idea would be to retain the order in which the classes are generated.

luisa-beerboom:
Closing bc it was done through #3273

PR #2949, opened 2025-05-16
Allow forwarding motions with attachments

### LINKED ISSUE #2889: Add option to forward motion with/without attachment
<s>This Issue should be worked on after the feature branch motion-forward was merged 
https://github.com/OpenSlides/openslides-backend/tree/feature/motion-forward</s>

**Current behaviour**

If a motion is forwarded then only the motion is forwarded the attechments go missing

**Wanted behaviour**
If you forward a motion then it should be optional to also forward the motion with the attachments.

The backend should have a boolean to allow the copying of files if a motion is forwarded

If a motion is forwarded with the files then...
- if the files are only in the meeting: the files should be copied to the other meeting and the motion should reference them 
- if the files are orga-published then the files should not be copied but referenced accordingly

vkrasnovyd:
Closes #2889


luisa-beerboom:
Alternatively use actions instead of writing directly into events, write a new internal one for cloning a mediafile into another meeting if necessary, that's going to be a lot easier and safer and won't require such a massive test.

openslides_backend/action/actions/motion/base_create_forwarded.py
```diff
@@ -423,218 +417,199 @@ def set_origin_ids(self, instance: dict[str, Any]) -> None:
...
-    ) -> dict[str, dict[int, dict[str, Any]]] | None:
+    def duplicate_mediafiles(self, action_data: ActionData) -> None:
         """
-        Builds a collections dictionary with the models to be created or updated.
+        Duplicates the meeting-wide mediafile models that must be forwarded to
+        the target motion along with the corresponding mediafiles in the
+        mediaservice.
+
+        Builds meeting_mediafiles replace map for each meeting where the motions
+        are forwarded.
```
luisa-beerboom, line 427: You seem to be doing this not only for the motions called in this action, but also for the amendments of those motions. These aren't necessarily forwarded even if `with_amendments` is set, however. To be forwarded an amendment also needs to be in a state with `allow_amendment_forwarding` set. I don't see you checking for that in this function, so I assume that if an amendment doesn't have this kind of state, you'll still forward the mediafiles, causing the mediafiles to be ported even if the amendment they are attached to isn't.

Also, you call this function in perform. Please take note that this is a base class for both the main action `motion.create_forwarded` and the sub-action `motion.create_forwarded_amendment`. So this function is going to be called both in the main action and in every single call of `create_forwarded_amendment` (potentially once for every call of the main actions `update_instance`).
To eliminate the chance of accidentally duplicating mediafiles multiple times into the same meeting, perhaps move this method to the create_forwarded and simply pass the map to the `create_forwarded_amendment` as a parameter when calling it. Alternatively you could leave the function here and just create the map for the main action motions in `create_forwarded`, pass it to the amendment sub-action, have it update the map with all the additional mediafiles it created with this same function and then have it return the map as a return object that the `create_forwarded` action can use for the next `create_forwarded_amendment` call. The latter option also gets rid of the requirement to check for that state, since the amendments given to `create_forwarded_amendment` are already filtered to be the ones that should be forwarded.

luisa-beerboom (changes requested):
Did not look at the tests yet, comments so far are purely on the function code.

The `duplicate_mediafile` function is pretty long, perhaps look into splitting some parts off into concisely named helper functions

tests/system/action/motion/test_create_forwarded.py
```diff
@@ -1675,3 +1678,1099 @@ def test_amendment_forwarding_different_states(self) -> None:
...
+                "meeting_id": 2,
+                "mediafile_id": 19,
+                "is_public": True,
+                "attachment_ids": ["motion/19"],
+            },
+            "meeting_mediafile/35": {
+                "meeting_id": 2,
+                "mediafile_id": 21,
+                "is_public": True,
+                "attachment_ids": ["motion/21", "motion/20"],
```
vkrasnovyd, line 2772: @luisa-beerboom is it possible and reasonable to sort the value in the `attachment_ids` field of the new meeting_mediafiles after creating the models so the attachments are listed in ascending order (by id)? If yes, what is the correct place for this logic? Currently the amendments are listed before their lead motions which seems to be not intuitive to me

luisa-beerboom: I wouldn't say there is much of a necessity to keep this sorted. Based on my knowledge I can say: As far as the client is concerned this field specifically is not even used. We only have it because our system requires back relations. Furthermore even if it was, it's pretty established that relation lists are unordered and that if the client wants things ordered it'll have to do so itself, with the backend only aiding that when necessary through weight fields to allow specific ordering. Any other sorting of relation lists in the backend is usually coincidental.

But just for the sake of answering the question:
There is no standard way of doing that as back-relation lists are usually generated automatically by the relation handling in the backend and the event handling in the datastore. If you really wanted it sorted, the most practical way is probably to overwrite key parts of the relation handling, which is generally not recommended because of the instability (specifically: ensure that the write events generated for the `meeting_mediafile` are replaced with entire list replacement events containing the new order).
Unless that is done you can generally expect lists that were calculated as back-relations to be ordered according to the order in which they were processed. Based on that I'd say keeping the motions attachment list sorted while also keeping the meeting_mediafiles attachment list sorted in the same way as before is not all too plausible.

luisa-beerboom (changes requested):

openslides_backend/action/actions/motion/create_forwarded_amendment.py
```diff
@@ -37,15 +38,32 @@ class MotionCreateForwardedAmendment(BaseMotionCreateForwarded):
...
     )
 
+    def perform(
+        self, action_data: ActionData, user_id: int, internal: bool = False
+    ) -> tuple[WriteRequest | None, ActionResults | None]:
+        amendment_data = action_data.pop("amendment_data", [])  # type: ignore[attr-defined]
+        self.forwarded_attachments = action_data.pop("forwarded_attachments", {})  # type: ignore[attr-defined]
+        self.meeting_mediafile_replace_map = action_data.pop(  # type: ignore[attr-defined]
+            "meeting_mediafile_replace_map", {}
+        )
```
luisa-beerboom, line 48: You could avoid all these `type: ignore`s by redefining `action_data` to be a single item list containing your dict.
Think `[{"amendment_data": [...], "forwarded_attachments": {...}, "meeting_mediafile_replace_map": {...}}]` instead of `{"amendment_data": [...], "forwarded_attachments": {...}, "meeting_mediafile_replace_map": {...}}`.
You could extend `MotionCreateForwardedAmendment` with the `SingularActionMixin` or add a check in the validate_instance to check if only one item is actually sent, then edit the `get_updated_instances` function to destruct your special format into the version of action data you actually want: Saving the contents of the `forwarded_attachments` and `meeting_mediafile_replace_map` keys in their respective attributes and only returning `amendment_data` as the actual instance list.

You'd have to rewrite the schema of course, nesting everything in your dict format, but I think that is doable. Alternatively you could still do all this in perform, then you wouldn't have to edit the schema.

In that case, if you want `MotionCreateForwardedAmendment` to still return a result_element, you'll need to use the `@original_instances` decorator on your overwritten `get_updated_instances` method.

luisa-beerboom (changes requested):

PR #2950, opened 2025-05-16
Lengthen migrate thread join timeout

luisa-beerboom:
Background: Recently there have been cases of failing tests in PRs because the threads in the code took too long to write data. I changed the timeout to be bigger, so this will perhaps be less common.

@m-schieder I would like you to tell me if it is okay if migration requests can potentially take 0.6 seconds longer

### DIFF
--- a/openslides_backend/migrations/migration_handler.py
+++ b/openslides_backend/migrations/migration_handler.py
@@ -13,7 +13,7 @@ from ..shared.interfaces.services import Services
 from . import MigrationWrapper
 
 # Amount of time that should be waited for a result from the migrate thread before returning an empty result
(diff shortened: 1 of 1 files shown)


hjanott (approved):
Just out of curiosity. Is there knowledge of why the delay increased or with which PR this got to be a problem?

luisa-beerboom:
> Just out of curiosity. Is there knowledge of why the delay increased or with which PR this got to be a problem?

The PR where I noticed this was the split models.yml PR. On the most recent commit, the tests had to be run 5 times before they were successful. I found nothing in the PR which would've caused this, however. Code-wise it only changes the generate-models script and the order within the models.py after all.

PR #2951, opened 2025-05-16
Make poll delete inform vote service on success

### LINKED ISSUE #2948: Inform the vote-service about deleted active polls
When a poll gets stopped, the backend sends a "clear" message to the vote service.

But when a active (started) poll gets deleted, the backend does currently not inform the vote-service about it. So the vote service does not know, that the poll does not exist anymore and thinks, that it is active forever.

Could the backend send a "clear" request, when an active poll gets deleted?

See also: https://github.com/OpenSlides/openslides-autoupdate-service/issues/1199

luisa-beerboom:
Closes #2948


hjanott (approved):
lgtm

PR #2984, opened 2025-06-12
Complete agenda_item tests

hjanott:
This PR will complete the missing data and create other fixes for the relational scheme in the agenda_item tests.
Also moves list sorting to database.

### DIFF
--- a/openslides_backend/services/database/database_writer.py
+++ b/openslides_backend/services/database/database_writer.py
@@ -1,7 +1,7 @@
 import threading
 from typing import Any
 
-from psycopg import Connection, rows, sql
+from psycopg import Connection, adapt, rows, sql
 from psycopg.errors import (
     CheckViolation,
     GeneratedAlways,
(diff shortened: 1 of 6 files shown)


openslides_backend/services/database/database_writer.py
```diff
@@ -547,28 +548,29 @@ def execute_sql(
...
+            in_table_block = False
             with open("meta/dev/sql/schema_relational.sql") as f:
                 for line in f:
                     # search only in the table block to prevent finding duplicates of other tables
-                    if "CREATE TABLE" in line and table in line:
-                        record = True
-                    elif record and constraint_name in line:
+                    if f"CREATE TABLE {table}" in line:
+                        in_table_block = True
+                    elif in_table_block and constraint_name in line:
```
luisa-beerboom: 1. Table name could still be substring of another table string if the end is not also checked against an empty space
2. There is no guarantee that the table create lines are always going to have the format `CREATE TABLE <table_name>`. In the future it might be necessary to use something like `CREATE TABLE IF NOT EXISTS`. Maybe use regex?
3. The `constraint_name` line has the same problem as the table line

luisa-beerboom (changes requested):

PR #2999, opened 2025-06-23
Add superadmin self-inactive setting check in user imports

### LINKED ISSUE #2993: Add superadmin self-inactive setting check in user imports
If a superadmin tries to set himself inactive via the user imports, there is no error message in the row. This causes the error message for this case only to be thrown at the end when the import action calls the user.update.

The user json_upload and import actions should all check for this and set an error on the is_active field when this is attempted.

vkrasnovyd:
Closes #2993


tests/system/action/user/test_participant_json_upload.py
```diff
@@ -1132,6 +1132,37 @@ def test_json_upload_update_locked_out_on_superadmin_error(self) -> None:
...
 
+    def test_json_upload_perm_superadmin_self_set_inactive_error(self) -> None:
+        """SUPERADMIN may not set himself inactive."""
+        self.create_committee()
+        response = self.request(
+            "participant.json_upload",
+            {
+                "meeting_id": 1,
+                "data": [
+                    {"username": "admin", "is_active": "False", "groups": ["admin"]}
```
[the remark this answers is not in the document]
vkrasnovyd: For participants import either group in the payload or default_group in the meeting is required: https://github.com/OpenSlides/openslides-backend/blob/374ef9ee854f5b2321348bf74d6e468f50e46d36/openslides_backend/action/actions/user/participant_json_upload.py#L83-L91

So, without the group test fails

tests/system/action/user/test_participant_json_upload.py
```diff
@@ -1134,13 +1134,12 @@ def test_json_upload_update_locked_out_on_superadmin_error(self) -> None:
 
     def test_json_upload_perm_superadmin_self_set_inactive_error(self) -> None:
         """SUPERADMIN may not set himself inactive."""
+        self.set_models({"group/1": {"default_group_for_meeting_id": 1}})
```
hjanott, line 1137: Yes that's one way to do it. I prefer to do it in the `setUp` method so that it exists for all other tests too.
What I would even prefer more because of the upcoming relational db, is to set the relation inside the meeting itself. The field should be called `default_group_id`.

tests/system/action/user/test_participant_json_upload.py
```diff
@@ -218,6 +220,7 @@ def test_json_upload_without_names_error(self) -> None:
         }
 
     def test_json_upload_no_default_group(self) -> None:
+        self.set_models({"group/1": {"default_group_for_meeting_id": None}})
```
hjanott, line 223: Good thinking. I'm wondering if we could throw this test out, since this would be an illegal state.
In the rel db branch it is gonna leave soon.

PR #3015, opened 2025-07-07
Rename user field 'guest' to 'external'

### LINKED ISSUE #3005: Rename user field "guest" to "external"
To have a consistent naming of the field it should be renamed from "guest" to "external". In the im-/export of users (account and participant) the raw field names are used and occur in the client. This brings confusion to users who use the import.

vkrasnovyd:
Closes #3005 
Needs https://github.com/OpenSlides/openslides-meta/pull/279


openslides_backend/migrations/migrations/0067_rename_guest_field.py
```diff
@@ -0,0 +1,17 @@
+from datastore.migrations import RenameFieldMigration
+
+
+class Migration(RenameFieldMigration):
```
luisa-beerboom, line 4: The RenameFieldMigration is based on the old EventMigration system that isn't supposed to be used anymore. I think this will fail in production at the latest. Please write this Migration based on the BaseModelMigration class

luisa-beerboom (changes requested):

reiterl (approved):
Migration in place. Okay.

PR #3016, opened 2025-07-07
Refactor permission_helper methods

### LINKED ISSUE #2995: Improve get_shared_committee_management_levels
In `permission_helper.py` we have the function `get_shared_committee_management_levels`. 
`expected_level` is unused.
This uses a condition check that is always false because a field is not retrieved from datastore.
It should be investigated if enabling this causes bugs.
It should be discussed if the function can reside in the user scope mixin and if the `self.permstore` can be used as this function is only used there.
The doc string should tell about the semantics of returned integers.

vkrasnovyd:
Closes #2995 
- Removed unused 'expected_level: CommitteeManagementLevel'. Found it in 2 more methods and removed there too;
- Enabling OML check doesn't cause test failures => enabled it;
- As discussed with Luisa, although this method is currently used only in the user_scope_mixin, it logically belongs to this file => no changes.
- Added description of the return value to the docstring.

### DIFF
--- a/openslides_backend/permissions/permission_helper.py
+++ b/openslides_backend/permissions/permission_helper.py
@@ -152,8 +152,7 @@ def has_committee_management_level(
     committee_id: int,
 ) -> bool:
     """
-    Checks whether a user has CommitteeManagementLevel 'can_manage'
-    in the given committee.
+    Checks whether a user is committee manager in the given committee.
     """
     if user_id > 0:
         user = datastore.get(
--- a/openslides_backend/permissions/permission_helper.py
+++ b/openslides_backend/permissions/permission_helper.py
@@ -183,7 +183,11 @@ def get_shared_committee_management_levels(
     user_id: int,
     committee_ids: list[int],
 ) -> list[int]:
-    """Checks wether a user has CommitteeManagementLevel 'can_manage'."""
+    """
+    Checks wether a user has CommitteeManagementLevel 'can_manage'.
+    Returns: list with the intersection of the given committee_ids and
+    ids of the committees where user has management rights.
+    """
     if user_id > 0:
         user = datastore.get(
             fqid_from_collection_and_id("user", user_id),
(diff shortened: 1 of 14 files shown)


openslides_backend/permissions/permission_helper.py
```diff
@@ -111,13 +110,11 @@ def has_organization_management_level(
...
     user_id: int,
-    expected_level: CommitteeManagementLevel,
     committee_ids: list[int],
 ) -> list[int]:
     """
-    Checks whether a user has the minimum necessary
-    CommitteeManagementLevel for the committees in the list and
-    returns the ids of all that fail.
+    Checks whether a user has CommitteeManagementLevel 'can_manage' for the committees
+    in the list and returns the ids of all that fail.
```
hjanott, line 117:
```suggestion
    Checks whether a user committee manager for the committees
    in the list and returns the ids of all that fail.
```

openslides_backend/permissions/permission_helper.py
```diff
@@ -152,10 +149,12 @@ def get_failing_committee_management_levels(
...
     datastore: DatastoreService,
     user_id: int,
-    expected_level: CommitteeManagementLevel,
     committee_id: int,
 ) -> bool:
-    """Checks whether a user has the minimum necessary CommitteeManagementLevel"""
+    """
+    Checks whether a user has CommitteeManagementLevel 'can_manage'
+    in the given committee.
+    """
```
hjanott, line 157:
```suggestion
    """
    Checks whether a user is committee manager in the given committee.
    """
```

openslides_backend/permissions/permission_helper.py
```diff
@@ -182,14 +181,17 @@ def has_committee_management_level(
...
     user_id: int,
-    expected_level: CommitteeManagementLevel,
     committee_ids: list[int],
 ) -> list[int]:
-    """Checks wether a user has the minimum necessary CommitteeManagementLevel"""
+    """
+    Checks wether a user has CommitteeManagementLevel 'can_manage'.
+    Returns: list with the intersection of the given committee_ids and
+    ids of the committees where user has management rights.
+    """
```
hjanott, line 190:
```suggestion
    """
    Checks whether a user is manager in the given committees.
    Returns a list where this is the case or all if the user is orga admin.
    """
```

hjanott (approved):

PR #3025, opened 2025-07-14
Update unique mediafile title suffixes

### LINKED ISSUE #3024: Fix/change behaviour for forward motions with attachments 2
**Describe the bug**
Follow up from https://github.com/OpenSlides/openslides-backend/issues/3008
```
Expected behavior
If the exact same name for the file already exists in the meeting then the file name should get a number in brackets (#)
```
Currently the files name will be ```bla bla (#0000)```

**Expected behavior**

files name should be ```bla bla (0000)```

The # was used to symbolize a number.

vkrasnovyd:
Closes #3024 

1. As requested, removed hashtag sign from the unique suffixes.
2. In the example of the desired behaviour numeration of the copies starts from 1, not from 2 like in the previous fix. Updated that.
3. Updated tests. Made sure that if the origin name already contains a suffix, it is treated as the part of the origin title.

### DIFF
--- a/openslides_backend/action/actions/mediafile/duplicate_to_another_meeting.py
+++ b/openslides_backend/action/actions/mediafile/duplicate_to_another_meeting.py
@@ -117,8 +117,8 @@ class MediafileDuplicateToAnotherMeetingAction(MediafileCreateMixin, CreateActio
             ).values()
         }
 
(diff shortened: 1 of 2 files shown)


hjanott (approved):
lgtm. ;)

PR #3027, opened 2025-07-14
Change database checker to no longer consider default fields required

### LINKED ISSUE #3020: Manually creating projectors causes database consistency check to fail
**Describe the bug**
Manually creating a projector in the client, without setting is_internal will cause the `check_database_all` presenter to return a `broken` verdict. This happens because the client sends `null` as the value of one field and the backend doesn't stop this.

**How To Reproduce**

First
```
[{"action":"projector.create","data":[{"meeting_id":2,"name":"pro jek tor","is_internal":null}]}]
```
then
```
[{"presenter":"check_database_all","data":{}}]
```

**Response from backend**
For first request
```
{"status_code": 200, "success": true, "message": "Actions handled successfully", "results": [[{"id": 5, "sequential_number": 3}]]}
```
For database check
```
[{"ok": false, "errors": "\tprojector/5: Missing fields is_internal"}]
```
**Expected behavior**
The `projector.create` action should not allow `None` as a field value for `is_internal`

Ergo:
1. ensure the backend-action does not allow `None` for the `is_internal` field.
2. write a migration to fix any old inconsistencies caused by this

It should also be ensured that projector.update doesn't have the same problem

luisa-beerboom:
Closes #3020

TODO:

- [ ] What should be done about the fields that the database code is assuming are always filled (like organization/default_language)
- [ ] Remove the extra code from generate_models

### DIFF
--- a/cli/generate_models.py
+++ b/cli/generate_models.py
@@ -164,6 +164,8 @@ class Model(Node):
             if field.get("calculated"):
                 continue
             self.attributes[field_name] = Attribute(field)
+            if not field.get("required") and field.get("default") is not None:
+                print(f"{self.collection}/{field_name}: {field.get('default')}")
 
     def get_code(self) -> str:
         verbose_name = " ".join(self.collection.split("_"))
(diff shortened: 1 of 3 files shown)


luisa-beerboom:
Closing bc PR is very old, will probably be hard to update and the lack of attention to the question in the issue may point to lack in necessity.
If this is still relevant, a new PR should be created.

PR #3041, opened 2025-07-15
Add fields for live voting

bastianjoel:
Needs https://github.com/OpenSlides/openslides-meta/pull/282

### DIFF
--- a/openslides_backend/action/actions/poll/create.py
+++ b/openslides_backend/action/actions/poll/create.py
@@ -70,6 +70,7 @@ class PollCreateAction(
             "backend",
         ],
         additional_optional_fields={
+            "live_voting_enabled": {"type": "boolean"},
             "publish_immediately": {"type": "boolean"},
             "amount_global_yes": decimal_schema,
             "amount_global_no": decimal_schema,
@@ -92,6 +93,10 @@ class PollCreateAction(
             )
             if not organization.get("enable_electronic_voting"):
                 raise ActionException("Electronic voting is not allowed.")
+
+        # check named and live_voting_enabled
+        if instance["type"] != Poll.TYPE_NAMED and "live_voting_enabled" in instance and instance["live_voting_enabled"]:
+            raise ActionException("live_voting_enabled only allowed for named polls.")
 
         # check entitled_group_ids and analog
         if instance["type"] == Poll.TYPE_ANALOG and "entitled_group_ids" in instance:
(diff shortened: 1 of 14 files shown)


Elblinator:
> new meeting settings for motions: label: "Set live voting enabled by default"; hint: "Only available for nominal voting"; position in settings: after "Default voting type"; default value: diasbled

This setting is needed from the backend

Please add this setting sth sth "live_voting_enabled" default is false

Elblinator:
https://github.com/OpenSlides/openslides-client/pull/5226#discussion_r2218300071

isMotionPoll needs to be added

openslides_backend/action/actions/poll/create.py
```diff
@@ -70,6 +70,7 @@ class PollCreateAction(
             "backend",
         ],
         additional_optional_fields={
+            "live_voting_enabled": {"type": "boolean"},
```
luisa-beerboom, line 73: This should be in the `optional_fields` list as it is a property of the model

openslides_backend/action/actions/poll/update.py
```diff
@@ -53,6 +53,7 @@ class PollUpdateAction(
         ],
         additional_optional_fields={
             "publish_immediately": {"type": "boolean"},
+            "live_voting_enabled": {"type": "boolean"},
```
luisa-beerboom, line 56: This should be in the optional_fields list as it is a property of the model

openslides_backend/action/actions/poll/update.py
```diff
@@ -114,6 +116,17 @@ def update_instance(self, instance: dict[str, Any]) -> dict[str, Any]:
                 "Following options are not allowed in this state and type: "
                 + ", ".join(not_allowed)
             )
+
+        # check named and live_voting_enabled
+        if instance.get("live_voting_enabled") and not (
+            poll["type"] == Poll.TYPE_NAMED
+            and poll.get("content_object_id")
+            and collection_from_fqid(poll["content_object_id"]) == "motion"
```
luisa-beerboom, line 124: `content_object_id` is a required field, therefore checking if `poll.get("content_object_id")` should be unnecessary

openslides_backend/action/actions/poll/create.py
```diff
@@ -83,6 +84,10 @@ def update_instance(self, instance: dict[str, Any]) -> dict[str, Any]:
         action_data = []
 
         state_change = self.check_state_change(instance)
+        is_motion_poll = (
+            instance.get("content_object_id")
```
luisa-beerboom, line 88: `content_object_id` is a required field, therefore checking if `instance.get("content_object_id")` should be unnecessary

luisa-beerboom (changes requested):

PR #3049, opened 2025-07-23
Rel-DB Gender tests

vkrasnovyd:
* For all actions removed relational field list from model_create and refactored short dictionaries that can fit into one line;
*  Tests for create and update were running even before the changes. In tests for delete action user.username was missing, added it.

### DIFF
--- a/tests/system/action/gender/test_delete.py
+++ b/tests/system/action/gender/test_delete.py
@@ -10,35 +10,25 @@ class GenderDeleteActionTest(BaseActionTestCase):
     def create_data(self) -> None:
         self.set_models(
             {
-                ONE_ORGANIZATION_FQID: {"gender_ids": [self.gender_id]},
-                "user/20": {"gender_id": 1},
-                "user/21": {"gender_id": self.gender_id},
+                ONE_ORGANIZATION_FQID: {"name": "test_organization1"},
+                "user/20": {"username": "human", "gender_id": 1},
+                "user/21": {"username": "fairy", "gender_id": self.gender_id},
                 "gender/1": {
                     "id": 1,
                     "name": "male",
                     "organization_id": 1,
-                    "user_ids": [20],
                 },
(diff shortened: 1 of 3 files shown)


tests/system/action/gender/test_delete.py
```diff
@@ -10,35 +10,25 @@ class GenderDeleteActionTest(BaseActionTestCase):
     def create_data(self) -> None:
         self.set_models(
             {
-                ONE_ORGANIZATION_FQID: {"gender_ids": [self.gender_id]},
-                "user/20": {"gender_id": 1},
-                "user/21": {"gender_id": self.gender_id},
+                ONE_ORGANIZATION_FQID: {"name": "test_organization1"},
+                "user/20": {"username": "human", "gender_id": 1},
+                "user/21": {"username": "fairy", "gender_id": self.gender_id},
```
hjanott, line 15: It would be better to choose a username different from `gender_name`.

tests/system/action/gender/test_delete.py
```diff
@@ -10,35 +10,25 @@ class GenderDeleteActionTest(BaseActionTestCase):
...
 
     def test_delete_correctly(self) -> None:
         self.create_data()
-        self.set_models(
-            {
-                "gender/6": {
-                    "name": "dragon",
-                    "organization_id": 1,
-                },
-                ONE_ORGANIZATION_FQID: {"gender_ids": [1, self.gender_id, 6]},
```
hjanott, line 39: It's okay to delete the organization. You could also delete the other occurrences, since the org is already set by the `setUp` method.

hjanott (approved):
Just some tiny remarks.

tests/system/action/gender/test_delete.py
vkrasnovyd: Replaced with unique value.

PR #3065, opened 2025-07-29
Make meeting is_admin check work for committee admins

### LINKED ISSUE #3045: Allow committee admins to edit in sub-committees meetings
**Describe the bug**
Some meeting-internal actions cannot be taken by committee admins when accessing meetings of subcommittees. Namely group.update in case of external ids being set and meeting.update.

**How To Reproduce**
Using the example of the group id:
Try to set an external_id of a group without any admin rights except a higher ranked committees management

```
[
  {
    "action": "group.update",
    "data": [
      {
        "id": 12,
        "name": "Delegates",
        "external_id": "Delegates-Com-3"
      }
    ]
  }
]
```

**Response from backend**
`Missing permission: Not admin of this meeting`

**Expected behavior**
It should work.

**Additional context**
I looked into it a bit and the is_admin function in the permission_helper.py doesn't seem to consider committee admin rights at all. The meeting.update also uses that function, so it should be the cause there as well
Necessary for https://github.com/OpenSlides/openslides-client/pull/5102

luisa-beerboom:
Closes #3045

### DIFF
--- a/tests/system/action/meeting/test_update.py
+++ b/tests/system/action/meeting/test_update.py
@@ -587,8 +587,8 @@ class MeetingUpdateActionTest(BaseActionTestCase):
         )
 
     def test_update_group_d_committee_parent_permissions(self) -> None:
-        self.create_committee(59)
         self.create_meeting()
+        self.create_committee(59)
         self.create_committee(60, parent_id=59)
         self.user_id = self.create_user("user")
         self.login(self.user_id)
--- a/tests/system/action/meeting/test_update.py
+++ b/tests/system/action/meeting/test_update.py
@@ -590,9 +590,8 @@ class MeetingUpdateActionTest(BaseActionTestCase):
         self.create_meeting()
         self.create_committee(59)
         self.create_committee(60, parent_id=59)
-        self.user_id = self.create_user("user")
-        self.login(self.user_id)
-        self.set_committee_management_level([59], self.user_id)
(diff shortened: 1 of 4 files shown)


tests/system/action/meeting/test_update.py
```diff
@@ -586,6 +586,30 @@ def test_update_group_d_permissions(self) -> None:
             },
         )
 
+    def test_update_group_d_committee_parent_permissions(self) -> None:
+        self.create_committee(59)
+        self.create_meeting()
+        self.create_committee(60, parent_id=59)
```
hjanott, line 592:
```suggestion
        self.create_meeting()
        self.create_committee(59)
        self.create_committee(60, parent_id=59)
```

tests/system/action/meeting/test_update.py
```diff
@@ -586,6 +586,30 @@ def test_update_group_d_permissions(self) -> None:
...
+        self.user_id = self.create_user("user")
+        self.login(self.user_id)
```
hjanott, line 594: Why not use `self.set_organization_management_level(None)`?

hjanott (approved):
logtm. Just one minor thing and an interest question.

tests/system/action/meeting/test_update.py
luisa-beerboom: done

PR #3070, opened 2025-07-30
[rel-db] Complete meeting user tests

hjanott:
Also remaining "Etc/UTC" to "UTC".

### DIFF
--- a/openslides_backend/action/actions/assignment_candidate/mixins.py
+++ b/openslides_backend/action/actions/assignment_candidate/mixins.py
@@ -40,12 +40,10 @@ class PermissionMixin(Action):
         )
         if meeting_user_id in (user.get("meeting_user_ids") or []):
             permission = Permissions.Assignment.CAN_NOMINATE_SELF
-            if not has_perm(self.datastore, self.user_id, permission, meeting_id):
-                missing_permission = permission
         else:
             permission = Permissions.Assignment.CAN_NOMINATE_OTHER
-            if not has_perm(self.datastore, self.user_id, permission, meeting_id):
-                missing_permission = permission
+        if not has_perm(self.datastore, self.user_id, permission, meeting_id):
+            missing_permission = permission
 
         if missing_permission:
             raise MissingPermission(missing_permission)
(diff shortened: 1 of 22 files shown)


tests/system/action/meeting_user/test_set_data_delegation.py
```diff
@@ -7,68 +7,36 @@
...
-                "committee/1": {"meeting_ids": [222]},
-                "meeting/222": {
-                    "name": "Meeting222",
-                    "is_active_in_organization_id": 1,
-                    "committee_id": 1,
+                "meeting/222": {"default_group_id": 11},
+                "meeting/225": {"default_group_id": 12},
+                "group/1": {
+                    "name": "groupy",
+                    "meeting_id": 222,
```
vkrasnovyd, line 18: Since these groups 1, 2, 11 and 12 are pretty simple and similar to the ones created by `self.create_meeting`, I'd suggest using the existing ones instead of them: 222, 224, 225 and 227. Not only does it reduce the amount of requests to db, but also makes tests with meetings created with `self.create_meeting` more predictable. What do you think about that?

There are also multiple re-definitions of the simple groups in meeting_user.test_create and test_update (group/21).

tests/system/action/meeting_user/test_set_data_delegation.py
```diff
@@ -164,11 +123,8 @@ def test_reverse_delegation(
     ) -> None:
         self.set_models(
             {
-                "meeting_user/14": {"vote_delegations_from_ids": [13]},
-                "meeting_user/13": {
-                    "vote_delegated_to_id": 14,
-                    "vote_delegations_from_ids": [],
-                },
+                "meeting_user/12": {"vote_delegated_to_id": None},
```
vkrasnovyd, line 126:
```suggestion
                "meeting_user/14": {"vote_delegated_to_id": None},
```

vkrasnovyd (changes requested):
Looks good. Just one typo and a suggestion.

tests/system/action/meeting_user/test_set_data_delegation.py
hjanott: No I don't think that's right. `meeting_user/14` has no vote delegated in the `setUp`. There 12 is delegating to 13 but the other direction wasn't rewritten before this change. For the action it was enough that it seemed like 13 wasn't receiving any delegations. Hence, just the back relation was emptied.

PR #3096, opened 2025-08-13
Split poll manage rights from content object management rights

### LINKED ISSUE #3095: Split poll manage permissions from content_object permissions
See OpenSlides/openslides-client#5120

The motion poll and assignment poll permissions should no longer be subordinate to the motion and assignment model can_manage rights, but should rather be top-level permissions.

There probably needs to be a migration to ensure users don't loose any implicit poll rights granted via can_manage by this switch

luisa-beerboom:
Closes #3095
Needs OpenSlides/openslides-meta#302

Don't merge until client is done


openslides_backend/migrations/migrations/0068_split_off_poll_permissions.py
```diff
@@ -0,0 +1,44 @@
...
+    This migration splits motion.can_manage_polls away from motion.can_manage
+    and gives the new permission assignment.can_manage_polls to any group with assignment.can_manage.
+    """
+
+    target_migration_index = 69
+
+    perms = ["motion.can_manage", "assignment.can_manage"]
+
+    def migrate_models(self) -> list[BaseRequestEvent]:
+        groups = self.reader.get_all("group", ["permissions"])
```
hjanott, line 18: Maybe filter for all groups with at least one permission set? Or is that a rare case?

openslides_backend/action/actions/meeting/create.py
```diff
@@ -148,6 +148,7 @@ def update_instance(self, instance: dict[str, Any]) -> dict[str, Any]:
                 "permissions": [
                     Permissions.AgendaItem.CAN_MANAGE,
                     Permissions.Assignment.CAN_MANAGE,
+                    Permissions.Assignment.CAN_MANAGE_POLLS,
                     Permissions.ListOfSpeakers.CAN_MANAGE,
                     Permissions.Mediafile.CAN_MANAGE,
                     Permissions.Meeting.CAN_SEE_FRONTPAGE,
```
hjanott, line 154: What about motion can manage polls?

hjanott (commented):
Do any of the motion docs need to be changed?

openslides_backend/action/actions/meeting/create.py
luisa-beerboom: Already exists and was merely moved

luisa-beerboom:
> Do any of the motion docs need to be changed? (Also tests failing)

No motion actions were changed, so why do you think the docs might need to be changed

luisa-beerboom:
> > No motion actions were changed, so why do you think the docs might need to be changed
> 
> Since the permissions of motion got changed, it shouldn't be possible to update a motion with a poll with only motion.can_manage any-more. (But maybe I am misinterpreting the purpose of this permission?) That's why I am thinking that the permissions section of those actions may need to be updated. Coming to think of it. We would also need a test for this new permission context.

I don't think so. 
Poll manage is for managing polls, updating the related motion/assignment changes nothing about the poll.
For the model delete actions? Maybe there's a slight case to be made there. After all, the deletion of the content-object should also delete the poll, but even then I don't think it is out of the ordinary to expect this to work for a regular motion/assignment admin.

hjanott (approved):
lgtm then.

PR #3097, opened 2025-08-14
[rel-db] Complete motion-related tests (Part 1)

vkrasnovyd:
**Deleted tests:**
* motion_workflow.test_delete:
    * `test_delete_with_states`, `test_delete_with_first_state`: after fixing test_data inconsistencies they became the duplicates of `test_delete_correct`, since `first_state_id` is a required unique field of `motion_workflow`
    * `test_delete_fail_case_default_2`: `motions_default_workflow_id` is a required field of the meeting, so described validation is unreachable in the db with consistent data. Also deleted the corresponding check in the action code
    * `test_delete_last_workflow`: looks like it was the duplicate of `test_delete_fail_case_default_2` and describes the case not possible in consistent db

**Renamed** `test_delete_fail_case_default_1` and `test_delete_fail_case_default_3` to make the names more descriptive.

**TODO** (together with similarly failing meeting and motion tests): `motion_change_recommendation.delete test_delete_correct`. Fails on self.events.extend(events) (updating history step).

3 more collections (`motion_editor`, `motion_submitter`, `motion_working_group_speaker`) will be handled in a separate PR, since their test classes extend not BaseActionTestCase directly but it's extended version. It would make sence to handle them together separately from the other collections.

### DIFF
--- a/openslides_backend/action/actions/motion_workflow/delete.py
+++ b/openslides_backend/action/actions/motion_workflow/delete.py
@@ -1,4 +1,4 @@
-from typing import Any, cast
+from typing import Any
 
 from ....models.models import MotionWorkflow
 from ....permissions.permissions import Permissions
@@ -43,12 +43,6 @@ class MotionWorkflowDeleteAction(DeleteAction):
             if instance["id"] == meeting.get("motions_default_amendment_workflow_id"):
                 raise ActionException(
                     "You cannot delete the workflow as long as it is selected as default workflow for new amendments in the settings. Please set another workflow as default in the settings and try to delete the workflow again."
-                )
-
-            workflow_ids = cast(list[int], meeting.get("motion_workflow_ids"))
-            if len(workflow_ids) == 1:
-                raise ActionException(
-                    "You cannot delete the last workflow of a meeting."
                 )
 
         return instance
--- a/tests/system/action/motion_state/test_create.py
+++ b/tests/system/action/motion_state/test_create.py
@@ -1,5 +1,3 @@
-from typing import Any
-
 from openslides_backend.permissions.permissions import Permissions
 from tests.system.action.base import BaseActionTestCase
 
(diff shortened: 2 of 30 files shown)


tests/system/action/motion_state/test_create.py
```diff
@@ -114,103 +78,71 @@ def test_create_as_new_first_state_of_second_workflow(self) -> None:
...
-                "name": "test_Xcdfgee",
-                "workflow_id": 42,
-            },
+            {"name": "test_Xcdfgee", "workflow_id": 1},
         )
         self.assert_status_code(response, 200)
-        self.assert_model_exists("motion_state/1", {"weight": 1})
+        self.assert_model_exists("motion_state/2", {"weight": 37})
+
         response = self.request(
```
hjanott, line 126: I think one time should be enough.

tests/system/action/motion_state/test_update.py
```diff
@@ -8,173 +6,136 @@ class MotionStateActionTest(BaseActionTestCase):
...
-                "workflow_id": 110,
-                "meeting_id": 1,
-            },
-        }
+        self.set_models(
+            {
+                "motion_workflow/1": {"name": "name_Ycefgee"},
+                "motion_state/1": {"name": "name_srtgb123"},
+            }
+        )
```
hjanott, line 14: Instead of updating these names we could also change the names in the assertion. The flow of actions shouldn't be impacted as far as I can see.

tests/system/action/motion_workflow/test_delete.py
```diff
@@ -3,198 +3,67 @@
...
-        self.assert_status_code(response, 400)
-        self.assert_model_exists("motion_workflow/111")
-        self.assertIn(
-            "You cannot delete the workflow as long as it is selected as default workflow for new motions in the settings.",
-            response.json["message"],
+        self.assertEqual(
+            "Model 'motion_workflow/3' does not exist.", response.json["message"]
         )
 
-    def test_delete_fail_case_default_2(self) -> None:
```
hjanott, line 95: I think it should still be reachable. The database will fail after the events were send. But the check happens before that.

tests/system/action/motion_workflow/test_update.py
```diff
@@ -3,59 +3,42 @@
 
 
 class MotionWorkflowSystemTest(BaseActionTestCase):
+    def setUp(self) -> None:
+        super().setUp()
+        self.create_meeting()
+        self.set_models({"motion_workflow/1": {"name": "name_srtgb123"}})
```
hjanott, line 9: Please check if we actually have to update this.

tests/system/action/motion_state/test_create.py
vkrasnovyd: Removed.

openslides_backend/action/actions/motion_workflow/delete.py
```diff
@@ -45,10 +45,4 @@ def update_instance(self, instance: dict[str, Any]) -> dict[str, Any]:
                     "You cannot delete the workflow as long as it is selected as default workflow for new amendments in the settings. Please set another workflow as default in the settings and try to delete the workflow again."
                 )
 
-            workflow_ids = cast(list[int], meeting.get("motion_workflow_ids"))
-            if len(workflow_ids) == 1:
-                raise ActionException(
-                    "You cannot delete the last workflow of a meeting."
```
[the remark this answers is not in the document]
vkrasnovyd: Discussed. This case is indeed unreachable: the last workflow of the meeting is also its default workflow and we already have an Exception that checks this case.

hjanott (approved):
lgtm.

PR #3110, opened 2025-08-21
[rel-db] Complete tag and theme tests

vkrasnovyd:
Since `theme/1` is set in `BaseSystemTestCase.setUp`, I was able to remove any `set_models` calls for this model.


hjanott (approved):
lgtm.

PR #3115, opened 2025-08-22
[rel-db] cleanup fix base action test case

hjanott:
Adapts to rel-db changes. Makes `set_user_groups` respect existing group relations.


vkrasnovyd (approved):
LGTM and works!

PR #3162, opened 2025-09-30
[rel-DB] Fix different tests

vkrasnovyd:
- Commits 1-3: fixed tests that were failing because of the 1-to-1 not_null trigger.
- Commit 4: while fixing these tests in the `user.test_merge_together.py` file, I've spot quite a bunch of back relations that were set manually. Cleaned up some of them but there are still a lot of them left. Do we want to remove them all too? Probably from the whole directory and in a separate PR? If that's the case, I'll move this commit to a new branch.

### DIFF
--- a/tests/system/action/user/test_merge_together.py
+++ b/tests/system/action/user/test_merge_together.py
@@ -15,7 +15,7 @@ from openslides_backend.shared.patterns import (
     CollectionField,
     fqid_from_collection_and_id,
 )
-from openslides_backend.shared.util import ONE_ORGANIZATION_FQID, ONE_ORGANIZATION_ID
+from openslides_backend.shared.util import ONE_ORGANIZATION_ID
 from tests.system.action.base import BaseActionTestCase
 from tests.system.action.poll.test_vote import BaseVoteTestCase
 from tests.util import Response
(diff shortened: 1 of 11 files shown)


tests/system/action/user/test_merge_together.py
```diff
@@ -801,34 +801,10 @@ def test_with_multiple_delegations(self) -> None:
...
-                    "present_user_ids": [2, 4],
-                    "assignment_ids": [1],
-                },
-                "meeting/4": {
-                    "present_user_ids": [3, 4],
-                    "motion_ids": [1],
-                    "motion_submitter_ids": [1],
-                },
-                "meeting/7": {
-                    "present_user_ids": [2, 3, 4],
```
hjanott, line 795: We shouldn't delete these.

tests/system/action/user/test_merge_together.py
```diff
@@ -1491,10 +1461,6 @@ def test_merge_with_motion_submitters_and_supporters(
                     f"meeting_user/{id_}": {"supported_motion_ids": ids}
                     for id_, ids in motion_ids_per_supporter.items()
                 },
```
hjanott, line 1463: Don't unpack this dict but use it directly.

hjanott (commented):
In general I think that as long as the tests are running we don't need to do the clean-ups now. But it makes sense to do them while we are editing the files anyway. So let's keep your changes here.

tests/system/action/user/test_merge_together.py
```diff
@@ -1169,42 +1187,11 @@ def add_assignment_or_motion_models_for_meetings(
...
-                            next_sub_model_id + len(meeting_user_id_list),
-                        )
-                    ),
-                    "sequential_number": next_model_id,
-                }
                 if collection == "motion":
-                    data[a_or_m_fqid].update({"state_id": meeting_id})
+                    self.create_motion(meeting_id, next_model_id)
+                else:
+                    self.create_assignment(next_model_id, meeting_id)
```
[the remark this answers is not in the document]
vkrasnovyd: Done.

tests/system/action/user/test_merge_together.py
vkrasnovyd: Done.

PR #3191, opened 2025-11-05
Fix topic history entry creation bug

luisa-beerboom:
1. Changes history writing process: It now throws entries written for non-history-models away.
2. Migration to ensure that any mistakenly written history entries are _deleted_.
3. That migrations code is skipped on IMMs, because
   - IMMs are currently only used during meeting imports and meeting imports don't get history data due to it being excluded from the export
   - The filter method used to initially find the broken entries is not implemented for IMMs, breaking the migration code when called in-memory
   - Which will likely not matter for future migrations because that entire IMM code will be yeeted when relDB gets merged into main

### DIFF
--- a/docs/actions/motion_change_recommendation.update.md
+++ b/docs/actions/motion_change_recommendation.update.md
@@ -10,6 +10,8 @@
     internal: boolean;
     type: number;
     other_description: string;
+    line_from: number;
+    line_to: number;
 }
 ```
 
--- a/openslides_backend/action/actions/motion_change_recommendation/update.py
+++ b/openslides_backend/action/actions/motion_change_recommendation/update.py
@@ -1,10 +1,19 @@
+from typing import Any
+
 from openslides_backend.action.mixins.extend_history_mixin import ExtendHistoryMixin
 from openslides_backend.action.util.register import register_action
 
 from ....models.models import MotionChangeRecommendation
 from ....permissions.permissions import Permissions
+from ....services.datastore.commands import GetManyRequest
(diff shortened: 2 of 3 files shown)


tests/system/migrations/test_0070_remove_mistakenly_created_history_entries.py
```diff
@@ -0,0 +1,1179 @@
...
+            "meta_deleted": True,
+            "entries": ["Topic created"],
+            "original_model_id": "topic/4",
+            "position_id": 2,
+        },
+    )
+
+
+# TODO: Write tests at least checking these cases:
+# - Test for other not-allowed models
```
hjanott: Remaining TODOs.

tests/system/migrations/test_0070_remove_mistakenly_created_history_entries.py
```diff
@@ -0,0 +1,1179 @@
...
+        return self.get_create_event(fqid, fields)
+
+    def get_write_user_event(
+        self, id_: int, username: str | None = None
+    ) -> dict[str, Any]:
+        fqid = f"user/{id_}"
+        self.context[fqid] = fields = {"id": id_, "username": username or f"user{id_}"}
+        return self.get_create_event(fqid, fields)
+
+    def get_write_topic_events(self, id_: int, meeting_id: int) -> list[dict[str, Any]]:
```
hjanott: Could be generalized with motion and assignment creation.

PR #3192, opened 2025-11-06
Add an orga setting for disable forward with attachments

### LINKED ISSUE #3189: New setting for attachments for motion forwarding
Client issue: https://github.com/OpenSlides/openslides-client/issues/5540

A new setting is needed to restrict if attachments can be forwarded with motions.

reiterl:
Resolve #3189 

Add the new orga setting.


openslides_backend/action/actions/motion/base_create_forwarded.py
```diff
@@ -332,6 +332,7 @@ def update_instance(self, instance: dict[str, Any]) -> dict[str, Any]:
                 "amendment_result_data": [],
             }
         if self.with_attachments:
+            self.check_can_with_attachments()
```
hjanott:
```suggestion
            self.check_can_forward_with_attachments()
```

PR #3216, opened 2025-11-24
[rel-DB] Complete checker tests

vkrasnovyd:
Changes to presenters and tests:
- As discussed with Raimund, preserved the logic of the presenters and deleted only the tests with inconsistent data. The remaining tests make sure that the checker does not scream when data is correct;
- Removed fields that are not in `export_helper.get_fields_for_export` in the `check_database_all` (logic similar to `export_meeting`). Reason: checker was screaming because of the calculated fields pulled from db (like `mediafile.owner_id_organization_id` or `projection.content_object_id_meeting_id`).

Changes to checker:
- Added more precise field type validation for GenericRelationField and GenericRelationListField to get errors about incorrect field value type;
- Remove TODO-comment without actually making any changes. reason: the 2 areas are making slightly different checks so I'd better leave optimizing them for rel-DB 2;
- Added some missing points, fixed number of whitespaces for better readability;
- Merged "state_extension" and "recommendation_extension" validation in check_special_fields(). Reasons:
    - Logic is completely the same.
    - Error message for recommendation_extension had duplications. Example:
```
motion/1/recommendation_extension: Relation Error: Found motion/3 in recommendation_extension but not in models.
motion/1/recommendation_extension: Relation Error: Found theme/1 but only motion is allowed.
motion/1/recommendation_extension: Relation Error: Found motion/3 in recommendation_extension but not in models.
motion/1/recommendation_extension: Relation Error: Found theme/1 but only motion is allowed.
```

### DIFF
--- a/openslides_backend/models/checker.py
+++ b/openslides_backend/models/checker.py
@@ -660,7 +660,7 @@ class Checker:
             assert collection
             if self.mode == "external" and collection not in self.allowed_collections:
                 raise CheckException(f"Fqid {fqid} has an invalid collection.")
-            return collection, id
+            return collection, _id
         except (ValueError, AttributeError, AssertionError, IndexError):
             raise CheckException(f"Fqid {fqid} is malformed")
 


openslides_backend/models/checker.py
```diff
@@ -83,6 +82,15 @@ def check_string(value: Any) -> bool:
     return value is None or isinstance(value, str)
 
 
+def check_fqid(value: Any) -> bool:
+    if value is None:
+        return True
+    if not isinstance(value, str) or "/" not in value:
+        return False
```
hjanott: In patterns.py we have some regex patterns. Wouldn't it be better to use `is_fqid` from there?

hjanott (changes requested):
I see many beautiful changes.

openslides_backend/models/checker.py
```diff
@@ -668,17 +652,18 @@ def check_reverse_relation(
         if error:
             self.errors.append(
                 f"{basemsg} points to {foreign_collection}/{foreign_id}/{actual_foreign_field},"
-                " but the reverse relation for it is corrupt"
+                " but the reverse relation for it is corrupt."
             )
 
     def split_fqid(self, fqid: str) -> tuple[str, int]:
         try:
             collection, _id = fqid.split("/")
```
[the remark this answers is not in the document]
vkrasnovyd: Replaced there and in the next method.

openslides_backend/models/checker.py
```diff
@@ -657,17 +658,17 @@ def check_reverse_relation(
 
     def split_fqid(self, fqid: str) -> tuple[str, int]:
         try:
-            collection, _id = fqid.split("/")
+            collection, _id = collection_and_id_from_fqid(fqid)
             id = int(_id)
```
hjanott:
```suggestion
```

hjanott (approved):
I just found one line that should be removed before merging.

vkrasnovyd:
@hjanott added some missing test cases, did a bit of cleanup in the checker itself and added a couple of todos for rel-db-2. Also re-enabled migration index check in the checker.

PR #3219, opened 2025-11-26
Add required majority fields

### LINKED ISSUE #3199: Motions/Elections: Required majorities
context: https://github.com/OpenSlides/openslides-client/issues/5420

**wanted behaviour in plain text:**
election and motion polls should be visually displayed differently they need a new field to "remember" in how they should be displayed

adidtionally the meeting settings need two new settings where you can save the new option for assignments and motion polls 

**Wanted behaviour in technical terms:**
in poll.create there shoud be a new optional variable 
    onehundred_percent_base: string,
   **required_mayority: string**
    backend: string,

And two new meeting settings: 

    motion_poll_default_required_mayority: string;
    motion_poll_default_required_mayority: string;

**additional info:** 
The client can only use four different options as required mayority, I do not know if the backend wants to implement to only accept these 4 options or if it should just be a string. Please implement in own discretion, and ask question if in doubt

reiterl:
Resolve #3199 

Add some tests and add docs. Update the test data in some tests.

@luisa-beerboom please review.


luisa-beerboom (commented):
I am guessing `no_majority` is the description for the system we had until now.
Given that it is also the default, why not just use `None` instead? I am guessing that would functionally work the same.
One enum value less, no default setting necessary since it will still be `None` if left empty (therefore one less loop the automatic action logic will have to go through), and no need to think about possibly having to migrate past polls.

reiterl:
I will look, if I can use the None case for 'no_majority'.

reiterl:
It shows that the missing of the 'no_majority' option complicates things in the client. So I revert it.
I want a clear "no_majoirty" option here.

luisa-beerboom:
Wouldn't you need a migration to fill in the field for old polls then?

luisa-beerboom (approved):
It seemed a bit strange to me that the `no_majority` option is there, despite the client apparently being able to handle `None`.

Regardless, the code seems fine

reiterl:
I have updated the code and meta. It should be okay now.

PR #3251, opened 2025-12-11
[rel-DB] Complete agenda item forward tests

vkrasnovyd:


tests/system/action/agenda_item/test_forward.py
```diff
@@ -364,22 +319,12 @@ def create_structure_levels(
...
         self.set_models(
             {
                 **{
                     f"meeting_user/{mu_id}": {"structure_level_ids": sl_ids}
                     for mu_id, sl_ids in mu_to_sl_ids.items()
-                },
-                **{
-                    f"structure_level/{sl_id}": {"meeting_user_ids": sorted(mu_ids)}
-                    for sl_id, mu_ids in sl_to_mu_ids.items()
+                    if sl_ids
```
hjanott: This prevents resetting the `structure_level_ids`. Can be wanted. Please check again.
I was also thinking that the function should probably be called. `set_structure_levels_for_meeting_users`

hjanott (changes requested):
Some small things.

PR #3287, opened 2026-01-23
[rel-db] Fix meeting.import

luisa-beerboom:
Rebase was being difficult, so I cherry picked the commits instead.

Replaces #3281

### DIFF
--- a/tests/system/action/meeting/test_clone.py
+++ b/tests/system/action/meeting/test_clone.py
@@ -1372,11 +1372,11 @@ class MeetingClone(BaseActionTestCase):
         self.assert_model_exists("meeting/2", {"name": long_name + " - Copy"})
 
     def test_meeting_name_too_long(self) -> None:
-        self.meeting_data["name"] = "A" * 100
+        self.meeting_data["name"] = "A" * 200
         self.set_test_data_with_admin()
         response = self.request("meeting.clone", {"meeting_id": 1})
         self.assert_status_code(response, 200)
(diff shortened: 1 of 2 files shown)


tests/system/action/meeting/test_clone.py
```diff
@@ -2134,3 +2134,22 @@ def test_clone_with_structured_published_orga_files(self) -> None:
...
         self.media.duplicate_mediafile.assert_not_called()
+
+    def test_clone_require_duplicate_from_allowed(self) -> None:
+        self.set_test_data_with_admin()
+        self.set_models(
+            {
+                "meeting/1": {"template_for_organization_id": 1, "name": "m1"},
+                "organization/1": {
+                    "template_meeting_ids": [1],
+                },
```
vkrasnovyd: `template_meeting_ids` can be skipped as it is a back relation.

tests/system/action/meeting/test_clone.py
```diff
@@ -2134,3 +2134,22 @@ def test_clone_with_structured_published_orga_files(self) -> None:
...
+                "user/1": {
+                    "organization_management_level": None,
+                    "committee_ids": [1],
+                    "committee_management_ids": [1],
+                },
+                "committee/60": {"user_ids": [1], "manager_ids": [1]},
+            }
```
vkrasnovyd: In this case `committee_ids` and `committee_management_ids` should be 60. Test works because these are back relation and a calculated field, so in the database these values are actually 60.
I'd suggest using `self.set_committee_management_level([60])` instead of manually setting these fields.

PR #3295, opened 2026-01-29
[rel-DB] Fix tests after nt:ntR and initial migration

vkrasnovyd:
Needs https://github.com/OpenSlides/openslides-meta/pull/377 for `set_user_groups`.

Mainly set missing required `meeting_user.group_ids`. Additionally:
* Updated base test methods:
    * `set_user_groups` now also deletes `meeting_users` without groups
    * `set_user_groups` builds list events instead of recalculating the `group.meeting_user_ids` and calling `set_models`
    * `assert_db_entries` ignores `version` table while counting db-entries in the databse tests. Reason: fix for the `assert_no_db_entry` method after initial migration PR
* Deleted impossible test cases (`meeting_user` without `group_ids`):
    * `personal_note`: 2 tests
    * `test_get_user_related_models.py`: 2 tests

### DIFF
--- a/tests/system/action/user/test_update.py
+++ b/tests/system/action/user/test_update.py
@@ -2761,14 +2761,7 @@ class UserUpdateActionTest(BaseActionTestCase):
     def test_update_history_add_multiple_groups_with_default_group(self) -> None:
         self.create_meeting()
         user_id = self.create_user(username="test")
-        self.set_models(
-            {
-                "meeting_user/1": {
-                    "meeting_id": 1,
-                    "user_id": user_id,
-                },
-            }
-        )
+        self.set_user_groups(user_id, [1])
 
         response = self.request(
             "user.update",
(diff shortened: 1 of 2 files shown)


tests/system/base.py
```diff
@@ -246,6 +246,9 @@ def get_update_events(self, fqid: str, data: dict[str, Any]) -> list[Event]:
         self.validate_fields(fqid, data)
         return [Event(type=EventType.Update, fqid=fqid, fields=data)]
 
+    def get_delete_events(self, fqid: str) -> list[Event]:
+        return [Event(type=EventType.Delete, fqid=fqid)]
```
hjanott: If it's not much effort to remove the `deleted` flag from the `get_create_events`, let's remove that functionality from there. With rel-db it doesn't make much sense and just creates unnecessary traffic.

PR #3298, opened 2026-02-02
Fix weird clone bug with mediafiles

luisa-beerboom:
I _might_ have found the problem with the clone bug @rrenkert @Elblinator @peb-adr can you test this on an actual instance?

### DIFF
--- a/data/initial-data.json
+++ b/data/initial-data.json
@@ -4,26 +4,22 @@
         "1":{
             "id": 1,
             "name": "male",
-            "organization_id": 1,
-            "user_ids":[]
+            "organization_id": 1
         },
         "2":{
             "id": 2,
             "name": "female",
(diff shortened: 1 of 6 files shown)


tests/system/action/meeting/test_clone.py
```diff
@@ -2161,3 +2161,454 @@ def test_clone_require_duplicate_from_allowed(self) -> None:
         self.set_organization_management_level(None)
         response = self.request("meeting.clone", {"meeting_id": 1})
         self.assert_status_code(response, 200)
+
+    def test_clone_with_very_complex_mediafiles(self) -> None:
+        self.create_meeting(1101)
+        self.set_models(
```
hjanott: Would it be possible to reduce the amount of media files in this setup?

Issue #3307, opened 2026-02-03
[relDB] Tests: get db connection in a cleaner way


luisa-beerboom:
See tests/conftest.py `db_connection` function

The BaseSystemTestCase Classes `connection` attribute is filled from the outside currently. Filling a class variable from the outside seems a bit improper. We should probably change that at some point

PR #3313, opened 2026-02-06
[rel-DB] Remove auth mockers, complete poll tests

vkrasnovyd:
Resolves https://github.com/OpenSlides/openslides-backend/issues/3307

Extra changes:

1. Due to the isolation level used in auth service, attempt to update the user after logging in as this user triggers error: `psycopg.errors.SerializationFailure: could not serialize access due to concurrent update`

Since the error occurs only in tests and not in the production setup, fixed by re-arrenging the order in which login and user update calls are made and replaced manual permission tests setup with `base_permission_test` where possible.

2. `user/test_merge_together.py/test_merge_with_polls_all_errors`: vote service sets relation between votes and users in db only on stop => no duplicated votes are present for `poll/1` because `vote_ids` == `None`.  Fix: moved "duplicated" vote to `poll/2`.

3. Iterating through `db_instance.items()` in the delete-action for getting delete events for the related instances lead to inconsistent order of the `history_entry` items assigned to the same `history_position`. Fixed by sorting the dictionary before iteration.

### DIFF
--- a/tests/system/action/motion/test_update.py
+++ b/tests/system/action/motion/test_update.py
@@ -676,8 +676,6 @@ class MotionUpdatePermissionTest(BaseMotionUpdateActionTest):
 
     def test_update_permission_metadata_and_submitter(self) -> None:
         self.setup_can_manage_metadata()
-        self.permission_test_models["meeting_user/1"] = {"motion_submitter_ids": [1]}
-        self.set_models(self.permission_test_models)
         self.set_models(
             {
                 "motion_submitter/1": {
--- a/openslides_backend/action/generics/delete.py
+++ b/openslides_backend/action/generics/delete.py
@@ -49,7 +49,7 @@ class DeleteAction(Action):
         # Update instance and set relation fields to None.
         # Gather all delete actions with action data and also all models to be deleted
         delete_actions: list[tuple[FullQualifiedId, type[Action], ActionData]] = []
-        for field_name, value in dict(sorted(db_instance.items())).items():
+        for field_name, value in sorted(db_instance.items()):
             if field_name == "id":
                 continue
             field = cast(BaseRelationField, self.model.get_field(field_name))
--- a/tests/system/action/base.py
+++ b/tests/system/action/base.py
@@ -245,6 +245,13 @@ class BaseActionTestCase(BaseSystemTestCase):
         user_groups: list[int] = [3],
         custom_error_message: str | None = None,
     ) -> None:
+        """
+        Checks that response.status_code == 403 when any of the conditions is true:
+            - `permission` is not provided
+            - `custom_error_message` is provided
+            - `fail` is set to True
+        In other cases asserts response.status_code to be 200.
+        """
         meeting_data = {"locked_from_inside": True} if lock_meeting else {}
         self.create_meeting(meeting_data=meeting_data)
         self.user_id = self.create_user("user")


tests/system/action/poll/test_vote.py
```diff
@@ -299,7 +305,9 @@ def test_vote_no_votes_total_check_by_YNA_max_votes_error(self) -> None:
         assert "You have to select between 1 and 1 options" in response.json["message"]
 
     def test_vote_no_votes_total_check_by_YN(self) -> None:
-        self.create_poll_with_3_options({"max_votes_per_option": 1})
+        self.create_poll_with_3_options(
+            {"max_votes_amount": 10, "max_votes_per_option": 1}
```
hjanott: Same here.

openslides_backend/services/postgresql/create_schema.py
```diff
@@ -55,6 +56,11 @@ def create_schema() -> None:
                 print(
                     "Assuming relational schema is applied, because table version exists.\n"
                 )
+                if not MigrationHelper.get_database_migration_index(cursor):
+                    MigrationHelper.set_database_migration_info(
+                        cursor, 100, MigrationState.FINALIZED
```
luisa-beerboom: Why set this to 100 and not dynamically whatever the most recent migration index is?

vkrasnovyd: @hjanott

openslides_backend/services/postgresql/create_schema.py
```diff
@@ -55,6 +56,11 @@ def create_schema() -> None:
...
+                    )
+                deactivate_notify_triggers(cursor)
```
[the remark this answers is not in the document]
luisa-beerboom: Are we disabling them _just_ for the backend here? This won't have an effect on prod, right?

tests/system/action/poll/test_vote.py
```diff
@@ -110,11 +110,12 @@ def test_vote_correct_pollmethod_Y(self) -> None:
                 },
             }
         )
+        self.login(user_id)
         response = self.request(
             "poll.vote", {"id": 1, "value": {"11": 1}}, stop_poll_after_vote=False
         )
         self.assert_status_code(response, 200)
-        self.login(user_id)
+        self.login(1)
```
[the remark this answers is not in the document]
vkrasnovyd: Seems to work the old way when executing `poll.stop` with `self.request` with the flag `internal=True`. Reverted these changes.

tests/system/action/user/test_merge_together.py
```diff
@@ -1146,19 +1122,15 @@ def test_merge_with_polls_and_subsequent_merges(self) -> None:
...
         self.assert_merge_with_polls_correct(password, 1)
 
-    @pytest.mark.skipif(
-        not isinstance("UserMergeTogether", BaseVoteTestCase),
-        reason="set base class to BaseVoteTestCase, if auth isn't mocked for polls anymore. Subsequently remove the skipifs.",
-    )
     def test_merge_with_polls_all_errors(self) -> None:
         self.set_up_polls_for_merge()
-        self.request_multi("poll.start", [{"id": i} for i in range(1, 7)])
+        self.request_multi("poll.start", [{"id": i} for i in [3, 4]])
```
vkrasnovyd: Now `poll.start` is also executed as the internal action and doesn't fail for regular users anymore. So there's no need to separately start the polls with the vote-requests below.

tests/system/action/motion/test_update.py
```diff
@@ -519,7 +519,7 @@ def setUp(self) -> None:
         self.permission_test_models: dict[str, dict[str, Any]] = {
             "meeting_user/1": {
                 "meeting_id": 1,
-                "user_id": 1,
+                "user_id": 2,
```
hjanott: Why did this change?

PR #3356, opened 2026-02-23
Migration 100 guided

peb-adr:


hjanott (changes requested):
I like the general approach and have some suggestions. :)

peb-adr:
The PR for documentation of the Update to 4.3.0 is now open in main repo

https://github.com/OpenSlides/OpenSlides/pull/7091/changes

It Includes the file `UPDATE_TO_4.3.md`.
So `LINK` can presumably be replaced with `https://github.com/OpenSlides/OpenSlides/blob/main/UPDATE_TO_4.3.md`.

After that the only remaining TODO is:
- [x] StringIO stream: replace `getvalue` with appropriate `read` / `truncate`
- [x] reset route: return [501 Not Implemented](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/501)

rrenkert:
@peb-adr Because of https://github.com/OpenSlides/openslides-client/issues/5949#issuecomment-4207007579 we need to set the timezone for the orga an d meetings. Can we use the ENV variables used here to fill the fields?

peb-adr:
> Can we use the ENV variables used here to fill the fields?

Basically, yes we can.

But if we want to have a more readable timezone string in those fields and not just a numeric UTC offset (though it should be an allowed value I think), we should rename `MIG0100_UTC_OFFSET` to `MIG0100_UTC_TIMEZONE`.

If still necessary (not sure at the moment) we can then derive the UTC offset from that.

openslides_backend/migrations/migration_manager.py
```diff
@@ -71,9 +71,14 @@ def get_migration_result(self) -> dict[str, Any]:
         if MigrationHelper.migrate_thread_stream and (
             output := MigrationHelper.migrate_thread_stream.getvalue()
         ):
+            # TODO: removed commented out lines after review
+            # MigrationHelper.write_line is hardly usable when only the last
+            # line is output.
```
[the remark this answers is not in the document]
hjanott: Changed it so that all lines are read since last time reading. Reads the same block again if nothing new was written.

hjanott:
> If still necessary (not sure at the moment) we can then derive the UTC offset from that.

Yes we could do that easily. But that must mean that the system was operating in UTC+00. We could also look at the systems time zone and then derive the offset from that. Or as a third option allow to override the calculation by the env variable `MIG0100_UTC_OFFSET`. It could be that I'm describing it wrongly.
As far as I can see there are two factors to consider. The time zone the OS operated in and the time zone the organization wants to operate in.

openslides_backend/migrations/migration_handler.py
```diff
@@ -305,10 +303,28 @@ def migrate(self) -> None:
...
+                    if callable(getattr(migration_module, "check_prerequisites", None)):
+                        pre_check_ok = migration_module.check_prerequisites(self.cursor)
+
+                    if not pre_check_ok:
+                        self.logger.info("failed.")
+                        # TODO: Is rasing an exception appropriate here?
+
+                        #raise MigrationException(
+                        #    f"pre check for migration {module_name} failed."
+                        #)
```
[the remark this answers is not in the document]
hjanott: Changed it so that a string with all errors is returned from the prerequisites check and thrown if truthy.

openslides_backend/http/http_exceptions.py
```diff
@@ -62,3 +63,7 @@ class NotFound(HTTPException, BaseNotFound):
 
 class InternalServerError(HTTPException, BaseInternalServerError):
     pass
+
+
+class InternalNotImplemented(HTTPException, BaseNotImplemented):
```
peb-adr, line 68: Why not use the regular `NotImplemented` exception?

openslides_backend/migrations/migrations/0100_init_reldb.py
```diff
@@ -235,68 +232,49 @@ def get_insert_intermediate_t_commands(
...
+    errors = ""
+    MigrationHelper.write_line(
+        "This is migration 100, part of the OpenSlides 4.3.0 release."
+    )
+    MigrationHelper.write_line(
+        "This migration will fundamentally restructure all data."
+    )
+    MigrationHelper.write_line(
+        "See https://github.com/OpenSlides/OpenSlides/blob/main/UPDATE_TO_4.3.md for more information."
+    )
```
peb-adr, line 246: Aesthetic personal preference. I realize this form was originally introduced by me.

```suggestion
    MigrationHelper.write_line(
        "For more information, see"
    )
    MigrationHelper.write_line(
        "  https://github.com/OpenSlides/OpenSlides/blob/main/UPDATE_TO_4.3.md"
    )
```

openslides_backend/migrations/migration_handler.py
```diff
@@ -555,6 +559,7 @@ def reset(self) -> None:
         """
         Resets the migrations currently in progress and restores the state before the migration.
         """
+        raise CommandNotImplemented("The reset route is not implemented yet.")
```
peb-adr, line 571: Another type of "not implemented". Are there good reasons for it?

peb-adr (commented):
Looked over the changes since 04e723c and documented some thoughts I had.
In general it looks good.
I will test the migration later.

openslides_backend/http/http_exceptions.py
hjanott: This is `werkzeug.exceptions.NotImplemented`.
I'm strictly following the pattern of how the other Exceptions are created here.

openslides_backend/migrations/migration_handler.py
hjanott: See https://github.com/OpenSlides/openslides-backend/pull/3356/changes#r3117186039 This one just holds the error code checked during error handling. 
Also I didn't want to name it exactly like the "normal" one, overriding it.

hjanott (approved):
Looks good to me. Thank you for the fix.

peb-adr:
Force-pushed back to f51d62f

I didn't think the rest through. Probably will open a new PR next week ..

PR #3375, opened 2026-02-27
Implement action retry

### LINKED ISSUE #3305: [relDB] Investigate safety of TreeSortMixin
In the `sort_tree` method there's a TODO: "Check if instances exist in DB and is not deleted."

There is the possibility that the instances are already guaranteed to exist due to being found by the filter in the new system.
That really only leaves the question if any cached models that are marked for deletion may be found, or something along those lines.

We should take a look at the `extended_database.py` and the `database_reader.py` to see how the filter acts precisely. If the filter doesn't guarantee existance, that should be fixed, either in the mixin, or maybe better even, in the filter function. If everything's fine, the TODO can be deleted and this issue closed.

### LINKED ISSUE #3372: Action retry in action-worker
The new database handling uses postgresql Isolation level `read for update` and rows that have to be locked are selected with `SELECT [...] FOR UDPATE`. This leads to concurrency errors if two transactions try to select-for-update the same row.

This is reproducable by reverting the changes made in 
https://github.com/OpenSlides/openslides-backend/pull/3313/changes#diff-91be329081a037edb9a2ae53ab4e7ebe89e081557c85df34875479c98eb606d5R32

To avoid immediate response to client with the error `could not serialize access due to concurrent update` we want to perform a retry of the called action on action-worker level.

The amount of retries should be configurable via ENV-variable (ACTION_RETRY_COUNT or something) with a default of 3 and there should be a wait between the retries also configurable via ENV (ACTION_RETRY_DELAY) with a default of 250ms.

The response to the client should be a _user_ readable message like "Can not perform the current data changes. Please try again later".

luisa-beerboom:
Resolves #3372 
Resolves #3305

Used the env settings 
- ACTION_MAX_RETRIES with default 3
- ACTION_RETRY_TIMEOUT with default 0.4

Names are different from the ones suggested in the issue to resemble the database retry settings

### DIFF
--- a/openslides_backend/action/action_handler.py
+++ b/openslides_backend/action/action_handler.py
@@ -125,7 +125,7 @@ class ActionHandler(BaseHandler):
                 except fastjsonschema.JsonSchemaException as exception:
                     raise ActionException(exception.message)
 
-            retry_count = int(self.env.ACTION_MAX_RETRIES) or 1
+            retry_count = int(self.env.ACTION_MAX_RETRIES or 1)
             retry_timeout = float(self.env.ACTION_RETRY_TIMEOUT or 0.4)
             for attempt in range(1, retry_count + 1):
                 try:
@@ -178,10 +178,10 @@ class ActionHandler(BaseHandler):
                     raise RelationException(
                         f"Relation violates required constraint: {e}"
                     )
-                except SerializationFailure as e:
+                except SerializationFailure:
                     if attempt == retry_count:
                         raise DatabaseException(
-                            f"Unexpected error reading from database: {e}"
+                            "Database operation failed due to concurring actions. Please try again later."
                         )
                     sleep(retry_timeout)
             raise BadCodingException("This code should never execute")


openslides_backend/action/action_handler.py
```diff
@@ -122,52 +125,66 @@ def handle_request(
...
-                    else:
-
-                        def transform_to_list(
-                            tuple: tuple[WriteRequest | None, ActionResults | None],
-                        ) -> tuple[list[WriteRequest], ActionResults | None]:
-                            return (
-                                [tuple[0]] if tuple[0] is not None else [],
-                                tuple[1],
+            retry_count = int(self.env.ACTION_MAX_RETRIES) or 1
+            retry_timeout = float(self.env.ACTION_RETRY_TIMEOUT or 0.4)
```
peb-adr: Not sure if `or` is better before or after cast.
If `self.env.VAR` is `None` or `''` if VAR is unset, I'd say `or` should happen before casting the result (as was done for `retry_timeout`).

In any case it should be uniform.

openslides_backend/shared/env.py
```diff
@@ -27,6 +27,8 @@ class Environment(Env):
 
     vars = {
         "ACTION_PORT": "9002",
+        "ACTION_MAX_RETRIES": "3",
+        "ACTION_RETRY_TIMEOUT": "0.4",
```
peb-adr: `ACTION_RETRY_TIMEOUT`
Not sure if I like this name .. To me it suggests that a retry would be aborted after this amount of time.
This is what a "timeout" usually refers to.

Maybe `ACTION_RETRY_INTERVAL` would be more fitting and possibly more in line with what we use in other places. F.e. the AU uses [`METRIC_INTERVAL`](https://github.com/OpenSlides/openslides-autoupdate-service/blob/main/environment.md) to set the time between two metric logs.

luisa-beerboom: I used those names to make it coherent with the names for the database retry env variables, which do the same thing as these, just for backend-internal database calls.
I'd prefer to keep it uniform. Is this really that important?

peb-adr: Hmm, I guess if there's already a pattern you're following I can accept that, although I still think the name is misleading.

PR #3408, opened 2026-03-20
Use real enums

vkrasnovyd:
Adjusts backend to the changes in https://github.com/OpenSlides/openslides-meta/pull/424

Database writer is now using the correcponding enum as a type for updating list fields with items.enum.


hjanott (changes requested):
Also see the general comment in the related meta issue.

hjanott (changes requested):
There is good improvements. Especially the new tests.

hjanott (approved):
Also this is done now. Just waiting for the vote service to get fixed.

PR #3410, opened 2026-03-22
Get everything post-reldb

peb-adr:


peb-adr (commented):
Should look as similar as possible to
https://github.com/OpenSlides/openslides-backend/blob/staging/4.2.30/cli/get_everything.py

hjanott (approved):
Added the self-requested changes and linters. Should be great now.

PR #3438, opened 2026-03-27
Add time_zone to committee json_upload

### LINKED ISSUE #3434: Add time_zone interpretation to committee import
Change committee import to have a `meeting_time_zone` field.
- [ ] Sets the meetings `time_zone` field
- [ ] If it is set, it will be used to interpret the `meeting_start_time` and `meeting_end_time` dates into timestamps.
- [ ] If it isn't set, but `meeting_start_time` and `meeting_end_time` are, add a warning about which timezone will be used for interpretation, which will be either the orga timestamp or UTC as a fallback.

luisa-beerboom:
Closes #3434 

Also added `get_valid_timezones` presenter with which one may get the valid timezones.


openslides_backend/services/database/interface.py
```diff
@@ -23,6 +25,9 @@
 FQID_MAX_LEN = 48  # collection + id
 COLLECTIONFIELD_MAX_LEN = 239  # collection + field
 
+SqlArguments = list[str | int]
+SqlArgumentsExtended = tuple[list[Id]] | SqlArguments
```
hjanott: Would adding the tuple of lists to the former break the code elsewhere?

reiterl (commented):
It looks okay to me. (I didn't check the database_reader stuff).

luisa-beerboom:
> Removed time_zone from orga update bc it would break meeting start- and end_times if there is no meeting/time_zone

Removed that commit again

hjanott (approved):
I don't see what changed so nothing for me to review.

PR #3489, opened 2026-04-15
Do not omit header when generating pot file

rrenkert:
Removed --omit-header to make pot file merging work in main repo

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -159,8 +159,8 @@ pip-check: | deprecation-warning
 	pip-check -H
 
 
-extract-translations: | deprecation-warning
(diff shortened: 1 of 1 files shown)


vkrasnovyd (changes requested):
Please also run the updated command to re-generate the pot-file with the header and merge it in the same PR.

rrenkert:
> Please also run the updated command to re-generate the pot-file with the header and merge it in the same PR.

This is done with the translations. Generating the pot file and not translating new strings will lead to inconsistencies.

vkrasnovyd (approved):
In this case looks good to me.

emanuelschuetze (approved):
tested

PR #3497, opened 2026-04-16
Set prepare_threshold to DB connection kwargs to no prepared statements

hjanott:
Removing prepared statements again to prevent errors with libpq version 15. Once on debian trixie or using psycopg[binary] we could allow them again. Those would provide libpq 17.


rrenkert:
Maybe update to trixie and psycopg[binary] with https://github.com/OpenSlides/OpenSlides/pull/7092

rrenkert:
We updated to trixie and use psycopg binary in #3502

PR #3501, opened 2026-04-16
Add new required constraints

vkrasnovyd:
Needs https://github.com/OpenSlides/openslides-meta/pull/468

### DIFF
--- a/openslides_backend/services/vote/adapter.py
+++ b/openslides_backend/services/vote/adapter.py
@@ -25,9 +25,7 @@ class VoteAdapter(VoteService, AuthenticatedService):
             self.logger.debug(message)
         elif response.status_code == 500:
             self.logger.error(message)
-            raise VoteServiceException(
-                "Vote service sends HTTP 500 Internal Server Error."
-            )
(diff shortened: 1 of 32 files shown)


openslides_backend/services/vote/adapter.py
```diff
@@ -25,9 +25,7 @@ def retrieve(self, endpoint: str, payload: dict[str, Any] | None = None) -> Any:
             self.logger.debug(message)
         elif response.status_code == 500:
             self.logger.error(message)
-            raise VoteServiceException(
-                "Vote service sends HTTP 500 Internal Server Error."
-            )
+            raise VoteServiceException(message)
```
hjanott: If we change it like this. Will it still be clear at first glance that this error stems from specifically a code 500 in the vote service?

hjanott (approved):
With a tiny change can be merged.

openslides_backend/services/vote/adapter.py
vkrasnovyd: We don't change this. Leftovers of debugging (was a setup problem). Reverted.

PR #3514, opened 2026-04-21
SAML meeting mapping always allow create if `meeting_user` not meeting member

### LINKED ISSUE #3496: Change SAML mapping behaviour for allow update and groups
Talked with @emanuelschuetze. We may need a bug fix for pre rel-db aka 4.2.29.
**Describe the bug**
If you create a SAML meeting mapping with `allow_update` and a user for a SAML user already exists he doesn't get added via the mapper to the meeting. For this we don't only need to check if `allow_update` is true. But changing" the info should also be allowed if the user doesn't exist in the meeting. It needs to be documented that this means with an active SAML mapping you cannot remove users from a meeting but you can only lock them out or deactivate them completely.

**How To Reproduce**
Create a mapping for a meeting.
Log in and out as a SAML user.
Create a mapping for a meeting with `allow_update`  false.
Log in as that SAML user and he won't be added.

Furthermore we want that if the IDP doesn't send any key for mapped groups or no mappings apart from default groups are set that group settings don't get updated even if `allow_update` is true. This change of behaviour also needs to be documented.

- [x] create fix for 4.2.29
- [ ] apply this fix to 4.3.0 rule set

hjanott:
resolves #3496 
Fix groups being removed with update.
Allow `meeting_user` creation if user is not member of meeting.
I added the group rule that the default group is only used when there no other is.

### DIFF
--- a/docs/actions/user.save_saml_account.md
+++ b/docs/actions/user.save_saml_account.md
@@ -37,7 +37,7 @@ Extras to do on creation:
 - The saml attribute mapping can have a list of 'meeting_mappers' that can be used to assign users meeting related data. (See example below. A full example can be found in the [organization.update.md](organization.update.md))
     - A mapper can be given a 'name' for debugging purposes.
     - The 'external_id' maps to the meeting and is required (logged as warning if meeting does not exist). Multiple mappers can map to the same meeting.
-    - If 'allow_update' is set to false, the mapper is only used if the user does not already exist. If it is not given it defaults to true.
+    - If 'allow_update' is set to false, the mapper is only used if the user does not already exist in the meeting. If it is not given it defaults to true. This option should only be used if you are sure that users aren't deleted from meetings. Alternatively they can be locked out from the meeting.
     - Mappers are only used if every condition in the list of 'conditions' resolves to true. For this 
         - the value for 'attribute' in the payload data has to match the string or regex given in 'condition'. 
         - if no condition is given this defaults to true. 
--- a/docs/actions/user.save_saml_account.md
+++ b/docs/actions/user.save_saml_account.md
@@ -37,7 +37,7 @@ Extras to do on creation:
 - The saml attribute mapping can have a list of 'meeting_mappers' that can be used to assign users meeting related data. (See example below. A full example can be found in the [organization.update.md](organization.update.md))
     - A mapper can be given a 'name' for debugging purposes.
     - The 'external_id' maps to the meeting and is required (logged as warning if meeting does not exist). Multiple mappers can map to the same meeting.
-    - If 'allow_update' is set to false, the mapper is only used if the user does not already exist in the meeting. If it is not given it defaults to true. This option should only be used if you are sure that users aren't deleted from meetings. Alternatively they can be locked out from the meeting.
+    - If 'allow_update' is set to false, the mapper is only used to create a new participant if the user does not already exist in the meeting. If it is not given it defaults to true. This option should only be used if you are sure that users aren't deleted from meetings. Alternatively they can be locked out from the meeting.
     - Mappers are only used if every condition in the list of 'conditions' resolves to true. For this 
         - the value for 'attribute' in the payload data has to match the string or regex given in 'condition'. 
         - if no condition is given this defaults to true. 
@@ -50,7 +50,7 @@ Extras to do on creation:
 - On conflict of multiple mappers mappings on a same meetings field the last given mappers data for that field is used. Exception to this are groups and structure levels. Their data is combined. 
 - Values for groups and structure levels can additionally be given in comma separated lists composed as a single string.
 - Values for groups are interpreted as their external ID and structure levels as their name within that meeting.
-- If no group exists for a meeting and no default is given, the meetings default group is used. (Logged as warning)
+- If no group exists for a newly created participant and no default is given, the meetings default group is used. (Logged as warning)
 - If a structure level does not exist, it is created.
 - Vote weights need to be given as 6 digit decimal strings.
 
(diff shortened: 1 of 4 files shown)


docs/actions/user.save_saml_account.md
```diff
@@ -37,7 +37,7 @@ Extras to do on creation:
 - The saml attribute mapping can have a list of 'meeting_mappers' that can be used to assign users meeting related data. (See example below. A full example can be found in the [organization.update.md](organization.update.md))
     - A mapper can be given a 'name' for debugging purposes.
     - The 'external_id' maps to the meeting and is required (logged as warning if meeting does not exist). Multiple mappers can map to the same meeting.
-    - If 'allow_update' is set to false, the mapper is only used if the user does not already exist. If it is not given it defaults to true.
+    - If 'allow_update' is set to false, the mapper is only used to create a new participant if the user does not already exist in the meeting. If it is not given it defaults to true. This option should only be used if you are sure that users aren't deleted from meetings. Alternatively they can be locked out from the meeting.
```
luisa-beerboom, line 40: I don't get how `Alternatively they can be locked out from the meeting` relates to the rest of the line.

openslides_backend/action/actions/user/save_saml_account.py
```diff
@@ -362,7 +362,18 @@ def apply_meeting_user_data(
...
+            db_meeting_user_exists = bool(
+                (
+                    get_meeting_user(
+                        self.datastore,
+                        meeting_id,
+                        user_id,
+                        ["group_ids"],
+                    )
+                    or {}
+                ).get("group_ids")
```
luisa-beerboom, line 374: I believe this `.get("group_ids")` line ought to be unnecessary ever since migration 0074.

Then again, I don't think this hurts either.

openslides_backend/action/actions/user/save_saml_account.py
```diff
@@ -373,10 +384,13 @@ def apply_meeting_user_data(
                     names = sorted(
                         instance_meeting_user.pop(saml_meeting_user_field, [])
                     )
-                    if saml_meeting_user_field == "groups":
-                        ids = self.get_group_ids(names, meeting)
-                    elif saml_meeting_user_field == "structure_levels":
-                        ids = self.get_structure_level_ids(names, meeting)
+                    match saml_meeting_user_field:
```
luisa-beerboom, line 387: wow, match case!

openslides_backend/action/actions/user/save_saml_account.py
```diff
@@ -403,11 +417,22 @@ def update_meeting_users_from_db(
                 ["id", "group_ids", "structure_level_ids"],
             ):
                 for field_name in ["group_ids", "structure_level_ids"]:
+                    # TODO use sets
```
luisa-beerboom, line 420: Then use sets

openslides_backend/action/actions/user/save_saml_account.py
```diff
@@ -504,14 +531,19 @@ def get_group_ids(self, group_names: list[str], meeting: dict) -> list[int]:
...
             external_meeting_id = meeting["external_id"]
             self.logger.warning(
                 f"save_saml_account found no group in meeting '{external_meeting_id}' for {group_names}, but used default_group of meeting"
             )
             return [default_group_id]
         else:
-            assert False
+            raise ActionException(
+                "Invalid data state: mapped meeting does not have default group."
+            )
```
luisa-beerboom, line 546: Since I belieeeveeeee `default_group_id` is a required field, I'd like to ask the question: Do we really need to have this else case in the first place?

tests/system/action/user/test_save_saml_account.py
```diff
@@ -1139,26 +1139,26 @@ def test_update_user_with_default_membership(self) -> None:
...
         )
 
-    def test_update_user_participant_already_in_group(self) -> None:
+    def test_update_user_participant_already_in_another_meeting(self) -> None:
         """
         Shows:
             * user stays in group 2
             * structure_level/1 left untouched structure_level/2 added
-            * second mapper is ignored due to being an update and allow_update being false
+            * second mapper is used in spite of allow_update being false since Kreistag meeting is a create
```
luisa-beerboom, line 1147:
```suggestion
            * second mapper is used in spite of allow_update being false since 'Kreistag meeting' is a create
```
Purely aesthetic ideas here, but I don't like text with random other language words thrown in the middle and this makes it clear at a glance that the German word is a test-data internal name.

luisa-beerboom: Same thing for other instances in this test.

luisa-beerboom (changes requested):

docs/actions/user.save_saml_account.md
hjanott: I hope this change explains it better.
```suggestion
    - If 'allow_update' is set to false, the mapper is only used to create a new participant if the user does not already exist in the meeting. If it is not given it defaults to true. This should only be left at true if you are sure that users aren't deleted from meetings as the user would get readded upon the next login. When set to true participants can be alternatively to removing from the meeting locked out from the meeting.
```

luisa-beerboom: maybe rather this
```suggestion
    - If 'allow_update' is set to false, the mapper is only used to create a new participant if the user does not already exist in the meeting. If it is not given it defaults to true. This should only be left at true if you are sure that users aren't deleted from meetings. When set to true participants can be locked out from the meeting, alternatively to removing them from the meeting.
```

hjanott: Yes. What do you think about `in preference` instead of `alternatively`?

openslides_backend/action/actions/user/save_saml_account.py
hjanott: Yes. That's why this should be removed in the main branch.

docs/actions/user.save_saml_account.md
luisa-beerboom: That's good, I'll use it in the newer PR

PR #3520, opened 2026-04-22
Create pre-migration 0081_check_equal_fields

### LINKED ISSUE #3507: Check for equal fields in pre reldb
We need a special migration to perform a check on the equal fields. In some instances there are error while executing migration 100 since https://github.com/OpenSlides/openslides-meta/pull/407 is merged.

Like other pre-migrations this "check-migration" should be based on the `staging/4.2.30` branch

luisa-beerboom:
Closes #3507

TODO:
- [x] Migration indices in json files
- [x] Fix tests in setup
   - [x] Check if test payloads are still intact after the fix
- [x] Give the tests a culling
- [x] Clean up the code, ya nasty

### DIFF
--- a/openslides_backend/migrations/migrations/0074_check_equal_fields.py
+++ b/openslides_backend/migrations/migrations/0074_check_equal_fields.py
@@ -29,7 +29,7 @@ class Migration(BaseModelMigration):
     It throws an aggregate error listing everything that is broken.
     """
 
-    target_migration_index = 82
+    target_migration_index = 75
 
     def migrate_models(self) -> list[BaseRequestEvent] | None:
         self.load_models()
--- a/openslides_backend/migrations/migrations/0074_check_equal_fields.py
+++ b/openslides_backend/migrations/migrations/0074_check_equal_fields.py
@@ -2,7 +2,7 @@ from typing import Any, cast
 
 from datastore.migrations import BaseModelMigration
 from datastore.reader.core import GetManyRequestPart
-from datastore.writer.core.write_request import BaseRequestEvent
+from datastore.writer.core.write_request import BaseRequestEvent, RequestUpdateEvent
 
 from openslides_backend.models.base import model_registry
 from openslides_backend.models.fields import (
@@ -10,7 +10,10 @@ from openslides_backend.models.fields import (
     GenericRelationField,
     GenericRelationListField,
 )
-from openslides_backend.shared.patterns import collection_and_id_from_fqid
+from openslides_backend.shared.patterns import (
+    collection_and_id_from_fqid,
+    fqid_from_collection_and_id,
+)
 
 from ...shared.filters import And, FilterOperator
 from ..exceptions import MigrationException
@@ -35,6 +38,7 @@ class Migration(BaseModelMigration):
         self.load_models()
         relations = self.get_relations()
         errors: set[str] = set()
+        events: list[BaseRequestEvent] = []
         for ((collection1, field1), (collection2, field2)), (
             eq_fields,
             field_def1,
@@ -60,20 +64,25 @@ class Migration(BaseModelMigration):
                 affected_models2,
             )
             for match in matches:
-                if error := self.check_equal_data(
+                if check_result := self.check_equal_data(
                     match,
                     collection1,
                     field1,
+                    field_def1,
                     affected_models1,
                     collection2,
                     field2,
+                    field_def2,
                     affected_models2,
                     eq_fields,
                 ):
-                    errors.add(error)
+                    if isinstance(check_result, str):
+                        errors.add(check_result)
+                    else:
+                        events.append(check_result)
         if len(errors):
             raise MigrationException(list(errors))
-        return None
+        return events
 
     def load_models(self) -> None:
         self.models: dict[str, dict[str, dict[str, Any]]] = {
@@ -356,12 +365,14 @@ class Migration(BaseModelMigration):
         match: tuple[int, int],
         collection1: str,
         field1: str,
+        field_def1: dict[str, Any],
         affected_models1: dict[int, dict[str, Any]],
         collection2: str,
         field2: str,
+        field_def2: dict[str, Any],
         affected_models2: dict[int, dict[str, Any]],
         equal_fields: tuple[str, ...],
-    ) -> str | None:
+    ) -> RequestUpdateEvent | str | None:
         """
         Checks equal_data between two entries.
         Returns an error-string if there is a difference
@@ -373,7 +384,15 @@ class Migration(BaseModelMigration):
         eq_data_tup2 = self.get_equal_data_tuple(
             collection2, id2, affected_models2, equal_fields
         )
-        if eq_data_tup1 != eq_data_tup2:
+        if eq_data_tup1 is None:
+            return self.get_cleanup_event(
+                collection2, id2, field2, field_def2, collection1, id1
+            )
+        elif eq_data_tup2 is None:
+            return self.get_cleanup_event(
+                collection1, id1, field1, field_def1, collection2, id2
+            )
+        elif eq_data_tup1 != eq_data_tup2:
             the_problem = " and ".join(
                 sorted(
                     [
@@ -385,14 +404,47 @@ class Migration(BaseModelMigration):
             return f"Detected different equal_fields: {the_problem} for equal_fields {equal_fields}"
         return None
 
+    def get_cleanup_event(
+        self,
+        collection1: str,
+        id1: int,
+        field1: str,
+        field_def1: dict[str, Any],
+        collection2: str,
+        id2: int,
+    ) -> RequestUpdateEvent:
+        if field_def1["is_list_relation"]:
+            return RequestUpdateEvent(
+                fqid=fqid_from_collection_and_id(collection1, id1),
+                fields={},
+                list_fields={
+                    "remove": {
+                        field1: [
+                            (
+                                fqid_from_collection_and_id(collection2, id2)
+                                if field_def1["is_generic"]
+                                else id2
+                            )
+                        ]
+                    }
+                },
+            )
+        else:
+            return RequestUpdateEvent(
+                fqid=fqid_from_collection_and_id(collection1, id1),
+                fields={field1: None},
+            )
+
     def get_equal_data_tuple(
         self,
         collection: str,
         id_: int,
         affected_models: dict[int, dict[str, Any]],
         sorted_eqfs: tuple[str, ...],
-    ) -> tuple[Any, ...]:
-        model = affected_models.get(id_, {})
+    ) -> tuple[Any, ...] | None:
+        model = affected_models.get(id_)
+        if model is None:
+            return None
         return tuple(
             [
                 (
--- a/openslides_backend/migrations/migrations/0081_check_equal_fields.py
+++ b/openslides_backend/migrations/migrations/0081_check_equal_fields.py
@@ -1,15 +1,16 @@
-from typing import Any
+from typing import Any, cast
 
 from datastore.migrations import BaseModelMigration
 from datastore.reader.core import GetManyRequestPart
 from datastore.writer.core.write_request import BaseRequestEvent
 
-from cli.util.util import open_yml_file
-from openslides_backend.shared.patterns import (
-    collection_and_id_from_fqid,
-    collection_from_collectionfield,
-    field_from_collectionfield,
+from openslides_backend.models.base import model_registry
+from openslides_backend.models.fields import (
+    BaseRelationField,
+    GenericRelationField,
+    GenericRelationListField,
 )
+from openslides_backend.shared.patterns import collection_and_id_from_fqid
 
 from ...shared.filters import And, FilterOperator
 from ..exceptions import MigrationException
@@ -109,8 +110,7 @@ class Migration(BaseModelMigration):
         return matches
 
     def get_relation_flags(self, field_def: dict[str, Any]) -> tuple[bool, bool]:
-        typ: str = field_def["type"]
-        return typ.startswith("generic-"), typ.endswith("-list")
+        return field_def["is_generic"], field_def["is_list_relation"]
 
     def get_affected_models(
         self,
@@ -204,11 +204,10 @@ class Migration(BaseModelMigration):
         collection2: str,
         affected_models2: dict[int, dict[str, Any]],
     ) -> list[int]:
-        typ: str = field_def1["type"]
-        is_generic = typ.startswith("generic-")
+        is_generic = field_def1["is_generic"]
         missing_ids: set[int] = set()
         for model in affected_models1.values():
-            if typ.endswith("-list"):
+            if field_def1["is_list_relation"]:
                 for entry in model.get(field1) or []:
                     if is_generic:
                         coll, id_ = collection_and_id_from_fqid(entry)
@@ -233,7 +232,30 @@ class Migration(BaseModelMigration):
         tuple[tuple[str, str], tuple[str, str]],
         tuple[set[str], dict[str, Any], dict[str, Any]],
     ]:
-        self.models = open_yml_file("./meta/models.yml")
+        self.models: dict[str, dict[str, dict[str, Any]]] = {
+            collection: {
+                field.own_field_name: (
+                    {
+                        "to": cast(BaseRelationField, field).to,
+                        "equal_fields": cast(BaseRelationField, field).equal_fields,
+                        "is_relation": True,
+                        "is_generic": isinstance(field, GenericRelationField)
+                        or isinstance(field, GenericRelationListField),
+                        "is_list_relation": cast(
+                            BaseRelationField, field
+                        ).is_list_field,
+                    }
+                    if hasattr(field, "to")
+                    else {
+                        "is_relation": False,
+                        "is_generic": False,
+                        "is_list_relation": False,
+                    }
+                )
+                for field in Model().get_fields()
+            }
+            for collection, Model in model_registry.items()
+        }
         # main_coll, main_field, back_coll, back_field -> equal_fields, main_field_def, back_field_def
         relations: dict[
             tuple[tuple[str, str], tuple[str, str]],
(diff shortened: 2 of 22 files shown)


m-schieder (approved):

luisa-beerboom:
Magnus noticed this:

```
ERROR:openslides_backend.shared.handlers.base_handler:2892
Traceback (most recent call last):
  File "/app/openslides_backend/migrations/migration_handler.py", line 102, in execute_migrate_command
    return handler.execute_command(command)
  File "/app/openslides_backend/migrations/migration_wrapper.py", line 76, in execute_command
    self.handler.finalize()
  File "/usr/local/lib/python3.10/site-packages/datastore/migrations/core/migration_handler.py", line 291, in finalize
    self.model_migrater.migrate()
  File "/usr/local/lib/python3.10/site-packages/datastore/migrations/core/migraters/model_migrater.py", line 29, in migrate
    events = migration.migrate(self.reader)
  File "/usr/local/lib/python3.10/site-packages/datastore/migrations/core/base_migrations/base_model_migration.py", line 16, in migrate
    return self.migrate_models()
  File "/app/openslides_backend/migrations/migrations/0081_check_equal_fields.py", line 63, in migrate_models
    if error := self.check_equal_data(
  File "/app/openslides_backend/migrations/migrations/0081_check_equal_fields.py", line 381, in check_equal_data
    eq_data_tup2 = self.get_equal_data_tuple(
  File "/app/openslides_backend/migrations/migrations/0081_check_equal_fields.py", line 403, in get_equal_data_tuple
    model = affected_models[id_]
KeyError: 2892
```

I'll look into it after

luisa-beerboom:
> I'll look into it after

Done

m-schieder (approved):

openslides_backend/migrations/migrations/0081_check_equal_fields.py
```diff
@@ -0,0 +1,413 @@
...
+    def get_relations(
+        self,
+    ) -> Relations:
+        relations: Relations = {}
+
+        for collection, fields in self.models.items():
+            for field, field_def in fields.items():
+                if eq := set(field_def.get("equal_fields", [])):
+                    if not field_def["is_relation"]:
+                        continue
```
vkrasnovyd: Is it really a possible case? In validate.py `equal_fields` is in valid_attributes only `if type in RELATION_TYPES`.

openslides_backend/migrations/migrations/0074_check_equal_fields.py
```diff
@@ -0,0 +1,413 @@
...
+    ) -> bool:
+        if field_def_a["is_generic"] != (generic_b := field_def_b["is_generic"]):
+            return generic_b
+        if field_def_a["is_list_relation"] != (
+            list_b := field_def_b["is_list_relation"]
+        ):
+            return list_b
+        if coll_a == coll_b:
+            return field_a <= field_b
+        return coll_a < coll_b
```
vkrasnovyd, line 194: It's so beautifully generic.

openslides_backend/migrations/migrations/0081_check_equal_fields.py
```diff
@@ -0,0 +1,413 @@
...
+                                    back_field_def,
+                                )
+                                or not back_field_def.get("equal_fields")
+                            ):
+                                full_eq_tup = tuple(sorted(full_eq))
+                                if collection == "meeting":
+                                    relations[
+                                        (back_collection, back_field),
+                                        (collection, field),
+                                    ] = (full_eq_tup, back_field_def, field_def)
```
vkrasnovyd: I didn't reach this code while running the migration_0081 tests with debugger. Are some cases missing or is it always not a main relation?

openslides_backend/migrations/migrations/0081_check_equal_fields.py
```diff
@@ -0,0 +1,413 @@
...
+    ) -> list[int]:
+        is_generic = field_def1["is_generic"]
+        missing_ids: set[int] = set()
+        for model in affected_models1.values():
+            if field_def1["is_list_relation"]:
+                for entry in model.get(field1) or []:
+                    if is_generic:
+                        coll, id_ = collection_and_id_from_fqid(entry)
+                        if coll != collection2:
+                            continue
```
vkrasnovyd: There seems to be no test case with:
- is_generic == True
- coll == collection2
- id_ not in affected_models2

openslides_backend/migrations/migrations/0074_check_equal_fields.py
```diff
@@ -0,0 +1,413 @@
...
+        if eq_data_tup1 != eq_data_tup2:
+            the_problem = " and ".join(
+                sorted(
+                    [
+                        f"{collection1}/{id1}/{field1}: {eq_data_tup1}",
+                        f"{collection2}/{id2}/{field2}: {eq_data_tup2}",
+                    ]
+                )
+            )
+            return f"Detected different equal_fields: {the_problem} for equal_fields {equal_fields}"
```
vkrasnovyd, line 404:
```suggestion
            return f"Detected different equal_fields: {the_problem} for equal_fields {equal_fields}\n"
```

Small improvement for readability in the console.

vkrasnovyd (changes requested):
Part 1. Still have to check the tests.

openslides_backend/migrations/migrations/0081_check_equal_fields.py
luisa-beerboom: Looked it up. It's not. Deleting these lines.

openslides_backend/migrations/migrations/0081_check_equal_fields.py
luisa-beerboom: Meeting seems to not have a single field marked with `equal_fields`, so `collection` can never be `"meeting"` and therefore these relations are only handled from the other side which will then get to this point using `not back_field_def.get("equal_fields")`.

I'm deleting this part of the code.

openslides_backend/migrations/migrations/0081_check_equal_fields.py
luisa-beerboom: coll in this case is the detected collection of the field value, therefore should be equal to at least one of the generic relations back collections.
The migration checks each collection/back-collection combo separately, so this line is to weed out any case that should be handled in another iteration of the relation-handling loop, for reason of relating to a collection that is not the currently processed back-collection.

Renaming it to `coll2` to make it more readable.

openslides_backend/migrations/migrations/0074_check_equal_fields.py
luisa-beerboom: They do get joined with an `\n` when the list of them is thrown in `MigrationException` is that not enough?

vkrasnovyd: Asked @m-schieder. Turns out, the way MigrationException is built can be not enough but this change also wouldn't make it better. So ignore this comment.

tests/system/migrations/test_0074_check_equal_fields.py
```diff
@@ -0,0 +1,700 @@
...
+    expect_errors_for: list[tuple[str, str, list[str]]] = []
+    collection_to_eq_fields: dict[str, set[str]] = defaultdict(set)
+    finished_combinations: set[tuple[tuple[str, str], tuple[str, str]]] = set()
+    for collection, fields in models.items():
+        id_ = collection_to_id[collection]
+        fqid = collection_to_fqid[collection]
+        for field, field_def in fields.items():
+            equal_fields = get_equal_fields(field_def)
+            if (not field_def["is_relation"] or not equal_fields) and not (
+                collection == "meeting" and field == "admin_group_id"
```
vkrasnovyd, line 123: Why is this a special case?

tests/system/migrations/test_0074_check_equal_fields.py
```diff
@@ -0,0 +1,700 @@
...
+        create_data[fqid][field] = val
+    else:
+        while (collection == check_back_collection and id_ == back_id) or (
+            field in create_data[fqid] and val != create_data[fqid][field]
+        ):
+            id_ += 1
+            fqid = f"{collection}/{id_}"
+            if collection == check_back_collection and id_ == back_id:
+                id_ += 1
+                fqid = f"{collection}/{id_}"
```
vkrasnovyd: Is this inner check needed? Above (inside of `if is_list:`) one increment is enough and I was not able to reach this inner increment with debugger.

tests/system/migrations/test_0074_check_equal_fields.py
```diff
@@ -0,0 +1,700 @@
...
+    use_other_model: bool = False,
+) -> list[str]:
+    created_fqids = []
+    if field in model or field not in models[collection]:
+        return created_fqids
+    field_def = models[collection][field]
+    assert field_def["is_relation"] and not field_def["is_list_relation"]
+    to_coll, to_field = list(field_def["to"].items())[0]
+    to_field_def = models[to_coll][to_field]
+    if not to_field_def["is_list_relation"] or to_field_def["is_generic"]:
```
vkrasnovyd: Is ` or to_field_def["is_generic"]` really relevant here?
GenericRelationField also has `is_generic = True`, although it's not a list. Both RelationListField and GenericRelationListField have `is_list_field = True`.
Since later you update `to_model[to_field]` as a list, just the first part of this check seems to be more precise. Am I missing somethig?

m-schieder:
If a relation refers to a model that does not exist, the relation should be deleted.

luisa-beerboom:
> If a relation refers to a model that does not exist, the relation should be deleted.

Is it fine if that only happens for relations with equal_fields or does this cleanup also need to happen for other relations?

tests/system/migrations/test_0074_check_equal_fields.py
```diff
@@ -0,0 +1,718 @@
...
+    collection: str,
+    id_: int,
+    field: str,
+    field_def: dict[str, Any],
+    model: dict[str, Any],
+    val: str | int,
+    back_combinations: dict[str, str],
+    create_data: dict[str, dict[str, Any]],
+) -> None | tuple[str, str, list[str]]:
+    if isinstance(val, str):
```
vkrasnovyd, line 360: Is there a reason why you check here the type of the value instead of field_def['is_generic'] like in the other places?

tests/system/migrations/test_0081_edit_duplicate_unique_values.py
```diff
@@ -36,13 +36,13 @@ def do_test(write, finalize, assert_model, test_data: TestData) -> None:
         )
     if isinstance(expect_or_error, list):
         try:
-            finalize("0080_edit_duplicate_unique_values")
-            raise pytest.fail("Expected migration to fail. It didn't.")
+            finalize("0081_edit_duplicate_unique_values")
+            raise pytest.fail("Expected migration 80 to fail. It didn't.")
```
vkrasnovyd:
```suggestion
            raise pytest.fail("Expected migration 81 to fail. It didn't.")
```

tests/system/migrations/test_0074_check_equal_fields.py
```diff
@@ -0,0 +1,718 @@
...
+        *[
+            {"type": "create", "fqid": fqid, "fields": model}
+            for fqid, model in create_data.items()
+        ]
+    )
+    if expected_errors:
+        try:
+            finalize("0074_check_equal_fields")
+            raise pytest.fail(
+                f"Expected migration 81 to fail for changed fqids {break_eq_fields_for_fqids}. It didn't."
```
vkrasnovyd:
```suggestion
                f"Expected migration 74 to fail for changed fqids {break_eq_fields_for_fqids}. It didn't."
```

tests/system/migrations/test_0074_check_equal_fields.py
```diff
@@ -0,0 +1,718 @@
...
+    val: str | int,
+    create_data: dict[str, dict[str, Any]],
+    front: bool,
+    from_list: bool = False,
+) -> None:
+    if isinstance(val, int):
+        back_id = val
+        back_collection, back_field = list(field_def["to"].items())[0]
+        back_fqid = fqid_from_collection_and_id(back_collection, back_id)
+    else:
```
vkrasnovyd, line 482: Same as in get_expected_error_for_single: is there a reason why you check here the type of the value instead of field_def['is_generic']?

tests/system/migrations/test_0074_check_equal_fields.py
```diff
@@ -0,0 +1,718 @@
...
+    write(
+        *[
+            {"type": "create", "fqid": fqid, "fields": model}
+            for fqid, model in create_data.items()
+        ]
+    )
+    try:
+        finalize("0074_check_equal_fields")
+        raise pytest.fail(
+            "Expected migration 81 to fail for changed projection/100. It didn't."
```
vkrasnovyd:
```suggestion
            "Expected migration 74 to fail for changed projection/100. It didn't."
```

tests/system/migrations/test_0074_check_equal_fields.py
```diff
@@ -0,0 +1,718 @@
...
+        *[
+            {"type": "create", "fqid": fqid, "fields": model}
+            for fqid, model in create_data.items()
+        ]
+    )
+    write({"type": "delete", "fqid": (fqid := collection_to_fqid["group"])})
+    try:
+        finalize("0074_check_equal_fields")
+        raise pytest.fail(
+            f"Expected migration 81 to fail for deleted {fqid}. It didn't."
```
vkrasnovyd:
```suggestion
            f"Expected migration 74 to fail for deleted {fqid}. It didn't."
```

vkrasnovyd (changes requested):
Part 2 of the review. Tests seem to cover all the important types of broken data and I like how you grouped them in the most recent commit. Just minor questions to this part.

tests/system/migrations/test_0074_check_equal_fields.py
luisa-beerboom: Checker expects a meeting admin_group_id when checking meeting_mediafile access groups. If there is none, the checker crashes, so I'm ensuring there always is one here.

tests/system/migrations/test_0074_check_equal_fields.py
luisa-beerboom: I wrote that code because there was a certain (low) chance of the initial `id_+=1` could move the value to the same id the back-relation was previously moved to.
However, that possibility ought to be covered by the while loop, so I'm removing this if clause

tests/system/migrations/test_0074_check_equal_fields.py
luisa-beerboom: Seems to be a leftover from old code versions, removing the latter part.

tests/system/migrations/test_0074_check_equal_fields.py
luisa-beerboom: Does the same thing and otherwise I'd have to `assert isinstance(val, str)` at the start of the clause anyway to stop linter from screaming. This is less code.

tests/system/migrations/test_0081_edit_duplicate_unique_values.py
luisa-beerboom: Done

tests/system/migrations/test_0074_check_equal_fields.py
luisa-beerboom: Done

tests/system/migrations/test_0074_check_equal_fields.py
luisa-beerboom: see above

tests/system/migrations/test_0074_check_equal_fields.py
luisa-beerboom: Done

tests/system/migrations/test_0074_check_equal_fields.py
luisa-beerboom: Deleted in `Clean up equal-fields-relations` commit

openslides_backend/migrations/migrations/0074_check_equal_fields.py
```diff
@@ -385,14 +404,47 @@ def check_equal_data(
...
+                            )
+                        ]
+                    }
+                },
+            )
+        else:
+            return RequestUpdateEvent(
+                fqid=fqid_from_collection_and_id(collection1, id1),
+                fields={field1: None},
+            )
```
vkrasnovyd, line 436: None of the generated test cases reaches this part. How about projector-projection relation?

vkrasnovyd (changes requested):
Code looks good. One final request.

openslides_backend/migrations/migrations/0074_check_equal_fields.py
luisa-beerboom: Done

tests/system/migrations/test_0074_check_equal_fields.py
```diff
@@ -0,0 +1,827 @@
...
+            {"type": "create", "fqid": fqid, "fields": model}
+            for fqid, model in create_data.items()
+        ]
+    )
+    fqid = collection_to_fqid["projector"]
+
+    def get_reduced_projector_ids(
+        collection: str, field: str, projector_field: str
+    ) -> dict[int, set[str]]:
+        return {id_: {field} for id_ in create_data[fqid][projector_field]}
```
vkrasnovyd: Is the transformation from `field: str` into `{field}` really needed even if we want to keep this as generic as possible?

luisa-beerboom: Changed it.

vkrasnovyd (approved):
Code looks good to me.

tests/system/migrations/test_0074_check_equal_fields.py
luisa-beerboom: Right, changing that.

m-schieder (approved):

PR #3526, opened 2026-04-27
Add prevent_updates trigger

vkrasnovyd:
Adjusts backend to the changes in https://github.com/OpenSlides/openslides-meta/pull/489:

1. Generate `constant = True` from both `constant: true` and `constant_legacy: true`.
2. Got rid of errors in the test data setup when we firstly create an instance with unchangeable data (mainly group.meeting_id, meeting_committee_id and different instances with constant meeting_user_id) and then update it.
3. Deleted tests for 1_1 not_null trigger on instance update: all of them are defined on constant_strict-fields and are now not possible. Added tests for prevent_updates instead.


tests/database/writer/system/test_update.py
```diff
@@ -215,57 +215,76 @@ def test_update_error_own_field_not_null(
     assert_model("user/1", {"id": 1, "username": "1", "first_name": "1"})
 
 
-def create_models_for_1_1_tests() -> None:
```
hjanott, line 218: I'm under the impression that there is no functional replacement for this test.

tests/system/action/user/test_delete.py
```diff
@@ -474,13 +474,15 @@ def test_delete_scope_organization_no_permission(self) -> None:
...
-        self.setup_admin_scope_permissions(UserScope.Organization)
         self.setup_scoped_user(UserScope.Organization)
+        self.setup_admin_scope_permissions(UserScope.Organization)
         response = self.request("user.delete", {"id": 111})
         self.assert_status_code(response, 200)
         self.assert_model_not_exists("user/111")
 
     def test_delete_scope_organization_permission_in_committee(self) -> None:
+        self.create_meeting()
+        self.create_meeting(4)
```
hjanott, line 485: Why?

tests/system/action/user/test_generate_new_password.py
```diff
@@ -116,15 +116,17 @@ def test_scope_organization_no_permission(self) -> None:
...
+        self.setup_admin_scope_permissions(UserScope.Organization)
         response = self.request("user.generate_new_password", {"id": 111})
         self.assert_status_code(response, 200)
         user = self.get_model("user/111")
         assert user.get("password") and user.get("default_password")
         self.assert_logged_in()
 
     def test_scope_organization_permission_in_committee(self) -> None:
+        self.create_meeting()
+        self.create_meeting(4)
```
hjanott, line 129: Also, why?

tests/system/action/user/test_set_password.py
```diff
@@ -158,6 +158,8 @@ def test_scope_organization_permission_in_organization(self) -> None:
         self.assert_logged_in()
 
     def test_scope_organization_permission_in_committee(self) -> None:
+        self.create_meeting()
+        self.create_meeting(4)
```
hjanott, line 162: Why?

tests/system/action/meeting/test_delete.py
```diff
@@ -10,11 +10,8 @@
 
 
 class MeetingDeleteActionTest(BaseActionTestCase):
-    def setUp(self) -> None:
-        super().setUp()
-        self.create_meeting()
```
vkrasnovyd, line 15: Was conflicting with data from `self.load_example_data()` in `test_delete_full_meeting()`.

tests/system/action/user/scope_permissions_mixin.py
```diff
@@ -15,8 +15,6 @@ def setup_admin_scope_permissions(
         If no scope is given, the user has no permissions.
         Additionally creates a default admin for meetings 1 and 4 - user/777.
         """
-        self.create_meeting()
-        self.create_meeting(4)
```
vkrasnovyd, line 19: Moved to setup_scoped_user() because in that method is defined in which committee should the meeting be created.

tests/system/action/user/test_merge_together.py
```diff
@@ -1578,13 +1578,9 @@ def test_merge_with_motion_submitters_and_supporters(
 
     def test_merge_with_personal_notes(self) -> None:
         # create personal notes, motions, and meetings
-        meeting_ids = list(range(1, 8, 3))
-        for meeting_id in meeting_ids:
-            self.create_meeting(meeting_id)
```
vkrasnovyd, line 1583: `self.create_meeting()` was overriding committee_id.

tests/system/action/user/test_participant_import.py
```diff
@@ -1412,7 +1412,7 @@ def test_json_upload_permission_as_locked_out(self) -> None:
         response = self.request("participant.import", {"id": 1, "import": True})
         self.assert_status_code(response, 403)
         self.assertIn(
-            "You are not allowed to perform action participant.import. Missing permissions: Permission user.can_manage in meeting 1 or OrganizationManagementLevel can_manage_organization in organization 1 or CommitteeManagementLevel can_manage in committee 1",
+            "You are not allowed to perform action participant.import. Missing permissions: Permission user.can_manage in meeting 1 or OrganizationManagementLevel can_manage_organization in organization 1 or CommitteeManagementLevel can_manage in committee 60",
```
vkrasnovyd, line 1415: Because we no longer override `committee_id` in `json_upload_dont_recognize_empty_name_and_email`.

tests/system/action/user/test_participant_json_upload.py
```diff
@@ -9,8 +9,7 @@
 
 
 class ParticipantJsonUpload(BaseActionTestCase):
-    def setUp(self) -> None:
-        super().setUp()
```
vkrasnovyd, line 13: Used to be conflicts about `group/7` between `setUp()` and `prepare_locked_out_test()`.

tests/system/action/group/test_delete.py
```diff
@@ -61,30 +61,24 @@ def test_delete_with_users(self) -> None:
 
     def test_delete_no_permissions(self) -> None:
         self.base_permission_test(
-            {
-                "group/24": {"name": "name_srtgb123", "meeting_id": 1},
-            },
+            {"group/4": {"name": "name_srtgb123", "meeting_id": 1}},
```
hjanott, line 64: Why couldn't you create group 24?

tests/system/action/meeting/test_delete.py
hjanott: I would prefer to extract that single test case into its own test class then. But I leave that for you to decide.

tests/system/action/poll/test_start.py
```diff
@@ -91,8 +91,27 @@ def test_start_poll(self) -> None:
...
+                    "votesvalid": Decimal("0.000000"),
+                    "votescast": Decimal("0.000000"),
+                    "backend": "fast",
+                    "pollmethod": "YNA",
+                    "type": Poll.TYPE_NAMED,
+                    "onehundred_percent_base": "YNA",
+                },
+            }
+        )
+        response = self.request("poll.start", {"id": 2})
```
hjanott, line 114: Why did all this have to be added?

tests/system/action/poll/test_update.py
```diff
@@ -769,18 +759,34 @@ def base_test_live_voting(
...
-        if is_motion_poll:
-            self.create_motion(1, 3)
-            self.set_models({"motion_state/1": {"allow_create_poll": True}})
-            poll_updates["content_object_id"] = "motion/3"
-        elif poll_changes:
-            poll_updates.update(poll_changes)
-        self.update_model(f"poll/{poll_id}", poll_updates)
+        poll_data: dict[str, Any] = {"type": poll_type, **poll_changes}
+
+        if poll_id == 1 and "content_object_id" not in poll_changes:
```
hjanott: Please write a comment why you behave differently for poll 1.

tests/system/action/speaker/test_update.py
```diff
@@ -65,15 +70,8 @@ def test_update_contribution_ok(self) -> None:
...
-        self.create_meeting()
-        self.set_models(
-            {
-                "user/1": {"organization_management_level": None},
-                "meeting_user/7": {"user_id": 1},
-            }
-        )
-        self.set_user_groups(1, [3])
-        self.set_group_permissions(3, [Permissions.ListOfSpeakers.CAN_SEE])
+        self.login(7)
```
hjanott, line 73: Why wasn't the user logged in before?

tests/system/action/user/test_create.py
```diff
@@ -78,24 +78,8 @@ def test_create_some_more_fields(self) -> None:
...
-                },
-                "meeting/114": {
-                    "name": "name_xXRGTLAJ",
-                    "committee_id": 79,
-                    "is_active_in_organization_id": 1,
-                },
-                "committee/78": {"name": "name_TSXpBGdt", "meeting_ids": [110]},
-                "committee/79": {"name": "name_hOldWvVF", "meeting_ids": [114]},
-            }
-        )
```
hjanott, line 98: There is a lot to do but it also made some places more beautiful.

tests/system/action/user/test_participant_json_upload.py
```diff
@@ -9,8 +9,7 @@
...
+    def set_up_meeting_1(self) -> None:
```
hjanott:
```suggestion
    def set_up_test_models(self) -> None:
```

tests/system/action/user/test_participant_json_upload.py
```diff
@@ -2041,7 +2060,7 @@ def json_upload_no_permissions_to_set_meeting_external_fields_on_superadmin(
                 },
                 "committee/60": {"meeting_ids": [1, 4]},
                 "meeting/1": {"meeting_user_ids": [11]},
-                "meeting/4": {"meeting_user_ids": [44], "committee_id": 60},
+                "meeting/4": {"meeting_user_ids": [44]},
```
hjanott: This can be deleted completely since the data is in `meeting_user`.

tests/system/action/user/test_assign_meetings.py
```diff
@@ -1,48 +1,27 @@
...
-
     def test_assign_meetings_correct(self) -> None:
+        meeting_data: dict[int, dict[str, Any]] = {
+            1: {"name": "success(existing)"},
+            4: {"name": "nothing"},
+            7: {"name": "success(added)", "committee_id": 63},
+            10: {"name": "standard", "committee_id": 63},
+            13: {"name": "success(added)", "committee_id": 63},
+        }
+        for i in range(1, 14, 3):
```
hjanott: Better to iterate over `meeting_data.items()`. Same below.

hjanott (changes requested):

tests/database/writer/system/test_update.py
vkrasnovyd: Yes, like I wrote in the PR description, there are no cases of not_null_1_1 triggers where none of the sides has `constant: true`.

tests/system/action/group/test_delete.py
vkrasnovyd: In setUp we are creating `meeting/22`. `group/24` belongs to it and can not be assigned to another meeting.

tests/system/action/user/test_delete.py
vkrasnovyd: We used to create these meetings in `setup_admin_scope_permissions()`, now we do it in `setup_scoped_user()` to avoid moving meeting/4 to the committee of meeting/1. We don't call `setup_scoped_user()`. Therefore, if we want to preserve the original test setup, meetings have to be created manually.

tests/system/action/user/test_generate_new_password.py
vkrasnovyd: Same as above.

tests/system/action/user/test_set_password.py
vkrasnovyd: Same as above.

tests/system/action/poll/test_start.py
vkrasnovyd: `poll.content_object_id` is constant. For the new `content_object_id` a separate poll has to be created.

tests/system/action/speaker/test_update.py
vkrasnovyd: Because we were changing `user_id` of the `meeting_user/7` (originally created in setUp) to have the same request user as instance user.

Another solution for these cases could be creating a new meeting_user but I prefer to re-use the existing one.

tests/system/action/user/test_participant_json_upload.py
vkrasnovyd: Done.

tests/system/action/poll/test_update.py
vkrasnovyd: Done.

tests/system/action/user/test_assign_meetings.py
vkrasnovyd: Agree, it's overcomplicated. Changed.

tests/system/action/user/test_participant_json_upload.py
vkrasnovyd: Removed these and other back relations.

tests/system/action/meeting/test_delete.py
vkrasnovyd: Done.

tests/database/writer/system/test_update.py
hjanott: I would disagree in terms of that it is still possible to trigger this during a create.

tests/system/action/poll/test_update.py
```diff
@@ -769,18 +759,38 @@ def base_test_live_voting(
...
```
hjanott: Now I get why you are doing this but I would say something like this may be beneficial.
```suggestion
        if poll_id == 1 and ("content_object_id" not in poll_changes or poll_changes["content_object_id"] == "assignment/1"):
```

tests/system/action/poll/test_update.py
```diff
@@ -769,18 +759,38 @@ def base_test_live_voting(
...
+            # and content_object_id is equal to poll/1/content_object_id
+            self.update_model("poll/1", poll_data)
+            poll_id = 1
+        else:
+            if is_motion_poll:
+                self.create_motion(1, 3)
+                self.set_models({"motion_state/1": {"allow_create_poll": True}})
+                poll_data["content_object_id"] = "motion/3"
+            if poll_id == 1:
+                poll_id = 2  # Default if if no custom poll_id is provided
```
hjanott:
```suggestion
                poll_id = 2  # Default if no custom poll_id is provided
```

hjanott (changes requested):
In test [test_assign_meetings.py](https://github.com/OpenSlides/openslides-backend/pull/3526/changes#diff-5dab2cb7ff886434cdf05bc088e75d1a0b3d5dacc69a96b0d7d4789a157f61d1) there is also a deleted `setUp`. Do you think its worthy to consider the same solution as in the other case?

tests/database/writer/system/test_update.py
vkrasnovyd: I didn't touch tests for insert and delete. Not null 1:1 trigger is still tested for these operations.

vkrasnovyd:
> In test [test_assign_meetings.py](https://github.com/OpenSlides/openslides-backend/pull/3526/changes#diff-5dab2cb7ff886434cdf05bc088e75d1a0b3d5dacc69a96b0d7d4789a157f61d1) there is also a deleted `setUp`. Do you think its worthy to consider the same solution as in the other case?

No, I'd rather keep it as it is now. We create `meeting/1` in a bunch of different ways in that file and we actully need different amount of meetings in different cases. Attempts to generalize such setup would make code more complicated.

tests/database/writer/system/test_update.py
hjanott: You are absolutely right. I was reading this as a test, which it is not.

hjanott (approved):
All green.

PR #3529, opened 2026-04-27
Use new trigger names in tests

hjanott:


tests/system/migrations/test_0100_init_reldb.py
```diff
@@ -220,7 +220,9 @@ def assert_content_not_none(
             if error_message:
                 assert result, error_message
             else:
-                assert result
+                assert (
+                    result
+                ), f"Database did not contain a result for this query.\n{query}"
```
vkrasnovyd, line 225: Why do we need this change now?

hjanott: It's just a nicer debugging output. Instead of `assert None` this line plus the used query will be on the console which makes it clear at a glance where the error comes from.

vkrasnovyd (approved):

PR #3531, opened 2026-04-27
Introduce base migration class

hjanott:
This adds a base migration class.

### DIFF
--- a/openslides_backend/migrations/migrations/base.py
+++ b/openslides_backend/migrations/migrations/base.py
@@ -17,8 +17,8 @@ class BaseMigration:
     @staticmethod
     def data_definition(curs: Cursor[DictRow]) -> None:
         """
-        Apllies all manual SQL DDL changes necessary.
-        Triggers are automatically recreated.
+        Applies all manual SQL DDL changes necessary.
+        Triggers and views are automatically recreated.
         """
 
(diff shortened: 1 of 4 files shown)


openslides_backend/migrations/migrations/base.py
```diff
@@ -0,0 +1,40 @@
...
+            cursor
+        """
+
+    @staticmethod
+    def cleanup(curs: Cursor[DictRow]) -> None:
+        """
+        Purpose:
+            Deletes leftovers of the migration.
+        Input:
+            cursor
```
luisa-beerboom: Descriptions should probably include that the functions should be overridden in the import class

luisa-beerboom: Could also use the @abstractmethod decorator for things that'll always need to be overridden

luisa-beerboom (changes requested):
No linked issue, no description. Please add at least one of these.

openslides_backend/migrations/migrations/base.py
hjanott: Will add that in the description. I don't think these need to always be overridden. That would lead to many empty function declarations.

luisa-beerboom (approved):

PR #3549, opened 2026-05-07
SAML meeting mapping always allow create if `meeting_user` not meeting member

### LINKED ISSUE #3496: Change SAML mapping behaviour for allow update and groups

luisa-beerboom:
Closes #3496 
Based on #3514 

TODO: 
- [x] Still need to do most of the requested changes in #3514


tests/system/action/user/test_save_saml_account.py
```diff
@@ -1136,26 +1136,26 @@ def test_update_user_with_default_membership(self) -> None:
...
         )
 
-    def test_update_user_participant_already_in_group(self) -> None:
+    def test_update_user_participant_already_in_another_meeting(self) -> None:
         """
         Shows:
             * user stays in group 2
             * structure_level/1 left untouched structure_level/2 added
-            * second mapper is ignored due to being an update and allow_update being false
+            * second mapper is used in spite of allow_update being false since 'Kreistag meeting' is a create
```
vkrasnovyd:
```suggestion
            * second mapper is used in spite of allow_update being false since 'Kreistag' meeting is a create
```

tests/system/action/user/test_save_saml_account.py
```diff
@@ -1173,6 +1173,160 @@ def test_update_user_participant_already_in_group(self) -> None:
...
+        self.assert_model_exists(
+            "structure_level/3", {"name": "structure2", "meeting_user_ids": [2]}
+        )
+        self.assert_model_not_exists("structure_level/4")
+
+    def test_update_user_participant_already_in_not_mapped_meeting(self) -> None:
+        """
+        Shows:
+            * user stays in group 2
+            * second mapper is used in spite of allow_update being false since 'Kreistag meeting' is a create
```
vkrasnovyd:
```suggestion
            * second mapper is used in spite of allow_update being false since 'Kreistag' meeting is a create
```

tests/system/action/user/test_save_saml_account.py
```diff
@@ -1136,26 +1136,26 @@ def test_update_user_with_default_membership(self) -> None:
             "group/5", {"meeting_user_ids": [2], "external_id": "Delegates"}
         )
 
-    def test_update_user_participant_already_in_group(self) -> None:
+    def test_update_user_participant_already_in_another_meeting(self) -> None:
```
vkrasnovyd, line 1139: Looks like now it's a create case, not an update.

tests/system/action/user/test_save_saml_account.py
```diff
@@ -1173,6 +1173,160 @@ def test_update_user_participant_already_in_group(self) -> None:
...
+        )
+        self.assert_model_exists(
+            "structure_level/2", {"name": "structure2", "meeting_user_ids": [1]}
+        )
+        self.assert_model_exists(
+            "structure_level/3", {"name": "structure2", "meeting_user_ids": [2]}
+        )
+        self.assert_model_not_exists("structure_level/4")
+
+    def test_update_user_participant_already_in_not_mapped_meeting(self) -> None:
```
vkrasnovyd, line 1221: Same here. Seems to be a create case, not update.

openslides_backend/action/actions/user/save_saml_account.py
```diff
@@ -504,14 +525,15 @@ def get_group_ids(self, group_names: list[str], meeting: dict) -> list[int]:
             )
             if len(groups) > 0:
                 return sorted(groups)
-        if default_group_id := meeting["default_group_id"]:
+        # Because we don't know the db instance here, we can't determine if the meeting_user has no groups which leads to remaining without any group.
```
vkrasnovyd: This comment seems to be describing the old database structure, before https://github.com/OpenSlides/openslides-backend/pull/3206.

tests/system/action/user/test_save_saml_account.py
vkrasnovyd: Test description does not make the purpose of the test completely clear: 'is a create' is not explained in this file and is not the term used in the action or its docs.

This docstring can be improved, for example, by mentioning the exact reason why second mapper is used (becuse user is not in 'Kreistag' meeting) or by refering to 'for_create' used in the action code. Maybe you have a better idea.

tests/system/action/user/test_save_saml_account.py
vkrasnovyd: Same as above about the docstring clarity.

tests/system/action/user/test_save_saml_account.py
```diff
@@ -1203,8 +1357,56 @@ def test_update_user_participant_already_in_group(self) -> None:
...
+        self.assert_model_exists(
+            "group/2", {"meeting_user_ids": [1], "external_id": "Delegates"}
+        )
+
     def test_update_user_add_group_to_existing_groups(self) -> None:
-        """group added, user created and logged in"""
+        """
+        Shows:
+            * group added
+            * user created and logged in
```
vkrasnovyd, line 1407: There's no login in this test case.

vkrasnovyd (changes requested):
Code looks okay. I'd like the documentation to be more precise.

tests/system/action/user/test_save_saml_account.py
luisa-beerboom: It's not creating a new user, just a new meeting_user. Since this is a user action, in this case that would be an update.

tests/system/action/user/test_save_saml_account.py
luisa-beerboom: See above

openslides_backend/action/actions/user/save_saml_account.py
luisa-beerboom: Deleted the comment

tests/system/action/user/test_save_saml_account.py
luisa-beerboom: I edited the description

tests/system/action/user/test_save_saml_account.py
luisa-beerboom: The logged in-part is a carry-over from before this PR (see previous description).

I'm guessing it's because the added user is `user/1` who carries out the test actions by default.

openslides_backend/action/actions/user/save_saml_account.py
```diff
@@ -381,6 +381,7 @@ def apply_meeting_user_data(
                     names = sorted(
                         instance_meeting_user.pop(saml_meeting_user_field, [])
                     )
+                    ids = []
```
vkrasnovyd, line 384: We don't need this fallback: both `get_group_ids` and `get_structure_level_ids` return either list[int] or empty list.

tests/system/action/user/test_save_saml_account.py
vkrasnovyd: Thanks. Much more understandable now.

tests/system/action/user/test_save_saml_account.py
vkrasnovyd: Not obvious but reasonable.

vkrasnovyd (changes requested):

openslides_backend/action/actions/user/save_saml_account.py
luisa-beerboom: This isn't really a fallback, this is because the checker screams `value possibly unbound`, due to the match case, which is setting `ids`, technically having branches that don't set it (due to having no default).
I could give the match_case a default branch that sets it to `[]` instead or turn the match case into an if-else-clause, but I didn't really see the point in those either.
What do you think?

vkrasnovyd: I agree with the chosen approach. Let's keep it this way.

vkrasnovyd (approved):

PR #3551, opened 2026-05-08
Cleanup duplicate mixins

### LINKED ISSUE #3351: Remove DuplicateCheckMixin
Backend currently has mixins with name `DuplicateCheckMixin` in 2 files: 
* `openslides_backend/action/actions/user/user_mixins.py`
* `openslides_backend/action/actions/topic/mixins.py`

At this point they are not used in any Action class.
- [x] Check whether all the checks performed in mixins are either implemented in other mixins or directly in actions or not relevant anymore
- [x] Delete these 2 mixins
- [ ] Remove `execute_presenter` method of `Action` (used only in `user.DuplicateCheckMixin`

luisa-beerboom:
Closes #3351


openslides_backend/action/actions/topic/mixins.py
```diff
@@ -1,18 +0,0 @@
...
-            for values in self.datastore.filter(
-                "topic", FilterOperator("meeting_id", "=", meeting_id), ["title"]
-            ).values()
-        }
-
-    def check_for_duplicate(self, title: str) -> bool:
-        result = title in self.all_titles_in_meeting
-        if not result:
-            self.all_titles_in_meeting.add(title)
-        return result
```
vkrasnovyd, line 18: In the import and json_upload actions I've found the check for the unique title: `self.topic_lookup.check_duplicate(title := entry["title"])`. However, it does not seem to exist for the create or update actions. In this case we probably don't want a database constraint for it, right?

luisa-beerboom: We don't.
This is just the internal logic of the imports, there is no actual implied uniqueness constraint for the topics and so on.

vkrasnovyd (approved):

PR #3566, opened 2026-05-12
Don't duplicate mediafiles in forward if it is disabled

### LINKED ISSUE #3225: Mediafiles get duplicated in mediaservice with organization.disable_forward_with_attachments
Occurs on attempt to forward motion in the organization with the setting `organization.disable_forward_with_attachments`.

**How to reproduce:**
In the end of the test `test_forward_with_attachment_true_and_forward_with_attachments_disabled` add check `self.media.duplicate_mediafile.assert_not_called()`.

**Current behaviour:**
```
AssertionError: Expected 'duplicate_mediafile' to not have been called. Called 2 times.
Calls: [call(3, 8), call(4, 9)].
```
Mediafiles don't get forwarded in the database but meeting-wide mediafiles get duplicated in the mediaservice.

**Expected behaviour:**
When mediafiles can not get forwarded, they should not be duplicated.

luisa-beerboom:
Closes #3225


vkrasnovyd (approved):
Great improvement.

PR #3572, opened 2026-05-13
compatibility changes for `osmanage` calling `migrations` route

peb-adr:
Revise output handling for stats and progress cmd

progress: output is obtained via `StringIO.read()`, giving output since last read
stats: output is obtained via `StringIO.getvalue()`, giving *all* output since beginning

Remove state `migrations_preparing`

Also remove the - albeit well-intended yet troublesome - feature of preserving output until next is written (using the bool flag `migrate_thread_stream_just_read`).
This way the caller of progress (in practice being `osmanage`) can naively just print whatever progress returns, building coherent output.


openslides_backend/migrations/migrations/0100_init_reldb.py
```diff
@@ -269,6 +270,7 @@ def check_prerequisites(curs: Cursor[DictRow]) -> str:
         if time_zone not in (pg_time_zones):
             errors += f"{time_zone} is no accepted value for MIG0100_TIMEZONE. Please refer to the documentation on how to obtain a full list of all options available."
         if errors:
+            MigrationHelper.write_line(errors)
```
hjanott, line 273: I would suggest to do this in the migration handler if this a behaviour you want for all migrations either way.

openslides_backend/migrations/migration_helper.py
```diff
@@ -71,35 +70,32 @@ class MigrationHelper:
...
     @staticmethod
-    def read_stream() -> str:
+    def read_stream(all: bool = False) -> str:
         """
-        Reads all lines since reading last.
-        Also signals the write process on the buffer that it was just read.
-        This is to preserve lines for read until new lines were written.
+        Reads all lines since the last time reading.
+        If `all` is set, all lines present in buffer are returned without
+        moving the cursor at all.
```
hjanott:
```suggestion
        If `all` is set, all lines present in buffer are returned without
        moving the cursor at all. Doesn't move `MigrationHelper.migrate_thread_stream_read_pos`.
```

openslides_backend/migrations/migration_helper.py
```diff
@@ -320,10 +316,9 @@ def get_migration_state(curs: Cursor[DictRow]) -> MigrationState:
         if not states_and_indices:
             return MigrationState.FINALIZED
         states = {elem.get("migration_state") for elem in states_and_indices}
-        # 1. migration, 2. finalization | -> failed > (preparing >) running > required
+        # 1. migration, 2. finalization -> failed > running > required
```
hjanott:
```suggestion
        # 1. migration, 2. finalization | both: failed > running > required
```

openslides_backend/migrations/migrations/0100_init_reldb.py
peb-adr: Don't think it should be like this for all migrations.

openslides_backend/migrations/migration_helper.py
peb-adr: Included requested info in a slightly different way.

luisa-beerboom (approved):

PR #3590, opened 2026-05-21
Delete poll with vote adapter

vkrasnovyd:
- Instead of calling the internal action, send request to the vote_service to delete the poll;
- Skip deletion of the models from collections managed by vote service on cascade delete. They will be handled here along with the poll: https://github.com/OpenSlides/openslides-vote-service/blob/e1140fec06bc92b08e36a87516d6688c0c73379c/vote/vote.go#L548-L555;
- A bit of cleanup.

### DIFF
--- a/openslides_backend/services/request_methods.py
+++ b/openslides_backend/services/request_methods.py
@@ -1,0 +1,6 @@
+from enum import StrEnum
+
+
+class REQUEST_METHOD(StrEnum):
+    POST = "post"
+    DELETE = "delete"
--- a/openslides_backend/services/request_methods.py
+++ b/openslides_backend/services/request_methods.py
@@ -1,6 +1,0 @@
-from enum import StrEnum
-
-
-class REQUEST_METHOD(StrEnum):
-    POST = "post"
-    DELETE = "delete"
--- a/openslides_backend/services/vote/adapter.py
+++ b/openslides_backend/services/vote/adapter.py
@@ -5,6 +5,7 @@ import simplejson as json
 
 from ...shared.exceptions import VoteServiceException
 from ...shared.interfaces.logging import LoggingModule
+from ..request_methods import REQUEST_METHOD
 from ..shared.authenticated_service import AuthenticatedService
 from .interface import VoteService
 
@@ -21,7 +22,9 @@ class VoteAdapter(VoteService, AuthenticatedService):
     def retrieve(
         self,
         endpoint: str,
-        request_method: Literal["post", "delete"] = "post",
+        request_method: Literal[
+            REQUEST_METHOD.POST, REQUEST_METHOD.DELETE
+        ] = REQUEST_METHOD.POST,
         payload: dict[str, Any] | None = None,
     ) -> Any:
         response = self.make_request(endpoint, request_method, payload)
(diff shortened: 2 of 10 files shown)


openslides_backend/services/vote/adapter.py
```diff
@@ -18,27 +18,46 @@ def __init__(self, vote_url: str, logging: LoggingModule) -> None:
...
             self.logger.error(message)
             raise VoteServiceException(message)
         if response.content:
             return response.json()
 
-    def make_request(self, endpoint: str, payload: dict[str, Any] | None = None) -> Any:
+    def make_request(
+        self,
+        endpoint: str,
+        request_method: Literal["post", "delete"],
```
luisa-beerboom: How about turning this into a type or StrEnum?

openslides_backend/shared/export_helper.py
```diff
@@ -239,31 +239,31 @@ def export_meeting(
...
                     ).append(ballot_id)
 
-        config_ids = [poll.get("config_id") for poll in polls.values()]
-
-        for config_fqid in config_ids:
-            collection, id_ = config_fqid.split("/")
-            instance = datastore.get(
-                config_fqid,
-                list(get_fields_for_export(collection)),
+    config_ids = [poll.get("config_id") for poll in polls.values()]
```
luisa-beerboom: If you are going to move this outside the if-clause, you might want to ensure polls is set by adding a default, or move this back

openslides_backend/shared/export_helper.py
```diff
@@ -239,31 +239,31 @@ def export_meeting(
...
+
+    for config_fqid in config_ids:
+        collection, id_ = config_fqid.split("/")
```
luisa-beerboom: I can see that this is not your code so this is just a suggestion, but you could use `collection_and_id_from_fqid` here

luisa-beerboom (commented):
From my understanding this isn't done, which is why I won't complain about the user_merge test yet.

vkrasnovyd:
> From my understanding this isn't done, which is why I won't complain about the user_merge test yet.

user_merge needs to be discussed and more seriously updated because of the changes that have happened in the collections. I will handle user_merge and tests for it separately.

openslides_backend/services/vote/adapter.py
vkrasnovyd: Good idea. Added REQUEST_METHOD type with currently only 2 methods that are being used in the code.

openslides_backend/shared/export_helper.py
vkrasnovyd: Reverted this part. Was part of some `work in progress` changes that don't even belong here.

openslides_backend/services/vote/adapter.py
```diff
@@ -42,14 +45,14 @@ def retrieve(
     def make_request(
         self,
         endpoint: str,
-        request_method: Literal["post", "delete"],
+        request_method: Literal[REQUEST_METHOD.POST, REQUEST_METHOD.DELETE],
```
luisa-beerboom: Can't you just write `request_method: REQUEST_METHOD,`?

openslides_backend/services/request_methods.py
```diff
@@ -0,0 +1,6 @@
+from enum import StrEnum
+
+
+class REQUEST_METHOD(StrEnum):
```
luisa-beerboom: I think `RequestMethod` may be a better name. We don't usually fully capitalize StrEnums.

vkrasnovyd: Done.

openslides_backend/services/vote/adapter.py
vkrasnovyd: Doing this with the StrEnum-class shared for all the services seems to be not safe to me (in case we ever have to update it). As we didn't need it for other services, moved it to the file with VoteAdapter and made the suggested change.

luisa-beerboom (approved):
Might be a good idea to create a follow-up issue for the user merge

luisa-beerboom:
Also maybe take a look at the linting before merging

vkrasnovyd:
> Also maybe take a look at the linting before merging

Checked. All of them are getting fixed in user.merge_together PR or on tests cleanup.

PR #3599, opened 2026-06-02
[vote] Update user.merge_together

vkrasnovyd:
## Changes in collections, reflected in the action code
Deleted fields:
- poll.entitled_users_at_stop (relevant for user.merge_together action)
- user.poll_candidate_ids (along with `poll_candidate_list` and `poll_candidate` collections).
    
Updated relational fields:
- user.poll_voted_ids -> meeting_user.poll_voted_id
- user.option_ids -> meeting_user.poll_option_ids (to: option -> poll_option)
- user.vote_ids -> meeting_user.represented_ballot_ids (to: vote.user_id -> poll_ballot.represented_meeting_user_id)
- user.delegated_vote_ids -> meeting_user.acting_ballot_ids (to: vote.delegated_user_id -> poll_ballot.acting_meeting_user_id)

## Other changes in user.merge_together action
- Merged `UserMergeTogether.check_polls` into `MeetingUserMergeMixin.check_polls_helper` and renamed it into `check_polls`.
- Removed `get_meeting_ids_per_user()` and replaced it with just `user.get("meeting_ids", [])`. Reason: `user.meeting_ids` are getting calculated from `meeting_user` instances in the database and meeting_users must be in the groups after https://github.com/OpenSlides/openslides-backend/pull/3206.

## Changes to the user.merge_together tests
All the test data is now being written directly into the db, without requests to vote service. Reason: new implementation of vote service relies on notify triggers for updating the cached data, which are disabled in the test framework.


openslides_backend/action/actions/user/base_merge_mixin.py
```diff
@@ -98,7 +98,8 @@ def add_collection_field_groups(
         self._all_collection_fields[collection] = [
             i
             for i in Class.__dict__.keys()
-            if i[:1] != "_" and i not in ["collection", "verbose_name", "id"]
+            if i[:1] != "_"
+            and i not in ["collection", "verbose_name", "id", "managed_by"]
```
luisa-beerboom, line 102: Is managed_by a field that every model has now?

openslides_backend/action/actions/user/merge_mixins.py
```diff
@@ -275,26 +280,37 @@ def get_merge_comparison_hash(
...
+                        "vote_delegated_to_id",
+                        "poll_voted_ids",
+                        "poll_option_ids",
+                        "acting_ballot_ids",
+                        "represented_ballot_ids",
                     ],
                 )
             ]
-        ).get("meeting_user", {})
+        )["meeting_user"]
```
luisa-beerboom, line 312: Is this going to work if the data has no meeting users?

openslides_backend/action/actions/user/merge_mixins.py
```diff
@@ -371,4 +387,71 @@ def check_polls_helper(self, meeting_user_ids: list[int]) -> list[str]:
...
-        return messages
+
+        ballot_poll_ids_per_user_id: dict[int, set[int]] = {}
+        option_poll_ids_per_user_id: dict[int, set[int]] = {}
+        for meeting_user in meeting_users.values():
+            if poll_voted_ids := meeting_user.get("poll_voted_ids"):
+                ballot_poll_ids_per_user_id.setdefault(
+                    meeting_user["user_id"], set()
+                ).update(set(poll_voted_ids))
+        for meeting_user in meeting_users.values():
```
luisa-beerboom: Why can't the content of the first loop happen in the beginning of the second?

openslides_backend/action/actions/user/merge_mixins.py
```diff
@@ -371,4 +387,71 @@ def check_polls_helper(self, meeting_user_ids: list[int]) -> list[str]:
...
+                            ]
+                        }
+                    )
+                )
+            ):
+                get_many_requests = [
+                    GetManyRequest("poll_option", o_ids, ["poll_id"]),
+                    GetManyRequest("poll_ballot", b_ids, ["poll_id"]),
+                ]
+                many_models = self.datastore.get_many(get_many_requests)
```
luisa-beerboom: Put the request data into the datastore call directly

openslides_backend/action/actions/user/merge_mixins.py
```diff
@@ -371,4 +387,71 @@ def check_polls_helper(self, meeting_user_ids: list[int]) -> list[str]:
...
+                ballot_poll_ids_per_user_id.setdefault(
+                    meeting_user["user_id"], set()
+                ).update({ballot["poll_id"] for ballot in ballot_data.values()})
+        ballot_conflicts = {
+            poll_id
+            for id1, poll_ids1 in ballot_poll_ids_per_user_id.items()
+            for id2, poll_ids2 in ballot_poll_ids_per_user_id.items()
+            for poll_id in poll_ids1.intersection(poll_ids2)
+            if id1 != id2
+        }
```
luisa-beerboom: I don't know if I like the two nested loops on `ballot_poll_ids_per_user_id.items()`.
Efficiency-wise, it may be better to loop across ballot_poll_ids_per_user_id once and count the usages of each poll_id across the dicts sets  in another (default)dict. Then loop across that dict and use any poll_id that is present more than once. Would turn this from a `n²*m` runtime to a `n*m + m` runtime, with n being the amount of users and m being the amount of polls.

luisa-beerboom: Same below btw.

luisa-beerboom (changes requested):
I have only reviewed the base and main class  and the merge_mixins so far.
The other three files will probably have to wait until tomorrow

openslides_backend/action/actions/user/base_merge_mixin.py
vkrasnovyd: Yes, it gets always generated by `generate_models`.

openslides_backend/action/actions/user/merge_mixins.py
vkrasnovyd: Good point. Since I've already did some cleanup in this action, can also improve this part.

openslides_backend/action/actions/user/merge_mixins.py
vkrasnovyd: Yes, it works without meeting_users. `get_many()` uses `defaultdict`.

openslides_backend/action/actions/user/merge_mixins.py
vkrasnovyd: Changed and moved to helper method.

openslides_backend/action/actions/user/merge_mixins.py
vkrasnovyd: Done.

openslides_backend/action/actions/user/merge_mixins.py
vkrasnovyd: It absolutely can. Changed it.

openslides_backend/action/actions/user/merge_mixins.py
```diff
@@ -455,3 +447,14 @@ def check_polls(self, into: PartialModel, other_models: list[PartialModel]) -> N
             raise ActionException(
                 f"Cannot carry out merge into user/{into['id']}, because {' and '.join(messages)}"
             )
+
+    def _get_conflicts_between_users(self, ids_map: dict[int, set[int]]) -> list[int]:
+        seen_ids = []
+        duplicates = []
```
luisa-beerboom: It would be better to use sets, the membership test for `seen_ids` is faster (because sets are implemented with hash tables) and you don't have to worry about duplicates in `duplicates`.

luisa-beerboom (changes requested):
Just this one thing, after that I'll gladly approve.

openslides_backend/action/actions/user/merge_mixins.py
vkrasnovyd: Done.

luisa-beerboom (approved):

PR #3608, opened 2026-06-04
[vote] Complete export_helper, tests and docs

vkrasnovyd:
Note that some of the old code (especially in the tests) is not from main but from https://github.com/OpenSlides/openslides-backend/pull/3167

Most common changes are based on the changed collections:
- (old) poll_config_option -> (new) poll_option
- (old) vote or ballot -> (new) poll_ballot
- (old) relation between option and poll through poll_config -> (new) poll_option has own `poll_id` field and has no direct relation to the poll_config
- for all poll_config_* collections: (old) no required fields -> (new) onehundred_percent_base is required
- moved fields from poll to config collections
- changes in available options for onehundred_percent_base
- deleted meeting.*_default_backend fields

Extra:
- started creating polls and related models manually, without requests to vote service, deleted unused methods (according to the concept, only delete requests can be sent from the backend to the new vote service);
- temporary commented out migration tests and decreased minimum coverage to 93% to be able to keep updating the branch while other services are still being developed. Marked both changes as those that need to be reverted in the PR for merging feature/vote into main;
- adjusted meeting export to correctly fetch the poll data, added a test with all the implemented config types;
- deleted `test_delete_with_poll_candidate` as there's no separate model for poll candidates now: they are being set up as values of poll_option.meeting_user:


tests/system/action/meeting/test_import.py
```diff
@@ -432,6 +433,44 @@ def get_meeting_mediafile_data(
...
+                    "visibility": Poll.VISIBILITY_MANUALLY,
+                    "state": Poll.STATE_STARTED,
+                    "meeting_id": 1,
+                    "content_object_id": f"assignment/{obj_id}",
+                }
+            },
+            "poll_config_approval": {
+                str(obj_id): {
+                    "id": obj_id,
+                    "poll_id": obj_id,
```
luisa-beerboom: The tests would probably be a bit cleaner if the objects from the different collections had different `ids`.
You add `100` to the base id to create the los id already, so why not do the same for the poll and maybe the `poll_config_approval`?

tests/system/action/test_migration_route.py
```diff
@@ -18,6 +20,7 @@
 from .test_internal_actions import BaseInternalPasswordTest, BaseInternalRequestTest
 
 
+@pytest.mark.skip(reason="Waiting for the migration framework.")
```
luisa-beerboom, line 23: This needs an issue to be created so we don't forget to delete this before main merge

setup.cfg
```diff
@@ -52,7 +52,7 @@ exclude_lines =
     raise NotImplementedError
     @abstractmethod
     ^\s*...$
-fail_under = 96
+fail_under = 93
```
luisa-beerboom: I don't think this is something we should just sneak in between hundreds of lines of code without comments. If we're going to compromise on test coverage by lowering the requirements, it ought to be discussed first. So either make sure test coverage is up, accept that everything isn't quite tested yet and therefore accept the failing CIs, or discuss this with @rrenkert at least.

luisa-beerboom (changes requested):

tests/system/action/test_migration_route.py
vkrasnovyd: Have already marked both of them in the PR for the feature branch: https://github.com/OpenSlides/openslides-backend/pull/3377

tests/system/action/meeting/test_import.py
vkrasnovyd: No need to add 100, actually. No other lists of speakers in the base request data. Simplified this part.

setup.cfg
vkrasnovyd: Discussed. We'll be accepting the failing CIs for now. reverted the change.

tests/system/action/meeting/test_import.py
luisa-beerboom: Having different ids is actually a good thing because it ensures that the action doesn't just accidentally write one type of id into relations pointing to the wrong collection

vkrasnovyd: Updated helper method that generates data and added the checks for the relations.

luisa-beerboom (approved):

PR #3615, opened 2026-06-10
[ReAct] Create prototype for database-direct actions

### LINKED ISSUE #3614: [ReAct] Create prototype for database-direct actions
- Create a new style of action that allows for closer interaction with the database, including direct writing.
- Get rid of costly things like the automatic relation handling during writing for efficiency's sake (this will complicate development somewhat, but will be worth it due to the added efficiency).
- Simplify process further
- Create a strategy for interactions between the new style and the old style
- Create the conditions for gradually replacing the old style actions with the new style actions
- Write some specific tests

luisa-beerboom:
Closes #3614

TODO
- [x] What to do about retry_on_db_failure
- [x] History still needs to be tested.
- [x] check_perm efficiency
- [x] check_archived efficiency
- [x] Transform some more Actions
- [x] Write some specific tests.
- [ ] Rename `Action` to something like `EventAction` and rename `DDAction` to `Action` as a last step


luisa-beerboom:
@hjanott 
This is still in draft, because the last step is not done, but considering that said last step smells like merge conflicts, I'd prefer to do it after the main functionality is reviewed already.
The way I see it, said main functionality is done now, so could you please review it?.

openslides_backend/services/database/database_writer.py
```diff
@@ -126,29 +129,36 @@ def write_events(
...
         event_return_fields = event.get("return_fields", ["id"])
-        if "id" not in event_return_fields:
-            event_return_fields.append("id")
+        return self.insert_model(collection, event_fields, id_, event_return_fields)
+
+    def insert_model(
+        self,
+        collection: Collection,
+        fields: dict[str, Any],
+        id_: Id | None = None,
```
hjanott, line 155: I'd prefer this order. Analogous to `update_model`.
```suggestion
        collection: Collection,
        id_: Id | None = None,
        fields: dict[str, Any],
```

openslides_backend/services/database/database_writer.py
```diff
@@ -185,8 +191,23 @@ def delete_model(
...
+            collection, id_, event_fields, list_fields, event_return_fields
+        )
+
+    def update_model(
+        self,
+        collection: Collection,
+        id_: Id,
+        event_fields: dict[str, Any],
+        list_fields: ListFields = {},
+        event_return_fields: list[str] = ["id"],
```
hjanott: Analogous to `insert_model`.
```suggestion
        return_fields: list[str] = ["id"],
```

openslides_backend/services/database/database_writer.py
luisa-beerboom: Can't place kwargs in front of normal args though

openslides_backend/services/database/database_writer.py
```diff
@@ -218,15 +236,15 @@ def update_model(
...
-                )
+        nm_relation_list_fields_set = set(nm_relation_list_fields)
+        if nm_relation_list_fields_set.intersection(add_dict):
+            self.write_to_intermediate_tables(
+                add_dict, nm_relation_list_fields, id_, collection
+            )
+        if nm_relation_list_fields_set.intersection(remove_dict):
+            self.delete_from_intermediate_tables(
+                remove_dict, nm_relation_list_fields, id_, collection, directly=True
+            )
```
hjanott, line 247: That's a serious improvement. But wouldn't it be even more efficient to check with `any()`?

openslides_backend/services/database/database_writer.py
```diff
@@ -464,6 +482,253 @@ def get_array_values(
...
+            columns=sql.SQL(", ").join(sql.Identifier(field) for field in match_on),
+            placeholders=sql.SQL(", ").join(placeholders),
+        )
+        return self.execute_multi_row_sql(
+            statement,
+            values,
+            return_fields=return_fields,
+        )
+
+    def execute_multi_row_sql(
```
hjanott, line 629: I guess these exceptions are copied over. There should be an efficient way to maintain these in one place. Probably a decorator would be great.

openslides_backend/services/database/database_writer.py
luisa-beerboom: ^^
Could you give an example for what you mean?

openslides_backend/action/actions/motion_workflow/create.py
```diff
@@ -40,6 +44,58 @@ def get_dependent_action_data(
...
+    Action to create a motion workflow.
+    """
+
+    model = MotionWorkflow()
+    schema = DefaultSchema(MotionWorkflow()).get_create_schema(["name", "meeting_id"])
+    permission = Permissions.Motion.CAN_MANAGE
+
+    def write_instances(self, action_data: ActionData) -> ActionResults | None:
+        ad = list(action_data)
+        ids = self.database.reserve_ids(self.model.collection, len(ad))
```
hjanott, line 59: So sad that we still need this here. I hope in the future we can just have a specialized function for this type of creation or just no circular dependencies.

openslides_backend/action/actions/motion_workflow/update.py
```diff
@@ -16,3 +17,12 @@ class MotionWorkflowUpdateAction(UpdateAction):
...
     permission = Permissions.Motion.CAN_MANAGE
+
+    def write_instances(self, action_data: ActionData) -> ActionResults | None:
+        return [
+            None
+            for _ in self.database.update_models(
+                self.model.collection,
+                list(action_data),
+            )
+        ]
```
hjanott, line 28: Wouldn't this be the standard case that should always happen per default? Suggesting moving this to the base class.

openslides_backend/action/actions/organization_tag/delete.py
```diff
@@ -15,3 +16,14 @@ class OrganizationTagDelete(DeleteAction):
...
     skip_archived_meeting_check = True
+
+    def write_instances(self, action_data: ActionData) -> ActionResults | None:
+        instances = list(action_data)
+        self.database.delete_models(
+            "gm_organization_tag_tagged_ids_t",
+            [{"organization_tag_id": instance["id"]} for instance in instances],
+            [],
+            match_on=["organization_tag_id"],
+        )
```
hjanott, line 27: Our intermediate tables already have on delete cascade set and already delete this kind of relation. We can skip doing this.
What is left can probably just be a base method.

openslides_backend/action/actions/theme/create.py
```diff
@@ -68,7 +65,13 @@ class ThemeCreate(CreateAction, CheckForArchivedMeetingMixin):
...
 
-    def update_instance(self, instance: dict[str, Any]) -> dict[str, Any]:
-        instance["organization_id"] = ONE_ORGANIZATION_ID
-        return instance
+    def write_instances(self, action_data: ActionData) -> ActionResults | None:
+        return list(
+            self.database.insert_models(
+                self.model.collection,
+                list(action_data),
+                [*THEME_REQ_FIELDS, *THEME_OPT_FIELDS],
```
hjanott, line 75: Why do these have to be sent manually now?

openslides_backend/action/action.py
```diff
@@ -138,6 +140,10 @@ def __init__(
         self.results = []
         self.cascaded_actions_history = {}
 
+    @property
+    def datastore(self) -> Database:
+        return self.database
```
hjanott, line 145: Great idea.

openslides_backend/action/actions/theme/update.py
```diff
@@ -18,3 +18,13 @@ class ThemeUpdate(UpdateAction, CheckForArchivedMeetingMixin):
...
+    skip_archived_meeting_check = True
+
+    def write_instances(self, action_data: ActionData) -> ActionResults | None:
+        return [
+            None
+            for _ in self.database.update_models(
+                self.model.collection,
+                list(action_data),
+            )
+        ]
```
hjanott, line 30: Returning None per entry should be default behaviour no?

openslides_backend/action/action_handler.py
```diff
@@ -321,31 +344,49 @@ def perform_action(
...
-                # write_request.locked_fields = self.datastore.locked_fields
-                # # reset locked fields, but not changed models - these might be needed
-                # # by another action
-                # self.datastore.reset(hard=False)
-
-            # add on_success routine
-            if on_success := action.get_on_success(action_data):
-                self.on_success.append(on_success)
+                if isinstance(action, Action):
+                    self.datastore.toggle_changed_models(True)
```
hjanott, line 351: Why don't we switch to using the property name here also?

openslides_backend/action/ddaction.py
```diff
@@ -0,0 +1,385 @@
...
+        services: Services,
+        database: Database,
+        relation_manager: RelationManager,
+        logging: LoggingModule,
+        env: Env,
+        skip_archived_meeting_check: bool | None = None,
+        use_meeting_ids_for_archived_meeting_check: bool | None = None,
+    ) -> None:
+        super().__init__(services, database, logging)
+        self.relation_manager = relation_manager
```
hjanott, line 85: Wouldn't it be a better idea to get rid of the migration manager from the beginning?

openslides_backend/action/ddaction.py
```diff
@@ -0,0 +1,385 @@
...
+                {
+                    "original_user_id": self.user_id,
+                    "user_id": self.user_id,
+                    "timestamp": datetime.now(),
+                },
+            )["id"]
+        else:
+            found = self.database.execute_custom_select(
+                sql.SQL(
+                    "id, model_id, entries FROM history_entry_t WHERE position_id = {position_id} AND model_id = ANY(%s)"
```
hjanott, line 318: The database reader implements this operator. `                elif filter_.operator == "in":` Why not use it?

openslides_backend/action/ddaction.py
```diff
@@ -0,0 +1,385 @@
...
+                    fqids=tuple(history_information),
+                    position_id=self.history_position_id,
+                ),
+                arguments=(list(history_information),),
+            )
+            for model in found:
+                update_history[model["id"]] = [
+                    *model.get("entries", []),
+                    *history_information[fqid := model["model_id"]],
+                ]
```
hjanott, line 329: Should be a tad more self-explanatory and efficient.
```suggestion
            for h_entry in found:
                entries = h_entry.get("entries", [])
                entries.extend(history_information[fqid := h_entry["model_id"]])
                update_history[h_entry["id"]] = entries
```

openslides_backend/permissions/permission_helper.py
```diff
@@ -1,49 +1,206 @@
...
-            "anonymous_group_id",
-            "enable_anonymous",
-            "locked_from_inside",
-            "committee_id",
-        ],
-        lock_result=False,
-    )
+    if database.enable_changed_models:
+        # Legacy clause: Delete once all old-style actions are gone
+        perm_data: dict[Collection, dict[Id, dict[str, Any]]] = {}
```
hjanott, line 153: Why not define this outside the if block?

openslides_backend/permissions/permission_helper.py
```diff
@@ -58,20 +215,31 @@ def has_perm(
...
-                return True
+        if database.enable_changed_models:
+            # Legacy clause: Delete once all old-style actions are gone
+            for group_permission in group.get("permissions", []):
+                if is_child_permission(permission, group_permission):
+                    return True
+        else:
+            for group_permission in group.get("permissions") or []:
+                if is_child_permission(permission, group_permission):
+                    return True
```
hjanott, line 242: If and else blocks are identical.

openslides_backend/permissions/permission_helper.py
```diff
@@ -168,15 +354,31 @@ def has_committee_management_level(
...
+) -> bool:
+    if user_id > 0:
+        user = perm_data["user"][user_id]
+        if user.get("organization_management_level") in (
+            OrganizationManagementLevel.SUPERADMIN,
+            OrganizationManagementLevel.CAN_MANAGE_ORGANIZATION,
+        ):
+            return True
+        if user["is_committee_manager"]:
+            return True
```
hjanott, line 378: Check the simple things first.

openslides_backend/permissions/permission_helper.py
```diff
@@ -1,49 +1,206 @@
...
+        "enable_anonymous",
+        "locked_from_inside",
+        "committee_id",
+        "admin_group_id",
+    ],
+    "group": ["permissions", "admin_group_for_meeting_id"],
+    "meeting_user": ["group_ids", "locked_out"],
+}
+
+WriteFields = tuple[str, str, str, list[str]]
```
hjanott, line 41: Is this something we could want in a more central place?

openslides_backend/services/database/database_writer.py
hjanott: Like this we could leave out the set creation.
```suggestion
        nm_relation_list_fields_set = set(nm_relation_list_fields)
        if any(elem in add_dict for elem in nm_relation_list_fields):
            self.write_to_intermediate_tables(
                add_dict, nm_relation_list_fields, id_, collection
            )
        if any(elem in remove_dict for elem in nm_relation_list_fields):
            self.delete_from_intermediate_tables(
                remove_dict, nm_relation_list_fields, id_, collection, directly=True
            )
```
I don't know how efficient the intersection calculation in comparison to `any()` plus set lookup is though.

openslides_backend/shared/exceptions.py
```diff
@@ -110,6 +110,14 @@ def __init__(self, fqid: str) -> None:
...
 
 
+class DoesNotExist(DatabaseException):
+    def __init__(self, identifier: str | None) -> None:
+        super().__init__(
+            f"Does not exist error on '{identifier or 'undefined operation'}'."
+        )
+        self.identifier = identifier
+
+
```
hjanott, line 120: Why not `ModelDoesNotExist`? What other identifiers could there be? If there is a good reason to keep this, let's inherit `ModelDoesNotExist` from this and generate a more uniform error message.

openslides_backend/action/action_handler.py
```diff
@@ -321,31 +344,49 @@ def perform_action(
...
+                        for element in results:
+                            if element and isinstance(element, dict):
+                                element[META_POST_EDIT_FN_KEY] = edit_fn
+                                self.post_edit_necessary = True
+                else:
+                    if self.prev_was_new:
+                        for write_request in self.write_requests:
+                            if not write_request.written:
+                                write_results = self.datastore.write(write_request)
+                                write_request.results = write_results
```
hjanott, line 366: Where is this being used?

hjanott (commented):
This review is a WIP but I'm publishing it so you can work on some points already.

openslides_backend/services/database/database_writer.py
luisa-beerboom: Done

openslides_backend/services/database/database_writer.py
luisa-beerboom: The contains lookup for sets and dicts is supposed to have a O(1) runtime in any usual case because of hashing, so your solution ought to be faster since the intersection would just do the same, but without stopping once the first element is found.

I'm using your version.

openslides_backend/action/actions/motion_workflow/create.py
luisa-beerboom: Wouldn't get my hopes up for the latter

openslides_backend/action/actions/motion_workflow/update.py
luisa-beerboom: I personally like the intentionality and, if we're going to move the writing to the actual action code, I think we _should_ do it explicitly in the actions.
I don't like that this statement needs to be wrapped in a list comprehension over so many lines here though, perhaps it ought to be possible to have the update_models method return None lists in the first place. Maybe that should be the standard.

What do you think?

openslides_backend/action/actions/theme/create.py
luisa-beerboom: It's because of how I needed to implement the database_writer.insert_rows method  in order for it to work. Let's talk about it in person, when we have the time.

openslides_backend/action/actions/theme/update.py
luisa-beerboom: See https://github.com/OpenSlides/openslides-backend/pull/3615#discussion_r3730183532

Same topic

openslides_backend/action/action_handler.py
luisa-beerboom: I'm assuming you mean `write_request.results`?

In line 244 ff.

luisa-beerboom:
TODO: Consider https://github.com/OpenSlides/openslides-backend/issues/3756#issuecomment-5679866694

PR #3625, opened 2026-06-18
Sql diff generator

hjanott:
Generates a diff from two versions of yaml files and translates that into an sql diff that can be applied during incremental migrations.

### DIFF
--- a/openslides_backend/migrations/sql_diff_generator.py
+++ b/openslides_backend/migrations/sql_diff_generator.py
@@ -124,28 +124,26 @@ def handle_add_tree(
     sql = ""
     for collection_name, collection_def in add_tree_dict.items():
         table_name = HelperGetNames.get_table_name(collection_name)
-        for field_name, field_def in collection_def[1]["fields"][0].items():
-            diff_control_part = dc_add_tree_dict[collection_name][1]["fields"][0][
-                field_name
-            ]
-            constraints_sql = generate_constraints_sql(
-                table_name, field_name, field_def, diff_control_part
-            )
-            sql += (
-                f"ALTER TABLE {table_name} ADD COLUMN {field_name}{constraints_sql};\n"
-            )
-            remove_empty(dc_add_tree_dict[collection_name][1]["fields"][0], field_name)
-        for field_name, field_def in collection_def[1]["fields"][1].items():
-            diff_control_part = dc_add_tree_dict[collection_name][1]["fields"][1][
-                field_name
-            ][0]
-            constraints_sql = generate_constraints_sql(
-                table_name, field_name, field_def[0], diff_control_part
-            )
-            sql += (
-                f"ALTER TABLE {table_name} ADD COLUMN {field_name}{constraints_sql};\n"
-            )
-            remove_empty(dc_add_tree_dict[collection_name][1]["fields"][1], field_name)
+        for fields_idx in [0, 1]:
+            fields = collection_def[1]["fields"][fields_idx]
+            dc_fields = dc_add_tree_dict[collection_name][1]["fields"][fields_idx]
+            for field_name, field_def in fields.items():
+                if fields_idx == 0:
+                    # field added
+                    constraints_sql = generate_constraints_sql(
+                        table_name, field_name, field_def, dc_fields[field_name]
+                    )
+                else:
+                    # field altered
+                    # TODO This needs ALTER COLUMN instead
+                    constraints_sql = generate_constraints_sql(
+                        table_name, field_name, field_def[0], dc_fields[field_name][0]
+                    )
+                sql += f"ALTER TABLE {table_name} ADD COLUMN {field_name}{constraints_sql};\n"
+                remove_empty(
+                    dc_add_tree_dict[collection_name][1]["fields"][fields_idx],
+                    field_name,
+                )
         remove_empty(dc_add_tree_dict[collection_name][1], "fields")
         remove_empty(dc_add_tree_dict, collection_name)
     return sql
--- a/openslides_backend/migrations/sql_diff_generator.py
+++ b/openslides_backend/migrations/sql_diff_generator.py
@@ -1,0 +1,184 @@
+import os
+import sys
+from argparse import ArgumentParser
+from copy import deepcopy
+from typing import Any
+
+from meta.dev.src.generate_sql_schema import Helper
+from meta.dev.src.helper_get_names import HelperGetNames
+from openslides_backend.migrations.yaml_diff_generator import dumpjson, generate_diff
+
+
+def main() -> int:
+    parser = ArgumentParser()
+    parser.add_argument("--dumpjson", action="store_true")
+    args = parser.parse_args()
+
+    diff = generate_diff()
+    diff_control: dict[str, Any] = deepcopy(diff)
+    if args.dumpjson:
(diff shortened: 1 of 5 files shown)


openslides_backend/migrations/yaml_diff_generator.py
```diff
@@ -0,0 +1,336 @@
...
+
+    def deep_copy(diff: Any) -> Any:
+        if isinstance(diff, dict):
+            return {k: deep_copy(v) for k, v in diff.items()}
+        elif isinstance(diff, list):
+            return [deep_copy(item) for item in diff]
+        elif isinstance(diff, tuple):
+            return tuple(deep_copy(item) for item in diff)
+        else:
+            return diff
```
luisa-beerboom: why not use python-internal deepcopy function?

openslides_backend/migrations/yaml_diff_generator.py
```diff
@@ -0,0 +1,336 @@
...
+        return [missing_entries, tree]
+    else:
+        return None
+
+
+def create_add_recursive(
+    prev_models: dict[str, Any],
+    curr_models: dict[str, Any],
+    renames_dict: dict[str, Any],
+) -> tuple[dict[str, Any], dict[str, Any]] | None:
```
luisa-beerboom, line 115: How deep can this recursion get?
If it is unlimited, using iteration would be better.

hjanott: It's only as deep as the yaml definitions we write. I think that's alright.

openslides_backend/migrations/yaml_diff_generator.py
hjanott: Yes you're right. I should use that. I was just mistrusting out of no specific reason.

openslides_backend/migrations/yaml_diff_generator.py
```diff
@@ -0,0 +1,316 @@
...
+            "w",
+        ) as f:
+            f.write(json.dumps(diff, indent=4))
+
+    diff_control_part: dict[str, Any]
+    table_name: str
+    field_name: str
+    value: str
+
+    def generate_constraints_sql(field_def: dict[str, Any]) -> str:
```
luisa-beerboom: This method is way too long.
You should find a way to subdivide it into multiple helper methods

openslides_backend/migrations/yaml_diff_generator.py
```diff
@@ -0,0 +1,316 @@
...
+            "data",
+            "0101",
+            "schema_diff.sql",
+        ),
+        "w",
+    ) as f:
+        # TODO create generate diff content functions in schema generator.
+        add = diff["add"]
+        if isinstance(add, tuple) and isinstance(add[0], dict):
+            pass
```
luisa-beerboom: This does nothing

openslides_backend/migrations/yaml_diff_generator.py
```diff
@@ -0,0 +1,316 @@
...
+    TODO This should only generate diffs for the leafs. Thus the structure should be reconsidered. Maybe flatter or integrating rename info.
+    """
+    edited_entries = {}
+    tree = {}
+    for key, curr_value in curr_models.items():
+        if key in list(renames_dict.values()):
+            print(key + " renamed -> skip for edit")
+            continue
+        if key not in prev_models:
+            pass
```
luisa-beerboom: I don't like non-operative if clauses for the sake of hanging on elif clauses

openslides_backend/migrations/yaml_diff_generator.py
```diff
@@ -0,0 +1,316 @@
...
+        elif isinstance(curr_value, dict):
+            result = create_add_recursive(
+                prev_models[key], curr_value, renames_dict.get(key, {})
+            )
+            if result is not None:
+                tree[key] = result
+    if additional_entries or tree:
+        return (additional_entries, tree)
+    else:
+        return None
```
luisa-beerboom, line 136:
```suggestion
    return  (additional_entries, tree) if additional_entries or tree else None
```
Would be shorter.
There's other places like this as well

luisa-beerboom (commented):
What about all the TODOs that are still in the code?

openslides_backend/migrations/yaml_diff_generator.py
hjanott: Yes it's definitely shorter in terms of loc. But not that much in terms of characters and I find my version a little better to read.

openslides_backend/migrations/yaml_diff_generator.py
hjanott: Changed

openslides_backend/migrations/yaml_diff_generator.py
hjanott: That's work in progress and should be done in a different PR. Added a TODO.

openslides_backend/migrations/yaml_diff_generator.py
hjanott: I suppose you mean the main function. Did that.

openslides_backend/migrations/sql_diff_generator.py
```diff
@@ -0,0 +1,184 @@
...
+                field_name
+            ][0]
+            constraints_sql = generate_constraints_sql(
+                table_name, field_name, field_def[0], diff_control_part
+            )
+            sql += (
+                f"ALTER TABLE {table_name} ADD COLUMN {field_name}{constraints_sql};\n"
+            )
+            if not any(dc_add_tree_dict[collection_name][1]["fields"][1][field_name]):
+                del dc_add_tree_dict[collection_name][1]["fields"][1][field_name]
```
luisa-beerboom: These two for loops look extremely similar.
Could this be unified in some way?

openslides_backend/migrations/sql_diff_generator.py
```diff
@@ -0,0 +1,184 @@
...
+        if not any(dc_add_tree_dict[collection_name][1]["fields"]):
+            del dc_add_tree_dict[collection_name][1]["fields"]
+        if not any(dc_add_tree_dict[collection_name]):
+            del dc_add_tree_dict[collection_name]
```
luisa-beerboom: This looks like some kind of cleanup code that is present in multiple methods, why not turn it into its own function?

openslides_backend/migrations/sql_diff_generator.py
hjanott: The main difference is where `diff_control_part` is taken from and where the `field_def` is taken from to pass to `generate_constraints_sql`.
Do you suggest a flag to switch on and moving the entire loop or just moving the part where the  `constraints_sql` is generated and moved?

openslides_backend/migrations/sql_diff_generator.py
hjanott: Now its own function.

openslides_backend/migrations/sql_diff_generator.py
hjanott: I've decided to iterate over the used indices and compress it that way a bit. I decided against an own function as the functionality may deviate quiet a bit as we develop this further.

luisa-beerboom (approved):

PR #3627, opened 2026-06-19
Handle health route outside of threads

### LINKED ISSUE #3619: Make health route independent of worker processes
**Describe the bug**
Currently the health route does not respond in time when creating max. action workers (set via OPENSLIDES_BACKEND_NUM_WORKERS). This leads to a restart of the backend container and possible not proper stopped worker processes.

**How To Reproduce**
Create long running processes (e.g. create 100 users in bulk) and request health route

**Expected behavior**
The health route should be independent of the worker processes.

luisa-beerboom:
Closes #3619 

Establishes custom thread worker that handles the health route outside the threads.

From my manual testing it does handle the requests separately and the instance is functional, however I haven't tested it on a large scale prod setup or a long running instance and I don't know what will happen if the connection has high latency (my theory is some health requests may be missed and fed into the regular process instead, nothing that can be done about it though), so @rrenkert could you try out your experiment with this version?

Also, since the base method and class that was overridden can be further changed by the gunicorn devs, this might break at some point later on.
The code changes _are_ pretty minimal, and, except for the most recent change 5 months ago, this method that was changed did not have any substantial changes for 12 years, but I still thought this ought to be mentioned at least.

@hjanott if you want to know what the original code looks like before reviewing: https://github.com/benoitc/gunicorn/blob/master/gunicorn/workers/gthread.py#L241

### DIFF
--- a/openslides_backend/http/thread_worker.py
+++ b/openslides_backend/http/thread_worker.py
@@ -8,7 +8,12 @@ class OpenSlidesThreadWorker(ThreadWorker):
         # Peek into the sockets content to be able to read out the path part of the header (which ought to be the part after the first whitespace).
         peek = str(conn.sock.recv(4096, MSG_PEEK)).split()
 
-        if len(peek) >= 2 and "system/action/health" in peek[1]:
+        if (
+            len(peek) >= 2
+            and (stripped_path := peek[1].strip("/")).startswith("system/")
+            and stripped_path.endswith("/health")
+            and len(stripped_path.split("/")) == 3
+        ):
             # If the health route is recognized, set data_ready to True to stop handle from returning _DEFER and force it into working instead.
             # Since the message could be read, the data is probably there already,
             # the health route doesn't take a payload to begin with.
--- a/openslides_backend/http/thread_worker.py
+++ b/openslides_backend/http/thread_worker.py
@@ -1,0 +1,14 @@
+from socket import MSG_PEEK
+
+from gunicorn.workers.gthread import TConn, ThreadWorker
+
+
+class OpenSlidesThreadWorker(ThreadWorker):
+    def enqueue_req(self, conn: TConn) -> None:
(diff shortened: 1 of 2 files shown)


openslides_backend/http/thread_worker.py
```diff
@@ -0,0 +1,20 @@
...
+    def enqueue_req(self, conn: TConn) -> None:
+        # Peek into the sockets content to be able to read out the path part of the header (which ought to be the part after the first whitespace).
+        peek = str(conn.sock.recv(4096, MSG_PEEK)).split()
+
+        if len(peek) >= 2 and "system/action/health" in peek[1]:
+            # If the health route is recognized, set data_ready to True to stop handle from returning _DEFER and force it into working instead.
+            # Since the message could be read, the data is probably there already,
+            # the health route doesn't take a payload to begin with.
+            conn.data_ready = True
+            # Call the base class methods used in the base method in order, don't enthread them.
```
hjanott, line 21:
```suggestion
            # Call the base class methods used in the base method in order, don't execute in a separate thread.
```

hjanott (changes requested):
While the action route works, the health route on the presenter does not.
Maybe this is because of a missing import in the presenter view (in contrast to the action view) that would otherwise initialize parts of some library?

openslides_backend/http/thread_worker.py
luisa-beerboom: I don't like that wording because it suggests that we're still  handling it in a thread.

luisa-beerboom:
> While the action route works, the health route on the presenter does not. Maybe this is because of a missing import in the presenter view (in contrast to the action view) that would otherwise initialize parts of some library?

Might be because the health route recognition used the action route path text exclusively.
I changed that, however after some testing I found that I sometimes get 502 errors when testing actions and presenters with the client now.
Idk if it's my code or if my instance is broken though.

PR #3633, opened 2026-06-23
Set tz=UTC when converting integer timestamps

### LINKED ISSUE #3632: Meeting date: selected date snap back to previous day
**Describe the bug**
During meeting creation or meeting edit, the date selection doesn't work properly. The selected date will snap back to the previous day. 

Example: Selected 10.06.2026 as meeting date. After saving it snaps to 09.06.2026.

**How to Reproduce**
1. open an instance
2. open committee > create a committee
3. open the committee > create a meeting > select a specific date in the future > save it. Take a look at the meeting date. It's snapped back to a previous date. 

**Expected behavior**
The selected date should be correctly selected after meeting creation or edit.

peb-adr:
Fixes #3632

Seems in the action implementation integer timestamps are converted using `datetime.fromtimestamp(value)` without specifying timezone information. This will correctly assume UTC for the value. However as stated in the [docs](https://docs.python.org/3/library/datetime.html#datetime.datetime.fromtimestamp):

    If optional argument tz is None or not specified, [...] the returned datetime object is naive.

This leads to shifts in the value, if the timezone in the postgres's users locale differs from UTC. This is not the case is the compose based dev setup but was observed in other environments.

### DIFF
--- a/openslides_backend/action/actions/committee/import_.py
+++ b/openslides_backend/action/actions/committee/import_.py
@@ -192,7 +192,9 @@ class CommitteeImport(BaseImportAction, CommitteeImportMixin):
                     if (value := action_data.get(field_name)) and isinstance(
                         value, int
                     ):
-                        action_data[field_name] = datetime.fromtimestamp(value, tz=timezone.utc)
+                        action_data[field_name] = datetime.fromtimestamp(
+                            value, tz=timezone.utc
+                        )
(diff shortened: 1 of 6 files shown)


Elblinator (approved):

peb-adr:
When approved, feel free to merge

luisa-beerboom (approved):

PR #3635, opened 2026-06-24
Change field validation for HTML fields and add timezone to action

### LINKED ISSUE #3611: Error: motion_comment/xxx/comment: Field required but empty.
We get the error `motion_comment/xxx/comment: Field required but empty.` during meeting clone.

In the database it looks like this:

```
SELECT * FROM motion_comment WHERE id = 13531;
  id   | comment | motion_id | section_id | meeting_id
-------+---------+-----------+------------+------------
   xxx |         |     yyy |        zzz  |         1
```

It looked like this before the migration:

`{"id": xxx, "comment": "", "motion_id": yyy, "meeting_id": 1, "section_id": zzz, "meta_deleted": false, "meta_position":  nnn}`

### LINKED ISSUE #3629: Meeting Duplication crash
**Describe the bug**
Duplication of meetings crashes with error 'motion_comment/13844/comment: Field required but empty'. 

**How To Reproduce**
I don't know how to build a repro, but I can show it on an instance. 

```
[
    {
        "action": "meeting.clone",
        "data": [
            {
                "meeting_id": 92
            }
        ]
    }
]
```
**Response from backend**
```
{
    "status_code": 202,
    "success": false,
    "message": "Action (meeting.clone) lasts too long. action_worker/2177 written to database. Get the result from database, when the job is done.",
    "results": [
        [
            {
                "fqid": "action_worker/2177",
                "name": "meeting.clone",
                "written": true
            }
        ]
    ]
}
```

**Expected behavior**
Dupe a meeting should not crash. 

**Important additional information**
This also happen with vote weight failures: vote/126729/weight: Field required but empty. vote/126730/weight: Field required but empty. vote/126916/weight: Field required but empty. - Which also must be fixed.

### LINKED ISSUE #3631: Meeting dupe: crash because of time zones
**Describe the bug**
During meeting dupe a failure occured, which crashes the duplication process.

Failure message: Action meeting.clone: data must not contain {'time_zone'} properties

**How To Reproduce**
Currently no reproduction. Please contact me, so I can show you, where it happens.

```
[{action: "meeting.clone",…}]
0: {action: "meeting.clone",…}
action: "meeting.clone"
data: [{committee_id: 55, meeting_id: 2, start_time: null, end_time: null, time_zone: "Europe/Berlin",…}]
0: {committee_id: 55, meeting_id: 2, start_time: null, end_time: null, time_zone: "Europe/Berlin",…}
admin_ids: [1]
committee_id: 55
end_time: null
external_id: null
location: ""
meeting_id: 2
name: "HB Landesverband (5)"
organization_tag_ids: []
set_as_template: false
start_time: null
time_zone: "Europe/Berlin"
```

**Response from backend**
```
{
    "success": false,
    "message": "Action meeting.clone: data must not contain {'time_zone'} properties",
    "action_error_index": 0,
    "action_data_error_index": 0
}
```

**Expected behavior**
Meeting dupe should work.

luisa-beerboom:
Closes #3611
Closes #3629
Closes #3631 

Fix for #3611 also does the same for `chat_message/content`

Everything except the timezone fix is probably more of a workaround.


tests/system/action/meeting/test_clone.py
```diff
@@ -2358,3 +2383,115 @@ def test_clone_with_folder_structure_but_no_access_groups(self) -> None:
...
+                    "meeting_id": 1,
+                },
+                "option/2": {
+                    "weight": 0,
+                    "poll_id": 1,
+                    "content_object_id": "motion/1",
+                    "meeting_id": 1,
+                },
+                "vote/3": {
+                    "weight": "0.000000",
```
hjanott: This still exists? Oh god ... The data from the issue even implies that it was copied over and over for many times.

tests/system/action/meeting/test_clone.py
```diff
@@ -2358,3 +2383,115 @@ def test_clone_with_folder_structure_but_no_access_groups(self) -> None:
...
+        response = self.request(
+            "meeting.clone",
+            {
+                "meeting_id": 1,
+                "external_id": "external_id",
+            },
+        )
+        self.assert_status_code(response, 200)
+
+    def test_clone_with_empty_motion_comment(self) -> None:
```
hjanott: Wouldn't it be enough to just add such a `motion_comment` to an existing test instead of creating two new ones? I would prefer that out of test running time reasons. On top of that: Shouldn't this mostly be checked in the `motion_comment` create tests?

hjanott (changes requested):
I understand why you think the two other cases are more like a workaround. Shall we have an issue for a migration so we can clean this up then?

tests/system/action/meeting/test_clone.py
luisa-beerboom: > Shouldn't this mostly be checked in the `motion_comment` create tests?

What would the motion_comment create test have to do with a bug that appears in the meeting.clone?
Meeting.clone doesn't use motion_comment.create

hjanott: > Meeting.clone doesn't use motion_comment.create

I always forget that meeting import creates the events directly.

hjanott (approved):
Looks good to me.

PR #3641, opened 2026-06-26
[dammi] Remove mig tables

hjanott:
This aims at improving the test setup and applying the already sql diff from file.
It removes the migration tables, the finalize command and unused migration states.
Instead it applies the schema diff and operates on the tables directly using only one transaction.
It slims down the replace tables to basically only the collection names.
It changes the directory structure and file naming convention for the migrations.
Decouples the migration and migration route test cases again as the route test cases do not need a reset of the schema and it would lead to other unwanted side effects. Thus we are generating some code redundancy.
There is still one error in `test_play_with_migrations` which is intended as the test_migration will get removed again.


openslides_backend/migrations/migration_handler.py
```diff
@@ -324,31 +143,28 @@ def migrate(self) -> None:
...
-                        )
+                        errors = f"Pre check for migration {mig_name} failed.\n{errors}"
                         self.logger.info(errors)
                         raise MigrationSetupException(errors)
                 MigrationHelper.write_line("migration started")
-                self.set_public_tables_read_only()
-                self.setup_migration_relations()
+                deactivate_notify_triggers(self.cursor)
                 self.execute_migrations()
+                activate_notify_triggers(self.cursor)
```
luisa-beerboom, line 163: If the migrations fail with an error, is it okay for the triggers to not be activated again?

openslides_backend/migrations/migration_handler.py
```diff
@@ -367,212 +183,19 @@ def execute_command(self, command: str) -> None:
...
-        Closes the migration threads io stream.
-        """
-        assert MigrationHelper.migrate_thread_stream
-
-        MigrationHelper.migrate_thread_stream.close()
-        MigrationHelper.migrate_thread_stream = None
-        MigrationHelper.migrate_thread_stream_can_be_closed = False
-        MigrationHelper.migrate_thread_exception = None
-
-    def finalize(self) -> None:
```
luisa-beerboom, line 427: Merging this into #3643 is going to be fuuuuun (x_x)

openslides_backend/migrations/migration_helper.py
```diff
@@ -109,22 +120,24 @@ def load_migrations() -> None:
...
-            if reMatch is not None:
-                migration_file = reMatch.groupdict()["migration"]
-                migration_number = int(migration_file[:4])
-                if migration_number >= 100:
-                    MigrationHelper.migrations[migration_number] = migration[:-3]
-        MigrationHelper.migrations = dict(sorted(MigrationHelper.migrations.items()))
+        for file_or_folder in files_and_folders:
+            if os.path.isdir(os.path.join(MIGRATIONS_PATH, file_or_folder)):
+                reMatch = match(r"mig_(?P<migration>\d{4}_.*)", file_or_folder)
+                # mig_ matches literaly and \d{4}_.*\.py : 4 digits, 1 underscore, any characters
```
luisa-beerboom: Comment is not quite right anymore. The `.py` is no longer part of this regex.

openslides_backend/migrations/migration_helper.py
```diff
@@ -109,22 +120,24 @@ def load_migrations() -> None:
...
-            # \d{4}_.*\.py : 4 digits, 1 underscore, any characters, [dot]py
-            if reMatch is not None:
-                migration_file = reMatch.groupdict()["migration"]
-                migration_number = int(migration_file[:4])
-                if migration_number >= 100:
-                    MigrationHelper.migrations[migration_number] = migration[:-3]
-        MigrationHelper.migrations = dict(sorted(MigrationHelper.migrations.items()))
+        for file_or_folder in files_and_folders:
+            if os.path.isdir(os.path.join(MIGRATIONS_PATH, file_or_folder)):
+                reMatch = match(r"mig_(?P<migration>\d{4}_.*)", file_or_folder)
```
luisa-beerboom: Camelcase variable?

openslides_backend/services/postgresql/create_schema.py
```diff
@@ -37,6 +37,23 @@ def drop_db() -> None:
...
 
+def fill_empty_version(curs: Cursor[rows.DictRow], mig_nmbr: int) -> None:
+    """
+    If the version table is empty:
+    Fills the version table with state 'finalized' from migration number 100 until backend_migration_index.
+    Missing migration states for rel-db indices (>= 100) will be set by the migration manager.
+    """
+    print("Migration info written:")
+    if not MigrationHelper.get_database_migration_index(curs):
+        for nmbr in range(100, MigrationHelper.get_backend_migration_index() + 1):
```
luisa-beerboom, line 48: You're setting info for a migration 0 in this. Is that intended?

openslides_backend/services/postgresql/utils.py
```diff
@@ -0,0 +1,51 @@
...
+                    table=sql.Identifier(trigger_dict["table_name"]),
+                    trigger=sql.SQL(trigger_dict["trigger_name"]),
+                )
+            )
+
+
+def activate_notify_triggers(cursor: Cursor[dict[str, Any]]) -> None:
+    """Activates all notify triggers present in the database."""
+    for table in MigrationHelper.get_public_tables(cursor):
+        to_disable_triggers = get_notify_names(cursor, table)
```
luisa-beerboom: Perhaps change the variable name

tests/system/migrations/base_migration_test.py
```diff
@@ -1,22 +1,234 @@
...
+        def _wait_for_lock(*args: Any, **kwargs: Any) -> mockdefault:
+            assert (
+                MigrationHelper.migrate_thread_stream
+            ), "migrate_thread_stream not initialized by migration framework."
+            if args[0] == "migration started":
+                MigrationHelper.migrate_thread_stream.write(args[0] + "\n")
+                indicator_lock.release()
+                wait_lock.acquire()
+            else:
+                MigrationHelper.migrate_thread_stream.write(args[0] + "\n")
```
luisa-beerboom:
```suggestion
            MigrationHelper.migrate_thread_stream.write(args[0] + "\n")
            if args[0] == "migration started":
                indicator_lock.release()
                wait_lock.acquire()
```

tests/system/migrations/base_migration_test.py
```diff
@@ -1,22 +1,234 @@
...
+    def migrate_previous(self) -> dict[str, Any]:
+        """
+        Executes the `migrate` command using a MigrationManager.
+        Waits for thread execution.
+        """
+        # Migrate to state before tested migration.
+        if filenames := [
+            f
+            for f in os.listdir(MIGRATIONS_PATH)
+            if re.match("mig_\\d+", f[:8]) and int(f[4:8]) < self.migration_number
```
luisa-beerboom: Wouldn't `"mig_\\d{4}"`be more accurate?

tests/conftest.py
```diff
@@ -78,9 +76,7 @@ def db_connection() -> Generator[Connection[DictRow], None, None]:
...
         with conn.cursor() as curs:
             curs.execute("SELECT init_table_contents();")
-        conn.commit()
-        yield conn
-        conn.commit()
-        with conn.cursor() as curs, suppress(AdminShutdown):
-            # AdminShutdown will happen when the database is dropped during first rel-db migration tests
+            conn.commit()
+            yield conn
+            conn.commit()
```
luisa-beerboom, line 81: Just out of curiosity, why commit twice here?

luisa-beerboom (changes requested):

openslides_backend/migrations/migration_handler.py
hjanott: Yes. Since the database stays in an unusable state until reset was called this is not needed.
Reset will then reactivate the notifies. I'm currently implementing this for an upcoming PR.

openslides_backend/migrations/migration_handler.py
hjanott: I'm sorry.

openslides_backend/migrations/migration_helper.py
hjanott: You're right. That's ugly.

openslides_backend/services/postgresql/create_schema.py
hjanott: This is setting the info for migration 100 until the current backend index.

tests/conftest.py
hjanott: This is to make all changes visible done by `init_table_contents` to other connections/transactions and to make their changes visible to `truncate_testdata_tables`. (Waiting at yield until the test finishes.)
Perhaps that could be achieved with a change of isolation level here. If you want, you could try that in your PR.

openslides_backend/migrations/migration_helper.py
hjanott: corrected

openslides_backend/services/postgresql/utils.py
hjanott: done

tests/system/migrations/base_migration_test.py
hjanott: Thanks!

tests/system/migrations/base_migration_test.py
hjanott: yes. did so

luisa-beerboom (approved):

PR #3643, opened 2026-06-29
[Dammi] Split off version table writing for migrations

luisa-beerboom:
TODO:
- [x] See if all those `self.cursor.connection.commit()` are really necessary
- [x] Write test to see what happens if a mig fails (is `migration_failed` written?)
- [x] remove `**kwargs` from `data_manipulation`
- [x] linting


openslides_backend/migrations/migration_handler.py
```diff
@@ -105,25 +105,26 @@ def execute_migrations(self) -> None:
                 mig_class.data_manipulation(self.cursor, stash)
                 mig_class.cleanup(self.cursor)
 
-                MigrationHelper.set_database_migration_info(
-                    self.cursor, index, MigrationState.MIGRATION_FINISHED
-                )
+                self.cursor.connection.commit()
```
hjanott: Shouldn't be necessary. Even worse. We would persist the migration changes of already completed migrations even though later migrations may fail.

openslides_backend/migrations/migration_manager.py
```diff
@@ -54,8 +54,12 @@ def __init__(
...
         self.handler: MigrationHandler | None
         MigrationHelper.load_migrations()
-        MigrationHelper.add_new_migrations_to_version()
-        self.target_migration_index = MigrationHelper.get_backend_migration_index()
+        with get_new_os_conn() as conn:
+            with conn.cursor() as cursor:
+                MigrationHelper.add_new_migrations_to_version(cursor)
+                self.target_migration_index = (
+                    MigrationHelper.get_backend_migration_index()
+                )
```
hjanott: This can happen outside the context.

openslides_backend/migrations/migration_handler.py
```diff
@@ -133,52 +134,55 @@ def migrate(self) -> None:
...
-                    mig_class = MigrationHelper.get_migration_class(package_name)
-                    mig_name = package_name[4:]
-                    self.logger.info("Pre check: " + mig_name + " ...")
-                    if errors := mig_class.check_prerequisites(self.cursor):
-                        if minimum_required_index:
-                            MigrationHelper.set_database_migration_info(
-                                self.cursor,
-                                minimum_required_index["min"],
-                                MigrationState.MIGRATION_REQUIRED,
+        with self.ver_conn.cursor() as cursor:
```
hjanott: I think that would make it more readable.
```suggestion
        with self.ver_conn.cursor() as ver_cursor:
```

openslides_backend/migrations/migration_handler.py
```diff
@@ -133,52 +134,55 @@ def migrate(self) -> None:
...
+                            cursor,
+                            minimum_required_index["min"],
+                            MigrationState.MIGRATION_RUNNING,
+                        )
+                    # Check prerequisites
+                    for index, package_name in MigrationHelper.migrations.items():
+                        mig_class = MigrationHelper.get_migration_class(package_name)
+                        mig_name = package_name[4:]
+                        self.logger.info("Pre check: " + mig_name + " ...")
+                        if errors := mig_class.check_prerequisites(self.cursor):
```
hjanott: Maybe even this.
```suggestion
                        if errors := mig_class.check_prerequisites(self.migration_cursor):
```

openslides_backend/migrations/migration_handler.py
```diff
@@ -133,52 +134,55 @@ def migrate(self) -> None:
...
-            case _:
-                raise MigrationException(
-                    f"{state} not allowed when executing migrate command."
-                )
+                            self.logger.info(errors)
+                            raise MigrationSetupException(errors)
+                    MigrationHelper.write_line("migration started")
+                    deactivate_notify_triggers(self.cursor)
+                    self.execute_migrations()
+                    activate_notify_triggers(self.cursor)
```
hjanott: Seems like this works also. I was thinking this should be the version cursor. But let's leave it at this. Probably better this way. I'm thinking if we should take the reactivate out of the reset command then and do the notify enabled check earlier on (before line 470) in the test.

tests/system/migrations/test_0100_init_reldb.py
```diff
@@ -392,12 +475,27 @@ def test_migration_reset(self, method_mock: Mock) -> None:
...
+            prev = failed_data[tablename]
+            reset = reset_data[tablename]
+            if tablename == "version":
+                assert len(prev) == 2
+                assert len(reset) == 2
+                assert prev[0] == reset[0]
+                assert reset[1] == {
+                    "migration_index": 100,
+                    "migration_state": MigrationState.MIGRATION_REQUIRED,
+                    "replace_tables": {k: {} for k in ORIGIN_COLLECTION_LIST},
```
hjanott: Get this another way so we don't have to assign it weirdly just for this test. Probably `Migration.ORIGIN_COLLECTIONS` will do the trick.

tests/system/migrations/test_0100_init_reldb.py
```diff
@@ -320,13 +323,17 @@ def test_migration_fail_time_zone(self) -> None:
...
-        MigrationHelper.migrate_thread_stream = StringIO()
         # 5) Call data_manipulation of module
-        with get_new_os_conn() as conn:
-            with conn.cursor() as curs:
-                handler = MigrationHandler(curs, Mock(), Mock(), self.app.logging)
-                handler.execute_command("migrate")
+        with get_new_os_conn() as version_conn:
+            with version_conn.cursor() as cursor:
+                MigrationHelper.add_new_migrations_to_version(cursor)
+            MigrationHelper.migrate_thread_stream = StringIO()
```
hjanott: Filling `migrate_thread_stream` doesn't need to be within the connection context.

tests/system/migrations/test_0100_init_reldb.py
```diff
@@ -346,6 +353,19 @@ def test_migration_manager(self) -> None:
         self.assert_indices_state(MigrationState.FINALIZED)
         self.check_data()
 
+    def get_all_data(self) -> dict[str, list[dict[str, Any]]]:
+        with get_new_os_conn() as conn:
+            with conn.cursor() as curs:
+                curs.execute(
+                    "SELECT table_name FROM information_schema.tables WHERE table_schema = current_schema() AND table_type = 'BASE TABLE'"
```
hjanott: Do you think this kind of request would be better for what is being used in `MigrationHelper.get_public_tables`?

openslides_backend/migrations/migration_manager.py
```diff
@@ -88,50 +93,53 @@ def get_stats(self) -> dict[str, Any]:
...
-                if state_per_mi[mi]
-                in (
-                    MigrationState.MIGRATION_REQUIRED,
-                    MigrationState.MIGRATION_RUNNING,
+            if current_migration_index >= 100:
+                all_tables = MigrationHelper.get_public_tables(cursor)
+            else:
+                all_tables = set()
+
+            def count(table: str, curs: Cursor[DictRow]) -> int:
```
hjanott:
```suggestion
            def count(table: str) -> int:
```

hjanott: Or we move it outside the connection context and actually pass the cursor.

hjanott: Could as well be part of the new `utils` module or the `MigrationHelper` as I imagine a usage within the `Migration` classes for outputting information about the migrated modules or so.

openslides_backend/migrations/migration_manager.py
```diff
@@ -225,25 +231,28 @@ def execute_migrate_command(self, command: str, verbose: bool) -> None:
...
-                    with conn.cursor() as curs:
-                        relevant_mis = MigrationHelper.get_unfinalized_indices(curs)
+                with get_new_os_conn() as version_conn:
+                    version_conn.set_isolation_level(IsolationLevel.READ_COMMITTED)
+                    with version_conn.cursor() as cursor:
+                        relevant_mis = MigrationHelper.get_unfinalized_indices(cursor)
                         for mi in relevant_mis:
                             MigrationHelper.set_database_migration_info(
-                                curs, mi, MigrationState.MIGRATION_FAILED
+                                cursor, mi, MigrationState.MIGRATION_FAILED
```
hjanott: An unnecessary rename.
```suggestion
                                ver_cursor, mi, MigrationState.MIGRATION_FAILED
```

hjanott (changes requested):

openslides_backend/migrations/migration_handler.py
luisa-beerboom: This is still in there because I got "could not serialize access due to concurrent update" when I tried deleting it.
If you're saying that this ought not happen, then perhaps there is still some sort of error in there that we're probably going to have to debug together.

luisa-beerboom: Is it really that much of a problem to have finished migrations already committed?
I would think this means they don't have to run again once the broken migration is fixed.

openslides_backend/migrations/migration_manager.py
luisa-beerboom: Done

openslides_backend/migrations/migration_handler.py
luisa-beerboom: Done

openslides_backend/migrations/migration_handler.py
luisa-beerboom: Done

tests/system/migrations/test_0100_init_reldb.py
luisa-beerboom: Done

tests/system/migrations/test_0100_init_reldb.py
luisa-beerboom: Done

tests/system/migrations/test_0100_init_reldb.py
luisa-beerboom: I think I like the other query better because its shorter. Using that one in my test.

openslides_backend/migrations/migration_manager.py
luisa-beerboom: > Could as well be part of the new `utils` module or the `MigrationHelper` as I imagine a usage within the `Migration` classes for outputting information about the migrated modules or so.

I'm against it in this instance.
While I agree a count method would be useful there, this function's written too specific for this case, we'd have to completely rewrite it.

luisa-beerboom: > Or we move it outside the connection context and actually pass the cursor.

Done

openslides_backend/migrations/migration_manager.py
luisa-beerboom: Done

tests/system/migrations/test_0100_init_reldb.py
```diff
@@ -320,13 +323,17 @@ def test_migration_fail_time_zone(self) -> None:
...
+        MigrationHelper.migrate_thread_stream = StringIO()
+        with get_new_os_conn() as version_conn:
+            with version_conn.cursor() as cursor:
+                MigrationHelper.add_new_migrations_to_version(cursor)
+            with get_new_os_conn() as conn:
+                with conn.cursor() as curs:
+                    handler = MigrationHandler(
+                        curs, Mock(), Mock(), self.app.logging, version_conn
+                    )
+                    handler.execute_command("migrate")
```
hjanott, line 336: Let's move the comment where it needs to go.
```suggestion
        MigrationHelper.migrate_thread_stream = StringIO()
        with get_new_os_conn() as version_conn:
            with version_conn.cursor() as cursor:
                MigrationHelper.add_new_migrations_to_version(cursor)
            # 5) Call data_manipulation of module
            with get_new_os_conn() as conn:
                with conn.cursor() as curs:
                    handler = MigrationHandler(
                        curs, Mock(), Mock(), self.app.logging, version_conn
                    )
                    handler.execute_command("migrate")
```

hjanott (changes requested):

openslides_backend/migrations/migration_handler.py
hjanott: > then perhaps there is still some sort of error in there that we're probably going to have to debug together.

Let's do it!

luisa-beerboom: Done

openslides_backend/migrations/migration_handler.py
luisa-beerboom: Done

tests/system/migrations/test_0100_init_reldb.py
luisa-beerboom: Done

hjanott (approved):
Yay! Good work!

PR #3644, opened 2026-06-29
[vote] Update meeting defaults for polls

vkrasnovyd:
Adjusts backend to the changes in https://github.com/OpenSlides/openslides-meta/pull/532

How the poll_default settings fields of the meeting will be migrated is described here: https://github.com/OpenSlides/openslides-vote-service/tree/feature/vote/Migration.md#meeting


openslides_backend/action/actions/meeting/update.py
```diff
@@ -332,6 +334,51 @@ def update_instance(self, instance: dict[str, Any]) -> dict[str, Any]:
         if (translations := instance.get("custom_translations")) is not None:
             instance["custom_translations"] = Jsonb(translations)
 
+        # Set poll defaults
+        poll_default_field_name_pattern = re.compile(
+            r"([a-z]+)_poll_default_([a-z_]+[a-z]+)"
```
luisa-beerboom: I think
```suggestion
            r"([a-z]+)_poll_default_(([a-z]+_)*[a-z]+)"
```
might be more accurate to the pattern you're trying to match, though I wonder if a regex isn't a bit overkill for now.

openslides_backend/action/actions/meeting_poll_default/create.py
```diff
@@ -0,0 +1,104 @@
...
+            field_name: value
+            for field_name in MeetingPollDefault.POLL_TYPE_FIELDS
+            if (value := instance.get(field_name))
+        }
+        self.check_exactly_one_of(instance, poll_types_fields)
+
+        poll_type_field_name, poll_type_field_value = next(
+            iter(poll_types_fields.items())
+        )
+        self.check_equal_fields(instance, poll_type_field_name, poll_type_field_value)
```
luisa-beerboom, line 49: This setup of grabbing the next value in a dict iteration and checking only that seems somewhat incomplete unless one knows that there can be only one of the POLL_TYPE_FIELDS in the instance.
I'd suggest either using a for loop or writing an explanatory comment.

luisa-beerboom (commented):
Idk what I think about this new collection, however the code looks mostly good to me.

openslides_backend/action/actions/meeting/update.py
vkrasnovyd: You are right. I've just realized that .split() would do the same job in this case.

openslides_backend/action/actions/meeting_poll_default/create.py
vkrasnovyd: The check above raises error if number of items in poll_types_fields is not 1. Descriptions of the helper methods were not really descriptive indeed. Added proper docstrings.

luisa-beerboom (approved):

PR #3661, opened 2026-07-13
[vote] Delete polls in backend

vkrasnovyd:
Needs https://github.com/OpenSlides/openslides-meta/pull/548

As discussed with Oskar, at this point vote service and backend service can and should function more independently. When it is backend that initiates the deletion of poll and poll related objects, then the deletion should be handled by backend.

This PR:
* Creates internal delete actions for poll and all the related objects, uses them in `DeleteAction`, adds docs
* Deletes vote adapter - before this PR was used only for the cascade deletes
* Removes `managed_by` from the Model - was used only to determine which models should be deleted by vote service
* Extends example_data to include all the poll related collections


docs/Actions-Overview.md
```diff
@@ -171,6 +171,17 @@ A more general format description see in [Action-Service](https://github.com/Ope
...
+## Poll related objects
+
+- [poll.delete](actions/poll.delete.md)
+- [poll_ballot.delete](actions/poll_ballot.delete.md)
+- [poll_config_approval.delete](actions/poll_config_approval.delete.md)
+- [poll_config_rating_approval.delete](actions/poll_config_rating_approval.delete.md)
+- [poll_config_rating_score.delete](actions/poll_config_rating_score.delete.md)
+- [poll_config_selection.delete](actions/poll_config_selection.delete.md)
+- [poll_config_stv_scottish.delete](actions/poll_config_stv_scottish.delete.md)
+- [poll_option.delete](actions/poll_option.delete.md)
```
luisa-beerboom, line 190: I'm very happy about the documentation.

luisa-beerboom (approved):
Very nice

PR #3664, opened 2026-07-13
Limit action worker names to below 256 characters

hjanott:


openslides_backend/action/action_worker.py
```diff
@@ -34,6 +34,24 @@ class ActionWorkerState(StrEnum):
     ABORTED = "aborted"
 
 
+def concatenate_action_names(payload: Payload) -> str:
+    result = ",".join(elem.get("action", "") for elem in payload)
+    if len(result) > 255:
+        action_names = result.split(",")
```
luisa-beerboom, line 39: To make it even safer, we could think about
```suggestion
        action_names = sorted(result.split(","))
```
Takes the complexity from O(n) to O(n log n), but will make some more complex calls work when previously they wouldn't.

Then again
1. It wouldn't preserve action order
2. If the client is going to send the same action multiple times and cannot even group the actions together, it may as well be their fault

I'll leave that decision up to you

luisa-beerboom (changes requested):
I think at some point we should ellipse the rest.
This just stalls the problem until someone sends a request with an even bigger amount of actions

luisa-beerboom (approved):

PR #3659, opened 2026-07-09
[vote] Write changed user fields in history_entry

vkrasnovyd:
Resolves https://github.com/OpenSlides/openslides-backend/issues/3650
Needs https://github.com/OpenSlides/openslides-meta/pull/546

PR #3665, opened 2026-07-14
[vote] Update history_entry.changed_fields

vkrasnovyd:
Continuation of https://github.com/OpenSlides/openslides-backend/pull/3659
Details discussed in https://github.com/OpenSlides/openslides-vote-service/issues/568
Needs: https://github.com/OpenSlides/openslides-meta/pull/550

Reverts the parts of the previous history PR where `changed_fields` were written for the `user` and moves them to the `meeting_user`.

## Format of history information for meeting_user

1. When meeting_user gets created:
- {'group_ids': {'added': [_ids_]}}
- If user is present in meeting before the request: {'is_present': true}

2. When user gets added to a group and/or removed from it:
- {'group_ids': {'added': [_ids_]}, 'removed': [_ids_]}

`added` or `removed` can be omited when its value is empty.

3. When meeting_user gets deleted:
- {'group_ids': {'removed': [_ids_]}}
- If user is present in meeting before the request: {'is_present': false}

4. On user.set_present. Only if meeting_user for this user exists in the meeting:
- {'is_present': _new value_}


openslides_backend/shared/history_events.py
```diff
@@ -67,7 +67,7 @@ def calculate_history_event_payloads(
         (
             model_fqid_to_entry_id[fqid],
             fqid,
-            data["entries"],
+            data.get("entries") or None,
```
luisa-beerboom, line 70: Wouldn't that `or None` be obsolete if you're using `get`?

luisa-beerboom (approved):
One minor remark

openslides_backend/shared/history_events.py
vkrasnovyd: Wrote it this way to avoid the risk of writing an empty list into this field.

PR #3669, opened 2026-07-14
[Dammi] Improve schema diff file handling

hjanott:


openslides_backend/migrations/migration_helper.py
```diff
@@ -220,6 +220,15 @@ def get_database_migration_index(curs: Cursor[DictRow]) -> int:
             "Requested migration indices are not available in version table."
         )
 
+    @staticmethod
+    def get_last_migration_directory() -> str:
+        """Returns the directory to the last migration. Currently only supports numbers until 1099."""
+        idx_as_string = str(MigrationHelper.get_backend_migration_index())
+        for name in os.listdir(MIGRATIONS_PATH):
```
vkrasnovyd, line 227: Will we keep the migrations directly in the `MIGRATIONS_PATH` and not in a nested `migrations` directory like before?

hjanott: It will be a subfolder of this shape and form for each migration. `mig_[migration number].*` This way we can keep many assets separate from the other migrations.

vkrasnovyd (approved):

PR #3671, opened 2026-07-14
[Dammi] Improve migration lookup

hjanott:
This also fixes some documentation and unnecessary things.

### DIFF
--- a/openslides_backend/migrations/migration_handler.py
+++ b/openslides_backend/migrations/migration_handler.py
@@ -1,5 +1,4 @@
 import os
-from typing import Any
 
 from psycopg import Cursor, sql
 from psycopg.rows import DictRow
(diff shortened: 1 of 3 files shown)


openslides_backend/migrations/migration_handler.py
```diff
@@ -39,8 +38,13 @@ def __init__(
         logging: LoggingModule,
     ) -> None:
         super().__init__(env, services, logging)
+        database_m_idx = MigrationHelper.get_database_migration_index(curs)
```
luisa-beerboom, line 41: Wouldn't this have to be ver_curs?

luisa-beerboom (approved):

openslides_backend/migrations/migration_handler.py
hjanott: Once your changes are integrated yes.

PR #3679, opened 2026-07-16
[dammi] More ground work for sql diff

hjanott:


openslides_backend/migrations/sql_diff_generator.py
```diff
@@ -27,19 +29,36 @@ def main() -> int:
...
-    sql = ""
+    sql = "-- REMOVE SECTION --\n"
     # TODO create generate diff content functions in schema generator.
+    # Using a lot of isinstance for pleasing mypy
+    remove = diff["remove"]
+    if isinstance(remove, list) and isinstance(remove[0], list):
+        for collection_name in remove[0]:
+            sql += f"DROP TABLE {collection_name}_t CASCADE;\n"
+            diff_control["remove"][0].remove(collection_name)
+    if isinstance(remove, list) and isinstance(remove_tree_dict := remove[1], dict):
```
vkrasnovyd, line 40: As far as I understand you generate `remove` only with `create_remove_recursive`. In this case if `isinstance(remove, list)` (or even is not None), then `remove[0]` is always a list and `remove[1]` is always a dictionary. Just sometimes empty ones.

openslides_backend/migrations/sql_diff_generator.py
```diff
@@ -110,26 +152,159 @@ def generate_constraints_sql(
...
+                        )
+                    case "timezone":
+                        constraints_sql += Helper.get_inline_timezone_constraint(
+                            table_name, field_name
+                        )
+                    case "relation" | "relation-list":
+                        # TODO
+                        pass
+                    case _:
+                        raise NotImplementedError(f"{constraint}: {value}")
```
vkrasnovyd: If I understand correctly, we'll get thir error for the valid types for which we don't generate constraint (like number or string). If that's the case, add a case to omit them.

openslides_backend/migrations/sql_diff_generator.py
vkrasnovyd, line 125: These seem to be values from `collection-meta.yml`. Its structure and functions are different from the collections files. Wouldn't it be more convenient to check it separately?

openslides_backend/migrations/sql_diff_generator.py
```diff
@@ -110,26 +152,159 @@ def generate_constraints_sql(
...
+                    continue
+                else:
+                    NotImplementedError(
+                        f"{constraint}: {value} is probably a view field or unmentioned in renames."
+                    )
+                # TODO
+                # recreate affected triggers
+                # recreate views
+            case _:
+                raise NotImplementedError(f"{constraint}: {value}")
```
vkrasnovyd, line 281: Like above, some properties should be omited here (like type for example).

openslides_backend/migrations/sql_diff_generator.py
```diff
@@ -110,26 +152,159 @@ def generate_constraints_sql(
...
+    return result
+
+
 def handle_add_tree(
     add_tree_dict: dict[str, tuple[dict[str, Any], dict[str, Any]]],
     dc_add_tree_dict: dict[str, tuple[dict[str, Any], dict[str, Any]]],
 ) -> str:
     sql = ""
     for collection_name, collection_def in add_tree_dict.items():
+        # TODO collection_def[0] What does that even mean for the fields? It didn't have any before?
```
vkrasnovyd: Probably that's the reason for this change: https://github.com/OpenSlides/openslides-meta/pull/420

openslides_backend/migrations/sql_diff_generator.py
```diff
@@ -110,26 +152,159 @@ def generate_constraints_sql(
...
+        # TODO what else can exist here?
```
vkrasnovyd: If you mean in collection_def, then `unique_together` and `unique_together_strict`.

vkrasnovyd (changes requested):

openslides_backend/migrations/sql_diff_generator.py
hjanott: Yes they will be checked another place. I just left it here for now to have all that need to be handled somewhere in one place.

openslides_backend/migrations/sql_diff_generator.py
hjanott: Yes, exactly.

openslides_backend/migrations/sql_diff_generator.py
hjanott: Done.

vkrasnovyd (approved):

PR #3680, opened 2026-07-17
Vote adapter cleanup

vkrasnovyd:
Cleanup after https://github.com/OpenSlides/openslides-backend/pull/3661

Since backend does not send requests to vote service anymore and it is planned to keep it this way, vote service can be removed from the development setup along with the related environment variables.


Janmtbehrens (commented):
The vote service is still present in multiple other files / places and should be removed there too:
- `continuous_integration.yml`as a cloned repository
- `dc.local.yml` as a service

hjanott (approved):

Janmtbehrens (approved):

PR #3686, opened 2026-07-21
[dammi] Handle remove

vkrasnovyd:
Needs https://github.com/OpenSlides/openslides-meta/pull/557

Special handling:
- Skipped from diff.json:
    - relational view fields. @hjanott will view_generation PR deal with deleted view fields or changed `to`?
    - `constant` defined on the view fields. Currently we don't generate triggers for such fields but it will change if we implement https://github.com/OpenSlides/openslides-meta/issues/542;
    - Fields from skipped_in_schema - no need to migrate;
    - `to` and `reference`: sql diff generator will raise `NotImplementedError`. They are required for relational fields and we are not changing type from relational field to something else in the first migrations. So let's skip it for now;
    - `type`: strictly required => will raise `BadCodingException`;
    - `equal_fields`: added a todo. I'll take care of it after rewriting equal_fields in sql generator.
- `edit`-changes triggered by removing lines in the collection-files:
    - Removing enum from the field attributes means changing the type from this enum to the type from the new field definition => `create_remove_recursive` now also updates `edit` section to add this change;
    - Enum can be safely removed only after it is not used as a type in the fields =>  `edit` operations should be executed before `remove`;
    - `maxLength`: removing it should lead to changing the field type => adding this to `diff["edit"]` as well.
- `log_triggers`, `equal_fields`: contain more details in `diff.json` than just the attribute name (needed to build the trigger names);
- `unique_together`, `sequence_scope`: generated separately under `unique_together` next to `fields`;
- changes from `collections-meta.yml`: generated separately under `_meta` next to `collections`;
- Enum types to be dropped that are defined on fields: generated separately under `enum_types` next to `collections`.

Triggers are dropped with `IF EXISTS` -> to avoid errors if trigger was already dropped within the cascade delete of a related column.

### DIFF
--- a/openslides_backend/migrations/sql_diff_generator.py
+++ b/openslides_backend/migrations/sql_diff_generator.py
@@ -312,7 +312,9 @@ def handle_remove_tree(
     for collection_name, field_lists in remove_tree_dict.items():
         fields = field_lists[1]["fields"]
         for field_name in fields[0]:
-            result += f"ALTER TABLE {collection_name}_t DROP COLUMN {field_name};\n"
+            result += (
+                f"ALTER TABLE {collection_name}_t DROP COLUMN {field_name} CASCADE;\n"
+            )
 
             dc_remove_tree_dict[collection_name][1]["fields"][0].remove(field_name)
             # TODO fields[1]
--- a/openslides_backend/migrations/sql_diff_generator.py
+++ b/openslides_backend/migrations/sql_diff_generator.py
@@ -79,7 +79,7 @@ def main() -> int:
 
 
 def remove_empty(dictionary: dict[str, Any], key: str) -> None:
-    if not any(dictionary[key]):
+    if dictionary[key] is None or not any(dictionary[key]):
         del dictionary[key]
 
 
--- a/openslides_backend/migrations/sql_diff_generator.py
+++ b/openslides_backend/migrations/sql_diff_generator.py
@@ -358,6 +358,32 @@ def handle_remove_tree(
                                             collection_name, field_name, "DEFAULT"
                                         )
                                     )
+                                case "required":
+                                    result += (
+                                        Helper.get_drop_column_attribute_statement(
+                                            collection_name, field_name, "NOT NULL"
+                                        )
+                                    )
+                                case "minimum" | "maximum" | "minLength" | "unique":
+                                    constraint_name_func = getattr(
+                                        HelperGetNames,
+                                        f"get_{attr.lower()}_constraint_name",
+                                    )
+                                    constraint_name = constraint_name_func(
+                                        collection_name,
+                                        (
+                                            [field_name]
+                                            if attr == "unique"
+                                            else field_name
+                                        ),
+                                    )
+                                    result += (
+                                        Helper.get_drop_table_constraint_statement(
+                                            collection_name, constraint_name
+                                        )
+                                    )
+                                case _:
+                                    continue
                             dc_remove_tree_dict[collection_name][1]["fields"][1][
                                 field_name
                             ][0].remove(attr)
--- a/openslides_backend/migrations/sql_diff_generator.py
+++ b/openslides_backend/migrations/sql_diff_generator.py
@@ -339,26 +339,51 @@ def handle_remove_tree(
 ) -> str:
     result = ""
     for collection_name, collection_data in remove_tree_dict.items():
-        fields_data = collection_data[1]["fields"]
-        for field_name in fields_data[0]:
-            result += Helper.get_drop_column_statement(collection_name, field_name)
-            dc_remove_tree_dict[collection_name][1]["fields"][0].remove(field_name)
-            remove_empty(dc_remove_tree_dict[collection_name][1], "fields")
-        for field_name, attrs in fields_data[1].items():
-            for attr in attrs[0]:
-                match attr:
-                    case "default":
-                        result += Helper.get_drop_column_attribute_statement(
-                            collection_name, field_name, "DEFAULT"
+        for key, data in collection_data[1].items():
+            match key:
+                case "fields":
+                    for field_name in data[0]:
+                        result += Helper.get_drop_column_statement(
+                            collection_name, field_name
                         )
-                dc_remove_tree_dict[collection_name][1]["fields"][1][field_name][
-                    0
-                ].remove(attr)
-                remove_empty(
-                    dc_remove_tree_dict[collection_name][1]["fields"][1], field_name
-                )
-            remove_empty(dc_remove_tree_dict[collection_name][1], "fields")
-        remove_empty(dc_remove_tree_dict, collection_name)
+                        dc_remove_tree_dict[collection_name][1]["fields"][0].remove(
+                            field_name
+                        )
+                    for field_name, attrs in data[1].items():
+                        for attr in attrs[0]:
+                            match attr:
+                                case "default":
+                                    result += (
+                                        Helper.get_drop_column_attribute_statement(
+                                            collection_name, field_name, "DEFAULT"
+                                        )
+                                    )
+                            dc_remove_tree_dict[collection_name][1]["fields"][1][
+                                field_name
+                            ][0].remove(attr)
+                            remove_empty(
+                                dc_remove_tree_dict[collection_name][1]["fields"][1],
+                                field_name,
+                            )
+                    remove_empty(dc_remove_tree_dict[collection_name][1], "fields")
+                    remove_empty(dc_remove_tree_dict, collection_name)
+                case "unique_together":
+                    for fields in data:
+                        result += Helper.get_drop_table_constraint_statement(
+                            collection_name,
+                            HelperGetNames.get_unique_constraint_name(
+                                collection_name,
+                                Helper.split_unique_together_fields(fields),
+                            ),
+                        )
+                        dc_remove_tree_dict[collection_name][1][
+                            "unique_together"
+                        ].remove(fields)
+                        remove_empty(
+                            dc_remove_tree_dict[collection_name][1],
+                            "unique_together",
+                        )
+                    remove_empty(dc_remove_tree_dict, collection_name)
     return result
 
 
--- a/openslides_backend/migrations/sql_diff_generator.py
+++ b/openslides_backend/migrations/sql_diff_generator.py
@@ -12,10 +12,12 @@ from openslides_backend.migrations.migration_helper import MigrationHelper
 from openslides_backend.migrations.yaml_diff_generator import (
     CollectionsRemoveList,
     EnumTypesRemoveDict,
+    FieldAttributes,
     RemoveDiffDict,
     dumpjson,
     generate_diff,
 )
+from openslides_backend.shared.exceptions import BadCodingException
 
 """
 This script works in conjunction with the yaml_diff_generator.py.
@@ -382,8 +384,32 @@ def handle_remove_tree(
                                             collection_name, constraint_name
                                         )
                                     )
+                                case "sql":
+                                    result += Helper.get_drop_view_statement(
+                                        collection_name
+                                    )
+                                case "constant":
+                                    result += Helper.get_drop_trigger_statement(
+                                        collection_name,
+                                        HelperGetNames.get_constant_field_trigger_name(
+                                            collection_name, field_name
+                                        ),
+                                    )
+                                case value if (
+                                    value in FieldAttributes.skipped_in_schema
+                                ):
+                                    pass
+                                case "type":
+                                    raise BadCodingException(
+                                        f"{collection_name}/{field_name}: '{attr}' is a required field attribute."
+                                    )
                                 case _:
-                                    continue
+                                    # "to" and "reference" are currently skipped. They can only be removed
+                                    # if type changes to not relational field which is not the case
+                                    # in the foreseeable future.
+                                    raise NotImplementedError(
+                                        f"{collection_name}/{field_name}: {attr}"
+                                    )
                             dc_remove_tree_dict[collection_name][1]["fields"][1][
                                 field_name
                             ][0].remove(attr)
--- a/openslides_backend/migrations/sql_diff_generator.py
+++ b/openslides_backend/migrations/sql_diff_generator.py
@@ -35,7 +35,13 @@ def main() -> int:
     if args.dumpjson:
         dumpjson(diff)
 
-    sql = "-- REMOVE SECTION --\n"
+    # Has to happen before remove: field types have to change before dropping the enum
+    sql = "-- EDIT SECTION --\n"
+    edit = diff["edit"]
+    if isinstance(edit, tuple) and isinstance(edit_dict := edit[1], dict):
+        sql += handle_edit_tree(edit_dict, diff_control["edit"][1])
+
+    sql += "\n-- REMOVE SECTION --\n"
     # TODO create generate diff content functions in schema generator.
     # Using a lot of isinstance calls here for pleasing mypy
     remove: RemoveDiffDict | None = diff["remove"]
@@ -54,11 +60,6 @@ def main() -> int:
         sql += generate_new_collection_sql(add[0], diff_control["add"][0])
     if isinstance(add, tuple) and isinstance(add_tree_dict := add[1], dict):
         sql += handle_add_tree(add_tree_dict, diff_control["add"][1])
-
-    sql += "\n-- EDIT SECTION --\n"
-    edit = diff["edit"]
-    if isinstance(edit, tuple) and isinstance(edit_dict := edit[1], dict):
-        sql += handle_edit_tree(edit_dict, diff_control["edit"][1])
 
     # TODO Do this in a sub folder migrations?
     with open(
--- a/openslides_backend/migrations/sql_diff_generator.py
+++ b/openslides_backend/migrations/sql_diff_generator.py
@@ -314,7 +314,7 @@ def handle_remove(remove: RemoveDiffDict, dc_remove_dict: dict[str, Any]) -> str
             collection_names := collections_remove_list[0], list
         ) and isinstance(dc_collection_names := dc_remove_dict["collections"][0], list):
             for collection_name in collection_names:
-                result += f"DROP TABLE {collection_name}_t CASCADE;\n"
+                result += Helper.get_drop_table_statement(collection_name)
                 dc_collection_names.remove(collection_name)
         if isinstance(
             remove_tree_dict := collections_remove_list[1], dict
@@ -340,9 +340,7 @@ def handle_remove_tree(
     for collection_name, field_lists in remove_tree_dict.items():
         fields = field_lists[1]["fields"]
         for field_name in fields[0]:
-            result += (
-                f"ALTER TABLE {collection_name}_t DROP COLUMN {field_name} CASCADE;\n"
-            )
+            result += Helper.get_drop_column_statement(collection_name, field_name)
 
             dc_remove_tree_dict[collection_name][1]["fields"][0].remove(field_name)
             # TODO fields[1]
@@ -359,10 +357,9 @@ def handle_remove_enum_types(
     result = ""
     for collection_name, field_names in remove_tree_dict.items():
         for field_name in field_names:
-            enum_name = HelperGetNames.get_enum_name_for_column(
+            result += Helper.get_drop_enum_type_statement_from_collection_and_column(
                 collection_name, field_name
             )
-            result += f"DROP TYPE {enum_name};\n"
             dc_remove_tree_dict[collection_name].remove(field_name)
         remove_empty(dc_remove_tree_dict, collection_name)
     return result
--- a/openslides_backend/migrations/sql_diff_generator.py
+++ b/openslides_backend/migrations/sql_diff_generator.py
@@ -444,6 +444,9 @@ def handle_remove_tree(
                                             result += Helper.get_drop_trigger_statement(
                                                 log_trigger["on_table"], trigger_name
                                             )
+                                # case "equal_fields":
+                                #     pass
+                                #     TODO: correctly handle deleted equal_fields
                                 case _:
                                     raise NotImplementedError(
                                         f"{collection_name}/{field_name}: {attr}"
--- a/openslides_backend/migrations/sql_diff_generator.py
+++ b/openslides_backend/migrations/sql_diff_generator.py
@@ -404,13 +404,6 @@ def handle_remove_tree(
                                             collection_name, field_name
                                         ),
                                     )
-                                case "sequence_scope":
-                                    result += Helper.get_drop_trigger_statement(
-                                        collection_name,
-                                        HelperGetNames.get_partitioned_sequence_trigger_name(
-                                            collection_name, field_name
-                                        ),
-                                    )
                                 case value if (
                                     value in FieldAttributes.skipped_in_schema
                                 ):
@@ -420,9 +413,9 @@ def handle_remove_tree(
                                         f"{collection_name}/{field_name}: '{attr}' is a required field attribute."
                                     )
                                 case _:
-                                    # "to" and "reference" are currently skipped. They can only be removed
-                                    # if type changes to not relational field which is not the case
-                                    # in the foreseeable future.
+                                    # Skipped as not likely to be removed in the foreseeable future:
+                                    # "to" and "reference": can only be removed if type changes to not relational field
+                                    # "sequence_scope": would turn a sequence field into a regular number field
                                     raise NotImplementedError(
                                         f"{collection_name}/{field_name}: {attr}"
                                     )
--- a/openslides_backend/migrations/sql_diff_generator.py
+++ b/openslides_backend/migrations/sql_diff_generator.py
@@ -9,7 +9,13 @@ import simplejson as json
 from meta.dev.src.generate_sql_schema import GenerateCodeBlocks, Helper
 from meta.dev.src.helper_get_names import HelperGetNames
 from openslides_backend.migrations.migration_helper import MigrationHelper
-from openslides_backend.migrations.yaml_diff_generator import dumpjson, generate_diff
+from openslides_backend.migrations.yaml_diff_generator import (
+    CollectionsRemoveList,
+    EnumTypesRemoveDict,
+    RemoveDiffDict,
+    dumpjson,
+    generate_diff,
+)
 
 """
 This script works in conjunction with the yaml_diff_generator.py.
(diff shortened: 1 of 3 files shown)


openslides_backend/migrations/yaml_diff_generator.py
```diff
@@ -24,6 +30,71 @@
...
+CollectionsRemoveList = list[list[str] | dict[str, Any]]
+EnumTypesRemoveDict = dict[str, list[str]]
+MetaAttributesRemoveList = list[list[str] | dict[str, Any]]
+
+
+class RemoveDiffDict(TypedDict):
+    collections: NotRequired[CollectionsRemoveList]
+    enum_types: NotRequired[EnumTypesRemoveDict]
+    _meta: NotRequired[MetaAttributesRemoveList]
+
```
hjanott: I like these new types.

openslides_backend/migrations/sql_diff_generator.py
```diff
@@ -304,24 +310,222 @@ def handle_rename(
...
+        remove_meta_attributes_list := remove["_meta"], list
+    ):
+        result += handle_remove_meta_attributes(
+            remove_meta_attributes_list, dc_remove_dict["_meta"]
+        )
+        remove_empty(dc_remove_dict, "_meta")
+    return result
+
+
 def handle_remove_tree(
```
hjanott: This is quiet long now. Maybe we can move a match case or two out of here.

openslides_backend/migrations/sql_diff_generator.py
```diff
@@ -304,24 +310,222 @@ def handle_rename(
...
+                                            collection_name, field_name, "NOT NULL"
+                                        )
+                                    )
+                                case "minimum" | "maximum" | "minLength" | "unique":
+                                    constraint_name_func = getattr(
+                                        HelperGetNames,
+                                        f"get_{attr.lower()}_constraint_name",
+                                    )
+                                    constraint_name = constraint_name_func(
+                                        collection_name,
```
hjanott: Clever! Ah the power of homogenous naming. :3

openslides_backend/migrations/sql_diff_generator.py
```diff
@@ -304,24 +310,222 @@ def handle_rename(
...
+                                        collection_name
+                                    )
+                                case "constant":
+                                    result += Helper.get_drop_trigger_statement(
+                                        collection_name,
+                                        HelperGetNames.get_constant_field_trigger_name(
+                                            collection_name, field_name
+                                        ),
+                                    )
+                                case "sequence_scope":
```
hjanott: In this case we also need to drop the sequence itself.

openslides_backend/migrations/sql_diff_generator.py
```diff
@@ -304,24 +310,222 @@ def handle_rename(
...
+                                            else field_name
+                                        ),
+                                    )
+                                    result += (
+                                        Helper.get_drop_table_constraint_statement(
+                                            collection_name, constraint_name
+                                        )
+                                    )
+                                case "sql":
+                                    result += Helper.get_drop_view_statement(
```
hjanott: This needs to be adjusted in my upcoming PR. If you want you can already introduce the `alter_views` variable. Like here in lines 30 and 266 https://github.com/OpenSlides/openslides-backend/compare/feature/dammi...hjanott:openslides-backend:view_generation#diff-91ac0503ffea24be723e20175d91b0831cc854f4cfac466c614eac16caa921c2R30

openslides_backend/migrations/sql_diff_generator.py
```diff
@@ -304,24 +310,222 @@ def handle_rename(
...
+    remove_tree_dict: dict[str, Any],
+    dc_remove_tree_dict: dict[str, Any],
+) -> str:
+    result = ""
+    for collection_name, collection_data in remove_tree_dict.items():
+        for key, data in collection_data[1].items():
+            match key:
+                case "fields":
+                    for field_name in data[0]:
+                        result += Helper.get_drop_column_statement(
```
hjanott: If you decide to introduce the `alter_views` already, please also add the other collection here if that ones field is a view field.

openslides_backend/migrations/yaml_diff_generator.py
```diff
@@ -81,30 +158,116 @@ def validate_renames(
...
-            missing_entries.append(key)
-        elif isinstance(prev_value, dict):
+            if key == "id" or len(path) >= 3 and path[2] == "id":
+                continue
+            if key in [
+                *CollectionAttributes.unique_together,
+                "log_triggers",
+                "equal_fields",
+            ]:
+                # Old definitions are needed to re-build the trigger difinitions names
```
hjanott:
```suggestion
                # Old definitions are needed to re-build the trigger definitions names
```

openslides_backend/migrations/yaml_diff_generator.py
```diff
@@ -172,5 +342,45 @@ def load_models(mig_data_path: str) -> dict[str, Any]:
...
+def is_view_field(
+    collection_name: str,
+    field_name: str,
+    field_data: dict[str, Any],
+    all_prev_models: dict[str, dict[str, Any]],
+) -> bool:
+    own = TableFieldType(collection_name, field_name, field_data)
+
+    new_models = InternalHelper.MODELS
+    InternalHelper.MODELS = all_prev_models
```
hjanott: In my branch I'm also developing `is_view_field`. The most interesting difference is that you are passing the to be used models definition and patching it from here. I on the other hand use `CURR_MODELS` directly to create own_field from fully specified field data.
Are these different use cases? Should yours be called `was_view_field` since you are looking into the past?
We will decide later if we want a flag depending on whether I use `get_view_field_state_write_fields` instead.

openslides_backend/migrations/yaml_diff_generator.py
```diff
@@ -24,6 +30,71 @@
...
+
+
+class RemoveDiffDict(TypedDict):
+    collections: NotRequired[CollectionsRemoveList]
+    enum_types: NotRequired[EnumTypesRemoveDict]
+    _meta: NotRequired[MetaAttributesRemoveList]
+
+
+class FieldAttributes:
+    # TODO: if the list changes, we might need to take the migration index into account
```
hjanott: I don't think so since this script will only be run for the current migration.

openslides_backend/migrations/yaml_diff_generator.py
```diff
@@ -81,30 +158,116 @@ def validate_renames(
...
+
 def create_remove_recursive(
     prev_models: dict[str, Any],
     curr_models: dict[str, Any],
     renames_dict: dict[str, Any],
-) -> list[list[str] | dict[str, Any]] | None:
+    all_prev_models: dict[str, Any],
+    secondary_edits: dict[str, Any] = {},
+    enum_tree: EnumTypesRemoveDict = {},
+    path: tuple[str, ...] = (),
```
hjanott, line 217: If this stays this needs better explanation.

openslides_backend/migrations/yaml_diff_generator.py
```diff
@@ -81,30 +158,116 @@ def validate_renames(
...
+        1
+    ].setdefault(field, [{}, {}])[0][attr] = value
+
+
 def create_remove_recursive(
     prev_models: dict[str, Any],
     curr_models: dict[str, Any],
     renames_dict: dict[str, Any],
-) -> list[list[str] | dict[str, Any]] | None:
+    all_prev_models: dict[str, Any],
```
hjanott: With my changes this will not have to be passed anymore.

openslides_backend/migrations/yaml_diff_generator.py
```diff
@@ -81,30 +158,116 @@ def validate_renames(
...
+    secondary_edits: dict[str, Any] = {},
```
hjanott: The way this is used it doesn't need a default value.

openslides_backend/migrations/yaml_diff_generator.py
```diff
@@ -81,30 +158,116 @@ def validate_renames(
...
+                    else:
+                        field_name = path[2]
+                        field_def = all_prev_models[path[0]][path[1]][path[2]]
+                    if "type" in field_def and is_relational_field(field_def["type"]):
+                        if not is_view_field(
+                            path[0], field_name, field_def, all_prev_models
+                        ):
+                            missing_entries.append(key)
+                    else:
+                        missing_entries.append(key)
```
hjanott:
```suggestion
                    if not (
                        "type" in field_def 
                        and is_relational_field(field_def["type"])
                        and is_view_field(path[0], field_name, field_def, all_prev_models)
                    ):
                        missing_entries.append(key)
```
Does it even need the `is_relational_field` check?

openslides_backend/migrations/yaml_diff_generator.py
```diff
@@ -81,30 +158,116 @@ def validate_renames(
...
+                "equal_fields",
+            ]:
+                # Old definitions are needed to re-build the trigger difinitions names
+                tree[key] = prev_value
+            elif key == "maxLength":
+                # Should be processed as type change
+                update_edits_tree(secondary_edits, path[0], path[2], "maxLength", None)
+            elif key == "sequence_scope":
+                missing_entries.append(key)
+                tree[key] = prev_value
```
hjanott: Could also be retrieved via upcoming `PREV_MODELS`.

hjanott (changes requested):
Lot of tough nuts to crack here. If https://github.com/OpenSlides/openslides-meta/pull/562/ gets merged first you can also do the equal fields drops here too.

openslides_backend/migrations/sql_diff_generator.py
vkrasnovyd: As discussed, not touching it for now as this case is relatively complex and is not likely to be used in the near future. Diff generator will be throwing NotImplementedError.

openslides_backend/migrations/yaml_diff_generator.py
vkrasnovyd: I agree with the reasoning. Removing this comment.

openslides_backend/migrations/yaml_diff_generator.py
vkrasnovyd: Added a small docstring

openslides_backend/migrations/yaml_diff_generator.py
vkrasnovyd: Removed this default.

openslides_backend/migrations/yaml_diff_generator.py
vkrasnovyd: Yes, it needs it. `InternalHelper.get_cardinality()` raises error if a not relational field is being passed.

openslides_backend/migrations/yaml_diff_generator.py
vkrasnovyd: Have taken these changes from your PR and used them here.

openslides_backend/migrations/yaml_diff_generator.py
vkrasnovyd: Removed it instead. Now it just gets added to missing_entries and error gets raised.

openslides_backend/migrations/sql_diff_generator.py
vkrasnovyd: Done.

openslides_backend/migrations/sql_diff_generator.py
```diff
@@ -42,16 +60,18 @@ def main() -> int:
...
-    sql = "-- REMOVE SECTION --\n"
+    # Has to happen before remove: field types have to change before dropping the enum
+    sql = "-- EDIT SECTION --\n"
+    edit = diff["edit"]
+    if isinstance(edit, tuple) and isinstance(edit_dict := edit[1], dict):
+        sql += handle_edit_tree(edit_dict, diff_control["edit"][1])
+
+    sql += "\n-- REMOVE SECTION --\n"
     # TODO create generate diff content functions in schema generator.
     # Using a lot of isinstance calls here for pleasing mypy
```
hjanott: I don't see the use for this TODO. Let's delete it.
Move the second comment before the `EDIT SECTION`.

openslides_backend/migrations/sql_diff_generator.py
```diff
@@ -120,6 +142,692 @@ def generate_new_collection_sql(add: dict[str, Any], dc_add: dict[str, Any]) ->
...
+                    # )
+                    # curr_equal_fields = set(
+                    #     GenerateCodeBlocks.get_equal_fields(
+                    #         curr_own_table_field, curr_foreign_table_field
+                    #     )
+                    # )
+                    # if added_equal_fields := (
+                    #     curr_equal_fields - prev_equal_fields
+                    # ):
+
```
hjanott, line 349: What is this commented out code for?

openslides_backend/migrations/sql_diff_generator.py
```diff
@@ -120,6 +142,692 @@ def generate_new_collection_sql(add: dict[str, Any], dc_add: dict[str, Any]) ->
...
+        prev_collectionfields = set(prev_foreign_table_fields.keys())
+        curr_collectionfields = set(curr_foreign_table_fields.keys())
+
+        removed_collectionfields = prev_collectionfields - curr_collectionfields
+        remaining_collectionfields = prev_collectionfields - removed_collectionfields
+
+        # if added_collectionfields := (
+        #     curr_collectionfields - prev_collectionfields
+        # ):
+
```
hjanott: What is this commented out code for?

openslides_backend/migrations/sql_diff_generator.py
```diff
@@ -120,6 +142,692 @@ def generate_new_collection_sql(add: dict[str, Any], dc_add: dict[str, Any]) ->
...
+            cls.checked_equal_fields[foreign_table_field.table].add(
+                foreign_table_field.column
+            )
+        cls.checked_equal_fields[collection_name].add(field_name)
+
+    @classmethod
+    def handle_alter_equal_fields(cls) -> str:
+        # TODO: This method includes commented out lines for cases when triggers have to added.
+        # handle_add and handle edit sould probably already handle all such cases. If it's true,
+        # the commented out lines should be removed.
```
hjanott, line 184: Then do so.

openslides_backend/migrations/sql_diff_generator.py
```diff
@@ -120,6 +142,692 @@ def generate_new_collection_sql(add: dict[str, Any], dc_add: dict[str, Any]) ->
...
+            foreign_table_field: TableFieldType = (
+                TableFieldType.get_definitions_from_foreign(
+                    own_field_def.get("to"),
+                    own_field_def.get("reference"),
+                )
+            )
+            state, primary, *_ = InternalHelper.check_relation_definitions(
+                own_table_field, [foreign_table_field]
+            )
+            condition = (
```
hjanott: What does this condition express? Maybe there is a better name.

openslides_backend/migrations/sql_diff_generator.py
```diff
@@ -120,6 +142,692 @@ def generate_new_collection_sql(add: dict[str, Any], dc_add: dict[str, Any]) ->
...
+                cls.equal_fields_diff[foreign_table_field.table].add(
+                    foreign_table_field.column
+                )
+            cls.checked_equal_fields[foreign_table_field.table].add(
+                foreign_table_field.column
+            )
+        cls.checked_equal_fields[collection_name].add(field_name)
+
+    @classmethod
+    def handle_alter_equal_fields(cls) -> str:
```
hjanott: I would prefer this not to be in the middle of `handle_generic_relations` and `handle_plain_relations`. Please move it up or down.

openslides_backend/migrations/sql_diff_generator.py
```diff
@@ -120,6 +142,692 @@ def generate_new_collection_sql(add: dict[str, Any], dc_add: dict[str, Any]) ->
...
+    @classmethod
+    def update_equal_fields_diff(cls, collection_name: str, field_name: str) -> None:
+        if field_name in cls.checked_equal_fields.get(collection_name, set()):
+            return
+
+        own_field_def = PREV_MODELS[collection_name]["fields"][field_name]
+        own_table_field = TableFieldType(collection_name, field_name, own_field_def)
+        type_ = own_field_def["type"]
+
+        if type_ in ["generic-relation", "generic-relation-list"]:
```
hjanott: Copied from `handle_alter_equal_fields`.
```suggestion
        if "generic" in type_:
```

hjanott: Could also be `type_.startswith("generic")`.

openslides_backend/migrations/sql_diff_generator.py
```diff
@@ -120,6 +142,692 @@ def generate_new_collection_sql(add: dict[str, Any], dc_add: dict[str, Any]) ->
...
+                    (own_table, own_trigger_name),
+                    (foreign_table, foreign_trigger_name),
+                ]
+            )
+            if intermediate_table not in RemoveHelper.intermediate_tables_to_remove:
+                to_drop.append((intermediate_table, intermediate_trigger_name))
+
+        return to_drop
+
+    # Helpers
```
hjanott: Helper inside Helper? Or was this meant to be above this helper class?

openslides_backend/migrations/sql_diff_generator.py
```diff
@@ -120,6 +142,692 @@ def generate_new_collection_sql(add: dict[str, Any], dc_add: dict[str, Any]) ->
...
+                        # "to" and "reference": can only be removed if type changes to not relational field
+                        # "sequence_scope": would turn a sequence field into a regular number field
+                        raise NotImplementedError(
+                            f"{collection_name}/{field_name}: {attr}"
+                        )
+                dc_remove_tree_dict[field_name][0].remove(attr)
+                remove_empty(dc_remove_tree_dict, field_name)
+            for attr, attr_data in attrs[1].items():
+                match attr:
+                    case "log_triggers":
```
hjanott, line 724: Would this mean that some default generated triggers need to be created?

openslides_backend/migrations/yaml_diff_generator.py
```diff
@@ -95,19 +161,42 @@ def dumpjson(diff: dict[str, Any]) -> None:
 
 
 def generate_diff() -> dict[str, Any]:
+    # Edits caused by adding or removing field attributes
+    secondary_edits: dict[str, Any] = {}
```
hjanott, line 165: How about using `defaultdict` with a lambda generator to simplify `update_edits_tree`?

openslides_backend/migrations/yaml_diff_generator.py
```diff
@@ -207,11 +346,48 @@ def create_edit_recursive(
...
+    return is_view_field or bool(write_fields)
+
+
+@contextmanager
+def prev_models_context() -> Iterator[None]:
+    curr_models = InternalHelper.MODELS
+    InternalHelper.MODELS = PREV_MODELS
+    try:
+        yield
+    finally:
```
hjanott, line 403: Does this need 'try finally'?

openslides_backend/migrations/yaml_diff_generator.py
```diff
@@ -122,18 +211,66 @@ def create_remove_recursive(
...
+                if len(path) == 3:
+                    if not (
+                        isinstance(prev_value, dict)
+                        and "type" in prev_value
+                        and prev_value.get("to")
+                        and was_view_field(path[0], key, prev_value)
+                    ):
+                        missing_entries.append(key)
+                else:
+                    missing_entries.append(key)
```
hjanott, line 262: Something like this should be possible.
```suggestion
elif curr_models:
    if (
            not len(path) == 3
            or not (isinstance(prev_value, dict)
            and "type" in prev_value
            and "to" in prev_value
            and was_view_field(path[0], key, prev_value))
        ):
        missing_entries.append(key)
```

hjanott: Or is the `constant` thing a subcase in this?

openslides_backend/migrations/yaml_diff_generator.py
```diff
@@ -122,18 +211,66 @@ def create_remove_recursive(
...
-        return [missing_entries, tree]
-    else:
-        return None
+        combined_result["collections"] = [missing_entries, tree]
+    if enum_tree:
+        combined_result["enum_types"] = enum_tree
+
+    if combined_result:
+        return combined_result
+    return None
```
hjanott: Please move this to another function.

hjanott (changes requested):

openslides_backend/migrations/sql_diff_generator.py
vkrasnovyd: Was suposed to be more functions for the method above. Will remove.

openslides_backend/migrations/sql_diff_generator.py
vkrasnovyd: The suggested syntax is more precise indeed. Will change.

openslides_backend/migrations/sql_diff_generator.py
vkrasnovyd: All the commented code in this class was originally the part of `handle_alter_equal_fields` which contains a comment what this code can be used for. In short: if it is not possible to retrieve the equal_fields trigger definitions generated with GenerateCodeBlocks inside of `handle_add` and/or `handle_edit`, this code should be uncommented and finalized. Otherwise - deleted.

openslides_backend/migrations/sql_diff_generator.py
vkrasnovyd: Will do.

openslides_backend/migrations/sql_diff_generator.py
vkrasnovyd: Only their names to drop the triggers.

openslides_backend/migrations/yaml_diff_generator.py
vkrasnovyd: Will do.

openslides_backend/migrations/yaml_diff_generator.py
vkrasnovyd: I'd rather keep it. This context manager changes the variable of the class that is used for different operations, not exclusively for the current migration. That's why even though we re-generate InternalHelper.MODELS in the current codebase, I want to make sure that we restore its state no matter what happens inside of the context manager.

openslides_backend/migrations/sql_diff_generator.py
vkrasnovyd: As we have discussed, in my opinion, the right point to decide whether to finalise this code or remove it is at the end of implementing `handle_add` and `handle_edit`.

openslides_backend/migrations/sql_diff_generator.py
vkrasnovyd: Will do.

openslides_backend/migrations/yaml_diff_generator.py
vkrasnovyd: No, `constant` most likely won't be a special case here.

openslides_backend/migrations/yaml_diff_generator.py
vkrasnovyd: This function also updates the tree of `handle_edit` and I'd rather keep this method more compex then change the definition of 2 dictionaries.

openslides_backend/migrations/yaml_diff_generator.py
vkrasnovyd: Checked the current logic again:
1. This check does not impact if the `constant` is generated in `diff.json` - it is always there.
2. Not only reading-writing side matters, but also the type of the field.

Since it gets a bit complex, I'll move constant-related logic and the shortened comment to `sql_diff_generator`.

openslides_backend/migrations/yaml_diff_generator.py
```diff
@@ -259,7 +253,13 @@ def create_remove_recursive(
             return [missing_entries, tree]
         else:
             return None
+    else:
+        return generate_remove_diff_dict(missing_entries, tree, enum_tree)
+
 
+def generate_remove_diff_dict(
+    missing_entries: list[str], tree: dict[str, Any], enum_tree: EnumTypesRemoveDict
+) -> RemoveDiffDict | None:
```
hjanott: I would find it more logical if this calls `create_remove_recursive` and composes the diff dict. Please see if that is possible.

openslides_backend/migrations/sql_diff_generator.py
hjanott: I like it. As discussed see what other conditions this may correlate to and if we could create a generalized function for this.

hjanott (changes requested):

openslides_backend/migrations/sql_diff_generator.py
```diff
@@ -120,6 +142,709 @@ def generate_new_collection_sql(add: dict[str, Any], dc_add: dict[str, Any]) ->
...
+            for attr in attrs[0]:
+                match attr:
+                    case "default":
+                        result += AlterSchemaHelper.get_drop_column_attribute_statement(
+                            collection_name, field_name, "DEFAULT"
+                        )
+                    case "required":
+                        result += AlterSchemaHelper.get_drop_column_attribute_statement(
+                            collection_name, field_name, "NOT NULL"
+                        )
```
hjanott: What do you think about this? I'm a little unsure.
```suggestion
                    case "default" | "required":
                        if attr == "default":
                            statement = "DEFAULT"
                        else:
                            statement = "NOT NULL"
                        result += AlterSchemaHelper.get_drop_column_attribute_statement(
                            collection_name, field_name, statement
                        )
```

openslides_backend/migrations/sql_diff_generator.py
vkrasnovyd: Can not be completely generalized now, but was able to simplify this part.

openslides_backend/migrations/yaml_diff_generator.py
vkrasnovyd: Done. After this restructuring was also able to switch from list to tuple and this way remove isinstance checks from hendle_remove-methods.

openslides_backend/migrations/sql_diff_generator.py
vkrasnovyd: I like the idea, it makes the syntax similar to the next case. Did it without introducing `statement` variable.

openslides_backend/migrations/sql_diff_generator.py
```diff
@@ -120,6 +138,669 @@ def generate_new_collection_sql(add: dict[str, Any], dc_add: dict[str, Any]) ->
...
+        for table in sorted(cls.intermediate_tables_to_remove):
+            result += AlterSchemaHelper.get_drop_table_statement(table)
+        result += EqualFieldsHelper.handle_alter_equal_fields()
+        return result
+
+    @classmethod
+    def handle_remove_fields(
+        cls,
+        remove_list: tuple[list[str], dict[str, Any]],
+        dc_remove_list: tuple[list[str], dict[str, Any]],
```
hjanott: This is still called `list`.

openslides_backend/migrations/sql_diff_generator.py
```diff
@@ -120,6 +138,669 @@ def generate_new_collection_sql(add: dict[str, Any], dc_add: dict[str, Any]) ->
...
+                    collection_name, field_name
+                )
+                dc_remove_tree_dict[collection_name].remove(field_name)
+            remove_empty(dc_remove_tree_dict, collection_name)
+        return result
+
+    @staticmethod
+    def handle_remove_meta_attributes(
+        remove_meta_attributes_list: MetaAttributesRemoveTuple,
+        dc_remove_meta_attributes_list: MetaAttributesRemoveTuple,
```
hjanott: This is still called `list`.

hjanott (changes requested):
Love the changes. Just one tiny thing I spotted.

hjanott (approved):
Just see to those 2 renames before merging.

openslides_backend/migrations/sql_diff_generator.py
vkrasnovyd: Updated.

openslides_backend/migrations/sql_diff_generator.py
vkrasnovyd: Updated.

PR #3699, opened 2026-07-30
[dammi] View generation

hjanott:
* centralize models variables
 * rename some functions
 * collect alter_views conditionally and recreate in the end
 * extract get_view_field_state_write_fields
 * debug main functions with make test-file

### DIFF
--- a/cli/util/util.py
+++ b/cli/util/util.py
@@ -8,6 +8,12 @@ import requests
 import yaml
 
 from meta.dev.src.helper_get_names import DEFAULT_COLLECTIONS_DIR as SOURCE_COLLECTIONS
+from meta.dev.src.helper_get_names import (
+    FieldSqlErrorType,
+    HelperGetNames,
+    InternalHelper,
+    TableFieldType,
+)
 
 ROOT = os.path.join(
     os.path.dirname(os.path.abspath(__file__)),
@@ -69,3 +75,88 @@ def assert_equal(stream: TextIOBase, destination: str) -> None:
     result.check_returncode()
     with open(destination) as f:
         assert f.read() == result.stdout
+
+
+def get_view_field_state_write_fields(
+    collection_name: str, field_name: str, value: dict[str, Any]
+) -> tuple[bool, bool, tuple[str, str, str, list[str]] | None]:
+    """
+    Purpose:
+        Checks whether a field is a view field and if other fields need to be written in an intermediate
+        table.
+    Input:
+    - collection_name
+    - field_name
+    - value : represents the definition of the field ( field_name in collection_name )
+    Returns:
+    - is_view_field : whether the field is a view field or not
+    - is_primary: wether the field is primary or not
+    - write_fields:
+        - None if no fields need to be written
+        - Tuple
+            table_name : name of the intermediate table
+            field1
+            field2
+            foreign_fields
+    """
+    # variable declaration
+    own: TableFieldType
+    field_type: str
+    state: FieldSqlErrorType
+    primary: bool
+    error: str
+    is_view_field: bool
+    foreign: TableFieldType
+    foreign_type: str
+    table_name: str = ""
+    field1: str = ""
+    field2: str = ""
+    write_fields: tuple[str, str, str, list[str]] | None = None
+
+    # create TableFieldType own out of collection_name, field_name, value as field_def
+    own = TableFieldType(collection_name, field_name, value)
+    field_type = own.field_def.get("type", "")
+
+    # get the foreign field list and check the relations
+    foreign_fields = InternalHelper.get_definitions_from_foreign_list(
+        value.get("to", None), value.get("reference", None)
+    )
+    state, primary, _, error = InternalHelper.check_relation_definitions(
+        own, foreign_fields
+    )
+    is_view_field = state == FieldSqlErrorType.SQL
+
+    if not value.get("sql"):
+        foreign = foreign_fields[0]
+        foreign_type = foreign.field_def.get("type", "")
+        if "relation-list" == field_type == foreign_type:
+            table_name = HelperGetNames.get_nm_table_name(own, foreign)
+            field1 = HelperGetNames.get_field_in_n_m_relation_list(own, foreign)
+            field2 = HelperGetNames.get_field_in_n_m_relation_list(foreign, own)
+            if field1 == field2:
+                field1 += "_1"
+                field2 += "_2"
+            write_fields = (table_name, field1, field2, [])
+        elif "generic-relation-list" in (field_type, foreign_type):
+            write_fields = get_write_fields_for_generic(own, foreign_fields, primary)
+
+    assert error == "", error
+
+    return is_view_field, primary, write_fields
+
+
+def get_write_fields_for_generic(
+    own: TableFieldType, foreign_fields: list[TableFieldType], primary: bool
+) -> tuple[str, str, str, list[str]] | None:
+    if primary:
+        table_name = HelperGetNames.get_gm_table_name(own)
+    else:
+        table_name = HelperGetNames.get_gm_table_name(foreign_fields[0])
+    field1 = f"{own.table}_{own.ref_column}"
+    field2 = own.intermediate_column
+    return (
+        table_name,
+        field1,
+        field2,
+        [f"{field2}_{field.table}_{field.ref_column}" for field in foreign_fields],
+    )
--- a/cli/util/util.py
+++ b/cli/util/util.py
@@ -7,6 +7,7 @@ from typing import Any, cast
 import requests
 import yaml
 
+from meta.dev.src.generate_sql_schema import Helper
 from meta.dev.src.helper_get_names import DEFAULT_COLLECTIONS_DIR as SOURCE_COLLECTIONS
 from meta.dev.src.helper_get_names import (
     FieldSqlErrorType,
@@ -78,7 +79,7 @@ def assert_equal(stream: TextIOBase, destination: str) -> None:
 
 
 def get_view_field_state_write_fields(
-    collection_name: str, field_name: str, value: dict[str, Any]
+    collection_name: str, field_name: str, field_def: dict[str, Any]
 ) -> tuple[bool, bool, tuple[str, str, str, list[str]] | None]:
     """
     Purpose:
@@ -87,7 +88,7 @@ def get_view_field_state_write_fields(
     Input:
     - collection_name
     - field_name
-    - value : represents the definition of the field ( field_name in collection_name )
+    - field_def : represents the definition of the field ( field_name in collection_name )
     Returns:
     - is_view_field : whether the field is a view field or not
     - is_primary: wether the field is primary or not
@@ -108,35 +109,26 @@ def get_view_field_state_write_fields(
     is_view_field: bool
     foreign: TableFieldType
     foreign_type: str
-    table_name: str = ""
-    field1: str = ""
-    field2: str = ""
     write_fields: tuple[str, str, str, list[str]] | None = None
 
-    # create TableFieldType own out of collection_name, field_name, value as field_def
-    own = TableFieldType(collection_name, field_name, value)
-    field_type = own.field_def.get("type", "")
+    # create TableFieldType own out of collection_name, field_name, field_def
+    own = TableFieldType(collection_name, field_name, field_def)
+    field_type = field_def.get("type", "")
 
     # get the foreign field list and check the relations
     foreign_fields = InternalHelper.get_definitions_from_foreign_list(
-        value.get("to", None), value.get("reference", None)
+        field_def.get("to", None), field_def.get("reference", None)
     )
     state, primary, _, error = InternalHelper.check_relation_definitions(
         own, foreign_fields
     )
     is_view_field = state == FieldSqlErrorType.SQL
 
-    if not value.get("sql"):
+    if not field_def.get("sql"):
         foreign = foreign_fields[0]
         foreign_type = foreign.field_def.get("type", "")
         if "relation-list" == field_type == foreign_type:
-            table_name = HelperGetNames.get_nm_table_name(own, foreign)
-            field1 = HelperGetNames.get_field_in_n_m_relation_list(own, foreign)
-            field2 = HelperGetNames.get_field_in_n_m_relation_list(foreign, own)
-            if field1 == field2:
-                field1 += "_1"
-                field2 += "_2"
-            write_fields = (table_name, field1, field2, [])
+            write_fields = Helper.get_nm_table_name_and_fields(own, foreign) + ([],)
         elif "generic-relation-list" in (field_type, foreign_type):
             write_fields = get_write_fields_for_generic(own, foreign_fields, primary)
 
--- a/cli/util/util.py
+++ b/cli/util/util.py
@@ -111,7 +111,6 @@ def get_view_field_state_write_fields(
     foreign_type: str
     write_fields: tuple[str, str, str, list[str]] | None = None
 
-    # create TableFieldType own out of collection_name, field_name, field_def
     own = TableFieldType(collection_name, field_name, field_def)
     field_type = field_def.get("type", "")
 
--- a/openslides_backend/migrations/sql_diff_generator.py
+++ b/openslides_backend/migrations/sql_diff_generator.py
@@ -68,7 +68,7 @@ def main() -> int:
     sql += "\n-- VIEWS UPDATE SECTION --\n"
     for collection_name in sorted(alter_views):
         sql += (
-            GenerateCodeBlocks.view_sql.get(collection_name, "")
+            GenerateCodeBlocks.view_sql[collection_name]
             .lstrip("\n")
             .replace("CREATE", "CREATE OR REPLACE")
         )
--- a/openslides_backend/migrations/sql_diff_generator.py
+++ b/openslides_backend/migrations/sql_diff_generator.py
@@ -6,10 +6,15 @@ from typing import Any, cast
 
 import simplejson as json
 
+from cli.util.util import get_view_field_state_write_fields
 from meta.dev.src.generate_sql_schema import GenerateCodeBlocks, Helper
 from meta.dev.src.helper_get_names import HelperGetNames
 from openslides_backend.migrations.migration_helper import MigrationHelper
-from openslides_backend.migrations.yaml_diff_generator import dumpjson, generate_diff
+from openslides_backend.migrations.yaml_diff_generator import (
+    CURR_MODELS,
+    dumpjson,
+    generate_diff,
+)
 
 """
 This script works in conjunction with the yaml_diff_generator.py.
@@ -17,6 +22,8 @@ To use this script create a folder 'previous_models' next to it and copy the unc
 It will generate the sql diff comparing it to the changes made to the model definitions present in the meta.
 The sql diff will be written to 'migrations/mig_[last migration number].*/schema_diff.sql'.
 """
+
+alter_views: set[str] = set()
 
 
 def main() -> int:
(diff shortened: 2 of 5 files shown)


openslides_backend/migrations/sql_diff_generator.py
```diff
@@ -58,6 +65,13 @@ def main() -> int:
     if isinstance(edit, tuple) and isinstance(edit_dict := edit[1], dict):
         sql += handle_edit_tree(edit_dict, diff_control["edit"][1])
 
+    sql += "\n-- VIEW UPDATE SECTION --\n"
```
vkrasnovyd:
```suggestion
    sql += "\n-- VIEWS UPDATE SECTION --\n"
```

Seems to be more descriptive this way.

openslides_backend/migrations/sql_diff_generator.py
```diff
@@ -58,6 +65,13 @@ def main() -> int:
...
+    for collection_name in alter_views:
+        sql += (
+            GenerateCodeBlocks.view_sql.get(collection_name, "")
```
vkrasnovyd:
```suggestion
            GenerateCodeBlocks.view_sql[collection_name]
```

Similar reasons as in your comments about `IF EXISTS` to my PR: as far as I understand if no view code is found for the collection from `alter_views` then we want an error and not a silent skip.

openslides_backend/migrations/sql_diff_generator.py
```diff
@@ -288,21 +328,52 @@ def handle_rename(
 ) -> str:
     result = ""
     for collection_name_old, value in rename.items():
+        table_name = f"{collection_name_old}_t"
         if isinstance(value, str):
-            result += f"ALTER TABLE {collection_name_old}_t RENAME TO {value}_t;\n"
-            # TODO Renaming views
+            result += f"ALTER TABLE {table_name} RENAME TO {value}_t;\n"
```
vkrasnovyd: I'd prefer `HelperGetNames.get_table_name()` instead of hardcoding `_t` - to keep it more safe and self-explanatory.

openslides_backend/migrations/sql_diff_generator.py
```diff
@@ -288,21 +328,52 @@ def handle_rename(
 ) -> str:
     result = ""
     for collection_name_old, value in rename.items():
+        table_name = f"{collection_name_old}_t"
         if isinstance(value, str):
```
vkrasnovyd: Maybe the structure similar to other diff blocks will be easier to handle:
- list[0] - dictionary of old and new collection names
- list[1] - dictionaries of old and new field names nested per collection

openslides_backend/migrations/sql_diff_generator.py
```diff
@@ -288,21 +328,52 @@ def handle_rename(
...
     return result
 
 
+def alter_views_conditionally(
+    collection_name: str, field_name: str, has_write_fields: bool, is_view_field: bool
+) -> None:
+    if has_write_fields or is_view_field:
+        alter_views.add(collection_name)
+    if has_write_fields or not is_view_field:
+        reference = CURR_MODELS[collection_name]["fields"][field_name]["reference"]
```
vkrasnovyd: You can directly pass the same `field_def` that you are already calculating for `get_view_field_state_write_fields` and get reference from it.

cli/util/util.py
```diff
@@ -69,3 +75,88 @@ def assert_equal(stream: TextIOBase, destination: str) -> None:
...
+    collection_name: str, field_name: str, value: dict[str, Any]
+) -> tuple[bool, bool, tuple[str, str, str, list[str]] | None]:
+    """
+    Purpose:
+        Checks whether a field is a view field and if other fields need to be written in an intermediate
+        table.
+    Input:
+    - collection_name
+    - field_name
+    - value : represents the definition of the field ( field_name in collection_name )
```
vkrasnovyd: How about renaming value to something more descriptive like field_def?

openslides_backend/migrations/sql_diff_generator.py
```diff
@@ -288,21 +328,52 @@ def handle_rename(
...
+            # TODO recreate dependend triggers
             del dc_rename_dict[collection_name_old]
         else:
             dc_collection = cast(dict, dc_rename_dict[collection_name_old])
             for field_name_old, field_name_new in value["fields"].items():
-                result += f"ALTER TABLE {collection_name_old}_t RENAME COLUMN {field_name_old} TO {field_name_new}_t;\n"
+                assert isinstance(field_name_new, str)
+                result += f"ALTER TABLE {table_name} RENAME COLUMN {field_name_old} TO {field_name_new};\n"
+                field_def = CURR_MODELS[collection_name_old]["fields"][field_name_new]
+                if field_def.get("to"):
```
vkrasnovyd: Is it for linter? Because `to` is strictly required for the relational fields and it's value is checked in the validator. If it's not a collection name ot list of collections, the meta PR won't pass the checks.

openslides_backend/migrations/sql_diff_generator.py
```diff
@@ -288,21 +328,52 @@ def handle_rename(
...
+            result += f"DROP VIEW {collection_name_old};\n"
+            alter_views.add(value)
+            # TODO recreate dependend triggers
             del dc_rename_dict[collection_name_old]
         else:
             dc_collection = cast(dict, dc_rename_dict[collection_name_old])
             for field_name_old, field_name_new in value["fields"].items():
-                result += f"ALTER TABLE {collection_name_old}_t RENAME COLUMN {field_name_old} TO {field_name_new}_t;\n"
+                assert isinstance(field_name_new, str)
+                result += f"ALTER TABLE {table_name} RENAME COLUMN {field_name_old} TO {field_name_new};\n"
```
vkrasnovyd: This code block is relevant for the table field. For view field it's enough to recreate the view. For a side of n-m relation the intermidiate table should be updated.

openslides_backend/migrations/sql_diff_generator.py
```diff
@@ -248,34 +273,49 @@ def generate_constraints_sql(
...
+
+                is_view_field, _, write_fields = get_view_field_state_write_fields(
+                    collection_name,
+                    field_name,
+                    CURR_MODELS[collection_name]["fields"][field_name],
+                )
+                alter_views_conditionally(
+                    collection_name, field_name, bool(write_fields), is_view_field
+                )
+                # TODO recreate affected triggers
```
vkrasnovyd, line 318: For view fields changes in `to` don't matter because we recreate the whole view. For table fields changes in `to` are handled within `handle_rename`. Are there other cases for which this code block is relevant?

openslides_backend/migrations/sql_diff_generator.py
vkrasnovyd: Also, why `reference` and not `to`? Not all the relations have `reference` defined plus eventually we want to remove this attribute from our collections.

openslides_backend/migrations/sql_diff_generator.py
```diff
@@ -103,16 +114,18 @@ def generate_new_collection_sql(add: dict[str, Any], dc_add: dict[str, Any]) ->
...
+        alter_views.add(collection_name)
     return sql
 
 
-def generate_constraints_sql(
+def handle_add_field_attributes(
     table_name: str,
     field_name: str,
     field_def: dict[str, Any],
     diff_control_part: dict[str, Any],
```
vkrasnovyd: I really like the pairs of parameters like `field_def` + `dc_field_def` that you use in other functions. Such pairs of names make the functions easier to understand. Since you already make the names of this method and `handle_edit_field_attributes` more precise I'd suggest to unify this part too.

openslides_backend/migrations/sql_diff_generator.py
```diff
@@ -18,6 +23,8 @@
 The sql diff will be written to 'migrations/mig_[last migration number].*/schema_diff.sql'.
 """
 
+alter_views: set[str] = set()
```
vkrasnovyd, line 29: I like this way to avoid code repetition.

vkrasnovyd (changes requested):
A lot of great and important changes here. Handling our views is a serious challenge.

cli/util/util.py
hjanott: Yes. Will do.

openslides_backend/migrations/sql_diff_generator.py
hjanott: Fully agree.

openslides_backend/migrations/sql_diff_generator.py
hjanott: You're right. Will do.

openslides_backend/migrations/sql_diff_generator.py
hjanott: It should also be possible that a non relational field gets renamed. For those an update of views isn't necessary.

openslides_backend/migrations/sql_diff_generator.py
hjanott: I will actually use `ALTER VIEW RENAME TO`-syntax here to change the names. That seems overseeable and neet.

openslides_backend/migrations/sql_diff_generator.py
hjanott: That's actually a good idea.

openslides_backend/migrations/sql_diff_generator.py
hjanott: Done.

openslides_backend/migrations/sql_diff_generator.py
hjanott: I commented out parts of `alter_views_conditionally` probably it will be so simplistic that I may delete it again.
Relying on rename wouldn't be a good idea because the user may decide to just drop and recreate the column. However this may be possible to alter the views from add and remove, I believe that looking at each side will be more oversee-able.

cli/util/util.py
```diff
@@ -108,35 +109,26 @@ def get_view_field_state_write_fields(
...
     foreign_type: str
-    table_name: str = ""
-    field1: str = ""
-    field2: str = ""
     write_fields: tuple[str, str, str, list[str]] | None = None
 
-    # create TableFieldType own out of collection_name, field_name, value as field_def
-    own = TableFieldType(collection_name, field_name, value)
-    field_type = own.field_def.get("type", "")
+    # create TableFieldType own out of collection_name, field_name, field_def
```
vkrasnovyd: Seems to be redundant after the rename.

openslides_backend/migrations/sql_diff_generator.py
vkrasnovyd: `field_name` is not used after this change and can be deleted.

openslides_backend/migrations/yaml_diff_generator.py
```diff
@@ -19,6 +20,7 @@
 # for multi layered renames it will have to have that many migrations
 # Maybe future versions of this will allow multi layered renames including other changes within
 """
+Renames = tuple[dict[str, str], dict[str, dict[str, str]]]
```
vkrasnovyd, line 22: I like the usage of tuple here.

openslides_backend/migrations/yaml_diff_generator.py
```diff
@@ -50,52 +108,28 @@ def dumpjson(diff: dict[str, Any]) -> None:
...
     prev_models: dict[str, Any],
     curr_models: dict[str, Any],
-    renames_dict: dict[str, Any],
+    renames: Renames | dict,
 ) -> list[list[str] | dict[str, Any]] | None:
     missing_entries = []
     tree = {}
+    if isinstance(renames, tuple):
+        renames_dict = renames[0]
+        recurse_renames = {"fields": renames[1]}
```
vkrasnovyd, line 115: Smart way to handle the new structure of renames.

openslides_backend/migrations/sql_diff_generator.py
```diff
@@ -294,12 +294,11 @@ def handle_edit_field_attributes(
...
-                directory = MigrationHelper.get_last_migration_directory()
-                renames = MigrationHelper.get_migration_class(directory).renames
-                if table_name in renames or field_name in renames.get(
+                if table_name in RENAMES[0] or field_name in RENAMES[1].get(
                     table_name, {}
-                ).get("fields", {}):
+                ):
+                    # Shouldn't be a case since this is already skipped in yaml diff generator.
+                    # TODO decide whether to fail or delete
                     print(f"Skipping {table_name} since it is renamed.")
```
vkrasnovyd: Can also be `field_name`, not only `table_name`.

openslides_backend/migrations/sql_diff_generator.py
```diff
@@ -244,65 +269,101 @@ def generate_constraints_sql(
...
-            for field_name_old, field_name_new in value["fields"].items():
-                result += f"ALTER TABLE {collection_name_old}_t RENAME COLUMN {field_name_old} TO {field_name_new}_t;\n"
-                del dc_collection["fields"][field_name_old]
-            # TODO Renaming and redefining constraints
-            remove_empty(dc_collection, "fields")
-            remove_empty(dc_rename_dict, collection_name_old)
+    collection_renames = renames[0]
+    field_renames = renames[1]
+
+    for collection_name_old, field_name_new in collection_renames.items():
```
vkrasnovyd:
```suggestion
    for collection_name_old, collection_name_new in collection_renames.items():
```

openslides_backend/migrations/sql_diff_generator.py
```diff
@@ -244,65 +269,101 @@ def generate_constraints_sql(
...
+        table_name = f"{HelperGetNames.get_table_name(collection_name_old)}"
```
vkrasnovyd: How about `table_name_old` instead of just `table_name` to make code more self-explanatory?

openslides_backend/migrations/sql_diff_generator.py
```diff
@@ -244,65 +269,101 @@ def generate_constraints_sql(
...
+
+    for collection_name_old, collection_diff in field_renames.items():
+        dc_collection = cast(dict, dc_rename_dict[1][collection_name_old])
+        for field_name_old, field_name_new in collection_diff.items():
+            assert isinstance(field_name_new, str)
+            field_def = CURR_MODELS[collection_name_old]["fields"][field_name_new]
+            is_view_field = False
+            if field_def.get("to"):
+                # This also includes all sql fields
+                is_view_field, _, write_fields = get_view_field_state_write_fields(
```
vkrasnovyd:
```suggestion
                is_view_field, *_ = get_view_field_state_write_fields(
```

`write_fields` is never used in this code block.

openslides_backend/migrations/sql_diff_generator.py
```diff
@@ -318,46 +317,43 @@ def handle_edit_field_attributes(
...
+                    collection_name_old, field_name_new, field_def
+                )
+            if is_view_field:
+                result += f"ALTER VIEW {collection_name_old} "
+            else:
+                result += f"ALTER TABLE {table_name} "
```
vkrasnovyd: We always have to rename the view column. Renaming table column does not automatically lead to the name change of the view column.

vkrasnovyd (changes requested):
A lot of great changes already. Requesting mainly the codestyle changes.

openslides_backend/migrations/sql_diff_generator.py
hjanott: Good point. Found another error just because of this hint.

cli/util/util.py
hjanott: Deleting the comment.

openslides_backend/migrations/sql_diff_generator.py
hjanott: Deleting.

openslides_backend/migrations/sql_diff_generator.py
hjanott: I'm not exactly sure what you want me to do but will specify the print line here.

openslides_backend/migrations/sql_diff_generator.py
hjanott: Will do.

openslides_backend/migrations/sql_diff_generator.py
hjanott: Well spotted!

openslides_backend/migrations/sql_diff_generator.py
hjanott: Good catch! Will change that and also create helper functions for this.

openslides_backend/migrations/sql_diff_generator.py
```diff
@@ -66,12 +67,11 @@ def main() -> int:
...
-        sql += (
-            GenerateCodeBlocks.view_sql[collection_name]
-            .lstrip("\n")
-            .replace("CREATE", "CREATE OR REPLACE")
-        )
+    view_sql = "".join(
+        GenerateCodeBlocks.view_sql[collection_name]
+        for collection_name in sorted(alter_views)
+    )
+    sql += view_sql.replace("CREATE", "CREATE OR REPLACE").lstrip("\n")
```
vkrasnovyd, line 74: You used to remove `\n` for each line and now do it only once at the beginning of the joined string. Is it an intended logic change?

openslides_backend/migrations/sql_diff_generator.py
vkrasnovyd: That's the change that I meant. More precise this way.

openslides_backend/migrations/sql_diff_generator.py
```diff
@@ -244,65 +269,110 @@ def generate_constraints_sql(
...
+            HelperGetNames.get_table_name(collection_name_new),
+        )
+        result += AlterSchemaHelper.get_rename_view(
+            collection_name_old, collection_name_new
+        )
+        # TODO recreate dependend triggers
+        del dc_rename_dict[0][collection_name_old]
+
+    for collection_name, collection_diff in field_renames.items():
+        table_name_old = f"{HelperGetNames.get_table_name(collection_name_old)}"
```
vkrasnovyd:
```suggestion
        table_name_old = HelperGetNames.get_table_name(collection_name_old)
```

openslides_backend/migrations/sql_diff_generator.py
```diff
@@ -326,39 +328,46 @@ def handle_rename(renames: Renames, dc_rename_dict: Renames) -> str:
...
+        result += AlterSchemaHelper.get_rename_view(
+            collection_name_old, collection_name_new
+        )
         # TODO recreate dependend triggers
         del dc_rename_dict[0][collection_name_old]
 
-    for collection_name_old, collection_diff in field_renames.items():
-        dc_collection = cast(dict, dc_rename_dict[1][collection_name_old])
+    for collection_name, collection_diff in field_renames.items():
+        table_name_old = f"{HelperGetNames.get_table_name(collection_name_old)}"
```
vkrasnovyd: `collection_name_old` is a variable from the previous loop. Inside of this one it should be just `collection_name`.

vkrasnovyd: Also, `table_name_old` is used only once. `HelperGetNames.get_table_name` could also be called directly in the `AlterSchemaHelper.get_rename_table_column` call like you do above for `AlterSchemaHelper.get_rename_table`.

vkrasnovyd (changes requested):

openslides_backend/migrations/sql_diff_generator.py
hjanott: Yes. I decided that it's more readable if there is a paragraph between views.

openslides_backend/migrations/sql_diff_generator.py
hjanott: Did choose this way.

hjanott: Fixed it.

openslides_backend/migrations/sql_diff_generator.py
hjanott: Fixed it the other way.

